PR #13, opened 2018-10-14
Clearing task groups' "active" flag should happen on error

smurfix:


agronholm:
Hm, I already fixed this for trio. You didn't add any tests that would indicate the necessity for the changes in asyncio/curio, so are you really sure they're necessary?

smurfix:
Well … looking at the code shows that it's possible to leave the taskgroup's context with `_active` still true, and that can't be correct.

agronholm:
How? Can you write a test to point that out?

Issue #22, opened 2018-10-21
asyncio: Multiple consecutive cancel scopes cause a crash


smurfix:
The problem is that asyncio does not run callbacks immediately when a Future is done, so `CancelScope._task_done` is queued twice for the same task, which fails the second time around.

```
import asyncio
import anyio

async def main():
    async with anyio.open_cancel_scope() as scope:
        pass
    async with anyio.open_cancel_scope() as scope:
        pass
    await asyncio.sleep(0.1)

loop = asyncio.get_event_loop()
loop.run_until_complete(main())

```

agronholm:
Opening a cancel scope should not add a callback for the host task, so this is something that needs to be fixed. It may even necessitate a larger refactoring inside the asyncio backend.

smurfix:
No refactoring necessary. ;-)

agronholm:
Well yes it is because cancel scopes in asyncio and curio, as they are now, cannot support the use cases needed for feature parity with trio.

smurfix:
OK, that's correct, but not because of *this* bug.

agronholm:
Indeed. Cancel scopes and their interaction with task groups is probably the hardest thing here to wrap one's head around. I've been trying to sort out the larger issue with this and so far I've made little progress. But I *will* get there ;)

PR #23, opened 2018-10-23
Fix removing a concluded task from a cancel scope

smurfix:
Simply use a WeakSet.
This fixes issue #22.


agronholm:
What I couldn't accomplish on my own was writing a test for this. That's something you could help with if you're willing to.

smurfix:
I didn't manage that either – looks like pytest does a few interesting things behind the scenes which give asyncio enough leeway to run the callback "too" soon.

PR #25, opened 2018-10-26
Fixes bug#24

smurfix:
"parent_scope" ended up being the same as "scope". That can't work.


agronholm:
Thanks for the heads up on the bug, but I fixed this along with the same issue on curio and added a test too in the same commit.

PR #42, opened 2019-02-13
Fix networking tests on MacOS

standy66:
This PR fixes `OSError: AF_UNIX path too long` error in tests on MacOS due to long path to temporary socket (> 104 chars, which is seems to be [the limit](https://unix.stackexchange.com/questions/367008/why-is-socket-path-length-limited-to-a-hundred-chars)).

It also fixes assertion for getsockopt considering [different semantics of returned value on MacOS](https://stackoverflow.com/questions/31834030/reading-tcp-nodelay-using-getsockopt-returning-a-weird-value), e.g. in MacOS, the following code passes assertions

```python
sock = socket.socket(socket.AF_INET)
sock.setsockopt(socket.IPPROTO_TCP, socket.TCP_NODELAY, 1)

# 4 is the value of TF_NODELAY in netinet/tcp_var.h
# https://github.com/apple/darwin-xnu/blob/master/bsd/netinet/tcp_var.h#L263
assert sock.getsockopt(socket.IPPROTO_TCP, socket.TCP_NODELAY) == 4
```

### DIFF
--- a/tests/test_networking.py
+++ b/tests/test_networking.py
@@ -16,7 +16,7 @@ class TestTCPStream:
     async def test_receive_some(self):
         async def server():
             async with await stream_server.accept() as stream:
-                assert stream._socket.getsockopt(socket.IPPROTO_TCP, socket.TCP_NODELAY) == 1
+                assert stream._socket.getsockopt(socket.IPPROTO_TCP, socket.TCP_NODELAY) != 0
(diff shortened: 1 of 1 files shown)


agronholm:
Thanks. I should definitely set up CI for macOS and Windows as well.

PR #52, opened 2019-04-12
Cancel a new scope when its parent is already cancelled.

### LINKED ISSUE #51: Difference in cancellation behavior: asyncio vs. trio
```
import anyio
async def one():
    async with anyio.open_cancel_scope() as sc:
        await anyio.sleep(0.1)
    raise RuntimeError("This should not be printed")

async def bar(tg):
    try:
        await anyio.sleep(1)
    finally:
        await tg.spawn(one)
    pass
async def foo():
        await tg.spawn(bar,tg)
anyio.run(foo, backend="trio") # OK
anyio.run(foo, backend="asyncio")  # fails
```

smurfix:
Fixes #51.

### DIFF
--- a/anyio/_backends/asyncio.py
+++ b/anyio/_backends/asyncio.py
@@ -402,6 +402,9 @@ class TaskGroup:
         task = current_task()
         try:
             await func(*args)
+        except CancelledError as exc:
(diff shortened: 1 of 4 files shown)


agronholm:
I think this PR is obsolete because the tests added here already pass against current master.

PR #53, opened 2019-04-18
 Use `sniffio.current_async_library` directly

rnovatorov:
`anyio._detect_running_asynclib` extends `sniffio.current_async_library` by adding `curio` detection. Missing functionality has been added by python-trio/sniffio#5, so the wrapper is no longer needed.


rnovatorov:
Seems like `sniffio` version on PyPI is out of date. See: python-trio/sniffio#7.

agronholm:
I accidentally wrote the code myself before remembering this PR. Sorry. Thanks for the contribution anyway!

PR #60, opened 2019-05-29
Add an optional "lock" parameter to create_condition

smurfix:
A Condition frequently must be created with an explicit lock.
All three libraries support this.


agronholm:
Thanks!

Issue #63, opened 2019-07-01
abc.TaskGroup must define __aenter__ and __aexit__ methods


daa:
Currently ``mypy`` can't know if an object returned from ``anyio.create_task_group()`` may be used as async context manager because ``anyio.abc.TaskGroup`` does not define ``__aenter__()`` and ``__aexit__()`` methods and produces following errors:
```
error: "TaskGroup" has no attribute "__aenter__"
error: "TaskGroup" has no attribute "__aexit__"
```

PR #64, opened 2019-07-01
Add asynchronous context manager methods to task group base class

daa:
It's better for ``TaskGroup`` base class to define methos of asynchronous context manager -- ``__aexit__()`` and ``__aenter__()`` -- because otherwise ``mypy`` can't know that objects returned from ``anyio.create_task_group()`` may be used in ``async with`` statement.

Closes: #63

### DIFF
--- a/anyio/abc.py
+++ b/anyio/abc.py
@@ -2,9 +2,9 @@ from abc import ABCMeta, abstractmethod
 from io import SEEK_SET
 from ipaddress import IPv4Address, IPv6Address
 from ssl import SSLContext
+from types import TracebackType
 from typing import (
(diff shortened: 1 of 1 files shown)


agronholm:
Please stick to the existing code style.

agronholm:
Are you okay with only annotating and documenting the `__aenter__` and `__aexit__` of the `TaskGroup` abc and leaving the others as they are?

daa:
I'm okay with it. It's better to define public interface fully and to not have missing important parts but I've not met the need to change other classes yet. Also it would be better to create another issue in case of such a requirement.

agronholm:
Thanks.

PR #72, opened 2019-09-18
async and await is missing

ledmonster:
async and await seems to be missing here.


agronholm:
There is nothing missing. It just returns the coroutine object from the superclass method.

smurfix:
Well, the way the (async) context management protocol works, there's no practical difference.
While the new version is formally correct, the old version is a little bit faster.

ledmonster:
I see. Thanks you.

PR #75, opened 2019-09-29
Fix bug when entering/exiting task group as part of different asyncio tasks

### LINKED ISSUE #74: pytest fixture that uses TaskGroup fails to teardown when pytest-asyncio is installed
Hey,

First, I wanted to thank you @agronholm for this very well crafted library. I decided to use `anyio` in [asgi-lifespan](https://github.com/florimondmanca/asgi-lifespan), mostly to validate the idea of distributing packages built on top of `anyio`, and see how they perform in non-primarily-`anyio` environments.

In the context of HTTPX though and this issue in particular: https://github.com/encode/httpx/issues/350, I came across an tricky bug that occurs when `pytest-asyncio` is installed along with `anyio`.

(Off topic: this setup is very likely to occur for users that use asyncio as well as a library that uses `anyio` under the hood (e.g. someone building an asyncio-based ASGI web app and using `asgi-lifespan` as well). Which makes me thinkperhaps `pytest-anyio` should be a thing?)

Anyway, here's the bug:

I have this setup:

```bash
pip install anyio pytest pytest-asyncio
```

```python
# tests.py
import anyio
import pytest


@pytest.fixture
async def client():
        yield "client"


async def test_something(client):
    pass

```

If you run `$ pytest tests.py`, the test will fail during teardown:

```console
$ pytest tests.py
===================================================== test session starts =====================================================
platform darwin -- Python 3.7.3, pytest-5.1.3, py-1.8.0, pluggy-0.13.0
rootdir: /Users/florimond/Desktop/pytest-anyio-proposal
plugins: asyncio-0.10.0, anyio-1.1.0
collected 1 item                                                                                                              

tests.py sE                                                                                                             [100%]

=========================================================== ERRORS ============================================================
_____________________________________________ ERROR at teardown of test_something _____________________________________________

    @pytest.fixture
    async def client():
>           yield "client"

tests.py:8: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
venv/lib/python3.7/site-packages/anyio/_backends/_asyncio.py:398: in __aexit__
    ignore_exception = await self.cancel_scope.__aexit__(exc_type, exc_val, exc_tb)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <anyio._backends._asyncio.CancelScope object at 0x10a818ba8>, exc_type = None, exc_val = None, exc_tb = None

    async def __aexit__(self, exc_type, exc_val, exc_tb):
        self._active = False
        if self._timeout_task:
            self._timeout_task.cancel()
    
        host_task = current_task()
        print(type(host_task), host_task)
        print(host_task in self._tasks)
>       self._tasks.remove(host_task)
E       KeyError: <unprintable KeyError object>

venv/lib/python3.7/site-packages/anyio/_backends/_asyncio.py:232: KeyError
-------------------------------------------------- Captured stdout teardown ---------------------------------------------------
<class '_asyncio.Task'> <Task pending coro=<pytest_fixture_setup.<locals>.wrapper.<locals>.finalizer.<locals>.async_finalizer() running at /Users/florimond/Desktop/pytest-anyio-proposal/venv/lib/python3.7/site-packages/pytest_asyncio/plugin.py:85> cb=[_run_until_complete_cb() at /Users/florimond/.pyenv/versions/3.7.3/lib/python3.7/asyncio/base_events.py:158]>
False
====================================================== warnings summary =======================================================
tests.py::test_something
  /Users/florimond/Desktop/pytest-anyio-proposal/venv/lib/python3.7/site-packages/_pytest/python.py:160: PytestUnhandledCoroutineWarning: async def functions are not natively supported and have been skipped.
  You need to install a suitable plugin for your async framework, for example:
    - pytest-asyncio
    - pytest-trio
    - pytest-tornasync
    warnings.warn(PytestUnhandledCoroutineWarning(msg.format(pyfuncitem.nodeid)))

-- Docs: https://docs.pytest.org/en/latest/warnings.html
=========================================== 1 skipped, 1 warnings, 1 error in 0.15s ===========================================
```

The print statements in there are mine — it shows that the task group's cancel scope is trying to remove the host task from its set of tasks, but it's not there, leading to a `KeyError` when calling `.remove()`.

I looked at what exactly `pytest-asyncio` is doing for async fixtures, and here's what it looks like ([code](https://github.com/pytest-dev/pytest-asyncio/blob/86cd9a6fd248e1dc0a7de9c6d60dbf3773aa1fa6/pytest_asyncio/plugin.py#L82-L102
)):

```python
            gen_obj = generator(*args, **kwargs)

            async def setup():
                res = await gen_obj.__anext__()
                return res

            def finalizer():
                """Yield again, to finalize."""
                async def async_finalizer():
                    try:
                        await gen_obj.__anext__()
                    except StopAsyncIteration:
                        pass
                    else:
                        msg = "Async generator fixture didn't stop."
                        msg += "Yield only once."
                        raise ValueError(msg)
                asyncio.get_event_loop().run_until_complete(async_finalizer())

            request.addfinalizer(finalizer)
            return asyncio.get_event_loop().run_until_complete(setup())
```

So indeed, `host_task` in `CancelScope.__aenter__()` and `CancelScope.__aexit__()` doesn't refer to the same coroutine. In one case it's `setup()`, in the other it's `async_finalizer()`. This explains the `KeyError` — we never added `async_finalizer()` to `self._tasks` in the first place.

I tried to bypass the issue using:

```python
if host_task in self._tasks:
    self._tasks.remove(host_task)
    _task_states[host_task].cancel_scope = self._parent_scope
...
```

But this resulted in teardown hanging indefinitely, because of this part in `TaskGroup.__aexit__()`:

```python
while self.cancel_scope._tasks:
        await asyncio.wait(self.cancel_scope._tasks)
```

Since we're not removing finished tasks from `self.cancel_scope._tasks` (e.g. `setup()` in this case), this loops indefinitely. I managed to fix it using:

```python
while self.cancel_scope._tasks:
    done, pending = await asyncio.wait(self.cancel_scope._tasks)
    for task in done:
        self.cancel_scope._tasks.remove(task)
```

Not sure this won't break anything else, so I'll happily open a PR with this change to see how it performs against the test suite. :+1:

florimondmanca:
Fixes #74

### DIFF
--- a/anyio/_backends/_asyncio.py
+++ b/anyio/_backends/_asyncio.py
@@ -227,8 +227,14 @@ class CancelScope:
             self._timeout_task.cancel()
 
         host_task = current_task()
-        self._tasks.remove(host_task)
-        _task_states[host_task].cancel_scope = self._parent_scope
+        if host_task not in self._tasks:
+            # `__aexit__()` was called from a different host task than `__aenter__()`.
+            # As a result, this particular `host_task` has never been registered,
+            # so we don't need to clean it up.
+            pass
+        else:
+            self._tasks.remove(host_task)
+            _task_states[host_task].cancel_scope = self._parent_scope
 
         exceptions = exc_val.exceptions if isinstance(exc_val, ExceptionGroup) else [exc_val]
         if all(isinstance(exc, CancelledError) for exc in exceptions):
@@ -400,7 +406,10 @@ class TaskGroup:
                 self._exceptions.append(exc_val)
 
         while self.cancel_scope._tasks:
-            await asyncio.wait(self.cancel_scope._tasks)
+            done, pending = await asyncio.wait(self.cancel_scope._tasks)
+            for task in done:
+                if task in self.cancel_scope._tasks:
+                    self.cancel_scope._tasks.remove(task)
 
         self._active = False
         if not self.cancel_scope._parent_cancelled():
--- a/tests/test_taskgroups.py
+++ b/tests/test_taskgroups.py
@@ -1,3 +1,5 @@
+import sys
+
 import curio
 import pytest
 import trio
@@ -466,3 +468,29 @@ async def test_nested_shield():
                     await tg.spawn(killer, scope)
                     async with fail_after(0.2):
                         await sleep(2)
+
+
+@pytest.mark.skipif(sys.version_info < (3, 6), reason="Requires Python 3.6 or higher.")
+def test_asyncio_enter_exit_different_tasks():
+    """Regression test for: https://github.com/agronholm/anyio/issues/74."""
+    import asyncio
+
+    # Don't instanciate here or it will try to sniff the async library.
+    task_group = None
+
+    async def setup():
+        nonlocal task_group
+        task_group = create_task_group()
+        await task_group.__aenter__()
+
+    async def teardown():
+        assert task_group is not None
+        await task_group.__aexit__(None, None, None)
+
+    # Run `.run_until_complete()` twice (enter then exit) on the same loop.
+    # This is what pytest-asyncio does to setup and teardown async fixtures,
+    # and what initially prompted a bug fix.
+    loop = asyncio.new_event_loop()
+    asyncio.set_event_loop(loop)
+    loop.run_until_complete(setup())
+    loop.run_until_complete(teardown())


agronholm:
I do not expect I will merge this, as I have my own fix underway which also tests all backends, not just asyncio.

agronholm:
This brings to light a fairly big issue: say task A enters the cancel scope and then hits a yield. Then another task resumes the generator. Next, a spawned task crashes. The host task must now be cancelled, but how do we get the host task now?

agronholm:
I tested this on trio and it simply crashes with a `Cancelled` exception when there is no host task to receive the exception.

agronholm:
Although anyio currently trips on `__aexit__()` every time, even if that is fixed, any tasks spawned from fixtures will cause undefined behavior if they crash.

agronholm:
This has been a problem for quite a while: https://discuss.python.org/t/preventing-yield-inside-certain-context-managers/109

agronholm:
There is already a bug open against pytest-asyncio about this issue: https://github.com/pytest-dev/pytest-asyncio/issues/124
I may have missed this before, but is there a specific reason you're running the test suite with pytest-asyncio installed?

smurfix:
Well, I'm not @florimondmanca but on my system I have a heap of normally-benign modules installed so that I can test various stuff while I develop (and thus break) things. Requiring the removal (or reinstallation) of pytest-asyncio is no fun. Worse, I can't test that the module I'm writing works in a "pure" asyncio environment *and* in one that's running under trio, asyncio+anyio, and/or whatever.

agronholm:
That is unfortunate, but there is little I can do about the invasive behavior of pytest-asyncio. You can of course use AnyIO's pytest plugin to test your pure async stuff, as it only acts as a light wrapper around `asyncio.run()`.

florimondmanca:
> There is already a bug open against pytest-asyncio about this issue: pytest-dev/pytest-asyncio#124

Nice, thanks. There’s even a pending PR attempting to solve it: https://github.com/pytest-dev/pytest-asyncio/pull/125

> I may have missed this before, but is there a specific reason you're running the test suite with pytest-asyncio installed?

I was providing a code snippet to someone who runs on pytest-asyncio, and one of the components in that snippet uses anyio under the hood. I would assume this will be a common situation unless we propagate the need to know about anyio (use the anyio plugin) up to anyone who depends on anything that uses anyio. The compatibility layer at the run function level is looking good already, this is just one edge case that I think needs fixing on the pytest-asyncio side.

agronholm:
Fixing #74 will at least let people use task groups inside async fixtures. But the handling of exceptions from tasks spawned from those task groups is still a tricky problem.

agronholm:
There is also the problem of async fixtures with the `autouse=True` flag set. They may not be bound to any specific test so the pytest plugin may not recognize them as fixtures it needs to handle.

agronholm:
I now understand the reason why the asyncio test run did not show as failing. This is because `asyncio.run()` shuts down async generators on exit...including the fixture.

florimondmanca:
@agronholm Exactly — the fixture gets torn down by running the second part of the async generator, but that fails and so the test is marked as errorred, when it would have ended up being marked as skipped otherwise.

agronholm:
No, I think you misunderstood. The reason why my test failed on asyncio is that the cancel scope is both entered **and** exited during test setup (due to the semantics of `asyncio.run()`. The `StopIteration` / `StopAsyncIteration` exception are ignored.

agronholm:
I need to look at both trio and curio codebases to see how they handle this.

agronholm:
I don't think trio shuts down async generators at the end, and neither does curio.
I experimented by calling `loop.run_until_complete()` directly instead of `asyncio.run()`. This change allowed the `KeyError` to pass through as expected.

florimondmanca:
Is there any branch or PR where the tests you’re mentioning are available to look at? Would help me better follow the discussion here. I thought you were mentioning the original issue in #74. :)

agronholm:
Closing in favor of #76.

PR #79, opened 2019-10-05
Capacity limiter improvements

agronholm:
This pull request contains two loosely related features:

1. The addition of the `CapacityLimiter.set_total_tokens()` method
1. The addition of the `anyio.current_default_thread_limiter()` function

I was torn between following trio's example and adding a setter method for the `total_tokens` property (and having to work around curio's limitations) vs adding a coroutine method for setting the property. I opted for the later approach because the curio workaround would have been pretty ugly.

### DIFF
--- a/docs/synchronization.rst
+++ b/docs/synchronization.rst
@@ -186,3 +186,6 @@ Example::
                 await tg.spawn(use_resource, num, limiter)
 
     run(main)
+
+To adjust the number of total tokens, you can use the
+:meth:`~anyio.abc.CapacityLimiter.set_total_tokens` method.
--- a/anyio/__init__.py
+++ b/anyio/__init__.py
@@ -247,6 +247,16 @@ def run_async_from_thread(func: Callable[..., Coroutine[Any, Any, T_Retval]], *a
     return asynclib.run_async_from_thread(func, *args)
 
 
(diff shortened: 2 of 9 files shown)


agronholm:
@florimondmanca Here's your first PR for reviewing.

docs/synchronization.rst
```diff
@@ -186,3 +186,10 @@ Example::
                 await tg.spawn(use_resource, num, limiter)
 
     run(main)
+
+To adjust the number of total tokens, you can use the
+:meth:`~anyio.abc.CapacityLimiter.set_total_tokens` method.
+
+.. note:: In Trio, the total number of tokens capacity limiters is adjusted by setting the value
+    of the ``total_tokens`` property. Because waking up waiting tasks requires an ``await`` in
+    the curio backend, this cannot be supported in AnyIO.
```
florimondmanca: It seems to me like this note could be removed entirely. When I first read it I wondered "so is adjusting the number of tokens not supported?", but what it says is really that you need to use `.set_total_tokens()` instead of setting `.total_tokens`. I think we don't need this information, especially because understanding it requires the user to know about subtleties of trio and curio.

agronholm: Removed the comment.

agronholm:
Thanks for the review!

PR #81, opened 2019-10-18
Added the "cancellable" parameter to run_in_thread()

agronholm:


anyio/_backends/_asyncio.py
```diff
@@ -394,31 +394,43 @@ def __init__(self):
...
                         limiter: Optional['CapacityLimiter'] = None) -> T_Retval:
     def thread_worker():
         try:
             with claim_worker_thread('asyncio'):
                 _local.loop = loop
                 result = func(*args)
         except BaseException as exc:
-            loop.call_soon_threadsafe(queue.put_nowait, (None, exc))
+            if not loop.is_closed():
+                asyncio.run_coroutine_threadsafe(limiter.release_on_behalf_of(task), loop)
```
florimondmanca: It seems possible to me that `limiter` might be `None` here?

agronholm: Under what circumstances?

florimondmanca: Oh, nevermind. I was saying that because of `limiter: Optional['CapacityLimiter'] = None` in the signature, but now I see `limiter` is defaulted to `_default_thread_limiter` further in this function.

agronholm:
Thanks for the review!

PR #84, opened 2019-11-22
Add type annotation for `CancelScope.cancel`

tkukushkin:
Current mypy output:
```
foo.py:218:19: error: Call to untyped function "cancel" of "CancelScope" in typed context  [no-untyped-call]
                      ^
```


agronholm:
Thanks.

PR #87, opened 2019-11-30
Fix test_concurrent_write on macOS

pquentin:
The test was flaky on macOS because the second `client.send_all()` did not always raise as it happened before the `client.send_all()` in `send_data`. Sending more data fixed the problem somehow.

### DIFF
--- a/tests/test_networking.py
+++ b/tests/test_networking.py
@@ -1,4 +1,3 @@
-import platform
 import socket
 import ssl
 import sys
@@ -235,8 +234,6 @@ class TestTCPStream:
                 client.setsockopt(socket.SOL_SOCKET, socket.SO_RCVBUF, 80000)
                 assert client.getsockopt(socket.SOL_SOCKET, socket.SO_RCVBUF) in (80000, 160000)
 
-    @pytest.mark.xfail(condition=platform.system() == 'Darwin',
-                       reason='Occasionally fails on macOS')
     @pytest.mark.anyio
     async def test_concurrent_write(self, localhost):
         async def send_data():
@@ -250,7 +247,7 @@ class TestTCPStream:
                     await wait_all_tasks_blocked()
                     try:
                         with pytest.raises(ResourceBusyError) as exc:
-                            await client.send_all(b'foo')
+                            await client.send_all(b'\x00' * 1024000)
 
                         exc.match('already writing to')
                     finally:


agronholm:
I'd really like to understand what is happening here. I tried to fix this by sending 10 megabytes with one `send()` in the background task, since I doubt sending 10 MB with one system call is possible. Did you check if the `send()` in the background task completes before the `send('foo')` starts? That'd be the key.

pquentin:
> Did you check if the send() in the background task completes before the send('foo') starts? That'd be the key

Do you mean before or after my "fix"?

agronholm:
Before. I have no clue why your fix worked (assuming you ran it multiple times without encountering issues).

agronholm:
I have my qemu mac working now and I can repro the issue there.

agronholm:
My tests on mac confirm that `await client.send_all(b'foo')` starts and finishes after `await send_all(b'\x00' * 1024000)`, in which case it should report an error because the previous send is still in progress.

agronholm:
After extensive debugging, I now have a fairly good idea of what's going on.
This is the debugging output I get:
```
send_data starting
sendall(1024000 bytes) starting
waiting for socket to become writable
foo starting
sendall(3 bytes) starting
sendall(3 bytes) done
foo done
```
So here's what's happening, on the Python level at least:

1. The background task starts sending data
1. `BaseSocket.send_all()` raises a `BlockingIOError`, so it starts waiting for the socket to become writable
1. The main test task starts writing `foo` to the socket
1. For whatever reason, the write does not block, so the 3 bytes are written to the socket

The test fails because `send(b'foo')` is supposed to block but it doesn't. The kernel buffer should be full since we have nothing actually reading from the socket. What also mystifies me is the fact that this only fails randomly on the asyncio and curio backends, not trio, even though anyio only uses the lowest possible API calls to wait for socket writability.

I will try to use strace to figure out what's going on with the system calls.

agronholm:
In all honesty though, we should probably be checking for an existing writability wait event at the beginning of each write operation. That would definitely fix the issue.

agronholm:
Bah, I couldn't figure out how to trace system calls on macOS.

agronholm:
> In all honesty though, we should probably be checking for an existing writability wait event at the beginning of each write operation. That would definitely fix the issue.

While this sounded like a nice idea, I don't know to do this with trio. I also don't know how trio avoids this same issue.

njsmith:
It sounds like the issue is that there are cases where the kernel will reject a large `send`, but accept a small one. I guess that makes some sense... if the send buffer has like 1 KB of space in it, and someone tries to do a 10 MB write, then there's not much point in accepting the first 1 KB. I think you can work around this by doing large writes until you get a `BlockingIOError`, and then switch to doing small writes until you get a `BlockingIOError`?

Possibly Trio is just getting lucky, since we happen to use 64 KiB chunks when filling a buffer, and maybe that's small enough that macOS doesn't trigger this path? https://github.com/python-trio/trio/blob/0ef5d7260945a30ecbe0fe943ab8d6c01de3fc6f/trio/_core/tests/test_io.py#L19

I've seen a similar issue before with pipes... not with sockets though: https://github.com/python-trio/trio/blob/0ef5d7260945a30ecbe0fe943ab8d6c01de3fc6f/trio/tests/test_unix_pipes.py#L30-L49

njsmith:
I still don't know why you're seeing different behavior on trio vs the other backends though :-/

njsmith:
I think this is how to do strace on macos: https://stackoverflow.com/questions/31045575/how-to-trace-system-calls-of-a-program-in-mac-os-x

agronholm:
So, I've been digging deeper into this. What seems to happen here is:

1. The `send_data()` task sends until the data cannot fit into the buffer anymore

agronholm:
I ran into the same issue with asyncio on Linux, and _reducing_ the send packet size seemed to help. I'm closing this for now, but I will revisit the issue if problems arise.

PR #89, opened 2019-12-07
Made the pytest plugin more flexible about the backend

agronholm:
This change makes it possible to specify on a per-test basis which backend(s) it should run against.

### DIFF
--- a/anyio/pytest_plugin.py
+++ b/anyio/pytest_plugin.py
@@ -85,7 +85,7 @@ def pytest_pyfunc_call(pyfuncitem):
     try:
         backend = pyfuncitem.callspec.getparam('anyio_backend')
     except (AttributeError, ValueError):
-        return False
+        return None
 
     if backend:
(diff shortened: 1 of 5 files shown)


anyio/pytest_plugin.py
```diff
@@ -58,45 +58,46 @@ def wrapper(*args, **kwargs):
...
-    if marker:
-        metafunc.fixturenames.append('anyio_backend')
-
-    if 'anyio_backend' in metafunc.fixturenames:
-        backends = metafunc.config.getoption('anyio_backends')
-        if backends == 'all':
-            backends = BACKENDS
-        else:
+    if marker or 'anyio_backend' in metafunc.fixturenames:
+        backends = marker.kwargs.get('backend') if marker else []
```
smurfix: why not "else None"?

agronholm: You're looking at an outdated commit. Another hook expected `backends` to be iterable.

PR #90, opened 2019-12-11
Correct TaskGroup.__aexit__ types

daa:
Recently I've noticed that ``__aexit__()`` argument types are ``Optional``s (https://mypy.readthedocs.io/en/latest/protocols.html#context-manager-protocols) so I propose this change.
It seems that I've missed this detail when I submitted my previous pull request adding type information to ``TaskGroup.__aexit__()`` method, sorry for that, and thank you very much for your work!

### DIFF
--- a/anyio/abc.py
+++ b/anyio/abc.py
@@ -240,8 +240,9 @@ class TaskGroup(metaclass=ABCMeta):
         """Enter the task group context and allow starting new tasks."""
 
     @abstractmethod
-    async def __aexit__(self, exc_type: Type[BaseException], exc_val: BaseException,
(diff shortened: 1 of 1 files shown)


agronholm:
Would you mind adding typing information to other `__aexit__()` methods as well? There are quite a few.

daa:
I was not sure if you meant adding types to ``abc`` objects only or to all so I added types to each object's ``__aexit__()`` methods . Also I had to change body of some of the backends` ``__aexit__()`` methods because ``mypy`` showed typing errors in them.

agronholm:
Thanks!

daa:
Thank you for merge!

PR #103, opened 2020-03-01
Fixed compatibility with Curio 1.0

### LINKED ISSUE #101: Curio backend needs to be updated to support Curio 1.0
Curio 1.0 has been released a few days ago with lots of breaking changes; for instance, the `report_crash=...` keyword argument of `curio.spawn()` [has been removed](https://github.com/dabeaz/curio/blob/master/CHANGES#L203), which breaks Anyio when used with Curio 1.0. Most of the unit tests also fail on my machine because of this. I'm not experienced in Curio myself so I'd rather not attempt fixing it on my own.

agronholm:
Fixes #101.

### DIFF
--- a/anyio/_backends/_curio.py
+++ b/anyio/_backends/_curio.py
@@ -1,10 +1,13 @@
+import inspect
 import math
 import socket
-from collections import OrderedDict
+from collections import OrderedDict, defaultdict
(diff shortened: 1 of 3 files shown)


agronholm:
Thanks for the review. I should be able to cut a release within the next few days.

PR #104, opened 2020-03-22
Added the address and peer_address properties to SocketStream

agronholm:


tzickel:
I am not versed in all this layers of inheritance, but it seems OK :)

PR #106, opened 2020-05-19
Fix trio 0.15 warnings

smurfix:


njsmith:
Diff looks plausible to me. You have test failures though, and GitHub isn't letting me see the logs on my phone, so maybe everything is broken, I dunno :-)

I'm curious why anyio needs to subclass `trio.Condition` instead of just using it.

agronholm:
> I'm curious why anyio needs to subclass `trio.Condition` instead of just using it.

Because the `notify` and `notify_all` methods are coroutine methods in anyio and regular methods in trio.

agronholm:
> Diff looks plausible to me. You have test failures though, and GitHub isn't letting me see the logs on my phone, so maybe everything is broken, I dunno :-)

`./anyio/_backends/_trio.py:250:1: W293 blank line contains whitespace`

anyio/_backends/_trio.py
```diff
@@ -240,12 +240,19 @@ async def wait(self):
         await self._event.wait()
 
 
-class Condition(trio.Condition):
+class Condition:
```
agronholm, line 243: Why is this change needed? If somebody is expecting anyio's Condition to be type compatible with trio's, this change breaks that assumption.

anyio/_backends/_trio.py
```diff
@@ -2,11 +2,11 @@
 from types import TracebackType
 from typing import Callable, Optional, List, Type, Union
 
-import trio.hazmat
+import trio.lowlevel
```
agronholm, line 5: This module does not exist in trio before v0.15. It would be better to use a conditional import for now.

anyio/_backends/_trio.py
smurfix: Because subclassing trio.Condition now triggers a warning.
That expectation would fail anyway whenever a backend other than `trio` is used.

njsmith: We deprecated inheriting from most of trio's public types. More details: https://github.com/python-trio/trio/issues/1044

anyio/_backends/_trio.py
smurfix: Done. (For some values of "done" …)

smurfix:
OK, the Lock argument was missing. No longer testing against Py3.5 with Trio 0.15 should probably be fixed by somebody more familiar with the test setup.

smurfix:
>  /opt/hostedtoolcache/Python/3.8.2/x64/lib/python3.8/site-packages/curio/pytest_plugin.py:60: PytestDeprecationWarning: direct construction of Function has been deprecated, please use Function.from_parent

Not my problem either, I'm afraid …

anyio/_backends/_trio.py
```diff
@@ -240,12 +246,19 @@ async def wait(self):
         await self._event.wait()
 
 
-class Condition(trio.Condition):
+class Condition:
+    def __init__(self, lock: trio.Lock = None):
```
agronholm, line 250: Use `Optional[trio.Lock]` for the type hint.

anyio/_backends/_trio.py
```diff
@@ -2,12 +2,18 @@
 from types import TracebackType
 from typing import Callable, Optional, List, Type, Union
 
-import trio.hazmat
+try:
```
agronholm, line 5: Move the conditional imports after all others, as per PEP 8. I'm also not in favor of altering the `trio` package, so maybe instead alias the import as `lowlevel` or `trio_lowlevel`.

anyio/_backends/_trio.py
smurfix: I was lazy because AFAIK `Optional` is auto-implied when the default is `None`.

smurfix: https://docs.python.org/3/library/typing.html#typing.get_type_hints 

> Optional[t] is added for function and method annotations if a default value equal to None is set.

But I added it anyway because you're doing it everywhere else. ;-)

pquentin:
Trio 0.15.0 no longer supports Python 3.5, what is AnyIO's policy here?

agronholm:
AnyIO 1.x still supports py3.5 but AnyIO 2.0 (in development) has dropped support. I will look at modifying the test suite to skip trio testing on py3.5.

agronholm:
Thanks!

PR #108, opened 2020-05-21
Keep consistent event loop across multiple asyncio runs

### LINKED ISSUE #107: New event loop is created across different runs in asyncio
I think this is not intentional - the implication is that setting new event loop policies will possibly lose the current event loop:

https://github.com/agronholm/anyio/blob/14daa2bad967bf6dd0f96d04ecffe8e383985195/anyio/_backends/_asyncio.py#L71-L86

This causes the pytest plugin to have different event loops across fixtures and tests:

```python
import asyncio

import pytest


@pytest.fixture
async def loop():
    return asyncio.get_event_loop()


@pytest.mark.anyio
async def test(loop):
    assert loop is asyncio.get_event_loop()  # fails here
```

In many cases, the fixture should share the same loop as the test, for example:

```python
import asyncpg

import pytest


@pytest.fixture
async def conn():
    return await asyncpg.connect("postgresql:///")


@pytest.mark.anyio
async def test(conn):
    assert await conn.fetchval("SELECT 123") == 123
```

```
E   RuntimeError: Task <Task pending name='Task-5' coro=<run.<locals>.wrapper() running at .../anyio/_backends/_asyncio.py:67> cb=[run_until_complete.<locals>.<lambda>()]> got Future <Future pending cb=[Protocol._on_waiter_completed()]> attached to a different loop
```

One possible solution is to cache all policy instances under different setups, and switch to use the right one. I'll create a PR with a more complete test.

fantix:
Fixes #107.

### DIFF
--- a/anyio/_backends/_asyncio.py
+++ b/anyio/_backends/_asyncio.py
@@ -59,8 +59,24 @@ _native_task_names = hasattr(asyncio.Task, 'get_name')
 # Event loop
 #
 
+# Must be explicitly set on Python 3.8 for now or wait_socket_(readable|writable) won't work
+if sys.platform == 'win32' and sys.version_info >= (3, 8):
+    _default_policy = asyncio.WindowsSelectorEventLoopPolicy()  # type: ignore
+else:
+    _default_policy = asyncio.DefaultEventLoopPolicy()
+_uvloop_policy = None
+# Use uvloop on CPython only
+if sys.implementation.name == 'cpython':
+    try:
+        import uvloop
+    except ImportError:
+        pass
+    else:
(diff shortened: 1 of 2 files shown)


agronholm:
Was your plan to only do this with uvloop?

fantix:
No, it's not uvloop-specific. Please kindly find comments in #107.

agronholm:
After reviewing this, I decided to turn it down. I don't want asyncio specific, much less uvloop specific, code in the test suite. The problem is real but these tests don't even test for it (say, being able to start a network service in a fixture and testing against it in the actual test). Even if I wanted to tackle it just for asyncio, the approach here is wrong. It should not replace the event loop policy.

PR #113, opened 2020-06-05
Don't die when there's no current task

smurfix:
asyncio's current_task() can be `None`.
weakrefs to `None` are not possible.

### DIFF
--- a/anyio/_backends/_asyncio.py
+++ b/anyio/_backends/_asyncio.py
@@ -686,13 +686,13 @@ class Semaphore(asyncio.Semaphore):
 
 
 class Queue(asyncio.Queue):
-    def get(self):
+    async def get(self):
         check_cancelled()
-        return super().get()
+        return await super().get()
 
-    def put(self, item):
+    async def put(self, item):
         check_cancelled()
-        return super().put(item)
+        return await super().put(item)
 
     def __aiter__(self):
         return self
--- a/anyio/_backends/_asyncio.py
+++ b/anyio/_backends/_asyncio.py
@@ -661,24 +661,24 @@ class Condition(asyncio.Condition):
     async def notify_all(self):
         super().notify(len(self._waiters))
 
-    def wait(self):
+    async def wait(self):
         check_cancelled()
-        return super().wait()
+        return await super().wait()
 
 
 class Event(asyncio.Event):
     async def set(self):
         super().set()
 
-    def wait(self):
+    async def wait(self):
         check_cancelled()
-        return super().wait()
+        return await super().wait()
 
 
 class Semaphore(asyncio.Semaphore):
-    def __aenter__(self):
+    async def __aenter__(self):
         check_cancelled()
-        return super().__aenter__()
+        return await super().__aenter__()
 
     @property
     def value(self):
@@ -697,9 +697,9 @@ class Queue(asyncio.Queue):
     def __aiter__(self):
         return self
 
-    def __anext__(self):
+    async def __anext__(self):
         check_cancelled()
-        return super().get()
+        return await super().get()
 
 
 class CapacityLimiter:


smurfix:
… don't ask me what mypy's problem is …

agronholm:
I'm curious: under what circumstances did this come up?

smurfix:
I hit that when porting xknx to anyio. That problem was triggered (and swallowed) by a test failure while some tasks were started by `asyncio.create_task` and others via a taskgroup.

Yes I know, I should have saved that code for later analysis / creation of a testcase. :-/ After completing the code's transition to anyio I can't reproduce it any more. I can try to recreate it using my intermediate check-ins.

agronholm:
> I can try to recreate it using my intermediate check-ins.

That would be appreciated.

smurfix:
Found it.
```
  File "/usr/lib/python3.8/asyncio/events.py", line 81, in _run
    self._context.run(self._callback, *self._args)
  File "/usr/lib/python3.8/asyncio/selector_events.py", line 1019, in _read_ready
    self._protocol.datagram_received(data, addr)
  File "/src/xknx/xknx/io/udp_client.py", line 56, in datagram_received
    self.data_received_callback(data)
  File "/src/xknx/xknx/io/udp_client.py", line 90, in data_received_callback
    self.handle_knxipframe(knxipframe)
  File "/src/xknx/xknx/io/udp_client.py", line 99, in handle_knxipframe
    callback.callback(knxipframe, self)
  File "/src/xknx/xknx/io/tunnel.py", line 70, in tunnel_reqest_received
    self.telegram_received_callback(telegram)
  File "/src/xknx/xknx/io/knxip_interface.py", line 168, in telegram_received
    self.xknx.telegrams.put(telegram))
  File "/src/anyio/anyio/_backends/_asyncio.py", line 694, in put
    check_cancelled()
  File "/src/anyio/anyio/_backends/_asyncio.py", line 260, in check_cancelled
    cancel_scope = _task_states[current_task()].cancel_scope
  File "/usr/lib/python3.8/weakref.py", line 383, in __getitem__
    return self.data[ref(key)]
TypeError: cannot create weak reference to 'NoneType' object
```
The code is at git@github.com:M-o-a-T/xknx.git, branch "anyio", commit f8ea576 started to exhibit the problem, it got fixed in commit 7dd48f3.

smurfix:
```
def telegram_received(self, telegram):
        """Put received telegram into queue. Callback for having received telegram."""
        loop = asyncio.get_event_loop()
        loop.create_task(
            self.xknx.telegrams.put(telegram))
```
`telegrams` is an anyio queue.

smurfix:
Thus, the reproducer:
```
import anyio
import asyncio

q = None
def cb():
    print("duh")
    loop = asyncio.get_event_loop()
    loop.create_task(q.put("Hello"))

async def main():
    global q
    q = anyio.create_queue(1)
    f = asyncio.Future()
    asyncio.get_event_loop().call_later(0.1, cb)
    print(await q.get())
asyncio.run(main())
```

smurfix:
OK, this should fix the actual problem.

agronholm:
Ok, it seems that the problem stems from using AnyIO code from asyncio-only code (outside an asyncio task, that is). Is the plan to make xknx compatible with other supported backends too?

agronholm:
By the way, `Queue.put()` is a coroutine function, so the above code would never work anyway. The fact that you're running into this particular problem is just an implementation side effect.

smurfix:
Umm, yes, I know that `Queue.put` is supposed to be a coroutine function, but before my patch it actually was not – it checked the cancel state synchronously.

The stack trace just looks totally wrong because the `create_task` call which wraps it is on the previous line … the only clue are the two closing parentheses. :-P

This can't bite in non-asyncio situations because anyio and trio always pass along coroutine functions, not coroutines.

I'm actually done with anyio-izing `xknx` (other than replacing some more stupid callbacks with `async for`-style loops), it now works perfectly with both asyncio and trio (assuming that the USB socket fix is applied). Next step, asking its maintainer to please consider merging all that work …

agronholm:
Ok. While we've established that the error came from wrongful use of the AnyIO API, I suppose there is no harm in merging this in, as the error message was not very indicative of the real problem.

agronholm:
Thanks.

smurfix:
Umm, IMHO that's not wrongful use of the AnyIO API, that is (or rather was) was AnyIO's violation of asyncio's assumptions.

To elaborate, with asyncio,
```
    coro = coro_func(args)
    await coro
```
implies that the first line will not actually execute **any** code in `coro_func`. You violate that assumption, your code may go splat. The problem doesn't exist with AnyIO (or Trio) because we always pass `coro_func` around (and its arguments) instead.

After applying my fix, the code which I quoted above works.

PR #115, opened 2020-06-07
Multicast support for UDP sockets

smurfix:
Most multicast options for UDP can be set by the application after `create_udp_socket` returns, but two cannot:
* Multicast requires specifying an address family for the socket, but you can't bind to an actual address.
* There may be more than one multicast user on the same host, so SO_REUSEADDR needs to be enabled.
With these changes, my anyio-enabled `xknx` fork works.

### DIFF
--- a/anyio/__init__.py
+++ b/anyio/__init__.py
@@ -520,7 +520,7 @@ async def create_udp_socket(
     if interface:
         interface, family, _v6only = await _networking.get_bind_address(interface)
     else:
-        interface, family = None, address_family
+        interface, family = None, address_family or socket.AF_UNSPEC
 
     if target_host:
         res = await run_in_thread(socket.getaddrinfo, target_host, target_port, family)
--- a/anyio/__init__.py
+++ b/anyio/__init__.py
@@ -533,10 +533,7 @@ async def create_udp_socket(
     raw_socket = socket.socket(family=family, type=socket.SOCK_DGRAM)
     sock = _get_asynclib().Socket(raw_socket)
     try:
-        if sys.platform == 'win32':
-            sock.setsockopt(socket.SOL_SOCKET, socket.SO_EXCLUSIVEADDRUSE, 1)
-        else:
-            sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+        sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
 
         if interface is not None or port is not None:
             await sock.bind((interface or '', port or 0))
--- a/anyio/__init__.py
+++ b/anyio/__init__.py
@@ -514,13 +514,12 @@ async def create_udp_socket(
     :param port: port to bind to
     :param target_host: remote host to set as the default target
     :param target_port: port on the remote host to set as the default target
+    :param address_family: the new socket's address family. Required if
+        neither ``interface`` nor ``target_host`` are specified.
     :return: a UDP socket
 
     """
-    if interface:
-        interface, family, _v6only = await _networking.get_bind_address(interface)
-    else:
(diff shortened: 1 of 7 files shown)


agronholm:
At the minimum I would require a new test to cover this use case. And fix the mypy error.

anyio/__init__.py
```diff
@@ -501,7 +501,8 @@ async def create_unix_server(
 
 async def create_udp_socket(
     *, interface: Optional[IPAddressType] = None, port: Optional[int] = None,
-    target_host: Optional[IPAddressType] = None, target_port: Optional[int] = None
+    target_host: Optional[IPAddressType] = None, target_port: Optional[int] = None,
+    address_family: Optional[int] = socket.AF_UNSPEC
```
agronholm, line 505: What would the address family be for multicast?

anyio/__init__.py
```diff
@@ -532,6 +533,11 @@ async def create_udp_socket(
     raw_socket = socket.socket(family=family, type=socket.SOCK_DGRAM)
     sock = _get_asynclib().Socket(raw_socket)
     try:
+        if sys.platform == 'win32':
```
agronholm, line 536: These can already be set after creation. Why should we change the default behavior?

anyio/__init__.py
agronholm: Never mind the previous question, but shouldn't the default be something more sensible, like `AF_INET`?

agronholm: Also, wouldn't you specify a target host/port when creating a socket for multicast?

smurfix: Well, there's AF_INET6 …
You can't specify a target because `connect()`ing to a UDB socket means that you only accept messages from that address. Multicast packets however cannot have a multicast source.

anyio/__init__.py
smurfix: Because they need to be set before calling `bind()`. Some OSes don't accept calling that twice.

agronholm: But I thought you said in the beginning that you can't `bind()` to an actual address? What am I missing here?

smurfix:
I have added a test that verifies that SO_REUSEADDR works. Actually testing multicast is not something I'd advocate because that frequently doesn't work in test environments.

smurfix:
mypy still fails but that's not my fault. :-/

anyio/__init__.py
smurfix: You can't bind to an (IPv4/IPv6) address, but you have to bind to the actual (UDP) port the multicast packets are addressed to. Otherwise you don't get them.

agronholm:
That mypy failure is strange. The asyncio backend has not been changed since the last successful run.

agronholm:
Found and fixed the problem. It was a missing parameter for the return value.

agronholm:
Meh, now that mypy fix caused tests to start failing. Maybe I should just add an ignore there.

agronholm:
Ok, now we're seeing some real failures, on both macos and windows.

anyio/__init__.py
```diff
@@ -532,6 +533,8 @@ async def create_udp_socket(
     raw_socket = socket.socket(family=family, type=socket.SOCK_DGRAM)
     sock = _get_asynclib().Socket(raw_socket)
     try:
+        sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
```
agronholm, line 536: This has to be optional, and off by default.

anyio/__init__.py
agronholm, line 505: Add this parameter to the docstring please. Emphasize it's only necessary if neither ìnterface` or `target_host` have been specified.

agronholm:
I needed to get the next release out the door so I moved this along on my own. I've credited you for the test code.

PR #116, opened 2020-06-14
Harmonize default task names across backends

agronholm:
This ensures that, in absence of explicitly given task names, the default task names match across the different backends.


agronholm:
Thanks for the review!

PR #117, opened 2020-06-14
Fixed fail_after() getting stuck on curio

agronholm:
The curio backend got stuck on fail_after() because the cancellation was not being delivered.


agronholm:
The original reported problem appears to have been fixed by this, as verified by practical testing.

PR #121, opened 2020-07-03
Added getaddrinfo() and getnameinfo() to the API

agronholm:
This also makes sure that we use IDNA 2008 for all DNS lookups which we did not do on all backends.

### DIFF
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -6,6 +6,7 @@ import socket
 import sys
 from collections import OrderedDict
 from functools import partial
+from socket import AddressFamily, SocketKind
 from threading import Thread
 from types import TracebackType
 from typing import (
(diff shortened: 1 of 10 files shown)


src/anyio/_backends/_asyncio.py
```diff
@@ -594,6 +597,29 @@ def _run_in_thread(self, func: Callable, *args):
         return run_in_thread(func, *args)
 
 
+async def getaddrinfo(host: str, port: int, *, family: Union[int, AddressFamily] = 0,
```
njsmith, line 600: `socket.getaddrinfo` accepts both `str` and `bytes`; do you want to match that?

agronholm: Yeah – the typeshed signature for that is `Union[bytearray, bytes, Text]` so I should probably add those too. Good catch!

PR #122, opened 2020-07-08
Refactored the abc module and added more stream classes

agronholm:
This paves way to a more trio-like stream hierarchy.


agronholm:
It's been 6 days and nobody has reviewed this. I need to move forward so I'm merging this.

PR #127, opened 2020-07-16
Fix Condition.wait on asyncio and curio

mjwestcott:
Fixes a bug introduced in https://github.com/agronholm/anyio/commit/f6987c116a86abbd66451ab44dcb239b683a57c9

`Condition` now inherits from an abstract base class (https://github.com/agronholm/anyio/commit/f6987c116a86abbd66451ab44dcb239b683a57c9#diff-d80f406d80e69d8ad05e4b5fbf988915R614), so the `super()` implementation is a no-op.

The new `test_wait_cancel` fails on master for `asyncio` and `curio`, and passes in this PR.

### DIFF
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -640,7 +640,7 @@ class Condition(abc.Condition):
 
     async def wait(self):
         await check_cancelled()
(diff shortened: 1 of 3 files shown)


agronholm:
Thanks a lot!

Issue #143, opened 2020-08-05
Object stream send is blocked after receive_nowait


mjwestcott:
This issue concerns `master@efa6071`

It appears that memory object stream sends may become blocked even if there is space in the buffer.

Here's a failing scenario:

```python
import anyio


async def main():
    send, receive = anyio.create_memory_object_stream(1)
    await send.send_nowait('hello')
    async with anyio.create_task_group() as tg:
        await tg.spawn(send.send, 'anyio')
        await anyio.wait_all_tasks_blocked()
        assert await receive.receive_nowait() == 'hello'
        await anyio.wait_all_tasks_blocked()
        assert len(receive._state.buffer) == 1


anyio.run(main)
```

This is unlike Trio, which passes the following:

```python
import trio


async def main():
    send, receive = trio.open_memory_channel(1)
    send.send_nowait('hello')
    async with trio.open_nursery() as n:
        n.start_soon(send.send, 'trio')
        await trio.sleep(0)
        assert receive.receive_nowait() == 'hello'
        await trio.sleep(0)
        assert receive.statistics().current_buffer_used == 1


trio.run(main)
```

mjwestcott:
Potential fix: https://github.com/agronholm/anyio/pull/144

PR #144, opened 2020-08-05
Fix object stream send blocked after receive_nowait

mjwestcott:
Intended to fix https://github.com/agronholm/anyio/issues/143

### DIFF
--- a/src/anyio/streams/memory.py
+++ b/src/anyio/streams/memory.py
@@ -43,15 +43,16 @@ class MemoryObjectReceiveStream(Generic[T_Item], ObjectReceiveStream[T_Item]):
         if self._closed:
             raise ClosedResourceError
 
+        if self._state.waiting_senders:
+            # Get the item from the next sender
+            send_event, item = self._state.waiting_senders.popitem(last=False)
+            await send_event.set()
+            self._state.buffer.append(item)
+
         if self._state.buffer:
             return self._state.buffer.popleft()
         elif not self._state.open_send_channels:
             raise EndOfStream
-        elif self._state.waiting_senders:
-            # Get the item from the next sender
-            send_event, item = self._state.waiting_senders.popitem(last=False)
-            await send_event.set()
-            return item
 
         raise WouldBlock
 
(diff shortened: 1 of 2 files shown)


src/anyio/streams/memory.py
```diff
@@ -43,15 +43,16 @@ async def receive_nowait(self) -> T_Item:
         if self._closed:
             raise ClosedResourceError
 
+        if self._state.waiting_senders:
+            # Get the item from the next sender
+            send_event, item = self._state.waiting_senders.popitem(last=False)
+            await send_event.set()
+            self._state.buffer.append(item)
```
agronholm, line 50: I would feel a lot safer if this swapped places with `send_event.set()`.

tests/streams/test_memory.py
```diff
@@ -56,6 +56,17 @@ async def test_send_then_receive_nowait():
...
 
+async def test_send_is_unblocked_after_receive_nowait():
+    send, receive = create_memory_object_stream(1)
+    await send.send_nowait('hello')
+    async with create_task_group() as tg:
+        await tg.spawn(send.send, 'anyio')
+        await wait_all_tasks_blocked()
+        assert await receive.receive_nowait() == 'hello'
+        await wait_all_tasks_blocked()
+        assert len(receive._state.buffer) == 1
```
agronholm, line 67: Rather than checking for the internal state of the buffer, I would prefer to have the task group surrounded with a `fail_after(1)` and then `assert receive.receive_nowait() == 'anyio'` outside of it.

mjwestcott: I believe that would pass in the current (broken) master, since the second `receive_nowait` works just fine. I think we need to test that the `send` is unblocked after the first `receive_nowait`.

I understand if you don't want to use internal state. Any other ideas?

mjwestcott: Oh, I think I was wrong, `fail_after` would will raise in master. I'll implement it.

agronholm:
Thanks! Nice to get this fixed before the release.

mjwestcott:
Great, thanks for the quick response. Looking forward to a release on PyPI (even just an alpha/rc1)

PR #156, opened 2020-09-05
Replaced individual stream/listener attributes with typed attributes

### LINKED ISSUE #133: Should we implement an asyncio-like get_extra_info() method on streams?
We have stream wrappers such as `TLSStream` and `BufferedByteReceiveStream` which expose a different set of attributes than the streams they wrap. I imagine it will be quite inconvenient for developers that just want the remote address (for example) to manually traverse the chain of wrapped transports to find the information they want. Asyncio has a pretty convenient solution to this: [get_extra_info()](https://docs.python.org/3/library/asyncio-protocol.html?#asyncio.BaseTransport.get_extra_info). I think it would be really useful provide a similar method so developers don't have to care how many layers down the information resides.

agronholm:
This is similar to asyncio's get_extra_info() (but more sophisticated).
Fixes #133.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -17,6 +17,13 @@ Asynchronous resources
 
 .. autoclass:: anyio.abc.AsyncResource
 
+Typed attributes
+----------------
+
+.. autofunction:: anyio.typed_attribute
+.. autoclass:: anyio.TypedAttributeSet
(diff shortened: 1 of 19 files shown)


smurfix (approved):

agronholm:
Thanks for the review!

Issue #167, opened 2020-10-18
Add support for wrappers of callback-based APIs


frederikaalund:
First off, thanks for writing and maintaining anyio. It's a useful and well-written library.

I'm in the process of writing an anyio wrapper around paho-mqtt. Here is an example from the [paho-mqtt](https://github.com/eclipse/paho.mqtt.python) readme:

```python
# The callback for when a PUBLISH message is received from the server.
def on_message(client, userdata, msg):
    print(msg.topic+" "+str(msg.payload))

client = mqtt.Client()
client.on_message = on_message
```

To bridge the gap between the old-school callbacks and the async domain, I use `create_memory_object_stream`:

```python
send_stream, receive_stream = create_memory_object_stream()

def _send_to_stream(client, userdata, msg):
    try:
        send_stream.send_nowait(msg)  # This is an async function so I can't call it in this context
    except WouldBlock:
        ...

client.on_message = _send_to_stream
```

This approach doesn't work, however, since [`send_nowait` is an async function](https://github.com/agronholm/anyio/blob/0455cd9b2b0e08ce916476e22e4f198c45a15f83/src/anyio/streams/memory.py#L121). From a cursory look, I get the impression that curio support is the culprit. Because `Event.set` is async in curio, `send_nowait` is also async.

Is there some way around this? Can I somehow specify that I don't need curio support and get the non-async version of `Event.set` and, in turn, the non-async version of `send_nowait`?

If there currently isn't such functionality, would you be interested in a pull request that adds this?

Many other old-school APIs are callback-based. [RxPy](https://github.com/ReactiveX/RxPY) is another example. I'd also like to write anyio helpers for RxPy to bridge the gap to the async domain. Currently, my efforts are blocked by the above-mentioned restriction.

This issue relates to #99.

agronholm:
Are you suggesting that I add an API to AnyIO that only works on some of the backends?

frederikaalund:
Thanks for the quick reply.

> Are you suggesting that I add an API to AnyIO that only works on some of the backends?

Not if there is a way around it.

I recently read about [`curio.UniversalEvent`](https://github.com/dabeaz/curio/blob/dc11ffefb2101b1b37be4c2183b5b3beac604fca/curio/sync.py#L56). This class has an non-async version of `Event.set`. Can we replace `Event` with `UniversalEvent` in the curio backend? In turn, this opens up for non-async versions of `send_nowait`.

smurfix:
[...]

PR #168, opened 2020-10-21
Add primitives for sync/async coordination

frederikaalund:
This pull request comes from the discussion in #167. Let's take this a couple of steps at a time.

So far, I added an `anyio.create_universal_event` function. Said function returns a `UniversalEvent`. A `UniversalEvent` is like a regular `Event` except that it's `set` method is sync:

```python
def set(self) -> None:
    """Set the flag, notifying all listeners."""
```

Behind the scenes, `UniversalEvent` maps as follows:
 * asyncio: `anyio.UniversalEvent` --> `asyncio.Event`
 * trio: `anyio.UniversalEvent` --> `trio.Event`
 * curio: `anyio.UniversalEvent` --> `curio.UniversalEvent`

If you don't like the "universal" term, I'm open to suggestions. I based it on curio's terminology.

I plan to add more commits to this pull request when we agree on the API design.

Let me know what you think so far.

### DIFF
--- a/src/anyio/_backends/_trio.py
+++ b/src/anyio/_backends/_trio.py
@@ -500,18 +500,25 @@ class Lock(abc.Lock):
         self._lock.release()
 
 
-class Event:
+class BaseEvent(abc.BaseEvent):
     def __init__(self):
         self._event = trio.Event()
-
-    async def set(self) -> None:
-        self._event.set()
 
     def is_set(self) -> bool:
         return self._event.is_set()
 
     async def wait(self):
         await self._event.wait()
+
+
+class Event(abc.Event, BaseEvent):
+    async def set(self) -> None:
+        self._event.set()
+
+
+class UniversalEvent(abc.UniversalEvent, BaseEvent):
+    def set(self) -> None:
+        self._event.set()
 
 
 class Condition(abc.Condition):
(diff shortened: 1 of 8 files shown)


src/anyio/_backends/_trio.py
```diff
@@ -500,20 +500,27 @@ async def release(self) -> None:
...
     def is_set(self) -> bool:
         return self._event.is_set()
 
     async def wait(self):
         await self._event.wait()
 
 
+class Event(abc.Event, BaseEvent):
+    async def set(self) -> None:
+        self._event.set()
```
belm0, line 516: A rule of trio api's is that async function should checpoint on all code paths, so add sleep(0). Original code had the same problem

frederikaalund: Good point. :+1: I guess that an `trio.lowlevel.checkpoint()` is even more descriptive than a raw `sleep(0)`.

frederikaalund:
I prefer the breaking change suggested in #179 over this pull request. :+1: When I made this pull request, I thought that you were opposed to such breaking changes at this point in time. Guess I was wrong (or something changed in the past couple of months). :)

A synchronous `Event.set` and `MemoryObjectSendStream.send_nowait` solves the original issue #167.

agronholm:
I'm turning this down in favor of making `Event.set()` synchronous (while returning an awaitable for compatibility). The only question now is whether I can do this in the 2.x series or if I have to bump it to 3.0. Thanks for the effort though!

PR #181, opened 2020-12-30
Changed several coroutine methods to synchronous methods

### LINKED ISSUE #179: synchronous Event set
I'd like to use anyio for compatibility across asyncio and Trio, but async `Event.set()` (due to Curio) is a non-starter.  Event is a widely use primitive.  I'm guessing that an async `set()` would double the number of functions needing to be async in my program, and hence greatly increase the race conditions I'd need to ponder.  It's advised in "The Function Colour Myth" to keep async function percentage to a minimum.

It's not like I'm explicitly opting to not support Curio either, because my intended use case depends on Quart, which itself only supports asyncio and Trio.

It seems like there could just be a `synchronous_set()` method in the ABC, and Curio could raise NotImplementedError.  Isn't there already precedent with `FileStreamWrapper.send_eof()`?.

agronholm:
Fixes #179.


graingert:
I think it would be nicer to leave the `async` versions and make sync versions with `_soon` suffixes

agronholm:
That would defeat part of the purpose of this PR because it would no longer align with Trio's API. Plus, the async versions are not really necessary.

agronholm:
Not to mention that the names would be misleading: `.release_soon()` for instance would release a lock instantly, not "soon".

agronholm:
Before the release I'll ponder how, and if backwards compatibility should be handled.

agronholm:
I've decided to add a couple functions that will let libraries and applications work with both v2 and v3. These compatibility functions will be added to both 2.x and 3.x release branches.

PR #182, opened 2020-12-30
Dropped Curio support

agronholm:
Curio has become a liability due to its policy of only allowing coroutine functions to interact with the curio kernel.
Dropping support will allow AnyIO to better match the Trio APIs and to allow more functionality for apps that still rely on synchronous callbacks due to existing asyncio libraries.

See #179 for discussion.


belm0:
Perhaps it's worth trying to say something objective about Curio in the readme or docs.

## Why not Curio?

Curio is another async framework which early versions of anyio tried to support along side of asyncio and Trio.  The Curio design is quite strict about every API call being async (`Event.set()`, `get_current_task()`, etc.), which hinders bridging to synchronous code and increases the general complexity of concurrent programs.  Lacking strong demand for Curio support, anyio elected to forgo it and allow the advantages of the partly-synchronous aspects of asyncio and Trio to be available in the API.

thedrow:
I completely agree with this decision.

docs/faq.rst
```diff
@@ -0,0 +1,26 @@
...
+==========================
+
+Why is Curio not supported as a backend?
+----------------------------------------
+
+Curio_ was supported in AnyIO before v3.0. Support for it was dropped for two reasons:
+
+#. Its interface allowed only coroutine functions to access the Curio_ kernel. This forced AnyIO to
+   follow suit in its own API design, making it difficult to adapt existing applications that
+   relied on synchronous callbacks to us AnyIO. It also interfered with the goal of matching Trio's
```
smurfix, line 11: to *use*

agronholm: Good catch. My "e" key has been working a bit erratically as of late.

PR #187, opened 2021-01-28
trio / pytest: Don't kill the test loop on exceptions

smurfix:
"Normal" Exceptions must not propagate to Trio.
    

Otherwise there's a race condition:
* main task [starts Trio and] queues an async call

  * Trio handles the call and steps through the call, or rather tells the main task to do so

* one of the steps in question raises an exception
* the exception is queued for the main task to process,
  but also re-raised and thus propagated to Trio

  * the Trio task starts shutting itself down

* main task reads the exception from the queue
* main task processes the exception and starts calling finalizers
* one of the finalizers calls into Trio (_nursery is still set)
* main task waits on the _call_queue

  * Trio calls the finalizer which clears _nursery
  * the Trio thread ends

* main task hangs indefinitely.

### DIFF
--- a/src/anyio/_backends/_trio.py
+++ b/src/anyio/_backends/_trio.py
@@ -674,6 +674,8 @@ class TestRunner(abc.TestRunner):
     async def _call_func(self, func, args, kwargs):
         try:
             retval = await func(*args, **kwargs)
+        except Exception as exc:
+            self._result_queue.append(Error(exc))
         except BaseException as exc:
             self._result_queue.append(Error(exc))
             raise


smurfix:
Bah. Please ignore the jumble, that was the wrong branch. I just force-pushed the correct one (92dc787)

agronholm:
If this causes problems, can you demonstrate?

smurfix:
It's part of the `aevent` test suite. I monkeypatched an appropriate replacement in and presto, no more hangs. ;-)
I didn't try to engineer a testcase that hits the race reliably.

agronholm:
I have not had an issue myself, and nobody else has complained either, so you need to convince me that this is a problem that needs to be fixed.

smurfix:
The reason for nobody having found anything is probably that it's in the test code and people aren't in the habit of crashing their test cases.

How to reproduce:

* Checkout `git@github.com:M-o-a-T/aevent.git`
* switch to "hang" branch
* `PYTHONPATH=src python3 -mpytest -sxv tests/test_basic.py`

The test hangs. Install my fix and it "works" (i.e. reports the error).

agronholm:
I wonder what's with coveralls now. It used to work just fine. The tests themselves appear to work fine.

agronholm:
There is no `test_basic.py` anywhere, and `test_time.py` does not reproduce the issue.

smurfix:
Meh. As usual I forgot to add the file. :-/

```
#
# Simple test that error handling works.
#

import pytest
import os

import anyio
import aevent

async def sleeper(t):
    time.sleep(t)

class TestMe:

    def setup(self):
        print("A")

    @pytest.mark.aevent
    def test_base(self):
        print("B")
        # testing this within Python is TODO via subprocess
        if os.environ.get("AEVENT_CRASHTEST",0):
            raise RuntimeError
    
    def teardown(self):
        print("C")

```

agronholm:
I'm running this but the tests do not hang.

smurfix:
Hmm, my instructions seem to be somewhat incomplete. Sorry about that. Multitasking is a bitch.

Missing:

* `python3 setup.py bdist_egg` (or anything else that teaches pytest about `aevent.pytest_plugin`)
* `export AEVENT_CRASHTEST=1`

If that doesn't do it I don't know how else to reproduce the problem.

agronholm:
I can reproduce the hang now. Thanks for the instructions.

agronholm:
I see, however, that `test_basic.py` does not use anyio's pytest plugin, but an altered derivative. Is there a way to reproduce this with anyio's own pytest plugin?

smurfix:
Not easily. The hang occurs because after the main test crashes, the pytest plugin immediately queues the teardown call. Doing the same sequence with another testcase doesn't work because the Trio task has too much time to shut down properly.

I tried to do it with an async fixture but that also failed. Would have to add some interesting cross-thread debug code to help the processes hit the race condition.

Anyway, this PR's change is (a) sufficiently obviously-correct (IM decidedly HO, and for some value of "obvious") and fixes the problem.

agronholm:
Alright, it seems harmless so I'll merge.

agronholm:
Thanks!

PR #188, opened 2021-01-28
Add network mark to getaddrinfo test

D4N:
Calling getaddrinfo in a builder without network access results in an error as
it cannot contact a DNS server. The network marker is used by convention to mark
such tests and allows this test to be skipped.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,6 +23,9 @@ max_line_length = 99
 addopts = "-rsx --tb=short"
 testpaths = ["tests"]
 filterwarnings = "always"
+markers = [
+    "network",
+]
 
 [tool.coverage.run]
 source = ["anyio"]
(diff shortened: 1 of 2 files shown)


agronholm:
Thanks.

agronholm:
Where is this mark declared by the way? Pytest raises a warning about this because it's not declared anywhere.

pyproject.toml
```diff
@@ -23,6 +23,9 @@ max_line_length = 99
 addopts = "-rsx --tb=short"
 testpaths = ["tests"]
 filterwarnings = "always"
+markers = [
+    "network",
```
D4N, line 27: @agronholm The marker should be declared here.

agronholm: Right; I'm not sure why I got this warning the other time. Not seeing it now. Thanks.

PR #193, opened 2021-02-10
Added the TaskGroup.start() method

### LINKED ISSUE #192: Add AnyIO equivalent of "await nursery.start()"
Trio has an alternative method of spawning tasks which requires the target coroutine function to take a specific keyword argument and use it to signal that the task has successfully initalized. This is ideal for inter-task coordination and starting background services.

So far AnyIO has resisted adding this feature because one can emulate this behavior:
```python3
async def service(*, start_event):
   ...  # initialize
   await start_event.set()
   ...  # do something else

    start_event = create_event()
    await tg.spawn(partial(service, start_event=start_event))
    await start_event.wait()
```
There are at least two problems with this, however:

1. Having to use `partial()` and having to explicitly create the event and wait on it is unnecessary clutter
2. It doesn't allow the task to pass a value to the caller (a typical use case would be passing a port number for a service that has been started)

Implementing this should be straightforward.

agronholm:
Closes #192.

### DIFF
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -378,6 +378,16 @@ class ExceptionGroup(BaseExceptionGroup):
         self.exceptions = exceptions
 
 
+class TaskStatus:
+    __slots__ = '_future'
+
+    def __init__(self, future: asyncio.Future):
+        self._future = future
+
+    def started(self, value=None) -> None:
+        self._future.set_result(value)
+
+
 class TaskGroup(abc.TaskGroup):
     __slots__ = 'cancel_scope', '_active', '_exceptions'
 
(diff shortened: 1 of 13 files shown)


tests/test_taskgroups.py
```diff
@@ -80,6 +80,59 @@ async def test_spawn_after_error():
...
+    exc.match('foo')
+    assert value == 2
+
+
+async def test_start_no_started_call():
+    async def taskfunc(*, task_status):
+        pass
+
+    async with create_task_group() as tg:
+        with pytest.raises(RuntimeError) as exc:
```
graingert, line 130:
```suggestion
        with pytest.raises(RuntimeError, match="Child exited")`

Comment (agronholm) [reply tests/test_taskgroups.py:130]: Trio starts the exception message with a "c" which is why I left out the first character. I'll probably do a mass refactoring later where I start using `match=...` in `pytest.raises()`.

File: src/anyio/_backends/_asyncio.py [hunk 2]
```diff
@@ -441,32 +449,73 @@ def _filter_cancellation_errors(exceptions: Sequence[BaseException]) -> List[Bas
...
+
+    def spawn(self, func: Callable[..., Coroutine], *args, name=None) -> None:
+        self._spawn(func, args, name)
+
+    async def start(self, func: Callable[..., Coroutine], *args, name=None) -> None:
+        future: asyncio.Future = asyncio.Future()
+        task = self._spawn(func, args, name, future)
+
+        # If the task raises an exception after sending a start value without a switch point
+        # between, the exception group is cancelled and this method never proceeds to process the
```
graingert, line 510:
```suggestion
        # between, the task group is cancelled and this method never proceeds to process the
```

graingert (approved):

Issue #195, opened 2021-02-11
should anyio.abc.BlockingPortal.__exit__ call anyio.abc.BlockingPortal.stop_from_external_thread ?


graingert:

graingert:
Fixed in #196

PR #196, opened 2021-02-11
Made start_blocking_portal() a context manager

agronholm:
* Ensures that the thread is joined when exiting
* Eliminates a potentially confusing regular context manager from BlockingPortal

### DIFF
--- a/src/anyio/_core/_threads.py
+++ b/src/anyio/_core/_threads.py
@@ -1,5 +1,7 @@
 import threading
-from typing import Any, Callable, Coroutine, Dict, Optional, TypeVar, cast
+from concurrent.futures import Future
+from contextlib import contextmanager
+from typing import Any, Callable, Coroutine, Dict, Generator, Optional, TypeVar
 
 from ..abc import BlockingPortal, CapacityLimiter
 from ._eventloop import get_asynclib, run, threadlocals
@@ -83,9 +85,10 @@ def create_blocking_portal() -> BlockingPortal:
     return get_asynclib().BlockingPortal()
 
 
+@contextmanager
 def start_blocking_portal(
         backend: str = 'asyncio',
-        backend_options: Optional[Dict[str, Any]] = None) -> BlockingPortal:
+        backend_options: Optional[Dict[str, Any]] = None) -> Generator[BlockingPortal, Any, None]:
     """
     Start a new event loop in a new thread and run a blocking portal in its main task.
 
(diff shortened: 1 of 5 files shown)


graingert:
Fixes https://github.com/agronholm/anyio/issues/195

src/anyio/_core/_threads.py
```diff
@@ -83,29 +85,36 @@ def create_blocking_portal() -> BlockingPortal:
...
-    portal: Optional[BlockingPortal]
-    event = threading.Event()
+    future: Future[BlockingPortal] = Future()
     kwargs = {'func': run_portal, 'backend': backend, 'backend_options': backend_options}
-    thread = threading.Thread(target=run, kwargs=kwargs)
+    thread = threading.Thread(target=run, kwargs=kwargs, daemon=True)
     thread.start()
-    event.wait()
-    return cast(BlockingPortal, portal)
+    portal = future.result()
```
graingert, line 111: calling `start_blocking_portal(backend="twisted")` hangs forever

graingert: I think you want a wrapper to deliver exceptions: https://gist.github.com/graingert/d730e3a4c368bcddfdeee836622ba376#file-start_blocking_portal-py-L29

src/anyio/_core/_threads.py
```diff
@@ -83,29 +85,36 @@ def create_blocking_portal() -> BlockingPortal:
...
+            future.set_result(portal_)
+            await portal_.sleep_until_stopped()
 
-    portal: Optional[BlockingPortal]
-    event = threading.Event()
+    future: Future[BlockingPortal] = Future()
     kwargs = {'func': run_portal, 'backend': backend, 'backend_options': backend_options}
-    thread = threading.Thread(target=run, kwargs=kwargs)
+    thread = threading.Thread(target=run, kwargs=kwargs, daemon=True)
     thread.start()
```
graingert, line 110: this thread is not cleaned up if SIGINT occurs during `portal = future.result()`

src/anyio/_core/_threads.py
```diff
@@ -83,29 +85,36 @@ def create_blocking_portal() -> BlockingPortal:
...
+        async with create_blocking_portal() as portal_:
+            future.set_result(portal_)
+            await portal_.sleep_until_stopped()
 
-    portal: Optional[BlockingPortal]
-    event = threading.Event()
+    future: Future[BlockingPortal] = Future()
     kwargs = {'func': run_portal, 'backend': backend, 'backend_options': backend_options}
-    thread = threading.Thread(target=run, kwargs=kwargs)
+    thread = threading.Thread(target=run, kwargs=kwargs, daemon=True)
```
graingert, line 109: this won't work in a sub-interpreter, and it looks like there's movement to remove the feature from the language

as @vstinner says: https://bugs.python.org/issue37266
> Daemon threads must die. That's a first step towards their death!


Per https://bugs.python.org/issue39812 using a `with concurrent.futures.ThreadPoolExecutor(max_workers=1) as pool:` should create threads that are scheduled to be cleaned up correctly

src/anyio/_core/_threads.py
agronholm: There is a test for this now.

src/anyio/_core/_threads.py
agronholm: I've switched to `ThreadPoolExecutor`, so this should not be an issue anymore.

graingert (approved):

graingert:
description is out of date `The thread is now started with daemon=True` this is true only on some python versions

PR #203, opened 2021-02-21
cancel native future tasks

### LINKED ISSUE #199: tasks that await a native asyncio.Future are not cancelled
```python
import asyncio
import anyio

async def amain():
    async def foo():
        await loop.create_future()

        await tg.spawn(foo)

anyio.run(amain)  # never completes
```

graingert:
fixes #199

### DIFF
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -10,8 +10,7 @@ from contextlib import contextmanager
 from dataclasses import dataclass
 from functools import partial, wraps
 from inspect import (
-    CORO_RUNNING, CORO_SUSPENDED, GEN_RUNNING, GEN_SUSPENDED, getcoroutinestate, getgeneratorstate,
-    isgenerator)
+    CORO_RUNNING, CORO_SUSPENDED, GEN_RUNNING, GEN_SUSPENDED, getcoroutinestate, getgeneratorstate)
 from io import IOBase
 from socket import AddressFamily, SocketKind, SocketType
 from threading import Thread
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -1243,6 +1243,7 @@ async def open_signal_receiver(*signals: int):
     queue = asyncio.Queue()  # type: asyncio.Queue[int]
     handled_signals = set()
     agen = process_signal_queue()
+    await checkpoint()
     try:
         for sig in set(signals):
             loop.add_signal_handler(sig, queue.put_nowait, sig)
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -8,7 +8,9 @@ from concurrent.futures import Future
 from contextlib import contextmanager
 from dataclasses import dataclass
 from functools import partial, wraps
-from inspect import isgenerator
+from inspect import (
+    CORO_RUNNING, CORO_SUSPENDED, GEN_RUNNING, GEN_SUSPENDED, getcoroutinestate, getgeneratorstate,
+    isgenerator)
 from socket import AddressFamily, SocketKind, SocketType
 from threading import Thread
 from types import TracebackType
@@ -26,6 +28,12 @@ from .._core._exceptions import WouldBlock
 from .._core._sockets import GetAddrInfoReturnType, convert_ipv6_sockaddr
 from .._core._synchronization import ResourceGuard
 from ..abc import IPSockAddrType, UDPPacketType
+
+if sys.version_info >= (3, 8):
+    get_coro = asyncio.Task.get_coro
+else:
+    def get_coro(task: asyncio.Task) -> Union[Coroutine, Generator]:
+        return task._coro
 
 if sys.version_info >= (3, 7):
     from asyncio import all_tasks, create_task, current_task, get_running_loop
@@ -121,6 +129,15 @@ def get_callable_name(func: Callable) -> str:
 # Event loop
 #
 
+def _task_started(task: asyncio.Task) -> bool:
+    """Return ``True`` if the task has been started and has not finished."""
+    coro = get_coro(task)
+    try:
+        return getcoroutinestate(coro) in (CORO_RUNNING, CORO_SUSPENDED)
+    except AttributeError:
+        return getgeneratorstate(coro) in (GEN_RUNNING, GEN_SUSPENDED)
+
+
 def _maybe_set_event_loop_policy(policy: Optional[asyncio.AbstractEventLoopPolicy],
                                  use_uvloop: bool) -> None:
     # On CPython, use uvloop when possible if no other policy has been given and if not
@@ -215,6 +232,7 @@ class CancelScope(abc.CancelScope):
         self._active = False
         if self._timeout_handle:
             self._timeout_handle.cancel()
+            self._timeout_handle = None
 
         assert self._host_task is not None
         self._tasks.remove(self._host_task)
@@ -231,9 +249,6 @@ class CancelScope(abc.CancelScope):
                     # This scope was directly cancelled
                     return True
 
-        if self._shield and self._parent_cancelled():
-            _start_cancelling(self._host_task)
-
         return None
 
     def _timeout(self):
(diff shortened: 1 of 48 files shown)


src/anyio/_backends/_asyncio.py
```diff
@@ -314,20 +321,23 @@ def shield(self) -> bool:
...
+            return True
+        if cancel_scope.shield:
+            return False
+        cancel_scope = cancel_scope._parent_scope
 
+
+async def checkpoint():
+    if _cancel_called(current_task()):
+        raise CancelledError
     await asyncio.sleep(0)
```
agronholm, line 341: An empty line after each control block.
```suggestion
        raise CancelledError

    await asyncio.sleep(0)
```

src/anyio/_backends/_asyncio.py
```diff
@@ -314,20 +321,23 @@ def shield(self) -> bool:
         return self._shield
 
 
-async def checkpoint():
+def _cancel_called(task):
```
agronholm, line 324: Needs type annotations.

src/anyio/_backends/_asyncio.py
```diff
@@ -314,20 +321,23 @@ def shield(self) -> bool:
...
         if cancel_scope.cancel_called:
-            raise CancelledError
-        elif cancel_scope.shield:
-            break
-        else:
-            cancel_scope = cancel_scope._parent_scope
+            return True
+        if cancel_scope.shield:
+            return False
+        cancel_scope = cancel_scope._parent_scope
```
agronholm, line 335: Empty lines after control blocks.
```suggestion

        if cancel_scope.shield:
            return False

        cancel_scope = cancel_scope._parent_scope
```

agronholm (changes requested):
Some fixes are needed.

src/anyio/_backends/_asyncio.py
```diff
@@ -314,19 +321,25 @@ def shield(self) -> bool:
...
-            break
-        else:
-            cancel_scope = cancel_scope._parent_scope
+            return True
+
+        if cancel_scope.shield:
+            return False
+
+        cancel_scope = cancel_scope._parent_scope
+
```
agronholm, line 338: The return statement is missing here (what happens if `cancel_scope` is falsy and the `while` loop ends?).

src/anyio/_backends/_asyncio.py
```diff
@@ -485,6 +493,8 @@ async def spawn(self, func: Callable[..., Coroutine], *args, name=None) -> None:
         _task_states[task] = TaskState(parent_id=id(current_task()), name=name,
                                        cancel_scope=self.cancel_scope)
         self.cancel_scope._tasks.add(task)
+        if self.cancel_scope._cancel_called:
+            create_task(_cancel_soon(task))
```
agronholm, line 497: Shouldn't `loop.call_soon()` be used here, and `_cancel_soon()` be made into a regular function?

src/anyio/_backends/_asyncio.py
```diff
@@ -216,8 +219,7 @@ async def timeout():
 
         if self._deadline != math.inf:
             if get_running_loop().time() >= self._deadline:
-                self._cancel_called = True
-                self._timeout_expired = True
+                self._timeout()
             else:
                 self._timeout_task = get_running_loop().create_task(timeout())
```
graingert, line 224:
```suggestion
                self._timeout_task = create_task(timeout())
```

tests/test_taskgroups.py
```diff
@@ -577,3 +577,26 @@ async def test_suppress_exception_context():
...
+@pytest.mark.parametrize('anyio_backend', ['asyncio'])
+async def test_cancel_native_future_tasks_cancel_scope():
+    async def wait_native_future():
+        async with anyio.open_cancel_scope():
+            loop = asyncio.get_event_loop()
+            await loop.create_future()
+
+    async with anyio.create_task_group() as tg:
+        await tg.spawn(wait_native_future)
+        await tg.cancel_scope.cancel()
```
graingert, line 602:
```suggestion
@pytest.mark.parametrize('anyio_backend', ['asyncio'])
async def test_cancel_native_future_tasks():
    async def wait_native_future():
        loop = asyncio.get_event_loop()
        await loop.create_future()

    async with anyio.create_task_group() as tg:
        tg.spawn(wait_native_future)
        tg.cancel_scope.cancel()


@pytest.mark.parametrize('anyio_backend', ['asyncio'])
async def test_cancel_native_future_tasks_cancel_scope():
    async def wait_native_future():
        with anyio.open_cancel_scope():
            loop = asyncio.get_event_loop()
            await loop.create_future()

    async with anyio.create_task_group() as tg:
        tg.spawn(wait_native_future)
        tg.cancel_scope.cancel()
```

src/anyio/_backends/_asyncio.py
```diff
@@ -190,11 +189,15 @@ def __init__(self, deadline: float = math.inf, shield: bool = False):
...
+        self._timeout_expired = True
+        self._cancel_called = True
+        self._cancel()
+
     def __enter__(self):
         async def timeout():
             await asyncio.sleep(self._deadline - get_running_loop().time())
-            self._timeout_expired = True
-            self.cancel()
+            self._timeout()
```
agronholm, line 200: Do we need this coroutine function here any more now that `cancel()` is a regular method?

src/anyio/_backends/_asyncio.py
```diff
@@ -190,11 +189,15 @@ def __init__(self, deadline: float = math.inf, shield: bool = False):
         self._host_task: Optional[asyncio.Task] = None
         self._timeout_expired = False
 
+    def _timeout(self):
+        self._timeout_expired = True
```
agronholm, line 193: I think this should check if the scope has already been cancelled. In such cases we don't want to set the `_timeout_expired` flag. In a later PR we should also check if the deadline has actually passed, since trio at least makes it possible to alter the deadline.

src/anyio/_backends/_asyncio.py
```diff
@@ -164,7 +164,6 @@ async def wrapper():
 #
 
 async def sleep(delay: float) -> None:
```
agronholm, line 166: I think this could simply be: sleep = asyncio.sleep

agronholm (changes requested):
Some questions, some change requests.

src/anyio/_backends/_asyncio.py
graingert: `self._timeout()` is a regular function

tests/test_signals.py
```diff
@@ -19,13 +24,26 @@ async def test_receive_signals():
...
     async with create_task_group() as tg:
         tg.spawn(signal_handler)
         tg.cancel_scope.cancel()
+
+
+async def test_task_group_cancellation_consume():
+    async def consume(sigiter):
+        async for v in sigiter:
+            pytest.fail()
+        pytest.fail()
```
graingert, line 44:
```suggestion
            pytest.fail("a SIGUSR1 should not be sent")
        pytest.fail("consume should have been cancelled")
```

tests/test_signals.py
```diff
@@ -19,13 +24,26 @@ async def test_receive_signals():
...
-@pytest.mark.skipif(sys.platform == 'win32', reason='Signal delivery cannot be tested on Windows')
-async def test_task_group_cancellation():
+async def test_task_group_cancellation_open():
     async def signal_handler():
         with open_signal_receiver(signal.SIGUSR1) as sigiter:
             async for v in sigiter:
-                assert False
+                pytest.fail()
+            pytest.fail()
+        pytest.fail()
```
graingert, line 33:
```suggestion
                pytest.fail("as SIGUSR1 should not be sent")
            pytest.fail("signal_handler should have been cancelled")
        pytest.fail("open_signal_receiver should not suppress cancellation")
```

src/anyio/_backends/_asyncio.py
agronholm: Yeah, so what? We'll just use `loop.call_later()`.

src/anyio/_backends/_asyncio.py
graingert: do you have a test case for that change?

agronholm: Probably not since the current code would not have done the right thing either. I'll take a look at the tests anyway.

src/anyio/_backends/_asyncio.py
graingert: seems to fail the tests: https://github.com/agronholm/anyio/pull/207/checks?check_run_id=1962019106#step:6:40

agronholm:
I'm going to first see what comes out of my original idea to solve this problem.

src/anyio/_backends/_asyncio.py
```diff
@@ -9,7 +9,9 @@
 from contextlib import contextmanager
 from dataclasses import dataclass
 from functools import partial, wraps
-from inspect import isgenerator
+from inspect import (
+    CORO_RUNNING, CORO_SUSPENDED, GEN_RUNNING, GEN_SUSPENDED, getcoroutinestate, getgeneratorstate,
+    isgenerator)
```
graingert, line 14:
```suggestion
    CORO_RUNNING, CORO_SUSPENDED, GEN_RUNNING, GEN_SUSPENDED, getcoroutinestate, getgeneratorstate)
```

src/anyio/_backends/_asyncio.py
```diff
@@ -1048,6 +1049,7 @@ async def accept(self) -> abc.SocketStream:
                     # Workaround for https://bugs.python.org/issue41317
                     try:
                         self._loop.remove_reader(self._raw_socket)
+                        print('removed reader')
```
graingert, line 1052:
```suggestion
```

src/anyio/_backends/_asyncio.py
```diff
@@ -1056,6 +1058,7 @@ async def accept(self) -> abc.SocketStream:
 
                     raise
                 finally:
+                    print('exiting accept()')
```
graingert, line 1061:
```suggestion
```

src/anyio/_backends/_asyncio.py
```diff
@@ -1068,6 +1071,13 @@ async def aclose(self) -> None:
 
         self._closed = True
         if self._accept_scope:
+            # Workaround for https://bugs.python.org/issue41317
+            try:
+                self._loop.remove_reader(self._raw_socket)
+                print('removed reader')
```
graingert, line 1076:
```suggestion
```

PR #204, opened 2021-02-22
Added backwards compatibility code and migration docs

agronholm:
These changes will be backported to the 2.x branch. They allow library authors to write code compatible with both major releases.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -146,6 +146,12 @@ Operating system signals
 
 .. autofunction:: anyio.open_signal_receiver
 
+Compatibility
+-------------
+
+.. autofunction:: anyio.maybe_async
+.. autofunction:: anyio.maybe_async_cm
+
 Testing and debugging
 ---------------------
 
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -22,6 +22,7 @@ The manual
    signals
    testing
    api
+   migration
    faq
    support
    contributing
--- a/docs/migration.rst
+++ b/docs/migration.rst
@@ -1,6 +1,8 @@
 Migrating from AnyIO 2 to AnyIO 3
 =================================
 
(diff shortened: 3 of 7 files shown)


smurfix:
Looks OK.

Now if there was a semantic patching tool for Python I could use to autoreplace all the calls that actually refer to anyio objects (and *only* those) I'd be even more happy. :-/

agronholm:
> Looks OK.
> 
> Now if there was a semantic patching tool for Python I could use to autoreplace all the calls that actually refer to anyio objects (and _only_ those) I'd be even more happy. :-/

Well, this is the best I can do. I've thought a lot about this and the alternatives are all inferior to this solution.

agronholm:
Thanks for the review!

smurfix:
> I've thought a lot about this and the alternatives are all inferior to this solution.

Depends on the metric you use for "inferior", but FWIW I tend to agree.

agronholm:
> > I've thought a lot about this and the alternatives are all inferior to this solution.
> 
> Depends on the metric you use for "inferior", but FWIW I tend to agree.

So far I've received just two alternative proposals:

#. Have the formerly-async functions return dummy awaitables. This falls apart when the functions return meaningful results, such as `current_time()` or `MemoryObjectStream.receive_nowait()`. It would end up some code working properly while other code would not.
#. Keep the async functions but add synchronous alternatives. While this would retain full backwards compatibility, it not only clutters the API significantly, but will also not achieve the second goal of this effort: to get closer to the trio API.

smurfix:
Well, the pieces which do return meaningful results are mostly-easily-greppable, so while that wouldn't be a 100% solution it'd get me personally most-if-not-all of the way there.

Still not ideal; I'm far from the last person to acknowledge that.

agronholm:
I'll think about it.

PR #208, opened 2021-02-23
Add missing 'await'

makkus:
Maybe I'm missing something obvious, but it seems to me the spawn should be awaited, no?

### DIFF
--- a/docs/tasks.rst
+++ b/docs/tasks.rst
@@ -30,7 +30,7 @@ Here's a demonstration::
     async def main():
         async with create_task_group() as tg:
             for num in range(5):
-                tg.spawn(sometask, num)
+                await tg.spawn(sometask, num)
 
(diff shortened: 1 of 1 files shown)


smurfix:
That's the master branch, so: no.

makkus:
Sorry, just realized this has changed in latest version. Apologies, readthedocs uses 'latest', which is why I was confused.

agronholm:
I'm confused. The docs on RTD default to "stable" which currently points to the v2.1.0 docs.

makkus:
Yes, that's correct. I arrived via a search engine, and that's probably why I somehow ended up on 'latest'. I know I didn't manually set the version, so I assume it was just some fluke that probably doesn't affect (m)any others (and nothing you could do about it anyway). Sorry again for the confusion.

PR #214, opened 2021-02-26
Forward all exceptions to the caller

smurfix:
Normal practice is to forward `Exception`s to the caller, but to handle
`BaseException`s yourself.

This does not apply here. Trio's guest mode is a special-purpose context
which must not terminate. Thus, we need to forward all exceptions to the
caller.

### DIFF
--- a/src/anyio/_backends/_trio.py
+++ b/src/anyio/_backends/_trio.py
@@ -497,11 +497,8 @@ class TestRunner(abc.TestRunner):
     async def _call_func(self, func, args, kwargs):
         try:
             retval = await func(*args, **kwargs)
-        except Exception as exc:
-            self._result_queue.append(Error(exc))
         except BaseException as exc:
             self._result_queue.append(Error(exc))
-            raise
         else:
             self._result_queue.append(Value(retval))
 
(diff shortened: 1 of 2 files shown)


agronholm:
This needs a test that would fail without this patch.

smurfix:
To be honest, this is partially a post-fact rationale.

My usecase is several tests hanging or failing with "not in async context" errors as I was getting `asyncdbus` and my `bleak` fork to run correctly, while dealing with DBus errors, `pytest.skip` calls, fixtures doing strange and interesting things, and whatnot.

The bottom line is that any other combination of handling (a subset of) `BaseException`s caused some of these tests to go splat.

I'd like to reproduce a testcase or two that illustrates this, but need to find the time for that somehow. :-/

agronholm:
Ok, but know that I won't merge this without a regression test.

smurfix:
I suspected as much. Will try to build a testcase for it, but it's not going to be easy. :-/

smurfix:
Found a corner case that demonstrates the problem well enough to justify the patch.

smurfix:
Meh. I should be more careful actually checking in the right pieces. Too much on my plate as usual.

agronholm:
I wanted the test in `test_pytest_plugin.py`. It was easier to commit it directly.

Thanks for the fix! I've credited you in the version history.

PR #218, opened 2021-02-27
Changed ExceptionGroup to match PEP 654

### LINKED ISSUE #17: Integrate with the exceptiongroup library
The current implementation of exception groups in anyio is quite poor. This has been left alone because of the emerging [exceptiongroup](https://github.com/python-trio/exceptiongroup) library which should provide a much better implementation. Once it's been released, we should add a dependency on it and remove our built-in implementation.

agronholm:
Closes #17.

### DIFF
--- a/src/anyio/_backends/_trio.py
+++ b/src/anyio/_backends/_trio.py
@@ -100,19 +100,19 @@ current_time = trio.current_time
 #
 
 class ExceptionGroup(BaseExceptionGroup, trio.MultiError):
-    def __new__(cls, message: str, errors: List[BaseException]):
-        return BaseExceptionGroup.__new__(cls, message, errors)
+    def __new__(cls, message: str, exceptions: List[BaseException]):
+        return BaseExceptionGroup.__new__(cls, message, exceptions)
 
-    def __init__(self, message: str, errors: List[BaseException]):
-        for i, exc in enumerate(errors):
+    def __init__(self, message: str, exceptions: List[BaseException]):
+        for i, exc in enumerate(exceptions):
             if isinstance(exc, trio.MultiError) and not isinstance(exc, ExceptionGroup):
-                excgroup = errors[i] = ExceptionGroup('multiple tasks failed', exc.exceptions)
+                excgroup = exceptions[i] = ExceptionGroup('multiple tasks failed', exc.exceptions)
                 excgroup.__cause__ = exc.__cause__
                 excgroup.__context__ = exc.__context__
                 excgroup.__traceback__ = exc.__traceback__
 
-        trio.MultiError.__init__(self, errors)
-        BaseExceptionGroup.__init__(self, message, errors)
+        trio.MultiError.__init__(self, exceptions)
+        BaseExceptionGroup.__init__(self, message, exceptions)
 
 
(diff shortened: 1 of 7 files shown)


src/anyio/_backends/_trio.py
```diff
@@ -64,7 +64,23 @@
...
+
+    def __init__(self, message: str, exceptions: List[BaseException]):
+        trio.MultiError.__init__(self, exceptions)
+        BaseExceptionGroup.__init__(self, message, exceptions)
+
+    @classmethod
+    def _get_exception_group(cls, exc: BaseException) -> Optional['BaseExceptionGroup']:
+        if isinstance(exc, trio.MultiError):
+            excgroup = ExceptionGroup('multiple tasks failed', exc.exceptions)
+            excgroup.__cause__ = excgroup.__cause__
```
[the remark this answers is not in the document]
agronholm: Oops! I've fixed this locally and will push once I fix another problem I found.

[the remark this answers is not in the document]
agronholm: Ok, fixed. I've also fixed both methods returning an empty group where they were supposed to return `None`.

agronholm:
I've fixed all the issues I've found. Do you want to continue reviewing this?

iritkatriel:
The PEP is not final yet, and may still change. Do you really want to change it now? Would that not break APIs for whoever is using this library?

agronholm:
> The PEP is not final yet, and may still change. Do you really want to change it now? Would that not break APIs for whoever is using this library?

This would go into the 3.0 release which has other breaking changes anyway. Of course the legitimate concern is that the API described in the PEP might change after AnyIO v3.0 final release.

src/anyio/_core/_exceptions.py
```diff
@@ -65,12 +66,20 @@ def __str__(self):
     def __repr__(self) -> str:
         return f'{self.__class__.__name__}({self.message!r}, {self.errors!r})'
 
+    @staticmethod
+    def check_direct_subclass(exc: BaseException, parents: Tuple[Type[BaseException]]) -> bool:
+        for cls in exc.__class__.mro()[:-1]:
```
graingert, line 71: https://docs.python.org/3/library/inspect.html#inspect.getmro ?

agronholm: Would that work differently here in some scenario?

graingert: https://docs.python.org/3/library/stdtypes.html?highlight=mro#class.mro yep, the mro method is for overriding the `__mro__` in a metaclass - not for getting the current `__mro__`

agronholm:
Lint is failing because Github Actions is checking against the wrong revision. It's complaining about line 1597 in `_backends/_asyncio.py` when the module is only 1365 lines long in revision 3d1001c which it's supposed to be checking. Instead, it checked out revision 2583a52725fcfd570ebe6d57bb6052e3f9f9447d which is not even any of the commits in this PR.

agronholm:
Superseded by use of the exceptiongroup backport and native (Base)ExceptionGroup, via 082169494be46f2f6d32db2c8950ba374504e048.

PR #220, opened 2021-02-27
Implemented the statistics() method for several synchronization classes

### LINKED ISSUE #216: Add the statistics() method to synchronization primitives
Trio's synchronization primitives allow one to observe their statistics:
* CapacityLimiter
* Condition
* Lock
* Event
* Semaphore

These ought to be supported on AnyIO as well, now that we have otherwise compatible implementations of these primitives in AnyIO.

### LINKED ISSUE #217: Add the statistics() method to memory channels
Trio's memory channels have a `statistics()` method: https://trio.readthedocs.io/en/stable/reference-core.html#trio.open_memory_channel

Some have already found this useful according to the Trio Gitter room. Thus AnyIO should get this as well.

agronholm:
Closes #216. Closes #217.

### DIFF
--- a/src/anyio/abc/_synchronization.py
+++ b/src/anyio/abc/_synchronization.py
@@ -1,8 +1,35 @@
 from abc import ABCMeta, abstractmethod
+from dataclasses import dataclass
 from types import TracebackType
 from typing import Optional, Type, TypeVar
 
 T_Retval = TypeVar('T_Retval')
+
+
+@dataclass(frozen=True)
+class EventStatistics:
+    """
+    :ivar int tasks_waiting: number of tasks waiting on :meth:`~.Event.wait`
+    """
+
+    tasks_waiting: int
+
+
+@dataclass(frozen=True)
+class CapacityLimiterStatistics:
+    """
+    :ivar int borrowed_tokens: number of tokens currently borrowed by tasks
+    :ivar float total_tokens: total number of available tokens
+    :ivar tuple borrowers: tasks or other objects currently holding tokens borrowed from this
+        limiter
+    :ivar int tasks_waiting: number of tasks waiting on :meth:`~.CapacityLimiter.acquire` or
+        :meth:`~.CapacityLimiter.acquire_on_behalf_of`
+    """
+
+    borrowed_tokens: int
+    total_tokens: float
+    borrowers: tuple
+    tasks_waiting: int
 
 
 class Event(metaclass=ABCMeta):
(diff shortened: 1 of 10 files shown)


src/anyio/abc/_synchronization.py
```diff
@@ -1,10 +1,37 @@
...
+    :ivar float total_tokens: total number of available tokens
+    :ivar tuple borrowers: tasks or other objects currently holding tokens borrowed from this
+        limiter
+    :ivar int tasks_waiting: number of tasks waiting on :meth:`~.CapacityLimiter.acquire` or
+        :meth:`~.CapacityLimiter.acquire_on_behalf_of`
+    """
+
+    borrowed_tokens: int
+    total_tokens: float
+    borrowers: tuple
```
graingert, line 31:
```suggestion
    borrowers: Tuple[object]
```

agronholm: Why? Borrowers can be any type, but this would mean only `object()` instances can go to `borrowers`.

graingert: every instance is an instance of object

agronholm: PEP 484 is pretty clear on this. A plain `tuple` which I had here, is equivalent to `Tuple[Any]` which is NOT the same as `Tuple[object]`.

agronholm: Reading up on it once more, it's not so *clear* why this is so, but the `Any` type was introduced specifically for this purpose.

graingert: sorry I meant:

```suggestion
    borrowers: Tuple[object, ...]
```

graingert: yes I know, using `tuple`, and alias of `Tuple[Any, ...]` mypy considers the following sound:

```python
capacity_limiter = anyio.create_capacity_limiter()
capacity_limiter.acquire_on_behalf_of(current_task())
borrowers: Tuple[int, ...] = capacity_limiter.statistics().borrowers
v: int = borrowers[0]  # woops v is actually a Task!
```

agronholm: Yeah but `Tuple[Any, ...]` is what we want and (I correct myself) `tuple` is equivalent to `Tuple[Any, ...]`.
The difference between `object` and `Any` comes down to type substitution rules which are not relevant here.
Bottom line is, `tuple` is correct.

PR #221, opened 2021-02-27
Made the cancel scope deadline property writable

### LINKED ISSUE #215: Allow changing the deadline in cancel scopes
Trio allows the deadline of a cancel scope to be changed at will while AnyIO does not (because of curio). Now that curio is no longer blocking the implementation of this feature, we should add it to further AnyIO's API compatibility with trio.

One thing to consider is the task timeout callback: it must now explicitly check if the deadline has indeed been reached, and if not, reschedule another callback.

agronholm:
Closes #215.

### DIFF
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -301,6 +301,16 @@ class CancelScope(abc.CancelScope):
     def deadline(self) -> float:
         return self._deadline
 
+    @deadline.setter
+    def deadline(self, value: float) -> None:
+        self._deadline = float(value)
+        if self._timeout_handle is not None:
+            self._timeout_handle.cancel()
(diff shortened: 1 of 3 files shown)


src/anyio/_backends/_asyncio.py
```diff
@@ -303,6 +301,12 @@ def cancel(self) -> None:
     def deadline(self) -> float:
         return self._deadline
 
+    @deadline.setter
+    def deadline(self, value: float) -> None:
+        self._deadline = float(value)
+        if self._active:
+            self._timeout()
```
graingert, line 308: don't you need to cancel the old `self._timeout_handle`?

agronholm: Yes, good catch!

graingert (approved):

PR #223, opened 2021-02-28
Implement worker processes

### LINKED ISSUE #219: Implement worker processes for CPU parallelism
Not all applications are consistently CPU-bound. To obtain CPU parallelism of arbitrary python code, it is currently necessary to use worker subprocesses, the easiest method being to build on top of `multiprocessing` or `concurrent.futures`.

Other IO libraries provide a shortcut for this. `asyncio` provides [`loop.run_in_executor`](https://github.com/python/cpython/blob/6c4c11763fad106e43cdcfdbe3bd33ea2765a13f/Lib/asyncio/base_events.py#L800-L814) which when combined with a `ProcessPoolExecutor` instance does the job. `curio` provides [`run_in_process`](https://github.com/dabeaz/curio/blob/bb4efc43175d454e1a55d415184db3ce0542b34b/curio/workers.py#L182-L216) which wraps `multiprocessing.Process`.  `trio` does not provide this feature but lists it as a [missing piece](https://github.com/python-trio/trio/issues/5).

I wrote [`trio-parallel`](https://github.com/richardsheridan/trio-parallel) to provide that missing piece, but the code works just as well as a [component of `trio` itself](https://github.com/python-trio/trio/pull/1781). @agronholm mentioned in [Gitter](https://gitter.im/python-trio/general?at=6032e8ca7ba8124e547e236c) that he would like to see a similar feature in `anyio`. I don't really fancy maintaining another library but if you would like to design an API, I can squeeze the internals of `trio-parallel` to fit it and provide a PR to `anyio`.

richardsheridan:
Closes #219. This is a straightforward transliteration of [`trio-parallel`](https://github.com/richardsheridan/trio-parallel), but using threads to portably do IPC and using a trick from [`curio`](https://github.com/dabeaz/curio/blob/bb4efc43175d454e1a55d415184db3ce0542b34b/curio/workers.py#L25-L53) to avoid depending on `outcome`.

### DIFF
--- a/.github/workflows/codeqa-test.yml
+++ b/.github/workflows/codeqa-test.yml
@@ -29,6 +29,7 @@ jobs:
 
   test:
     needs: [lint]
+    timeout-minutes: 10
     strategy:
       fail-fast: false
       matrix:
--- a/src/anyio/__init__.py
+++ b/src/anyio/__init__.py
@@ -14,7 +14,8 @@ __all__ = ('maybe_async', 'maybe_async_cm', 'run', 'sleep', 'current_time', 'get
            'wait_all_tasks_blocked', 'run_sync_in_worker_thread', 'run_async_from_thread',
            'run_sync_from_thread', 'current_default_worker_thread_limiter',
            'create_blocking_portal', 'start_blocking_portal', 'typed_attribute',
-           'TypedAttributeSet', 'TypedAttributeProvider')
+           'TypedAttributeSet', 'TypedAttributeProvider', 'run_sync_in_worker_process',
+           'current_default_worker_process_limiter', 'BrokenWorkerError')
 
 from ._core._compat import maybe_async, maybe_async_cm
 from ._core._eventloop import current_time, get_all_backends, get_cancelled_exc_class, run, sleep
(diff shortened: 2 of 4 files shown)


richardsheridan:
Looks like it is working except for a hang on Pypy. I'll need to debug that another time but the rest of the code should be good for review.

agronholm:
This looks fairly complex, and as its trio counterpart, uses multiprocessing. I'm going to explore my own alternative and see what I come up with before making a decision about this.

I've started work on my own executor pooling situation (threads at first), but encountered an awkward design problem: when do we shut down the workers? AnyIO code can be launched at any point in a library and application.

richardsheridan:
The contract as it stands is for the workers to be bounded by the main process lifetime. Since process startup takes such a long time, it is a big optimization to cache them this way, not knowing which thread or task or library may dispatch work.  In the worst case, a trio and asyncio run both fill up their capacity limiters once, leaving `cpu_count`x2 workers to die after 10 minutes. On unix this will leave zombies behind until the next `prune()` or the main process ends.

agronholm:
I've added a thread pool implementation for asyncio in #227. I'm going to try and see what I can do with process pooling.

agronholm:
I now have a rudimentary implementation of worker process pooling that utilizes anyio's own subprocess handling and does not involve multiprocessing. I will send a PR once I have polished the code.

agronholm:
The biggest challenge is still the management of the pool lifecycle – how to make sure it stops with the event loop. With trio this could possibly be done using its instrumentation facilities (`after_run`?) or a system task.

agronholm:
A system task seems to work fine on trio. The biggest remaining problem is asyncio when not using `asyncio.run()` or `anyio.run()`.

richardsheridan:
please also see my [unmerged branch of trio-parallel](https://github.com/richardsheridan/trio-parallel/pull/19) where I use contextvars to start and stop pools "on demand".

agronholm:
Are you referring to the pruning of dead workers? Or are you saying that's stopping workers at the end of the event loop run?

agronholm:
I'm thinking I could pull a rabbit from a hat by using some weakref magic to trigger last-resort finalization of the worker processes at exit.

PR #224, opened 2021-03-02
Added support for sending and receiving file descriptors on a UNIX socket stream

### LINKED ISSUE #201: Using "recvmsg" with anyio
Using `wait_socket_readable` fails when using asyncio.

Seems that asyncio decides to add a socket reader as soon as the socket is opened – which causes `BaseSelectorEventLoop._ensure_fd_no_transport` to go splat when I try to check for readability.

This test program succeeds with trio, but not with asyncio:
```python3
#!/usr/bin/python3

# boring stuff
import anyio
PORT=59493
async def main():
    async def handler(sock):
        await sock.send(b'123')
        await anyio.sleep(9999)
    async def server(evt):
        listener = await anyio.create_tcp_listener(local_port=PORT)
        await evt.set()
        await listener.serve(handler)

        evt = anyio.create_event()
        await tg.spawn(server, evt)
        await evt.wait()
# boring stuff ends

        async with await anyio.connect_tcp('localhost', PORT) as client:
            sock = client.extra(anyio.abc.SocketAttribute.raw_socket)
            await anyio.wait_socket_readable(sock.fileno())
            data, aux, *_ = sock.recvmsg(4096,1024)
            assert data == b'123', data

```
I'm doing this because depending on the options it's called with, the library I'm anyio-izing may need to pass file descriptors along.

agronholm:
Closes #201.

### DIFF
--- a/docs/networking.rst
+++ b/docs/networking.rst
@@ -101,6 +101,53 @@ And the listener::
 
     run(main)
 
+Sending and receiving file descriptors
+++++++++++++++++++++++++++++++++++++++
+
+UNIX sockets can be used to pass open file descriptors (sockets and files) to another process.
+The receiving end can then use either :func:`os.fdopen` or :func:`socket.socket` to get a usable
+file or socket object, respectively.
+
+The following is an example where a client connects to a UNIX socket server and receives the
+descriptor of a file opened on the server, reads the contents of the file and then prints them on
+standard output.
+
+Client::
+
+    import os
+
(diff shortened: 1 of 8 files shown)


smurfix (approved):
While I don't have time to do any in-depth testing of this right now, the code does look reasonable enough ;-)

agronholm:
That's understandable. Thanks for taking the time to peek at it at least!

PR #225, opened 2021-03-02
Added an exception handler in the asyncio test runner

### LINKED ISSUE #205: during asyncio tests set a default loop exception_handler that warns (similar to -p unraisablehook)
consider the following test:
```python
import pytest
import asyncio

@pytest.mark.anyio
@pytest.mark.parameterize("anyio_backend", ["asyncio"])
async def test_foo():
    def raise_():
        raise Exception

    asyncio.get_running_loop().call_soon(raise_)
    await asyncio.sleep(0)
```

it passes cleanly.

However if you explicitly collect any errors:

```python
import pytest
import asyncio

@pytest.mark.anyio
@pytest.mark.parameterize("anyio_backend", ["asyncio"])
async def test_foo():
    old_exception_handler = loop.get_exception_handler()
    exceptions = []

    def exception_handler(*args, **kwargs):
        exceptions.append((args, kwargs))

    loop.set_exception_handler(exception_handler)
    try:
        def raise_():
            raise Exception

        asyncio.get_running_loop().call_soon(raise_)
        await asyncio.sleep(0)
        assert exceptions == []
    finally:
        loop.set_exception_handler(old_exception_handler)
```

```
=================================== test session starts ====================================
platform linux -- Python 3.9.2, pytest-6.2.2, py-1.10.0, pluggy-0.13.1
rootdir: /home/graingert/projects/anyio, configfile: pyproject.toml
plugins: hypothesis-6.2.0, asyncio-0.14.0, anyio-2.1.0.post21+dirty
collected 1 item                                                                           

ham.py F                                                                             [100%]

========================================= FAILURES =========================================
_________________________________________ test_foo _________________________________________
ham.py:20: in test_foo
    assert exceptions == []
E   AssertionError: assert [((<_UnixSele...py:15'}), {})] == []
E     Left contains one more item: ((<_UnixSelectorEventLoop running=True closed=False debug=False>, {'exception': Exception(), 'handle': <Handle test_fo...y:15>, 'message': 'Exception in callback test_foo.<locals>.raise_() at /home/graingert/projects/anyio/ham.py:17'}), {})
E     Use -v to get the full diff
==================================== 1 failed in 0.04s ====================================
```

pytest-dev/pytest-asyncio#205
https://github.com/pytest-dev/pytest/blob/master/src/_pytest/unraisableexception.py#L62-L78

agronholm:
It will now raise all the capture exceptions if the test function completes without raising any exceptions of its own.

Closes #205.

### DIFF
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -1023,6 +1023,9 @@ class UNIXSocketStream(abc.SocketStream):
 
 
 class TCPSocketListener(abc.SocketListener):
+    _accept_scope: Optional[CancelScope] = None
+    _closed = False
+
     def __init__(self, raw_socket: socket.SocketType):
         self.__raw_socket = raw_socket
         self._loop = cast(asyncio.BaseEventLoop, get_running_loop())
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -1235,7 +1235,8 @@ class _SignalReceiver:
 
     def _deliver(self, signum: int) -> None:
         self._signal_queue.append(signum)
-        self._future.set_result(None)
+        if not self._future.done():
+            self._future.set_result(None)
 
     def __aiter__(self):
         return self
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -835,7 +835,9 @@ class SocketStream(abc.SocketStream):
             try:
                 self._transport.write(item)
             except RuntimeError as exc:
-                if self._closed:
+                if self._protocol.write_future.exception():
+                    await self._protocol.write_future
+                elif self._closed:
                     raise ClosedResourceError from None
                 elif self._transport.is_closing():
                     raise BrokenResourceError from exc
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -1333,10 +1333,15 @@ async def wait_all_tasks_blocked() -> None:
 class TestRunner(abc.TestRunner):
     def __init__(self, debug: bool = False, use_uvloop: bool = True,
                  policy: Optional[asyncio.AbstractEventLoopPolicy] = None):
+        self._exceptions: List[Exception] = []
         _maybe_set_event_loop_policy(policy, use_uvloop)
         self._loop = asyncio.new_event_loop()
         self._loop.set_debug(debug)
         asyncio.set_event_loop(self._loop)
+        self._loop.set_exception_handler(self._exception_handler)
+
+    def _exception_handler(self, loop: asyncio.AbstractEventLoop, context: Dict[str, Any]) -> None:
+        self._exceptions.append(context['exception'])
 
     def _cancel_all_tasks(self):
         to_cancel = all_tasks(self._loop)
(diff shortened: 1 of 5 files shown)


src/anyio/_backends/_asyncio.py
```diff
@@ -1333,10 +1333,15 @@ async def wait_all_tasks_blocked() -> None:
 class TestRunner(abc.TestRunner):
     def __init__(self, debug: bool = False, use_uvloop: bool = True,
                  policy: Optional[asyncio.AbstractEventLoopPolicy] = None):
+        self._exceptions: List[Exception] = []
```
graingert, line 1336:
```suggestion
        self._exceptions: List[BaseException] = []
```

?

agronholm: BaseExceptions crash the event loop; the handler is never called with such an exception.

graingert: Cool

src/anyio/_backends/_asyncio.py
```diff
@@ -1364,4 +1369,10 @@ def close(self) -> None:
             self._loop.close()
 
     def call(self, func: Callable[..., Awaitable], *args, **kwargs):
-        return self._loop.run_until_complete(func(*args, **kwargs))
+        retval = self._loop.run_until_complete(func(*args, **kwargs))
+        if len(self._exceptions) == 1:
+            raise self._exceptions[0]
+        elif self._exceptions:
+            raise ExceptionGroup(self._exceptions)
```
graingert, line 1376: If you raise this is a fixture teardown pytest will report the test result and this failure separately

agronholm: Isn't that what we want though?

graingert: Yes but it's not what's in the changelog

> exceptions raised in synchronous callbacks will be raised then, essentially failing the test

pytest can report a test passing and then failing the teardown

agronholm: I'll update the changelog then.

There's another problem: since `self._loop.run_until_complete(...)` raises `StopAsyncIteration` on fixture teardown, it won't look at the accumulated exceptions. And even if it did, there's *another* problem: the exception list is not cleared between runs, so the same exceptions would be raised again!

tests/test_pytest_plugin.py
```diff
@@ -68,7 +83,8 @@ async def test_skip_inline(some_feature):
     )
 
     result = testdir.runpytest('-v')
-    result.assert_outcomes(passed=3 * len(get_all_backends()) + 1, skipped=len(get_all_backends()))
+    result.assert_outcomes(passed=3 * len(get_all_backends()) + 1, skipped=len(get_all_backends()),
+                           failed=1)
```
graingert, line 87: I'd expect test_asyncio_exception_handler to pass, then fail in teardown:

```suggestion
    result.assert_outcomes(passed=3 * len(get_all_backends()) + 2, skipped=len(get_all_backends()),
                           failed=1)
```

agronholm: Why? The exception is raised during the actual test run (albeit in a callback), not during fixture teardown.

src/anyio/_backends/_asyncio.py
agronholm: I've updated the changelog to match the actual behavior.

agronholm:
I'll merge this tomorrow unless you have some concerns left.

PR #226, opened 2021-03-03
Added low level checkpoint functions to the public API

agronholm:
This part of the API now closely matches that of trio.

### DIFF
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -320,11 +320,20 @@ class CancelScope(abc.CancelScope):
         return self._shield
 
 
-async def checkpoint():
+async def checkpoint() -> None:
+    await checkpoint_if_cancelled()
+    await asyncio.sleep(0)
+
+
+async def checkpoint_if_cancelled() -> None:
+    task = current_task()
+    if task is None:
+        return
+
     try:
-        cancel_scope = _task_states[current_task()].cancel_scope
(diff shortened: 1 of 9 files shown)


src/anyio/_backends/_asyncio.py
```diff
@@ -320,11 +320,20 @@ def shield(self) -> bool:
         return self._shield
 
 
-async def checkpoint():
+async def checkpoint() -> None:
+    await checkpoint_if_cancelled()
+    await asyncio.sleep(0)
+
+
+async def checkpoint_if_cancelled() -> None:
```
graingert, line 328: trio this is documented as:
```
    This is either a no-op, or else it allow other tasks to be scheduled and
    then raises :exc:`trio.Cancelled`.
```

so it needs an `await asyncio.sleep(0)` somewhere

agronholm: Now it has (I needed to merge the changes from master to make it work).

src/anyio/_backends/_asyncio.py
```diff
@@ -337,10 +337,35 @@ def shield(self) -> bool:
         return self._shield
 
 
-async def checkpoint():
+async def checkpoint() -> None:
+    await checkpoint_if_cancelled()
```
graingert, line 341:
```suggestion
```

graingert (approved):

PR #227, opened 2021-03-03
Added thread pooling for asyncio

### LINKED ISSUE #222: Use thread pools in run_sync_in_worker_thread()
Right now, the asyncio implementation of this starts a new thread every time this function is called. We should be using thread pools instead, like trio does.

agronholm:
In the current implementation, every time `run_sync_in_worker_thread()` is called on asyncio, it spawns a dedicated worker thread for that call. This can be wasteful, so this PR adds thread pooling that works just like trio's (AFAIK). A special task is launched on the first call to `run_sync_in_worker_thread()` which launches a task that, when cancelled at the event loop exit, commands the workers to shut themselves down and waits until they have.

Closes #222.

### DIFF
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -653,14 +653,15 @@ def _thread_pool_worker(work_queue: Queue, workers: Set[Thread],
         work_queue.task_done()
 
 
+_threadpool_work_queue: RunVar[Queue] = RunVar('_threadpool_work_queue')
+_threadpool_idle_workers: RunVar[Set[Thread]] = RunVar('_threadpool_idle_workers')
+_threadpool_workers: RunVar[Set[Thread]] = RunVar('_threadpool_workers')
+
+
 def _loop_shutdown_callback(f: asyncio.Future) -> None:
     """This is called when the root task has finished."""
-    for _ in range(len(threadlocals.threadpool_workers)):
-        threadlocals.threadpool_work_queue.put_nowait((None, None, None))
-
-    del threadlocals.threadpool_work_queue
-    del threadlocals.threadpool_idle_workers
-    del threadlocals.threadpool_workers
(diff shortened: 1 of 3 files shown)


agronholm:
The workers shut themselves down after 10 seconds of inactivity. Is this okay?

agronholm:
I just realized another (minor) problem: the shutdown task only works if the event loop was launched using `asyncio.run()` or `anyio.run()` since it relies on being auto-cancelled at event loop exit. If instead the event loop was launched using `loop.run_until_complete()` or (heaven forbid) `loop.run_forever()`, the process will be stuck until the last worker times out.

agronholm:
The (ugly) solution was to find the root task and attach a callback to it which will prompt the worker threads to quit.

smurfix (approved):

agronholm:
Thanks for the review!

src/anyio/_backends/_asyncio.py
```diff
@@ -1536,11 +1595,16 @@ def statistics(self) -> CapacityLimiterStatistics:
...
-    return _default_thread_limiter
+_default_thread_limiter: RunVar[CapacityLimiter] = RunVar('_default_thread_limiter')
 
 
-_default_thread_limiter = CapacityLimiter(40)
+def current_default_thread_limiter():
+    try:
+        return _default_thread_limiter.get()
+    except LookupError:
+        limiter = CapacityLimiter(40)
```
amirouche, line 1605: @agronholm Does this means that the default behavior will allow to spawn at most 40 threads? Did you consider `os.cpu_count()` that is what is used in `asyncio.to_thread`?

ref: https://github.com/python/cpython/blob/main/Lib/concurrent/futures/thread.py#L142

agronholm: This is in line with trio.to_thread which also allows a maximum of 40 threads by default.

PR #229, opened 2021-03-06
run asyncio selector and proactor reactors on win32

graingert:


agronholm:
Is this really necessary? Both event loop policies are tested because we run tests against Python 3.6 and 3.9 on Windows, and those have different default event loop policies.

graingert:
> Is this really necessary? Both event loop policies are tested because we run tests against Python 3.6 and 3.9 on Windows, and those have different default event loop policies.

Python 3.6 is EOL at the end of the year so this would be needed then

agronholm:
> Python 3.6 is EOL at the end of the year so this would be needed then
So what? The default only changed in Python 3.8.

PR #232, opened 2021-03-07
test that cancelling shielded tasks spends more time sleeping than ca…

graingert:
…ncelling


agronholm:
I've committed this on your behalf already.

PR #234, opened 2021-03-08
Restored partial backwards compatibility with 2.x

agronholm:
The list of truly breaking changes has been narrowed down quite a bit.

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -12,7 +12,6 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
   * ``current_time()``
   * ``current_effective_deadline()``
-  * ``get_running_tasks()``
   * ``MemoryObjectReceiveStream.receive_nowait()``
   * ``MemoryObjectSendStream.send_nowait()``
 
@@ -20,6 +19,7 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
   so will emit a deprecation warning):
 
   * ``get_current_task()``
+  * ``get_running_tasks()``
   * ``CancelScope.cancel()``
   * ``CapacityLimiter.acquire_nowait()``
   * ``CapacityLimiter.acquire_on_behalf_of_nowait()``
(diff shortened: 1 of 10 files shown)


smurfix:
Nice. Saves me some work. *g*

agronholm:
Can you do a review/approval here?

docs/versionhistory.rst
```diff
@@ -12,26 +12,30 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
   * ``current_time()``
   * ``current_effective_deadline()``
+  * ``get_current_task()``
```
smurfix, line 15: For even more improved fake-await-ery we could add ```
def __await__(self):
    if False: yield
    warn("Don't do this")
    return self
```
method to `ansio.TaskInfo`. That'd fix `get_current_task`. A list with that method tacked on is almost as simple, for `get_running_tasks`.

Comment (agronholm) [reply docs/versionhistory.rst:15]: Ok, that's fair.

**Comment by agronholm** [comment]
As a bonus, `get_running_tasks()` is also now backwards compatible.

**Review by smurfix — APPROVED** [review]

File: src/anyio/_core/_compat.py [hunk 2]
```diff
@@ -76,6 +76,16 @@ def __await__(self):
             yield
 
 
+class DeprecatedAwaitableList(Generic[T], List[T], DeprecatedAwaitable):
```
smurfix, line 79: Huh. Do we actually need the `Generic[T]` part here?

agronholm: Yes, it's needed for any class that requires a type argument.

graingert: I thought if you already inherited from a GenericAlias it worked:

https://mypy-play.net/?mypy=latest&python=3.9&gist=fdc3ca168923e66527952336ccf7f320

agronholm: I couldn've sworn that I tried this already. I guess I left out the type argument.

PR #236, opened 2021-03-10
test that get_current_task in spawn callbacks is the outer task

graingert:


agronholm (changes requested):
The code looks good, but would you mind adding a note about the behavior change? You can mention that this brings asyncio behavior in line with trio.

docs/versionhistory.rst
```diff
@@ -74,6 +74,8 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 - Changed the asyncio test runner to capture unhandled exceptions from asynchronous callbacks and
   unbound native tasks which are then raised after the test function (or async fixture setup or
   teardown) completes
+- Changed, the asyncio ``TaskGroup.spawn()`` now calls the new task function immediately to be
+  consistent across backends.
```
agronholm, line 78:
```suggestion
- Changed the asyncio ``TaskGroup.spawn()`` method to call the target function immediately before spawning the
  task, for consistency across backends
```

agronholm (approved):

PR #238, opened 2021-03-14
Use callbacks instead of coroutine wrappers on Python 3.8+

agronholm:
The asyncio event loop is allergic to BaseExceptions being raised by tasks on Python 3.7 and earlier, so a separate code path is left for these older versions. An effort was also made to hide the coroutine wrapper in tracebacks in frameworks that support it.


src/anyio/_backends/_asyncio.py
```diff
@@ -315,6 +314,7 @@ def _deliver_cancellation(self) -> None:
                     cancellable_tasks.add(task)
 
         for task in cancellable_tasks:
+            print('cancelling task', id(task))
```
graingert, line 317: Print

PR #246, opened 2021-03-22
Improved the sales pitch

agronholm:


lewoudar:
One thing that amazed me with trio is how we can write less to achieve the same result as opposed to other frameworks, so one thing you can do is to link to an example showing how asyncio can be tricky to use compared to anyio, for example an **happy eye ball implementation** or an **udp service**.

agronholm:
Asyncio does have a [happy eyeballs implementation](https://bugs.python.org/issue33530) though (in Python 3.8+).

agronholm:
But yeah, but are probably worth mentioning in the feature list.

lewoudar:
yeah, I think that you should write one to demonstrate the advantage of anyio over asyncio :D

smurfix:
Asyncio now has a 150+ line Happy Eyeballs implementation which, while shorter/more correct/more comprehensive than previous attempts to write this, still contains a lot of code that is a "poor man's version of curio's TaskGroup or trio's nursery" (this is a direct quote from their code) and which happens to drop exceptions on the floor left and right – instead of, like, a real taskgroup.

agronholm:
I've mentioned both now, but frankly I am starting to feel too mentally exhausted at this point to write examples in the README (and I'm not 100% convinced that it's the right place to do it).

lewoudar:
I don't also think that the examples must be written in the README, I was thinking of creating a folder in the repository with examples and put a link to it

agronholm:
The documentation has a number of examples in it but one would have to go looking for them specifically.

Anyway, I'm gonna merge this and get back to writing more examples if I can figure out where they should go so that prospecting users would easily find them.

PR #249, opened 2021-03-29
Restructured thread/process related functions

### LINKED ISSUE #248: Move worker thread/process APIs to trio-like module names
Trio has the modules `trio.to_thread` and `trio.from_thread` where you can find the corresponding functions. It would be nice for AnyIO to follow this example too. For worker processes, `anyio.to_process` would be the logical choice.

agronholm:
Added new modules: anyio.to_thread, anyio.to_process, anyio.from_thread

This is another step towards a trio-like API.

Closes #248.


tests/test_to_process.py
```diff
@@ -0,0 +1,82 @@
...
+
+pytestmark = pytest.mark.anyio
+
+
+@pytest.fixture(autouse=True)
+def check_compatibility(anyio_backend_name):
+    if anyio_backend_name == 'asyncio':
+        if platform.system() == 'Windows' and sys.version_info < (3, 8):
+            pytest.skip('Python < 3.8 uses SelectorEventLoop by default and it does not support '
+                        'subprocesses')
```
richardsheridan, line 19: If Python < 3.8 is not supported on windows, will an informative error be raised if `to_process.run_sync` is called?

agronholm: Asyncio raises a `NotImplementedError`. I wonder if we need a better error message. Maybe we'll cross that bridge when we come to it? It may be that nobody even cares :)

richardsheridan:
Maybe not! I think a one-sentence note in the docs would cover it.

agronholm:
I added that note in e27be394 (sorry, pushed to master by mistake).

PR #253, opened 2021-04-06
Fix semaphore race condition

### LINKED ISSUE #252: Semaphore race condition
I unexpectedly got a 'WouldBlock' error while not using any *_nowait() calls. I'm using the latest version of the master branch. The problem goes away if I replace https://github.com/agronholm/anyio/blob/master/src/anyio/_core/_synchronization.py#L305 with `await self.acquire()`, but I'm not sure if that's actually the best solution. I think the problem is that another task increments the semaphore (by calling acquire()) before the Event.wait()-handling coroutine is scheduled to run.

Full traceback:

```
    async with self._pool.read_transaction() as tx:
/usr/lib/python3.8/contextlib.py:171: in __aenter__
    return await self.gen.__anext__()
tests/test_replicate.py:39: in read_transaction
    async with self.read_semaphore, self._transaction() as conn:
venv/lib/python3.8/site-packages/anyio/_core/_synchronization.py:282: in __aenter__
    await self.acquire()
venv/lib/python3.8/site-packages/anyio/_core/_synchronization.py:306: in acquire
    self.acquire_nowait()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <anyio.Semaphore object at 0x7f434698fb50>

    def acquire_nowait(self) -> None:
        """
        Acquire the underlying lock, without blocking.
    
        :raises ~WouldBlock: if the operation would block
    
        """
        if self._value == 0:
>           raise WouldBlock
E           anyio.WouldBlock

venv/lib/python3.8/site-packages/anyio/_core/_synchronization.py:316: WouldBlock
```

smurfix:
Fixes this problem:

* task A acquires S
* task B waits for S
* task C is scheduled
* task A releases S, scheduling task B
* task C runs and successfully acquires S, ahead of B
* task B's `acquire` calls `acquire_nowait` which raises `WouldBlock`

Closes #252.

### DIFF
--- a/src/anyio/_core/_synchronization.py
+++ b/src/anyio/_core/_synchronization.py
@@ -275,6 +275,7 @@ class Semaphore:
                 raise ValueError('max_value must be equal to or higher than initial_value')
 
         self._value = initial_value
+        self._waiting = 0
         self._max_value = max_value
         self._waiters: Deque[Event] = deque()
 
@@ -302,6 +303,7 @@ class Semaphore:
 
                 raise
 
+            self._waiting -= 1
             self.acquire_nowait()
 
     def acquire_nowait(self) -> None:
@@ -311,7 +313,7 @@ class Semaphore:
         :raises ~WouldBlock: if the operation would block
 
         """
-        if self._value == 0:
+        if self._value <= self._waiting:
             raise WouldBlock
 
         self._value -= 1
@@ -324,6 +326,7 @@ class Semaphore:
         self._value += 1
         if self._waiters:
             self._waiters.popleft().set()
+            self._waiting += 1
 
         return DeprecatedAwaitable(self.release)
 
--- a/tests/test_synchronization.py
+++ b/tests/test_synchronization.py
@@ -1,8 +1,8 @@
 import pytest
 
 from anyio import (
-    CancelScope, Condition, Event, Lock, Semaphore, WouldBlock, create_task_group, to_thread,
-    wait_all_tasks_blocked)
+    CancelScope, Condition, Event, Lock, Semaphore, WouldBlock, create_task_group,
+    get_current_task, to_thread, wait_all_tasks_blocked)
 from anyio.abc import CapacityLimiter
 
 pytestmark = pytest.mark.anyio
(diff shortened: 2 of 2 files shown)


smurfix:
I have no idea how to build a reliable testcase for this, given that scheduling is essentially random.

agronholm:
I wonder if it'd be possible to add a test for this.

jab:
Does anything in https://trio.readthedocs.io/en/stable/reference-testing.html help (which IIRC allows Trio programs to have these kinds of tests), even if only as inspiration for something similar anyio could do?

smurfix:
> Does anything in https://trio.readthedocs.io/en/stable/reference-testing.html help 

Not that I can see. The problem is that you truly do need a race of two runnable tasks to trigger this.

**Idea*: Implement the sequence I described in my commit message. Monkeypatch `event.wait()`, when called by task B, to (a) do the original wait (b) unblock task C (c) sleep a few milliseconds.

smurfix:
Ugh, testing the unfixed anyio doesn't reliably trigger the problem. Will investigate.

smurfix:
Being stupid (on my part) (didn't push from the correct machine) doesn't help. The test now works.

smurfix:
I verified that the problem is reliably triggered on a not-fixed anyio. It doesn't raise `WouldBlock` as in the bug report but it does mis-order the tasks.

agronholm:
I need to take a good look at both the problem and the proposed solution to see if this is the fix I want. The new test should help in any case.

agronholm:
Now that I fully understand the problem, I managed to write a simpler fix that also passes the new test. I'm currently trying to write a less complicated test to accompany it.

agronholm:
Thanks for the effort but my own fix and tests were substantially simpler so I'm going to go with those instead.

smurfix:
> Thanks for the effort but my own fix and tests were substantially simpler so I'm going to go with those instead.

The point of all of this is to get the bugs fixed. If I had a problem with you taking a shortcut instead of the long and winding road I proposed I wouldn't be here in the first place.

PR #254, opened 2021-04-06
Added TaskGroup.start_soon() and changed spawn() back into a coroutine function

agronholm:
One more step towards a trio-like API.

### DIFF
--- a/docs/cancellation.rst
+++ b/docs/cancellation.rst
@@ -69,7 +69,7 @@ To accomplish this, open a new cancel scope with the ``shield=True`` argument::
     async def main():
         async with create_task_group() as tg:
             with CancelScope(shield=True) as scope:
-                tg.spawn(external_task)
+                tg.start_soon(external_task)
                 tg.cancel_scope.cancel()
                 print('Started sleeping in the host task')
                 await sleep(1)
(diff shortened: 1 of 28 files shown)


agronholm:
This will break code written against the release candidates. The fix should be easy though.

agronholm:
I'm only just wondering if we could emit deprecation warnings from the renamed methods. How would AnyIO 2/3 compatible code look like?

smurfix:
Why doesn't `spawn` emit a DeprecationWarning?

agronholm:
I'm trying to give library developers a way to use AnyIO 2.x and 3.x in the same code base without running into deprecation warnings. With free functions you could just conditionally import them but with methods it's not as simple. Can you suggest a solution?

smurfix:
Hmm. Tricky, sounds like "defer adding a deprecation warning to `spawn` until anyio 3.1" is the best (or at least easiest-to-implement) idea here.

agronholm:
I made them emit deprecation warnings after all, and pointed out that cross-compatible code could simply suppress the deprecation warnings. Good enough?

smurfix:
Perfect.

smurfix (approved):

agronholm:
Thanks for the review!

PR #258, opened 2021-04-14
Refactored the asyncio thread pooling code

agronholm:
This fixes the asyncio idle thread race condition.

Idle worker threads are also now pruned on every call to run_sync_in_worker_thread() instead of the worker threads timing out.

### DIFF
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -12,9 +12,9 @@ from functools import partial, wraps
 from inspect import (
     CORO_RUNNING, CORO_SUSPENDED, GEN_RUNNING, GEN_SUSPENDED, getcoroutinestate, getgeneratorstate)
 from io import IOBase
-from queue import Empty, Queue
+from queue import Queue
 from socket import AddressFamily, SocketKind, SocketType
-from threading import Thread, current_thread
+from threading import Thread
 from types import TracebackType
 from typing import (
     Any, Awaitable, Callable, Collection, Coroutine, Deque, Dict, Generator, List, Optional,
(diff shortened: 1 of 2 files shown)


agronholm:
I wrote a test for this too but I'm not sure if it should be included because it relies on the previous implementation in order to fail consistently.

uSpike:
Can you elaborate on why you don't want to include your test?  Does it rely on specific library code that doesn't exist anymore?  Do you think it would not be useful for catching regressions?

agronholm:
It only catches the problem like 30% of the time. To make it catch the problem 100% of the time I would need to monkeypatch the `remove()` method of the asyncio `idle_workers` deque. This solution only applies to the specific implementation that existed before this patch.

uSpike:
Then it seems reasonable to me to omit the test for this.

agronholm:
Does the code look sane otherwise?

src/anyio/_backends/_asyncio.py
```diff
@@ -726,12 +711,28 @@ async def run_sync_in_worker_thread(
...
             work_queue.put_nowait((func, args, future))
             if not idle_workers:
-                args = (work_queue, workers, idle_workers)
-                thread = Thread(target=_thread_pool_worker, args=args, name='AnyIO worker thread')
+                thread = Thread(target=_thread_pool_worker, args=(work_queue, get_running_loop()),
+                                name='AnyIO worker thread')
                 workers.add(thread)
                 thread.start()
+            else:
+                thread, idle_since = idle_workers.pop()
```
uSpike, line 719: `idle_since` seems unused.  Is that intentional?

agronholm: Well, it does come out when I `pop()` from `idle_workers`, and I figured this serves as ad-hoc documentation of _what_ comes out. Doesn't matter either way I guess?

uSpike: OK that makes sense, we only care about the `idle_since` value of workers remaining.

uSpike (approved):

uSpike:
Looks good to me!  Thanks for the fix.

agronholm:
Thanks for looking through the code.

Issue #257, opened 2021-04-10
Implement `CancelScope.cancelled_caught`


agronholm:
The `CancelScope` implementation on AnyIO mirrors that of trio 1:1, except for this particular property.

Relevant trio documentation [here](https://trio.readthedocs.io/en/stable/reference-core.html#trio.CancelScope.cancelled_caught).

agronholm:
@johnzeringue do you intend to work on this any further?

johnzeringue:
> @johnzeringue do you intend to work on this any further?

I probably don't have the time to continue working on it. I was unable to find a way to make this work.

agronholm:
Closing in favor of #591, as it introduces a more useful property, `deadline_reached`.

PR #267, opened 2021-04-23
Add CancelScope.cancelled_caught

johnzeringue:
Added `CancelScope.cancelled_caught` to match the Trio API. Relevant
Trio documentation
[here](https://trio.readthedocs.io/en/stable/reference-core.html#trio.CancelScope.cancelled_caught).

Closes agronholm/anyio#257

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ __pycache__
 .cache
 .local
 .pre-commit-config.yaml
+.vscode/
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -254,13 +254,18 @@ class CancelScope(BaseCancelScope):
         self._deadline = deadline
         self._shield = shield
         self._parent_scope: Optional[CancelScope] = None
+        self._cancelled_caught = False
         self._cancel_called = False
+        self._cancelled = False
         self._active = False
         self._timeout_handle: Optional[asyncio.TimerHandle] = None
         self._cancel_handle: Optional[asyncio.Handle] = None
         self._tasks: Set[asyncio.Task] = set()
         self._host_task: Optional[asyncio.Task] = None
         self._timeout_expired = False
+
+    def _child_tasks(self):
+        return self._tasks - set([self._host_task])
 
     def __enter__(self):
         if self._active:
@@ -284,8 +289,7 @@ class CancelScope(BaseCancelScope):
         self._active = True
         return self
 
-    def __exit__(self, exc_type: Optional[Type[BaseException]], exc_val: Optional[BaseException],
-                 exc_tb: Optional[TracebackType]) -> Optional[bool]:
+    def _close(self, exc_val: Optional[BaseException]) -> Optional[bool]:
         if not self._active:
             raise RuntimeError('This cancel scope is not active')
         if current_task() is not self._host_task:
@@ -315,6 +319,8 @@ class CancelScope(BaseCancelScope):
         if exc_val is not None:
             exceptions = exc_val.exceptions if isinstance(exc_val, ExceptionGroup) else [exc_val]
             if all(isinstance(exc, CancelledError) for exc in exceptions):
+                self._cancelled_caught = self._cancel_called and not self._parent_cancelled()
+
                 if self._timeout_expired:
                     return True
                 elif not self._cancel_called:
@@ -325,6 +331,10 @@ class CancelScope(BaseCancelScope):
                     return True
 
         return None
+
+    def __exit__(self, exc_type: Optional[Type[BaseException]], exc_val: Optional[BaseException],
+                 exc_tb: Optional[TracebackType]) -> Optional[bool]:
+        return self._close(exc_val)
 
     def _timeout(self):
         if self._deadline != math.inf:
@@ -391,16 +401,25 @@ class CancelScope(BaseCancelScope):
 
         return False
 
-    def cancel(self) -> DeprecatedAwaitable:
-        if not self._cancel_called:
+    def _do_cancel(self, *, cancel_called) -> DeprecatedAwaitable:
+        if not self._cancelled:
             if self._timeout_handle:
                 self._timeout_handle.cancel()
                 self._timeout_handle = None
 
-            self._cancel_called = True
+            self._cancel_called |= cancel_called
+            self._cancelled = True
             self._deliver_cancellation()
 
         return DeprecatedAwaitable(self.cancel)
+
+    def cancel(self) -> DeprecatedAwaitable:
+        return self._do_cancel(cancel_called=True)
+
+    def _cancel_exc(self, exc_val: BaseException) -> None:
+        maybe_native_cancellation = isinstance(
+            exc_val, CancelledError) and not self._parent_cancelled()
+        self._do_cancel(cancel_called=not maybe_native_cancellation)
 
     @property
     def deadline(self) -> float:
@@ -415,6 +434,10 @@ class CancelScope(BaseCancelScope):
 
         if self._active and not self._cancel_called:
             self._timeout()
+
+    @property
+    def cancelled_caught(self) -> bool:
+        return self._cancelled_caught
 
     @property
     def cancel_called(self) -> bool:
@@ -535,16 +558,17 @@ class TaskGroup(abc.TaskGroup):
     async def __aexit__(self, exc_type: Optional[Type[BaseException]],
                         exc_val: Optional[BaseException],
                         exc_tb: Optional[TracebackType]) -> Optional[bool]:
-        ignore_exception = self.cancel_scope.__exit__(exc_type, exc_val, exc_tb)
         if exc_val is not None:
-            self.cancel_scope.cancel()
+            self.cancel_scope._cancel_exc(exc_val)
             self._exceptions.append(exc_val)
 
-        while self.cancel_scope._tasks:
+        while self.cancel_scope._child_tasks():
             try:
-                await asyncio.wait(self.cancel_scope._tasks)
-            except asyncio.CancelledError:
-                self.cancel_scope.cancel()
+                await asyncio.wait(self.cancel_scope._child_tasks())
+            except asyncio.CancelledError as exc:
+                self.cancel_scope._cancel_exc(exc)
+
+        ignore_exception = self.cancel_scope._close(ExceptionGroup(self._exceptions))
 
         self._active = False
         if not self.cancel_scope._parent_cancelled():
@@ -601,7 +625,7 @@ class TaskGroup(abc.TaskGroup):
         except BaseException as exc:
             if task_status_future is None or task_status_future.done():
                 self._exceptions.append(exc)
-                self.cancel_scope.cancel()
+                self.cancel_scope._cancel_exc(exc)
             else:
                 task_status_future.set_exception(exc)
         else:
@@ -632,7 +656,7 @@ class TaskGroup(abc.TaskGroup):
             if exc is not None:
                 if task_status_future is None or task_status_future.done():
                     self._exceptions.append(exc)
-                    self.cancel_scope.cancel()
+                    self.cancel_scope._cancel_exc(exc)
                 else:
                     task_status_future.set_exception(exc)
             elif task_status_future is not None and not task_status_future.done():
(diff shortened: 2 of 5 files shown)


src/anyio/_backends/_asyncio.py
```diff
@@ -304,6 +305,8 @@ def __exit__(self, exc_type: Optional[Type[BaseException]], exc_val: Optional[Ba
         if exc_val is not None:
             exceptions = exc_val.exceptions if isinstance(exc_val, ExceptionGroup) else [exc_val]
             if all(isinstance(exc, CancelledError) for exc in exceptions):
+                self._cancelled_caught = not self._parent_cancelled()
```
graingert, line 308: I think cancel_caught should be false on native cancellation

agronholm:
It seems that there are no tests for cases when `cancelled_caught` differs from `cancel_called`.

src/anyio/_backends/_asyncio.py
agronholm: Indeed, and there should be a test for this.

johnzeringue: Not sure if I understood native cancellation correctly. I took it to mean "a scope that's been cancelled without calling .cancel()", which seems to only happen when `Cancelled` is thrown inside the scope's context manager.

I added a test to that effect. Let me know if I've missed something.

johnzeringue:
@agronholm I added asserts for when `.cancel()` has been called but you're still inside the cancel scope's context manager. I think that's the only time that `cancelled_caught` and `cancel_called` should differ.

tests/test_taskgroups.py
```diff
@@ -884,3 +906,12 @@ async def exit_scope(scope):
     with pytest.raises(RuntimeError):
         async with create_task_group() as tg:
             tg.start_soon(exit_scope, scope)
+
+
+async def test_native_cancellation():
```
graingert, line 911: native cancellation is when someone uses `task = asyncio.create_task(...)` and `task.cancel()` eg:

```python
@pytest.mark.parametrize('anyio_backend', ['asyncio'])
async def test_start_native_host_cancelled():
    async def taskfunc(*, task_status):
        nonlocal started, finished
        started = True
        await sleep(2)
        finished = True

    async def start_another():
        nonlocal tg
            await tg.start(taskfunc)

    tg = None
    started = finished = False
    task = asyncio.get_event_loop().create_task(start_another())
    task.cancel()
    with pytest.raises(asyncio.CancelledError):
        await task

    assert started
    assert not finished

    assert not tg.cancel_scope.cancelled_caught
    assert not tg.cancel_scope.cancel_called
```

graingert: Hmm, the cancel should propagate and the other tasks should be cancelled, but the scope shouldn't be cancelled? Not sure

agronholm: The native cancellation should be treated as any other exception I think.

johnzeringue: > The native cancellation should be treated the same way as any other exception I think.

@agronholm is that at odds with 
> I think cancel_caught should be false on native cancellation

?

What would you expect the last two asserts to be in `test_start_native_host_cancelled`?

agronholm: The asserts look fine. If an exception is raised in a cancel scope, the scope won't be cancelled on its own.

agronholm: You can see for yourself:
```
from trio import CancelScope, run


async def main():
    try:
        with CancelScope() as scope:
            raise RuntimeError
    except RuntimeError:
        print(scope.cancel_called)
        print(scope.cancelled_caught)


run(main)
```

graingert: ah right no need for a task group here

johnzeringue: Independent of `cancelled_caught`, @graingert's `test_raised` passes and `test_start_native_host_cancelled` fails.

Is this because `CancelledError = asyncio.CancelledError`, so we don't know whether to treat `asyncio.CancelledError` as a special case or like any other exception? Or are we just not tracking state correctly?

agronholm: The reason `test_raises` fails is because there's a task group involved. When any exception falls through from a task in a task group, all other tasks are cancelled as well by means of cancelling the task group's cancel scope.

johnzeringue: Ah I see. I think I'm finally starting to understand.

I fixed `test_raises` and created a new version of `test_start_native_host_cancelled` that doesn't use a task group.

.gitignore
```diff
@@ -15,3 +15,5 @@ __pycache__
 .cache
 .local
 .pre-commit-config.yaml
+.vscode/
+.python-version
```
agronholm, line 19: What is `.python-version`? I've never encountered this before.

graingert: It's created by `pyenv local`

I'd recommend @johnzeringue and other pyenv uses add it to their global gitignore

tests/test_taskgroups.py
```diff
@@ -179,6 +181,34 @@ async def start_another():
     assert not finished
 
 
+@pytest.mark.parametrize('anyio_backend', ['asyncio'])
```
agronholm, line 184: Could you explain the purpose of this test?

tests/test_taskgroups.py
```diff
@@ -165,9 +165,11 @@ async def taskfunc(*, task_status):
         finished = True
 
     async def start_another():
+        nonlocal tg
```
agronholm, line 168: Why have you added this?

tests/test_taskgroups.py
```diff
@@ -582,7 +621,11 @@ async def child(fail):
 
 
 async def test_cancel_cascade():
+    tg2 = None
```
agronholm, line 624: Move this to the top of the main code block (after the function definitions) for consistency.

tests/test_taskgroups.py
```diff
@@ -456,6 +490,8 @@ async def test_nested_move_on_after():
 
     assert not sleep_completed
     assert not inner_scope_completed
+    assert outer_scope.cancelled_caught
+    assert not inner_scope.cancelled_caught
     assert outer_scope.cancel_called
     assert not inner_scope.cancel_called
```
agronholm, line 496: I would prefer this ordering (easier on my eyes at least).
```suggestion
    assert outer_scope.cancelled_caught
    assert outer_scope.cancel_called
    assert not inner_scope.cancelled_caught
    assert not inner_scope.cancel_called
```

tests/test_taskgroups.py
```diff
@@ -478,6 +514,8 @@ async def cancel_when_ready():
 
     assert inner_sleep_completed
     assert not outer_sleep_completed
+    assert tg.cancel_scope.cancelled_caught
+    assert not inner_scope.cancelled_caught
     assert tg.cancel_scope.cancel_called
     assert not inner_scope.cancel_called
```
agronholm, line 520:
```suggestion
    assert tg.cancel_scope.cancelled_caught
    assert tg.cancel_scope.cancel_called
    assert not inner_scope.cancelled_caught
    assert not inner_scope.cancel_called
```

tests/test_taskgroups.py
```diff
@@ -884,3 +936,15 @@ async def exit_scope(scope):
     with pytest.raises(RuntimeError):
         async with create_task_group() as tg:
             tg.start_soon(exit_scope, scope)
+
+
+async def test_raised():
+    class SomeExc(Exception):
```
agronholm, line 942: You don't really need a custom exception here, do you?

tests/test_taskgroups.py
agronholm, line 624: Move this to just before the assertions (for consistency with other code).

tests/test_taskgroups.py
```diff
@@ -582,7 +621,11 @@ async def child(fail):
 
 
 async def test_cancel_cascade():
+    tg2 = None
+
     async def do_something():
+        nonlocal tg2
+
```
agronholm, line 628:
```suggestion
        nonlocal tg2
```

agronholm (changes requested):
Some questions, and code style nits.

.gitignore
johnzeringue: Okay I removed this line

tests/test_taskgroups.py
johnzeringue: This is the only test that fails if you don't account for native cancellation in the asyncio `cancelled_caught`. It's the test you requested here: https://github.com/agronholm/anyio/pull/267#discussion_r619647657

tests/test_taskgroups.py
johnzeringue: I think I added some assertions to the test before removing them. I've removed this line

tests/test_taskgroups.py
johnzeringue: Done

tests/test_taskgroups.py
johnzeringue: Done

tests/test_taskgroups.py
johnzeringue: Changed to a `ValueError`

tests/test_taskgroups.py
johnzeringue: Done

johnzeringue:
Okay I think I've replied to or fixed everything

tests/test_taskgroups.py
agronholm: I just don't see why it has to be this complicated. Isn't starting a single task, natively cancelling it and then checking the status of its task scope enough? Why does the test check if it has started or finished?

tests/test_taskgroups.py
```diff
@@ -925,3 +973,12 @@ async def taskfunc(*, task_status):
             tg.start_soon(sleep, 5)
             await wait_all_tasks_blocked()
             task.cancel('blah')
+
+
+async def test_raised():
+    with pytest.raises(ValueError):
```
agronholm, line 979: What does this test test for that wouldn't be caught by previous tests?

agronholm (changes requested):
Some comments, and pointed out a problematic logic branch.

src/anyio/_backends/_asyncio.py
```diff
@@ -333,6 +336,8 @@ def __exit__(self, exc_type: Optional[Type[BaseException]], exc_val: Optional[Ba
                 elif not self._parent_cancelled():
                     # This scope was directly cancelled
                     return True
+        elif self._cancel_called:
+            self._cancelled_caught = True
```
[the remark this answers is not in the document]
agronholm: I have tried to do the same refactoring quite a few times and it always ended up in failure. Let's see if your effort will fare better.

tests/test_taskgroups.py
johnzeringue: Removed

tests/test_taskgroups.py
johnzeringue: I had copied the code above. I've now simplified it.

tests/test_taskgroups.py
```diff
@@ -583,16 +620,26 @@ async def child(fail):
...
         await wait_all_tasks_blocked()
         tg.cancel_scope.cancel()
 
+        assert not tg.cancel_scope.cancelled_caught
+        assert tg.cancel_scope.cancel_called
+
+    assert tg.cancel_scope.cancelled_caught
+    assert tg.cancel_scope.cancel_called
+    assert not tg2.cancel_scope.cancelled_caught
+    assert tg2.cancel_scope.cancel_called
```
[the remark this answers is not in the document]
johnzeringue: Moved the new logic to a new test.

src/anyio/_backends/_asyncio.py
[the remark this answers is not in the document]
johnzeringue: Okay, I pushed an attempt at that refactoring. It wasn't too terrible, so I'm afraid I might have missed something 🙈

agronholm:
Do the test suites pass for you locally?

johnzeringue:
Tests do pass locally, although until a final tweak, I was seeing issues with `test_shielded_cancel_sleep_time` being flakey locally.

agronholm:
This PR includes changes I'm not comfortable having in the (imminent) v3.1 release, so I'm deferring this to v3.2, minimum.

PR #270, opened 2021-04-29
Add example and docs link to the README

lordmauve:
I believe it's good practice to include a docs link in the README for users who arrive at the GitHub page and want to understand how to proceed.

I also added an example as a quick expression of the value proposition. This has pushed the detailed features list down the page a little but I think that is acceptable; if users arrive with no idea about what AnyIO does need to understand the value proposition before diving into the full features list.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -19,35 +19,10 @@ Applications and libraries written against AnyIO's API will run unmodified on ei
 trio_. AnyIO can also be adopted into a library or application incrementally – bit by bit, no full
 refactoring necessary. It will blend in with native libraries of your chosen backend.
 
-A small example
----------------
-
-Write an async function using ``anyio.sleep()``::
-
-  import anyio
-
-  async def main():
-      print('Hello...')
-      await anyio.sleep(2)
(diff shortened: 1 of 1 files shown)


agronholm:
There is a documentation link at the top, but I suppose there is no harm in adding more explicit link. But I'm not sure about the given example. It doesn't really demonstrate any added value over asyncio/trio. If such an example was to be added, I think it should come after the feature list.

lordmauve:
Yes, it's hard to write a useful example that is brief enough to give a useful taste. The example I added was condensed from https://anyio.readthedocs.io/en/stable/basics.html

Shall I just delete the example?

agronholm:
> Yes, it's hard to write a useful example that is brief enough for a README. The example I added was condensed from https://anyio.readthedocs.io/en/stable/basics.html
> 
> Shall I just delete the example?

Yeah, I can at least merge the rest.

agronholm (approved):

agronholm:
Thanks.

PR #273, opened 2021-04-30
Handle partial asyncio future callback

### LINKED ISSUE #272: Handle partial asyncio future callbacks
As discussed on gitter,

```
    def find_root_task() -> asyncio.Task:
        try:
            root_task = _root_task.get()
        except LookupError:
            for task in all_tasks():
                if task._callbacks:
                    for cb in _get_task_callbacks(task):
>                       if cb is _run_until_complete_cb or cb.__module__ == 'uvloop.loop':
E                       AttributeError: 'functools.partial' object has no attribute '__module__'
```

Full trace: https://pastebin.com/hhqMvhpT, It happens roughly 10% of the time I run my tests.

Triggered by: https://github.com/vinissimus/async-asgi-testclient/blob/f3243ea332b0a91f3990676aa850fdf68676ae93/async_asgi_testclient/utils.py#L75

```python
def create_monitored_task(coro, send):
    future = asyncio.ensure_future(coro)
    future.add_done_callback(partial(_callback, send))
    return future
```

Anyio needs to handle the case where future callbacks are partial objects, and probably any kind of callable.

uSpike:
If using the asyncio backend where the root task cannot be determined,
better handle partial functions and other callable objects used as
future callbacks.

Fixed #272

### DIFF
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -141,7 +141,8 @@ def find_root_task() -> asyncio.Task:
         for task in all_tasks():
             if task._callbacks:
                 for cb in _get_task_callbacks(task):
-                    if cb is _run_until_complete_cb or cb.__module__ == 'uvloop.loop':
+                    if (cb is _run_until_complete_cb
+                            or getattr(cb, '__module__', None) == 'uvloop.loop'):
                         _root_task.set(task)
                         return task
 
--- a/tests/test_to_thread.py
+++ b/tests/test_to_thread.py
@@ -1,4 +1,5 @@
 import asyncio
+from functools import partial
 import sys
 import threading
 import time
@@ -148,3 +149,21 @@ def test_asyncio_no_root_task(asyncio_event_loop):
         if t.name == 'AnyIO worker thread':
             t.join(2)
             assert not t.is_alive()
+
+
+def test_asyncio_future_callback_partial(asyncio_event_loop):
+    """
+    Regression test for #272.
+
+    Ensures that futures with partial callbacks are handled correctly when the root task
+    cannot be determined.
+    """
+    def func(future):
+        pass
+
+    async def sleep_sync():
+        return await to_thread.run_sync(time.sleep, 0)
+
+    task = asyncio_event_loop.create_task(sleep_sync())
+    task.add_done_callback(partial(func))
+    asyncio_event_loop.run_until_complete(task)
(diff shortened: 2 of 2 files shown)


src/anyio/_backends/_asyncio.py
```diff
@@ -141,7 +141,9 @@ def find_root_task() -> asyncio.Task:
         for task in all_tasks():
             if task._callbacks:
                 for cb in _get_task_callbacks(task):
-                    if cb is _run_until_complete_cb or cb.__module__ == 'uvloop.loop':
+                    if isinstance(cb, partial):
+                        cb = cb.func
```
graingert, line 145: we know that `_run_until_complete_cb` won't be used with partial, and 'uvloop.loop' doesn't use it either

uSpike: Good point, I updated to `continue` the for loop if it's a partial.

src/anyio/_backends/_asyncio.py
```diff
@@ -141,7 +141,9 @@ def find_root_task() -> asyncio.Task:
         for task in all_tasks():
             if task._callbacks:
                 for cb in _get_task_callbacks(task):
-                    if cb is _run_until_complete_cb or cb.__module__ == 'uvloop.loop':
+                    if isinstance(cb, partial):
+                        continue
```
graingert, line 145:
```suggestion
```

I think the getattr is all we need here, thoughts @agronholm ?

uSpike: I pushed changes to only use the `getattr` since that seems cleaner to me.

tests/test_to_thread.py
```diff
@@ -148,3 +149,21 @@ async def run_in_thread():
...
+    """
+    def func(future):
+        pass
+
+    async def sleep_sync():
+        return await to_thread.run_sync(time.sleep, 0)
+
+    task = asyncio_event_loop.create_task(sleep_sync())
+    task.add_done_callback(partial(func))
+    asyncio_event_loop.run_until_complete(task)
```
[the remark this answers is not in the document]
uSpike: I didn't need that to trigger the error when I tested it 10 times in a row.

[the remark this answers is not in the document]
agronholm: Yeah it's not needed here.

agronholm:
You need to fix the import ordering. I've added an option to `isort` that should help you see the problem if you don't want to run isort yourself.

uSpike:
@agronholm fixed the import ordering

agronholm (approved):

agronholm:
Thanks!

PR #275, opened 2021-05-02
Added the sleep_until() and sleep_forever() functions

### LINKED ISSUE #271: add sleep_until()
Often it's less code to use `sleep_until()` rather than `sleep()`, since it handles times in the past.

example:
```python3
elapsed = trio.current_time() - t_start
await trio.sleep(max(0, period - elapsed))
```
vs.
```python3
await trio.sleep_until(t_start + period)
```

agronholm:
Closes #271.


src/anyio/__init__.py
```diff
@@ -1,24 +1,26 @@
...
+           'run_process', 'open_process', 'create_lock', 'CapacityLimiter',
+           'CapacityLimiterStatistics', 'Condition', 'ConditionStatistics', 'Event',
+           'EventStatistics', 'Lock', 'LockStatistics', 'Semaphore', 'SemaphoreStatistics',
+           'create_condition', 'create_event', 'create_semaphore', 'create_capacity_limiter',
+           'open_cancel_scope', 'fail_after', 'move_on_after', 'current_effective_deadline',
+           'TASK_STATUS_IGNORED', 'CancelScope', 'create_task_group', 'TaskInfo',
+           'get_current_task', 'get_running_tasks', 'wait_all_tasks_blocked',
+           'run_sync_in_worker_thread', 'run_async_from_thread', 'run_sync_from_thread',
+           'current_default_worker_thread_limiter', 'create_blocking_portal',
+           'start_blocking_portal', 'typed_attribute', 'TypedAttributeSet',
```
belm0, line 18: perhaps one entry per line is better to minimize changed lines and conflicts

belm0 (approved):

belm0:
as seen on Windows 😢 :

```
tests\test_eventloop.py:11: in test_sleep_until
    assert current_time() >= deadline
E   assert 211.281 >= 211.28400000000002
E    +  where 211.281 = current_time()
```

agronholm:
Is the asyncio `sleep` function defective, or what could possibly cause this?

belm0:
> Is the asyncio `sleep` function defective, or what could possibly cause this?

I think it's valid for sleep to be imprecise in both "too late" and "too early" directions.

Tests using wall time are easily prone to mistakes and flakey behavior.  That's why it's important to have some support like Trio's `autojump_clock`.  (In my experience it's been extremely useful and makes testing much easier.)

src/anyio/__init__.py
agronholm: I've made this change in master. Thanks for the suggestion.

agronholm:
> > Is the asyncio `sleep` function defective, or what could possibly cause this?
> 
> I think it's valid for sleep to be imprecise in both "too late" and "too early" directions.
> 
> Tests using wall time are easily prone to mistakes and flakey behavior. That's why it's important to have some support like Trio's `autojump_clock`. (In my experience it's been extremely useful and makes testing much easier.)

It's not using wall time though, but `time.monotonic()` which AFAIK relies on the high performance timer of the host machine.

agronholm:
> It's best not to have this test depend on wall time. In the absence of autojump_clock, the way to do that would be to monkeypatch anyio.sleep() with a mock and ensure that sleep_until() calls it with the expected duration.

I was hoping to avoid that, but in the light of the test failures (why does it only fail on Windows?), it seems I have to go this route.

tests/test_eventloop.py
```diff
@@ -0,0 +1,42 @@
...
+    mocker.patch('anyio._core._eventloop.current_time', return_value=fake_current_time)
+    return mocker.patch('anyio._core._eventloop.sleep', AsyncMock())
+
+
+async def test_sleep_until(fake_sleep):
+    sleep_time = 500.102352
+    deadline = fake_current_time + sleep_time
+    mock = AsyncMock()
+    await mock()
+    mock.assert_called_once()
```
belm0, line 28: it looks like some scratch code?

tests/test_eventloop.py
```diff
@@ -0,0 +1,42 @@
...
+    await sleep_until(deadline)
+    fake_sleep.assert_called_once_with(pytest.approx(sleep_time))
```
belm0, line 30: I'd expect it to work with the exact value

tests/test_eventloop.py
```diff
@@ -0,0 +1,42 @@
...
+
+@pytest.fixture
+def fake_sleep(mocker):
+    mocker.patch('anyio._core._eventloop.current_time', return_value=fake_current_time)
+    return mocker.patch('anyio._core._eventloop.sleep', AsyncMock())
+
+
+async def test_sleep_until(fake_sleep):
+    sleep_time = 500.102352
+    deadline = fake_current_time + sleep_time
```
belm0, line 25: it seems more suitable for this test to define `deadline`, and derive `sleep_time` from that (which doesn't need to be named)

```python3
deadline = fake_current_time + 500.102352
...
fake_sleep.assert_called_once_with(deadline - fake_current_time)
```

tests/test_eventloop.py
```diff
@@ -0,0 +1,42 @@
...
+    mock = AsyncMock()
+    await mock()
+    mock.assert_called_once()
+    await sleep_until(deadline)
+    fake_sleep.assert_called_once_with(pytest.approx(sleep_time))
+
+
+async def test_sleep_until_in_past(fake_sleep):
+    sleep_time = 500.102352
+    deadline = fake_current_time - sleep_time
```
belm0, line 35: same, no need for explicit `sleep_time`
```python
deadline = fake_current_time - 500.102352
```

belm0 (commented):
👍

tests/test_eventloop.py
agronholm: It didn't.

tests/test_eventloop.py
agronholm: Oops, removing.

tests/test_eventloop.py
agronholm: After the other change you suggested, it no longer needs `approx()`.

belm0 (approved):

PR #276, opened 2021-05-03
simplify CancelScope._deliver_cancellation()

belm0:


agronholm:
I'm trying really hard to remember why I didn't cancel tasks right away, but most likely the code was ultimately refactored in such a way that collecting them for later cancellation is not necessary anymore.

agronholm (approved):

agronholm:
Thanks.

PR #277, opened 2021-05-03
allow mutation of CancelScope.shield

### LINKED ISSUE #274: make CancelScope.shield mutable
it should be possible to change the `shield` property of a CancelScope instance

belm0:
Fixes #274

TODO: test coverage


tests/test_taskgroups.py
```diff
@@ -542,6 +542,28 @@ async def cancel_when_ready():
...
+            await wait_all_tasks_blocked()
+            scope.shield = True
+            event.set()
+            await sleep(.1)
+            completed = True
+            scope.shield = False
+            await sleep(0)
+            assert False
+
+    assert completed
```
agronholm, line 564: A more concise test for this. Passes on trio.
```suggestion
async def test_shielding_mutate():
    async def task(*, task_status):
        with CancelScope(shield=True) as scope:
            task_status.started()
            await sleep(.1)
            scope.shield = False
            await sleep(1)
            pytest.fail('Execution should not reach this point')

    async with create_task_group() as tg:
        await tg.start(task)
        tg.cancel_scope.cancel()
```

src/anyio/_backends/_asyncio.py
```diff
@@ -416,6 +416,14 @@ def cancel_called(self) -> bool:
     def shield(self) -> bool:
         return self._shield
 
+    @shield.setter
+    def shield(self, value: bool) -> None:
+        if self._shield != value:
+            self._shield = value
+            # start cancellation effort if necessary when shield is removed
+            if not value and self._cancel_called and self._cancel_handle is None:
```
agronholm, line 424: The logic here is wrong because we want to restart the relevant cancellation delivery effort in a _parent_ cancel scope (may not be the immediate parent). Restarting it in the _current_ scope is pointless because shielding doesn't protect from direct cancellation anyway.

tests/test_taskgroups.py
belm0: The coverage is not quite equivalent, as the original code was also confirming that a shield transition to True from the body of the cancel scope is honored (and to make extra sure, not immediately on entrance of the scope).

I used Event to minimize reliance on wall time.

belm0: I've adopted your solution but added use of fail_after() to avoid a long sleep in the normal case

agronholm: These changes completely defeat the purpose of the test. That `sleep(1)` was supposed to be cancelled due to `shield=False`. Now the task has no more checkpoints where it could be cancelled. Was that the long sleep you referred to?

belm0: > Now the task has no more checkpoints where it could be cancelled.

I believe the test is correct.  See comments below.

```python
  async def task(task_status):
        with CancelScope() as scope:
            await sleep(.1)
            scope.shield = True
            task_status.started()
            await sleep(.1)  # Parent's start() call returns, at which point it issues cancel().
                             # This sleep() completes because we're shielded.
            scope.shield = False
            done.set()

    with fail_after(1):
            await tg.start(task)
            tg.cancel_scope.cancel()
            await done.wait()
```

belm0: I see-- there was a bug that done.wait() would not happen because the tg was cancelled.

Fixed that and confirmed that the test fails if `scope.shield = True` is commented out.

src/anyio/_backends/_asyncio.py
```diff
@@ -420,6 +425,14 @@ def cancel_called(self) -> bool:
     def shield(self) -> bool:
         return self._shield
 
+    @shield.setter
+    def shield(self, value: bool) -> None:
+        if self._shield != value:
+            self._shield = value
+            # start cancellation effort in nearest cancelled parent when shield is removed
```
agronholm, line 432: No need to duplicate the docstring here.
```suggestion
```

src/anyio/_backends/_asyncio.py
```diff
@@ -376,6 +366,21 @@ def _deliver_cancellation(self) -> None:
         else:
             self._cancel_handle = None
 
+    def _deliver_cancellation_to_parent(self) -> None:
+        """Start cancellation effort in the nearest directly cancelled parent scope"""
+
```
agronholm, line 371: To match the code style elsewhere.
```suggestion
```

tests/test_taskgroups.py
```diff
@@ -542,6 +542,26 @@ async def cancel_when_ready():
...
 
+async def test_shielding_mutate():
+    done = False
+
+    async def task(task_status):
+        nonlocal done
+        with CancelScope() as scope:
+            await sleep(.1)
+            scope.shield = True
+            task_status.started()
```
agronholm, line 553: Since cancellation is only attempted after this line, why not create the cancel scope with `shield=True`?

tests/test_taskgroups.py
```diff
@@ -542,6 +542,26 @@ async def cancel_when_ready():
...
+            await sleep(.1)
+            scope.shield = False
+            done = True
```
agronholm, line 556: The point is to check if `scope.shield=False` cancels the task. If the task exits right after setting `shield` to `False`, nothing is tested.

agronholm (changes requested):
Code nits, and a request to fix the test.

tests/test_taskgroups.py
belm0: My intention is to have coverage for the False -> True transition while inside the scope.  Even though we know this case isn't significant for the implementation, ideally the test shouldn't assume that.

tests/test_taskgroups.py
belm0: that's right-- I've reverted to your suggested code, thanks for your patience

tests/test_taskgroups.py
belm0: My intention is to have coverage for the False -> True transition while inside the scope.  Even though we know this case isn't significant for the implementation, ideally the test shouldn't assume that.

agronholm (changes requested):
Almost there!

tests/test_taskgroups.py
```diff
@@ -542,6 +542,23 @@ async def cancel_when_ready():
...
 
 
+async def test_shielding_mutate():
+
+    async def task(task_status):
+        with CancelScope() as scope:
+            await sleep(.1)
+            scope.shield = True
+            task_status.started()
+            await sleep(.1)
```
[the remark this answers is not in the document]
belm0: thank you-- I went with a `completed` flag, representing the work that a real shield would encompass.  (Catching BaseException also seems less than ideal.)

tests/test_taskgroups.py
```diff
@@ -542,6 +542,23 @@ async def cancel_when_ready():
     assert not sleep_completed
 
 
+async def test_shielding_mutate():
+
+    async def task(task_status):
+        with CancelScope() as scope:
+            await sleep(.1)
```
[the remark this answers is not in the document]
belm0: the intent of the sleep is to have a checkpoint before the shield False -> True transition.  Otherwise the intended distinction with CancelScope(shield=True) is diminished.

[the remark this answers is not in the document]
agronholm: Not really – it does test that the setter works as intended when the flag is flipped. Since the scope is only cancelled after `task_status.started()`, the checkpoint doesn't do anything useful.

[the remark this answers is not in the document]
belm0: I agree that the checkpoint doesn't do anything useful given our knowledge of the current implementation.  But if I were to write a test to validate a range of known and unknown implementations, then enabling the shield some time after the cancel scope is entered seems more likely to shake out bugs.

agronholm:
Thank you!

PR #280, opened 2021-05-05
fix typing on CancelScope.__enter__

belm0:
otherwise type check and autocomplete don't work for the typical use

Screen Shot 2021-05-05 at 7 18 50 PM

### DIFF
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -245,7 +245,7 @@ sleep = asyncio.sleep
 CancelledError = asyncio.CancelledError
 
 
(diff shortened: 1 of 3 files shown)


belm0:
I'm not sure how T is supposed to get bound to a type, since `DeprecatedAsyncContextManager` doesn't derive from `Generic`.

```python3
class DeprecatedAsyncContextManager(metaclass=ABCMeta):
    @abstractmethod
    def __enter__(self) -> T:
        pass
```

agronholm (approved):
LGTM.

agronholm:
Thanks!

Issue #282, opened 2021-05-06
Is @asynccontextmanager exempt from the "do not yield in an async generator" rule?


frederikaalund:
Sorry if this is the wrong place to post a question like this. Feel free to redirect me to the proper channels (if they exist).

---

In the [anyio docs](https://anyio.readthedocs.io/en/stable/cancellation.html#avoiding-cancel-scope-stack-corruption) I read that:

> However, in certain situations, cancel scope stack corruption might still occur:
>  * [...]
>  * Yielding in an async generator while enclosed in a cancel scope

Furthermore:

> Therefore, do NOT do this:
>
>     async def some_generator():
>     async with create_task_group() as tg:
>         tg.start_soon(foo)
>         yield
>

I want to do *exactly* that but within an `@asynccontextmanager`. Is this okay? I know that Trio has a [similar rule with an exemption for `@asynccontextmanager`](https://trio.readthedocs.io/en/stable/reference-core.html#cancel-scopes-and-nurseries).

graingert:
@frederikaalund yes in an `@asynccontextmanager` is fine

frederikaalund:
Great, thanks for the clarification. :+1:

graingert:
> Sorry if this is the wrong place to post a question like this. Feel free to redirect me to the proper channels (if they exist).

@frederikaalund https://gitter.im/python-trio/AnyIO I think is the preferred channel for questions

PR #283, opened 2021-05-06
document the @asynccontextmanager do-not-yield exemption

graingert:


docs/cancellation.rst
```diff
@@ -153,3 +153,19 @@ The problem with this code is that it violates structural concurrency: what happ
 task raises an exception? The host task would be cancelled as a result, but the host task might be
 long gone by the time that happens. Even if it weren't, any enclosing ``try...except`` in the
 generator would not be triggered. In other words: it's a Really Bad Idea to do this!
+
+Async generators decorated with ``@asynccontextmanager`` to serve as the template for an async context
```
graingert, line 157: lifted from trio https://trio.readthedocs.io/en/stable/reference-core.html#cancel-scopes-and-nurseries

graingert:
related to #282

frederikaalund:
Glad to see this as part of the documentation. :+1: 

I suggest that we add an "accepted use" example as well. Something like:

> You may do this:
>
>     @asynccontextmanager
>     async def some_cm():
>         async with create_task_group() as tg:
>             tg.start_soon(foo)
>             yield
>
> and use it like this:
>
>     async with some_cm():
>          ...
>
> but do **NOT** do this:
>
>     async def some_generator():
>         async with some_cm():
>             yield

agronholm:
I would likely rewrite that explanation. I don't think it really explains well enough _why_ async context managers are okay in this regard (and of course, when using `__aenter__()` and `__aexit__()` explicitly, all bets are off with those too).

agronholm:
I have opened an alternative PR as #294.

agronholm:
Closing in favor of the merged PR #294.

Issue #291, opened 2021-05-14
Support keyword arguments in callbacks


alexdelorenzo:
When supplying sync functions and coroutine functions as callbacks, `anyio` should accept and supply keyword arguments when calling them.

The following functions accept and supply `*args` to callbacks, but not keyword arguments:
 - `anyio.to_thread.run_sync()`
 - `anyio.to_process.run_sync()`
 - `anyio.from_thread.run_sync()`

agronholm:
The relevant trio discussion is [here](https://github.com/python-trio/trio/issues/470). I much prefer solving this in concert with trio.

PR #292, opened 2021-05-14
Support keyword arguments in callbacks

alexdelorenzo:
This is a fix for https://github.com/agronholm/anyio/issues/291.

The following functions accept and supply `*args` and `**kwargs` to callbacks:
 - `anyio.to_thread.run_sync()`
 - `anyio.to_process.run_sync()`
 - `anyio.from_thread.run_sync()`
 - `anyio.from_thread.run()`

### DIFF
--- a/src/anyio/from_thread.py
+++ b/src/anyio/from_thread.py
@@ -1,4 +1,5 @@
 import threading
+from functools import partial
 from asyncio import iscoroutine
 from concurrent.futures import FIRST_COMPLETED, Future, ThreadPoolExecutor, wait
 from contextlib import AbstractContextManager, contextmanager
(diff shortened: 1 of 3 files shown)


agronholm:
There is a reason neither asyncio or trio have added this: 1) a keyword argument named `func` cannot be used (and this is not obvious), 2) we could never add **any** keyword arguments, like options, to these functions. The trio debate can be found here: https://github.com/python-trio/trio/issues/470
There are various solutions to this, but for now, `functools.partial()` has been deemed "good enough". Personally I am favoring a `kwargs` argument (without the `**`). But this PR is clearly a no-go.

alexdelorenzo:
Thanks for taking the time to explain the situation behind the lack of kwargs support. I wouldn't have ever guessed that there was a years-long debate behind it. Thanks for the link to the trio issue, as well.

PR #293, opened 2021-05-16
add @wrapped_async_generator

belm0:
TODO: add to documentation

cc: @smurfix

### DIFF
--- a/src/anyio/__init__.py
+++ b/src/anyio/__init__.py
@@ -61,6 +61,7 @@ __all__ = (
     'get_current_task',
     'get_running_tasks',
     'wait_all_tasks_blocked',
+    'wrapped_async_generator',
     'run_sync_in_worker_thread',
     'run_async_from_thread',
     'run_sync_from_thread',
@@ -81,6 +82,7 @@ from ._core._exceptions import (
     DelimiterNotFound, EndOfStream, ExceptionGroup, IncompleteRead, TypedAttributeLookupError,
     WouldBlock)
 from ._core._fileio import AsyncFile, open_file
+from ._core._generators import wrapped_async_generator
 from ._core._resources import aclose_forcefully
 from ._core._signals import open_signal_receiver
 from ._core._sockets import (
--- a/src/anyio/_core/_generators.py
+++ b/src/anyio/_core/_generators.py
@@ -1,0 +1,85 @@
+import functools
+import sys
+from contextlib import asynccontextmanager
+
+from ._exceptions import BrokenResourceError
+from ._streams import create_memory_object_stream
+from ._tasks import create_task_group
+
+
+class _aclosing:
+    def __init__(self, aiter):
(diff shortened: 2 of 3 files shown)


agronholm:
This still has a task group straddling a `yield` which invalidates it in situations where cancel scope stack corruption can happen (e.g. pytest fixtures). And in other situations it's not necessary. So, please elaborate?

belm0:
> This still has a task group straddling a `yield` which invalidates it in situations where cancel scope stack corruption can happen (e.g. pytest fixtures). And in other situations it's not necessary. So, please elaborate?

I'm not an expert on this code (it's based on a prototype by Joshua Oreman), but I think the answer is this:

> This is still technically suspending an async generator while it has nurseries/cancel scopes open, but the task that's iterating the async generator doesn't exit or enter any nurseries or cancel scopes while the async generator is suspended, and the exception forwarding effectively puts send_channel.send() into the same context as the async generator's yield.

https://github.com/python-trio/trio/issues/638#issuecomment-431954073

This wrapper absolutely has real world use cases, and the application at my workplace relies on it.

agronholm:
> This wrapper absolutely has real world use cases, and the application at my workplace relies on it.

I would be interested in seeing this use case if you don't mind. If it makes a strong case then I will think about merging this PR.

belm0:
> > This wrapper absolutely has real world use cases, and the application at my workplace relies on it.
> 
> I would be interested in seeing this use case if you don't mind. If it makes a strong case then I will think about merging this PR.

The problem and need is clearly stated in the Trio documentation.

https://trio.readthedocs.io/en/stable/reference-core.html#cancel-scopes-and-nurseries

agronholm:
I am quite aware of the problem (as I am currently writing an expansion of the relevant AnyIO documentation) but I am interested in whatever **real world uses cases** this PR solves, and whether other workarounds could be better.

belm0:
It's not hard to imagine cases where you want to use a nursery on cancel scope within an async generator, and people are hitting this cases in real life (we did and Michael Elsdörfer who filed the trio bug did).

The case in our codebase is basically "async generator needs to call out to an external service for each item, and we want to apply a timeout to the call".

agronholm:
> It's not hard to imagine cases where you want to use a nursery or cancel scope within an async generator, and people are hitting these cases in real life (we did, and Michael Elsdörfer who filed the trio bug did).

Maybe for you it isn't, but I have never written code like that which is why I'm asking to see a real world example. Do I have to ask a third time?

o-fedorov:
Hi. I suppose this code would be a solution for my question asked in StackOwerflow here: https://stackoverflow.com/q/70037576/2069071

The example provided there is quite simplified, in reality there are multiple calls to other APIs that provide paginated responses, so the responses are wrapped in async generators. Though, without task groups the responses are currently provided sequentially.

PR #300, opened 2021-05-25
Add task_status type to docs

uSpike:
This type doesn't seem to be documented other than https://anyio.readthedocs.io/en/stable/api.html?highlight=taskstatus#anyio.abc.TaskStatus.  The `port` argument of `start_some_service` is typed but `task_status` is not.


agronholm (approved):
I was following trio docs on this, but I suppose it's better to have explicit types everywhere.

agronholm:
Thanks!

PR #302, opened 2021-05-30
Remove usage of socket.SocketType

JelleZijlstra:
SocketType is an alias for the private `_socket.socket` class, a superclass of `socket.socket`. It is better to just always use `socket.socket` in types. See https://bugs.python.org/issue44261 and python/typeshed#5545 for some context.

### DIFF
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -1268,8 +1268,10 @@ class UNIXSocketStream(abc.SocketStream):
                 try:
                     # The ignore can be removed after mypy picks up
                     # https://github.com/python/typeshed/pull/5545
-                    self.__raw_socket.sendmsg([message],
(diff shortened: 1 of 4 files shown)


JelleZijlstra:
Looks like this exposes a few new mypy complaints. I will fix those.

JelleZijlstra:
This should be good now. I checked both the current release of mypy and mypy+typeshed master.

A few tests in `test_to_process.py` fail for me locally on 3.7 (but not 3.8 or 3.9), but that seems unrelated to the changes in the PR.

agronholm:
This still fails the linter tests. I suggest that you run `tox -p` locally first.

JelleZijlstra:
Oops, sorry for missing that. Looks like I was on an old version of tox, because it passed for me locally.

agronholm (approved):

agronholm:
Thanks!

PR #303, opened 2021-06-02
Change calling TaskStatus.started() a second time to raise RuntimeError

graingert:


src/anyio/_backends/_asyncio.py
```diff
@@ -519,7 +519,10 @@ def __init__(self, future: asyncio.Future):
         self._future = future
 
     def started(self, value: object = None) -> None:
-        self._future.set_result(value)
+        try:
+            self._future.set_result(value)
+        except asyncio.InvalidStateError:
+            raise RuntimeError("called 'started' twice on the same task status") from None
```
agronholm, line 525:
```suggestion
            raise RuntimeError("Called 'started' twice on the same task status") from None
```

tests/test_taskgroups.py
```diff
@@ -96,6 +96,19 @@ async def taskfunc(*, task_status):
         assert value is None
 
 
+async def test_start_called_twice():
+    async def taskfunc(*, task_status):
+        task_status.started()
+
+        with pytest.raises(RuntimeError,
+                           match="called 'started' twice on the same task status"):
```
agronholm, line 104:
```suggestion
                           match="Called 'started' twice on the same task status"):
```

agronholm (changes requested):
Looks good; just a nit on matching exception message conventions to the rest of AnyIO.

tests/test_taskgroups.py
graingert:
```suggestion
                           match="[Cc]alled 'started' twice on the same task status"):
```

agronholm:
I realized that the reason you had this form in the asyncio version was because the trio exception was the same, and it only occurred to me after I saw your amended test.

agronholm (approved):

agronholm:
Thanks!

PR #309, opened 2021-06-15
avoid unclosed ResourceWarning when failing to connect to unix socket

graingert:


agronholm:
That warning complains about an unclosed TCP socket which would be weird in a UDP-only test. The only possibility is that the socket is left over from some other test.

agronholm:
The PR itself looks good, I just want to get this failure sorted before I merge.

agronholm (approved):

agronholm:
Thanks!

PR #319, opened 2021-06-17
Fixed to_thread.run_sync() using a busy worker thread

### LINKED ISSUE #310: asyncio worker threads are prematurely shut-down on SIGINT
```python
import anyio
import time

def thread():
    print("thread start", time.time())
    time.sleep(5)
    print("thread end", time.time())

async def main():
        tg.start_soon(anyio.to_thread.run_sync, thread)
        try:
            await anyio.sleep_forever()
        finally:
            await anyio.to_thread.run_sync(thread)

```
If i run this and press CTRL-C while sleeping, it does this:

```
thread start 1623770620.0201666
^Cthread end 1623770625.0241108
thread start 1623770625.0242603
thread end 1623770630.028106
```

with `backend="trio"` I get what I expect:
```
thread start 1623770659.6473012
^Cthread start 1623770660.5315428
thread end 1623770664.652128
thread end 1623770665.5361295
```

One workaround for this issue is to use a signal receiver:
```python
import anyio
import time
import signal

def thread():
    print("thread start", time.time())
    time.sleep(5)
    print("thread end", time.time())

async def main():
        tg.start_soon(anyio.to_thread.run_sync, thread)
        try:
            with anyio.open_signal_receiver(signal.SIGINT) as signals:
                async for signum in signals:
                    break
        finally:
            await anyio.to_thread.run_sync(thread)

```

Another workaround (which I'm using) is to spawn threads manually (`threading.Thread(...).start()`) for tear-down.  Both of these are asyncio-specific workarounds, and I think that anyio `to_thread.run_sync` should behave the same if running on asyncio or trio, if possible.

My use case is this is working with something that is synchronous and must use worker threads.  I must call the appropriate `close()` methods on the synchronous API from a thread so as to not block my async application.

agronholm:
Fixes #310.

### DIFF
--- a/.github/workflows/codeqa-test.yml
+++ b/.github/workflows/codeqa-test.yml
@@ -33,12 +33,14 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        python-version: [3.6, 3.7, 3.8, 3.9, pypy3]
+        python-version: [3.6, 3.7, 3.8, 3.9, 3.10.0-beta.2, pypy3]
         exclude:
         - os: macos-latest
           python-version: 3.7
         - os: macos-latest
           python-version: 3.8
+        - os: macos-latest
+          python-version: 3.10.0-beta.2
         - os: macos-latest
           python-version: pypy3
         - os: windows-latest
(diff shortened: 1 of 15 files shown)


agronholm:
This PR still lacks a regression test.

graingert (approved):

PR #321, opened 2021-06-19
Spawn worker threads with daemon=True on asyncio

davidbrochart:
This was added in https://github.com/agronholm/anyio/pull/81/commits/635e25adf8e35b920110cdc0f0c07112c2ade8ff:
https://github.com/agronholm/anyio/blob/635e25adf8e35b920110cdc0f0c07112c2ade8ff/anyio/_backends/_asyncio.py#L413
And removed in #227:
https://github.com/agronholm/anyio/blob/77759c8bc303b70934c423bea2ef0267795b182b/src/anyio/_backends/_asyncio.py#L717

Fixes https://github.com/jupyterlab/jupyterlab/issues/10442

### DIFF
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -704,7 +704,7 @@ class WorkerThread(Thread):
 
     def __init__(self, root_task: asyncio.Task, workers: Set['WorkerThread'],
                  idle_workers: Deque['WorkerThread'],):
-        super().__init__(name='AnyIO worker thread')
+        super().__init__(name='AnyIO worker thread', daemon=True)
         self.root_task = root_task
         self.workers = workers
         self.idle_workers = idle_workers


graingert:
daemon=True isn't supported with subinterpreters and skips cleanup

agronholm:
I removed `daemon=True` once I learned the above. Why does JupyterLab hang on server shutdown? Does it not clean up the tasks it creates?

agronholm:
There is also the matter of issue #323 which I just fixed (and which will require a prompt bugfix release). Could it be related?

graingert:
https://github.com/python-trio/trio/issues/2046

davidbrochart:
> There is also the matter of issue #323 which I just fixed (and which will require a prompt bugfix release). Could it be related?

It doesn't seem to be related, https://github.com/agronholm/anyio/commit/49f666ed5ef2f70efdd11006d4815c0a68cb8863 doesn't fix the JupyterLab hang.

> Why does JupyterLab hang on server shutdown? Does it not clean up the tasks it creates?

The issue is actually in Jupyter Server, which uses Tornado. Shutting down simply stops Tornado's event loop, and waits for each task to finish. But I think it is because of a task in a worker thread which is not thread-safe. I opened https://github.com/jupyter-server/jupyter_server/pull/545 which should fix the issue.

I thought `daemon=True` was removed by mistake, but anyway this doesn't seem to be the root cause of our issue, so it is fine to close this PR. Thanks for your help!

PR #325, opened 2021-06-24
Handle missing trio gracefully

mgorny:
Make it possible to use anyio without actually having to install trio.
This involves modifying get_all_backends() to only return the backends
that are actually present, and teaching the tests to skip trio if it
is not importable.

### DIFF
--- a/src/anyio/_core/_eventloop.py
+++ b/src/anyio/_core/_eventloop.py
@@ -10,8 +10,6 @@ import sniffio
 # This must be updated when new backends are introduced
 from ._compat import DeprecatedAwaitableFloat
 
-BACKENDS = 'asyncio', 'trio'
-
 T_Retval = TypeVar('T_Retval')
 threadlocals = threading.local()
 
@@ -106,8 +104,15 @@ def current_time() -> DeprecatedAwaitableFloat:
 
 
 def get_all_backends() -> Tuple[str, ...]:
-    """Return a tuple of the names of all built-in backends."""
-    return BACKENDS
+    """Return a tuple of the names of all available built-in backends."""
+    backends = ['asyncio']
+    try:
+        import trio
+    except ImportError:
+        pass
+    else:
+        backends.append(trio)
+    return tuple(backends)
 
 
 def get_cancelled_exc_class() -> Type[BaseException]:
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -8,6 +8,11 @@ import trustme
 from _pytest.fixtures import SubRequest
 from trustme import CA
 
+try:
+    import trio
+except ImportError:
+    trio = None
+
 uvloop_marks = []
 uvloop_policy = None
 try:
@@ -30,7 +35,9 @@ pytest_plugins = ['pytester']
                  id='asyncio'),
     pytest.param(('asyncio', {'debug': True, 'policy': uvloop_policy}), marks=uvloop_marks,
                  id='asyncio+uvloop'),
-    pytest.param('trio')
+    pytest.param('trio',
+                 marks=[pytest.mark.skipif(trio is None,
+                                           reason='trio is not available')])
 ])
 def anyio_backend(request: SubRequest) -> Tuple[str, Dict[str, Any]]:
     return request.param
--- a/tests/test_pytest_plugin.py
+++ b/tests/test_pytest_plugin.py
@@ -135,7 +135,7 @@ def test_asyncio(testdir: Testdir) -> None:
     )
 
     result = testdir.runpytest('-v')
-    result.assert_outcomes(passed=2, failed=1, errors=2)
+    result.assert_outcomes(passed=2, failed=1, errors=len(get_all_backends()))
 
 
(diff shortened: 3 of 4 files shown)


agronholm:
Testing is the only thing where missing some of the backends would result in an error, so "possible to use anyio without actually having to install trio" is misleading at best.

The proposed change to `get_all_backends()` feels dubious. I also feel that it violates the contract stated in the docstring, and would present a backwards incompatible behavioral change.

As for the tests skipping trio if it's not importable, that feels wrong to me for a couple reasons:

- For libraries, you would normally want to test on *all* backends, and it should, by default, raise an error if one of the supported backends was not available. Skipping one backend when it's not available could lead to the test suite falsely passing if they forgot to include the right dependencies.
- For applications, it does usually not make sense to test against multiple backends, and you would end up with a huge number of pointless "skipped" markers. As such, [specifying the backend manually](https://anyio.readthedocs.io/en/stable/testing.html#specifying-the-backends-to-run-on) is the correct thing to do.

mgorny:
Well, long story short we aren't packaging trio in Gentoo. We need to package anyio because some random packages need it. By its very definition, anyio makes it possible to use them without packaging trio. So the only piece of the puzzle that's bringing breakage is anyio requiring trio itself.

agronholm:
Use, yes. Test? Not without modifying the test suite. Trio is a *testing* dependency, but not a run-time dependency.
How are you running the test suite?

mgorny:
If I'm suppose to use something, I want to know whether it works first. In order to know that, I need to test it first. I want to test it in the same environment that it will be used. Otherwise, the tests might falsely indicate that things work in ideal environment but break in the real one.

I'm testing via:

    pytest -vv -ra -l -Wdefault -m 'not network' -p no:django

Also, isn't trio implicitly forced by the pytest plugin? This effectively forces it on all packages that use the plugin.

agronholm:
The default configuration tests on all available backends, but a project using the pytest plugin can freely alter this by defining its own `anyio_backend` fixture which will override the default. So no, a trio dependency is not being forced on any project wanting to use AnyIO.

agronholm:
If you really want to test w/o trio, you could patch the `tests/conftest.py` file on your end to exclude it. That seems to be the easiest solution here.

mgorny:
That's insufficient. I have to also prevent pytest plugin from grabbing all backends.

agronholm:
Then patch the pytest plugin to exclude trio. But for the reasons I mentioned above, I don't feel it is a wise decision for this project to adopt such changes upstream.

mgorny:
But patching the pytest plugin implies that Gentoo will now install non-standard version of the plugin. I'd really prefer if we could work out a solution that works for everyone.

agronholm:
How about modifying your pytest command line to skip the trio tests, with `-k "not [trio"`?

njsmith:
(BTW, is there some reason that makes packaging Trio so hard that it's easier to carry patches to other packages? If so it would be nice to know so we can fix it :-))

agronholm:
> (BTW, is there some reason that makes packaging Trio so hard that it's easier to carry patches to other packages? If so it would be nice to know so we can fix it :-))

I think it's just because none of the packages they support have a hard dependency on it.

mgorny:
> How about modifying your pytest command line to skip the trio tests, with `-k "not [trio"`?

Sure but I don't see how to make pytest tests respect that.

mgorny:
> (BTW, is there some reason that makes packaging Trio so hard that it's easier to carry patches to other packages? If so it would be nice to know so we can fix it :-))

I'm literally maintaining hundreds of NIH packages already, so I am naturally reluctant to having to maintain even one more. Python package maintenance costs are very high.

graingert:
```sh
pytest -vv -ra -l -Wdefault -m 'not network' -p no:django -k "not [trio"
```

agronholm:
> Sure but I don't see how to make pytest tests respect that.

Maybe I'm not getting it. Are you talking about Gentoo-packaged, AnyIO dependent Python packages? The same switch will work there too.

mgorny:
Well, thanks for that idea but there are two problems with that:

1. `tests/test_taskgroups.py` imports `trio` in global scope. I could skip the whole file but it feels wrong if only one test needs it. If you agree with changing it, I suppose I could make it parametrized via string, and import the module inside.
2. `tests/test_pytest_plugin.py` doesn't pass the `-k` inside the test, so it still runs trio variants.

agronholm:
> tests/test_taskgroups.py imports trio in global scope. I could skip the whole file but it feels wrong if only one test needs it. If you agree with changing it, I suppose I could make it parametrized via string, and import the module inside.

I could agree to this.

> tests/test_pytest_plugin.py doesn't pass the -k inside the test, so it still runs trio variants.

If pytest exposes that information somewhere, I could pass it on to the "nested" tests. Not sure this is a good idea, but I'm willing to consider it.

agronholm:
I've pushed changes that should make this easier for downstream packagers. The test suite no longer has a hard requirement for the trio import, although you still need to disable some tests in order for the suite to pass.

PR #327, opened 2021-06-26
Added asynchronous Path class

### LINKED ISSUE #296: anyio.Path like trio.Path support
could probably lift the implementation from https://github.com/python-trio/trio/blob/master/trio/_path.py but substitute anyio.to_thread.run

agronholm:
Fixes #296.

### DIFF
--- a/src/anyio/__init__.py
+++ b/src/anyio/__init__.py
@@ -19,6 +19,7 @@ __all__ = (
     'TypedAttributeLookupError',
     'WouldBlock',
     'AsyncFile',
+    'Path',
     'open_file',
     'aclose_forcefully',
     'open_signal_receiver',
@@ -72,6 +73,7 @@ __all__ = (
     'TypedAttributeProvider'
 )
 
+# isort: off
 from ._core._compat import maybe_async, maybe_async_cm
 from ._core._eventloop import (
     current_time, get_all_backends, get_cancelled_exc_class, run, sleep, sleep_forever,
@@ -80,7 +82,6 @@ from ._core._exceptions import (
     BrokenResourceError, BrokenWorkerProcess, BusyResourceError, ClosedResourceError,
     DelimiterNotFound, EndOfStream, ExceptionGroup, IncompleteRead, TypedAttributeLookupError,
     WouldBlock)
-from ._core._fileio import AsyncFile, open_file
 from ._core._resources import aclose_forcefully
 from ._core._signals import open_signal_receiver
 from ._core._sockets import (
@@ -97,6 +98,7 @@ from ._core._tasks import (
     move_on_after, open_cancel_scope)
 from ._core._testing import TaskInfo, get_current_task, get_running_tasks, wait_all_tasks_blocked
 from ._core._typedattr import TypedAttributeProvider, TypedAttributeSet, typed_attribute
+from ._core._fileio import AsyncFile, Path, open_file
 
 # Re-exported here, for backwards compatibility
 # isort: off
--- a/src/anyio/__init__.py
+++ b/src/anyio/__init__.py
@@ -73,7 +73,6 @@ __all__ = (
     'TypedAttributeProvider'
 )
 
-# isort: off
 from ._core._compat import maybe_async, maybe_async_cm
 from ._core._eventloop import (
     current_time, get_all_backends, get_cancelled_exc_class, run, sleep, sleep_forever,
@@ -82,6 +81,7 @@ from ._core._exceptions import (
     BrokenResourceError, BrokenWorkerProcess, BusyResourceError, ClosedResourceError,
     DelimiterNotFound, EndOfStream, ExceptionGroup, IncompleteRead, TypedAttributeLookupError,
     WouldBlock)
+from ._core._fileio import AsyncFile, Path, open_file
 from ._core._resources import aclose_forcefully
 from ._core._signals import open_signal_receiver
 from ._core._sockets import (
@@ -98,7 +98,6 @@ from ._core._tasks import (
     move_on_after, open_cancel_scope)
 from ._core._testing import TaskInfo, get_current_task, get_running_tasks, wait_all_tasks_blocked
 from ._core._typedattr import TypedAttributeProvider, TypedAttributeSet, typed_attribute
-from ._core._fileio import AsyncFile, Path, open_file
 
 # Re-exported here, for backwards compatibility
 # isort: off
--- a/src/anyio/_core/_fileio.py
+++ b/src/anyio/_core/_fileio.py
@@ -162,19 +162,34 @@ class Path:
         return self._path.__hash__()
 
     def __eq__(self, other: object) -> bool:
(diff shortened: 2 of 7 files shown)


src/anyio/_core/_fileio.py
```diff
@@ -114,3 +124,208 @@ async def open_file(file: Union[str, PathLike, int], mode: str = 'r', buffering:
...
+
+        await to_thread.run_sync(self._path.rename, target)
+
+    async def replace(self, target: Union[str, pathlib.PurePath, 'Path']) -> None:
+        if isinstance(target, Path):
+            target = target._path
+
+        await to_thread.run_sync(self._path.replace, target)
+
+    async def resolve(self) -> 'Path':
```
graingert, line 280: resolve has a strict kwarg

agronholm: Yeah, I was gonna add that, just slipped my mind. But thanks for the reminder.

agronholm:
This PR needs a whole bunch of additional tests before it's ready.

agronholm:
What this is still missing is some documentation of which operations require `await`.

src/anyio/__init__.py
```diff
@@ -72,6 +73,7 @@
     'TypedAttributeProvider'
 )
 
+# isort: off
```
graingert, line 76: Do you still need this?

src/anyio/_core/_fileio.py
```diff
@@ -114,3 +125,346 @@ async def open_file(file: Union[str, PathLike, int], mode: str = 'r', buffering:
...
+    __slots__ = '_path'
+
+    def __init__(self, *args: Union[str, PathLike]) -> None:
+        self._path: Final[pathlib.Path] = pathlib.Path(*args)
+
+    def __fspath__(self) -> str:
+        return self._path.__fspath__()
+
+    def __str__(self) -> str:
+        return self._path.__str__()
```
graingert, line 202: These dunders need to be called via the class with `type(v).__str__(v)` or just `str(v)`

src/anyio/_core/_fileio.py
```diff
@@ -114,3 +125,346 @@ async def open_file(file: Union[str, PathLike, int], mode: str = 'r', buffering:
...
+    def with_stem(self, stem: str) -> 'Path':
+        return Path(self._path.with_name(stem + self._path.suffix))
+
+    def with_suffix(self, suffix: str) -> 'Path':
+        return Path(self._path.with_suffix(suffix))
+
+    async def write_bytes(self, data: bytes) -> int:
+        return await to_thread.run_sync(self._path.write_bytes, data)
+
+    async def write_text(self, data: str, encoding: Optional[str] = None,
```
graingert, line 464: Why not to_thread the write_text directly?

src/anyio/_core/_fileio.py
smurfix: I don't understand this comment. `str(x)` or `x.__str__()` or `type(x).__str__(x)` are all equivalent.

graingert: @smurfix they're different:

```python
class Foo:
    def __init__(self) -> None:
        self.__str__ = lambda: "x.__str__()"

    def __str__(self) -> str:
        return "type(x).__str__(x)"

print(f"{str(Foo())=}")  # str(Foo())='type(x).__str__(x)'
print(f"{Foo().__str__()=}")   # Foo().__str__()='x.__str__()'
```

src/anyio/_core/_fileio.py
agronholm: Because on py < 3.10, the `newline` parameter doesn't exist.

src/anyio/_core/_fileio.py
smurfix: Yes I know that `str()` takes a shortcut and doesn't do a lookup on the instance. But which real-world class assigns to `self.__str__`? `pathlib.Path`certainly doesn't.
```
>>> import pathlib as p
>>> x=p.Path("a/b/c")
>>> x
PosixPath('a/b/c')
>>> str(x)
'a/b/c'
>>> x.__str__()
'a/b/c'
>>> 
>>> type(x).__str__(x)
'a/b/c'
>>>
```

agronholm: I'm also not understanding why this needs to be changed. Is there a potential problem with this way of doing things?

src/anyio/__init__.py
agronholm: Apparently not – I've removed it now.

src/anyio/_core/_fileio.py
```diff
@@ -114,3 +125,346 @@ async def open_file(file: Union[str, PathLike, int], mode: str = 'r', buffering:
...
+    * :meth:`~pathlib.Path.write_text`
+
+    Additionally, the following methods return an async iterator yielding :class:`~.Path` objects:
+
+    * :meth:`~pathlib.Path.glob`
+    * :meth:`~pathlib.Path.iterdir`
+    * :meth:`~pathlib.Path.rglob`
+    """
+
+    __slots__ = '_path'
```
graingert, line 193:
```suggestion
    __slots__ = '_path', '__weakref__'
```

graingert (changes requested):

src/anyio/_core/_fileio.py
graingert: Would it be better to do this in one call to to_thread?

src/anyio/_core/_fileio.py
graingert: Ah I forgot pathlib.Path was frozen. It's usually a problem but not in this specific case

src/anyio/_core/_fileio.py
agronholm: The `pathlib.Path` instance is not weak referable. That is why I didn't include this. I wonder if the lack of a `__weakref__` slot is intentional there.

graingert: Oh that's very surprising

graingert: Sounds like a cPython bug that anyio should copy. Might be worth a test

agronholm: Interestingly, on PyPy3 they *are* weakrefable. I don't think there is any harm done adding the weakref slot, so I'll do that.

graingert: Pypy makes everything weakrefable

agronholm: I'm not sure what to do about this, but I don't think we should test that it's *not* weakrefable.

src/anyio/_core/_fileio.py
agronholm: Done

graingert (approved):
Nice

agronholm:
Thanks for the review! As a side note, we really need to get rid of the flakiness of the TLS tests on Windows.

PR #331, opened 2021-06-28
drop async_generator dep

graingert:
inspect.isasyncgenfunction got added in 3.6 not 3.7


agronholm (changes requested):
I think this warrants a changelog entry. Otherwise, LGTM.

agronholm (approved):

PR #334, opened 2021-07-03
make BlockingPortal.stop sync

graingert:


agronholm (changes requested):
If this method is made synchronous, we need to ensure that people dont try to use it from external threads. At the very least, a warning against that should be added to the docstring. The `event.set()` call should raise a clear enough error but I feel that the event loop thread id member set should be moved to after the event set, to be safe.

src/anyio/from_thread.py
```diff
@@ -154,11 +155,13 @@ async def stop(self, cancel_remaining: bool = False) -> None:
             finish before returning
 
         """
-        self._event_loop_thread_id = None
         self._stop_event.set()
+        self._event_loop_thread_id = None
```
agronholm, line 159:
```suggestion
        if self._event_loop_thread_id is None:
            return
        if threading.get_ident() != self._event_loop_thread_id:
            raise RuntimeError('This method must be called from the event loop thread')

        self._event_loop_thread_id = None
        self._stop_event.set()
```

src/anyio/from_thread.py
```diff
@@ -154,11 +155,18 @@ async def stop(self, cancel_remaining: bool = False) -> None:
             finish before returning
 
         """
+        if self._event_loop_thread_id is None:
```
graingert, line 158: ah this is wrong - we need to handle cancel_remaining

PR #336, opened 2021-07-05
Allow re-entry into the event loop using tokens #256

### LINKED ISSUE #256: Allow re-entry into the event loop using tokens
Trio allows using [trio.from_thread.run](https://trio.readthedocs.io/en/stable/reference-core.html#trio.from_thread.run) and [trio.from_thread.run_sync](https://trio.readthedocs.io/en/stable/reference-core.html#trio.from_thread.run_sync) to be called from unrelated threads as long as a token (from trio.low_level.current.token) is provided.

On asyncio this token could be the event loop object itself, or another object that wraps it.

This change would likely allow us to remove the `__new__()` hack from `BlockingPortal`.

graingert:
Fixes #256

### DIFF
--- a/src/anyio/from_thread.py
+++ b/src/anyio/from_thread.py
@@ -35,7 +35,7 @@ def run(func: Callable[..., Coroutine[Any, Any, T_Retval]], *args: object,
         try:
             asynclib = threadlocals.current_async_module
         except AttributeError:
-            raise RuntimeError('This function can only be run from an AnyIO worker thread')
+            raise RuntimeError("This thread wasn't created by an async library, pass kwarg asynclib_token=...")
 
         return asynclib.run_async_from_thread(func, *args)
 
--- a/src/anyio/from_thread.py
+++ b/src/anyio/from_thread.py
@@ -70,6 +70,7 @@ def run_sync(func: Callable[..., T_Retval], *args: object,
             asynclib = threadlocals.current_async_module
         except AttributeError:
             raise RuntimeError('This function can only be run from an AnyIO worker thread')
+
         return asynclib.run_sync_from_thread(func, *args)
 
     if isinstance(asynclib_token, AbstractEventLoop):
--- a/src/anyio/from_thread.py
+++ b/src/anyio/from_thread.py
@@ -36,6 +36,7 @@ def run(func: Callable[..., Coroutine[Any, Any, T_Retval]], *args: object,
             asynclib = threadlocals.current_async_module
         except AttributeError:
             raise RuntimeError('This function can only be run from an AnyIO worker thread')
+
         return asynclib.run_async_from_thread(func, *args)
 
     if isinstance(asynclib_token, AbstractEventLoop):
--- a/src/anyio/from_thread.py
+++ b/src/anyio/from_thread.py
@@ -72,7 +72,10 @@ def run_sync(func: Callable[..., T_Retval], *args: object,
         try:
             asynclib = threadlocals.current_async_module
         except AttributeError:
-            raise RuntimeError('This function can only be run from an AnyIO worker thread')
+            raise RuntimeError(
+                "This thread wasn't created by an async library, "
+                "pass kwarg asynclib_token=..."
+            )
 
         return asynclib.run_sync_from_thread(func, *args)
 
--- a/src/anyio/from_thread.py
+++ b/src/anyio/from_thread.py
@@ -35,7 +35,10 @@ def run(func: Callable[..., Coroutine[Any, Any, T_Retval]], *args: object,
         try:
             asynclib = threadlocals.current_async_module
         except AttributeError:
-            raise RuntimeError("This thread wasn't created by an async library, pass kwarg asynclib_token=...")
+            raise RuntimeError(
+                "This thread wasn't created by an async library, "
+                "pass kwarg asynclib_token=..."
+            )
 
         return asynclib.run_async_from_thread(func, *args)
 
--- a/src/anyio/from_thread.py
+++ b/src/anyio/from_thread.py
@@ -1,5 +1,5 @@
 import threading
-from asyncio import iscoroutine
+from asyncio import AbstractEventLoop, iscoroutine
 from concurrent.futures import FIRST_COMPLETED, Future, ThreadPoolExecutor, wait
 from contextlib import AbstractContextManager, contextmanager
 from types import TracebackType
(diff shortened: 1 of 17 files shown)


src/anyio/from_thread.py
```diff
@@ -114,10 +135,8 @@ def started(self, value: object = None) -> None:
 class BlockingPortal:
     """An object that lets external threads run code in an asynchronous event loop."""
 
-    def __new__(cls) -> 'BlockingPortal':
-        return get_asynclib().BlockingPortal()
-
     def __init__(self) -> None:
+        self.token = current_token()
```
graingert, line 139: @agronholm I think this should be public so I can run


```python
anyio.from_thread.run_sync(whatever, asynclib_token=portal.token)
```

graingert: I can always get it from:

`asynclib_token = portal.call(anyio.current_token)`

agronholm (commented):
I still don't know what to do about the backend specific code, but here are some other comments.

src/anyio/from_thread.py
```diff
@@ -27,12 +31,20 @@ def run(func: Callable[..., Coroutine[Any, Any, T_Retval]], *args: object) -> T_
...
-    try:
-        asynclib = threadlocals.current_async_module
-    except AttributeError:
-        raise RuntimeError('This function can only be run from an AnyIO worker thread')
+    if asynclib_token is None:
+        try:
+            asynclib = threadlocals.current_async_module
+        except AttributeError:
+            raise RuntimeError('This function can only be run from an AnyIO worker thread')
+        return asynclib.run_async_from_thread(func, *args)
```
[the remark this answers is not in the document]
graingert:
```suggestion

        return asynclib.run_async_from_thread(func, *args)
```

src/anyio/from_thread.py
```diff
@@ -51,12 +64,20 @@ def run_sync(func: Callable[..., T_Retval], *args: object) -> T_Retval:
...
-    try:
-        asynclib = threadlocals.current_async_module
-    except AttributeError:
-        raise RuntimeError('This function can only be run from an AnyIO worker thread')
+    if asynclib_token is None:
+        try:
+            asynclib = threadlocals.current_async_module
+        except AttributeError:
+            raise RuntimeError('This function can only be run from an AnyIO worker thread')
+        return asynclib.run_sync_from_thread(func, *args)
```
graingert, line 73:
```suggestion

        return asynclib.run_sync_from_thread(func, *args)
```

src/anyio/from_thread.py
```diff
@@ -27,12 +31,20 @@ def run(func: Callable[..., Coroutine[Any, Any, T_Retval]], *args: object) -> T_
...
     """
-    try:
-        asynclib = threadlocals.current_async_module
-    except AttributeError:
-        raise RuntimeError('This function can only be run from an AnyIO worker thread')
+    if asynclib_token is None:
+        try:
+            asynclib = threadlocals.current_async_module
+        except AttributeError:
+            raise RuntimeError('This function can only be run from an AnyIO worker thread')
```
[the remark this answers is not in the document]
graingert:
```suggestion
            raise RuntimeError("This thread wasn't created by AnyIO, pass kwarg anyio_token=...")
```

based on https://github.com/python-trio/trio/blob/5b0b8388067d4edfdde64482953da4a3761f2f31/trio/_threads.py#L224-L227

src/anyio/lowlevel.py
```diff
@@ -108,8 +109,9 @@ def _current_vars(self) -> Dict[str, T]:
                 # Happens when token isn't weak referable (TrioToken).
                 # This workaround does mean that some memory will leak on Trio until the problem
                 # is fixed on their end.
-                token = _TokenWrapper(token)
-                self._token_wrappers.add(token)
+                wrapper = _TokenWrapper(token)
```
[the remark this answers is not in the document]
graingert: I use `anyio[trio]` and would happily let it upgrade my trio

src/anyio/from_thread.py
```diff
@@ -27,12 +31,21 @@ def run(func: Callable[..., Coroutine[Any, Any, T_Retval]], *args: object) -> T_
...
     """
-    try:
-        asynclib = threadlocals.current_async_module
-    except AttributeError:
-        raise RuntimeError('This function can only be run from an AnyIO worker thread')
+    if asynclib_token is None:
+        try:
+            asynclib = threadlocals.current_async_module
+        except AttributeError:
+            raise RuntimeError("This thread wasn't created by an async library, pass kwarg asynclib_token=...")
```
graingert, line 38:
```suggestion
            raise RuntimeError(
                "This thread wasn't created by an async library, "
                "pass kwarg asynclib_token=..."
            )
```

src/anyio/from_thread.py
```diff
@@ -51,12 +68,21 @@ def run_sync(func: Callable[..., T_Retval], *args: object) -> T_Retval:
...
     """
-    try:
-        asynclib = threadlocals.current_async_module
-    except AttributeError:
-        raise RuntimeError('This function can only be run from an AnyIO worker thread')
+    if asynclib_token is None:
+        try:
+            asynclib = threadlocals.current_async_module
+        except AttributeError:
+            raise RuntimeError('This function can only be run from an AnyIO worker thread')
```
graingert, line 75:
```suggestion
            raise RuntimeError(
                "This thread wasn't created by an async library, "
                "pass kwarg asynclib_token=..."
            )
```

src/anyio/from_thread.py
```diff
@@ -51,12 +65,21 @@ def run_sync(func: Callable[..., T_Retval], *args: object) -> T_Retval:
...
 
-    return asynclib.run_sync_from_thread(func, *args)
+        return asynclib.run_sync_from_thread(func, *args)
+
+    if isinstance(asynclib_token, AbstractEventLoop):
+        return get_asynclib("asyncio").run_sync_from_thread(func, *args, loop=asynclib_token)
+
+    if type(asynclib_token).__module__.startswith("trio."):
+        return get_asynclib("trio").run_sync_from_thread(func, *args, trio_token=asynclib_token)
```
agronholm, line 86: I think that the fundamental problem here is that tokens aren't being treated as opaque AnyIO specific objects here. If we did that, we could embed the backend name in them and not have to write code like this.

graingert: I think it's worth it to be able to accept a trio.lowlevel.current_token() here

agronholm: I don't agree with you here. Just as trio has its own opaque tokens, AnyIO could have its own. The fact that the "token" involves an asyncio event loop object or a trio token is just an implementation detail.

graingert:
Sounds good, thanks!

PR #337, opened 2021-07-06
remove old note about wrap_async_context_manager using multiple tasks

graingert:


graingert:
although I think you'd need to wrap the taskgroup in a contextmanager factory:

```python
@contextlib.asynccontextmanager
async def create_task_group_soon():
        yield tg

with portal.wrap_async_context_manager(create_task_group_soon()) as tg:
    portal.call(tg.start_soon, async_fn)
```

agronholm:
Maybe that entire note needs to be removed instead?

docs/threads.rst
```diff
@@ -178,5 +178,5 @@ managers as a synchronous one::
 .. note:: You cannot use wrapped async context managers in synchronous callbacks inside the event
           loop thread.
 
-.. note:: The ``__aenter__()`` and ``__aexit__()`` methods will be called from different
-          tasks so a task group as the async context manager will not work here.
+.. note:: The ``__aenter__()`` and ``__aexit__()`` methods will be called in the same task
+          so a task group as the async context manager will work here.
```
graingert, line 182:
```suggestion
```

graingert: How about just removing this bit?

agronholm: Wouldn't that be the assumption for new users, since the docs explicitly forbid doing it any other way?

docs/threads.rst
```diff
@@ -178,5 +178,4 @@ managers as a synchronous one::
 .. note:: You cannot use wrapped async context managers in synchronous callbacks inside the event
           loop thread.
 
-.. note:: The ``__aenter__()`` and ``__aexit__()`` methods will be called from different
-          tasks so a task group as the async context manager will not work here.
+.. note:: The ``__aenter__()`` and ``__aexit__()`` methods will be called in the same task
```
graingert, line 181:
```suggestion
```

agronholm (approved):

agronholm:
Thanks.

PR #338, opened 2021-07-06
Fixed generics in AsyncFile

### LINKED ISSUE #306: AsyncFile type incorrect
currently some methods take/return `str | bytes` and others only take `bytes`

Fix the type annotations of `AsyncFile.__aiter__`, `readline`, `write` to also accept/return `str`
  and also fixed `AsyncFile.writelines` to take an `Iterable[str|bytes]` rather than `bytes`.

It should probably be an `AsyncFile[AnyStr]` or a full copy of the IOBase hierarchy, but async

agronholm:
Fixes #306.

### DIFF
--- a/src/anyio/_core/_fileio.py
+++ b/src/anyio/_core/_fileio.py
@@ -7,8 +7,8 @@ from dataclasses import dataclass
 from functools import partial
 from os import PathLike
 from typing import (
-    IO, Any, AsyncIterator, Callable, Generic, Iterator, List, Optional, Sequence, Tuple, TypeVar,
-    Union, cast, overload)
+    IO, Any, AnyStr, AsyncIterator, Callable, Generic, Iterable, Iterator, List, Optional,
+    Sequence, Tuple, Union, cast, overload)
 
 from .. import to_thread
 from ..abc import AsyncResource
@@ -17,8 +17,6 @@ if sys.version_info >= (3, 8):
     from typing import Final, Literal
 else:
     from typing_extensions import Final, Literal
-
-T_IO = TypeVar("T_IO", bytes, str)
 
 # Copied from typeshed
 ReadableBuffer = Union[bytes, bytearray, memoryview, array.array, mmap.mmap]
@@ -102,7 +100,7 @@ OpenBinaryModeReading = Literal["rb", "br", "rbU", "rUb", "Urb", "brU", "bUr", "
 OpenBinaryMode = Union[OpenBinaryModeUpdating, OpenBinaryModeReading, OpenBinaryModeWriting]
 
 
-class AsyncFile(AsyncResource, Generic[T_IO]):
+class AsyncFile(AsyncResource, Generic[AnyStr]):
     """
     An asynchronous file object.
 
@@ -134,18 +132,18 @@ class AsyncFile(AsyncResource, Generic[T_IO]):
                 print(line)
     """
 
-    def __init__(self, fp: IO[T_IO]) -> None:
+    def __init__(self, fp: IO[AnyStr]) -> None:
         self._fp: Any = fp
 
     def __getattr__(self, name: str) -> object:
         return getattr(self._fp, name)
 
     @property
-    def wrapped(self) -> IO[T_IO]:
+    def wrapped(self) -> IO[AnyStr]:
         """The wrapped file object."""
         return self._fp
 
-    async def __aiter__(self) -> AsyncIterator[T_IO]:
+    async def __aiter__(self) -> AsyncIterator[AnyStr]:
         while True:
             line = await self.readline()
             if line:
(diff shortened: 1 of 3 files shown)


src/anyio/_core/_fileio.py
```diff
@@ -91,10 +174,10 @@ async def readinto(self, b: Union[bytes, memoryview]) -> bytes:
     async def readinto1(self, b: Union[bytes, memoryview]) -> bytes:
         return await to_thread.run_sync(self._fp.readinto1, b)
 
-    async def write(self, b: bytes) -> int:
+    async def write(self, b: T_IO) -> int:
         return await to_thread.run_sync(self._fp.write, b)
 
-    async def writelines(self, lines: bytes) -> None:
+    async def writelines(self, lines: T_IO) -> None:
```
graingert, line 180:
```suggestion
    async def writelines(self, lines: Iterable[T_IO]) -> None:
```

src/anyio/_core/_fileio.py
```diff
@@ -73,16 +156,16 @@ async def __aiter__(self) -> AsyncIterator[bytes]:
...
-    async def read1(self, size: int = -1) -> Union[bytes, str]:
+    async def read1(self, size: int = -1) -> bytes:
         return await to_thread.run_sync(self._fp.read1, size)
 
-    async def readline(self) -> bytes:
+    async def readline(self) -> T_IO:
         return await to_thread.run_sync(self._fp.readline)
 
-    async def readlines(self) -> bytes:
+    async def readlines(self) -> T_IO:
```
graingert, line 168:
```suggestion
    async def readlines(self) -> Iterable[T_IO]:
```

src/anyio/_core/_fileio.py
```diff
@@ -73,16 +156,16 @@ async def __aiter__(self) -> AsyncIterator[bytes]:
     async def aclose(self) -> None:
         return await to_thread.run_sync(self._fp.close)
 
-    async def read(self, size: int = -1) -> Union[bytes, str]:
+    async def read(self, size: int = -1) -> T_IO:
         return await to_thread.run_sync(self._fp.read, size)
 
-    async def read1(self, size: int = -1) -> Union[bytes, str]:
+    async def read1(self, size: int = -1) -> bytes:
```
graingert, line 162:
```suggestion
    async def read1(self: AsyncFile[bytes], size: int = -1) -> bytes:
```

src/anyio/_core/_fileio.py
```diff
@@ -73,16 +156,16 @@ async def __aiter__(self) -> AsyncIterator[bytes]:
...
 
-    async def readline(self) -> bytes:
+    async def readline(self) -> T_IO:
         return await to_thread.run_sync(self._fp.readline)
 
-    async def readlines(self) -> bytes:
+    async def readlines(self) -> T_IO:
         return await to_thread.run_sync(self._fp.readlines)
 
     async def readinto(self, b: Union[bytes, memoryview]) -> bytes:
```
graingert, line 171:
```suggestion
    async def readinto(self, b: WritableBuffer) -> bytes:
```

src/anyio/_core/_fileio.py
```diff
@@ -91,10 +174,10 @@ async def readinto(self, b: Union[bytes, memoryview]) -> bytes:
     async def readinto1(self, b: Union[bytes, memoryview]) -> bytes:
```
graingert, line 174:
```suggestion
    async def readinto1(self, b: WritableBuffer) -> int | None:
```

src/anyio/_core/_fileio.py
agronholm: I'll go with `List[AnyStr]` instead since typeshed defines `readlines()` to return `List`.

src/anyio/_core/_fileio.py
agronholm: I've fixed this now.

src/anyio/_core/_fileio.py
agronholm: Done.

src/anyio/_core/_fileio.py
agronholm: Also done.

src/anyio/_core/_fileio.py
agronholm: I used `Iterable[ReadableBuffer]` as typeshed did.

src/anyio/_core/_fileio.py
```diff
@@ -73,28 +154,28 @@ async def __aiter__(self) -> AsyncIterator[bytes]:
...
-    async def readinto(self, b: Union[bytes, memoryview]) -> bytes:
+    async def readinto(self: 'AsyncFile[bytes]', b: WriteableBuffer) -> bytes:
         return await to_thread.run_sync(self._fp.readinto, b)
 
-    async def readinto1(self, b: Union[bytes, memoryview]) -> bytes:
+    async def readinto1(self: 'AsyncFile[bytes]', b: WriteableBuffer) -> bytes:
         return await to_thread.run_sync(self._fp.readinto1, b)
 
-    async def write(self, b: bytes) -> int:
+    async def write(self, b: ReadableBuffer) -> int:
```
graingert, line 175: something like this?

```python
@overload
async def write(self: 'AsyncFile[bytes]', b: ReadableBuffer) -> int: ...

@overload
async def write(self: 'AsyncFile[str]', b: str) -> int: ...

async def write(self, b: AnyStr) -> int: ...
```

src/anyio/_core/_fileio.py
```diff
@@ -73,28 +154,28 @@ async def __aiter__(self) -> AsyncIterator[bytes]:
...
         return await to_thread.run_sync(self._fp.write, b)
 
-    async def writelines(self, lines: bytes) -> None:
+    async def writelines(self, lines: Iterable[ReadableBuffer]) -> None:
```
graingert, line 178:
```
@overload
async def write(self: 'AsyncFile[bytes]', b: Iterable[ReadableBuffer]) -> None: ...

@overload
async def write(self: 'AsyncFile[str]', b: Iterable[str]) -> None: ...

async def write(self, b: Iterable[AnyStr]) -> None: ...
```

src/anyio/_core/_fileio.py
agronholm: Fixed.

src/anyio/_core/_fileio.py
agronholm: Fixed.

graingert (approved):

src/anyio/_core/_fileio.py
```diff
@@ -1,25 +1,106 @@
+import array
+import mmap
```
graingert, line 2: I suspect mmap might not always exist

```
>>> import mmap
>>> mmap
<module 'mmap' from '/usr/lib/python3.8/lib-dynload/mmap.cpython-38-x86_64-linux-gnu.so'>
```

graingert (changes requested):

src/anyio/_core/_fileio.py
graingert: use this instead of copying from _typeshed I think

```
if TYPE_CHECKING:
    from _typeshed import ReadableBuffer, WritableBuffer
else:
    ReadableBuffer = WritableBuffer = object
```

agronholm: I'm no longer importing these modules at run time.

graingert (approved):

agronholm:
Thanks for the review!

PR #346, opened 2021-07-22
If a worker is being stopped, do not register it in idle_workers

### LINKED ISSUE #323: asyncio backend sometimes schedules sync tasks on worker threads that are about to shut down
The `asyncio` backend of AnyIO can sometimes schedule sync tasks on worker threads that have already been designated to shut down.

The erroneous code path is in `anyio._backends._asyncio`, in the `run_sync_in_worker_thread` method. When there are at least two workers in `idle_workers` that have been idle for more than `WorkerThread.MAX_IDLE_TIME`, the code will first pop off one of the workers from `idle_workers` [here](https://github.com/agronholm/anyio/blob/3cada46968b74568190abc71bc4aaf0c535a8c40/src/anyio/_backends/_asyncio.py#L787), and stores it in the `worker` variable. Then, the function goes on to check whether there are any _other_ workers that have been idle long enough. If there is at least one such worker, it will be asked to stop [here](https://github.com/agronholm/anyio/blob/3cada46968b74568190abc71bc4aaf0c535a8c40/src/anyio/_backends/_asyncio.py#L797) -- but the value of the `worker` variable is overwritten in the process, effectively putting the new task in the queue of the already-stopped worker, after the `None` item. The end result is that the worker stops and the future corresponding to the task will never get resolved.

I believe this can be fixed easily by renaming the `worker` variable in [this loop](https://github.com/agronholm/anyio/blob/3cada46968b74568190abc71bc4aaf0c535a8c40/src/anyio/_backends/_asyncio.py#L791), but I haven't had the time to test it yet.

jlaine:
The asyncio backend for thread pools sometimes schedules tasks on
workers for which a shutdown has been requested. This results in the
task never completing.

Fixes #323.

### DIFF
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -711,11 +711,13 @@ class WorkerThread(Thread):
         self.loop = root_task._loop
         self.queue: Queue[Union[Tuple[Callable, tuple, asyncio.Future], None]] = Queue(2)
         self.idle_since = current_time()
+        self.stopping = False
 
     def _report_result(self, future: asyncio.Future, result: Any,
                        exc: Optional[BaseException]) -> None:
-        self.idle_since = current_time()
-        self.idle_workers.append(self)
+        if not self.stopping:
+            self.idle_since = current_time()
+            self.idle_workers.append(self)
         if not future.cancelled():
             if exc is not None:
                 future.set_exception(exc)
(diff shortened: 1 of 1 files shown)


jlaine:
Commit af2b13880ca5785af9a9f98bcc20865c779ec9c1 does the same thing + tests, closing

PR #358, opened 2021-08-16
Fixed RuntimeError on unclosed async generators on py3.6/asyncio

agronholm:


src/anyio/_backends/_asyncio.py
```diff
@@ -126,6 +128,14 @@ def current_task(loop: Optional[asyncio.AbstractEventLoop] = None) -> Optional[a
 
         return asyncio.Task.current_task(loop)
 
+
+# Needed for the workaround for async generator shutdown on Python 3.6
+async def _dummy_asyncgen():
+    yield
+
+async_generator_athrow = _dummy_asyncgen().athrow(Exception).__class__
```
graingert, line 136: Doesn't this throw an unawaited Coroutine warning?

src/anyio/_backends/_asyncio.py
```diff
@@ -1857,7 +1875,9 @@ def _cancel_all_tasks(self) -> None:
             return
 
         for task in to_cancel:
-            task.cancel()
+            # Workaround for async generator shutdown on Python 3.6
+            if not isinstance(get_coro(task), async_generator_athrow):
+                task.cancel()
```
graingert, line 1880:
```suggestion
            # Workaround for async generator shutdown on Python 3.6
            if (
                sys.version_info >= (3, 6)
                and not isinstance(get_coro(task), async_generator_athrow)
            ):
                task.cancel()
```
you do want to cancel athrow tasks on 3.7+

src/anyio/_backends/_asyncio.py
```diff
@@ -65,7 +66,9 @@ def _cancel_all_tasks(loop):
                 return
 
             for task in to_cancel:
-                task.cancel()
+                # Workaround for async generator shutdown on Python 3.6
+                if not isinstance(get_coro(task), async_generator_athrow):
+                    task.cancel()
```
graingert, line 71: you can `sleep(0)` and then cancel these tasks

src/anyio/_backends/_asyncio.py
agronholm: Fixed that one.

src/anyio/_backends/_asyncio.py
agronholm: How does `sleep(0)` help?

graingert: the task will have called `self._coro.send(None)` after sleep(0) finishes

graingert:
Also the issue title is slightly wrong: the problem also applies to correctly aclosed async generators

src/anyio/_backends/_asyncio.py
```diff
@@ -65,7 +66,9 @@ def _cancel_all_tasks(loop):
                 return
 
             for task in to_cancel:
-                task.cancel()
+                # Workaround for async generator shutdown on Python 3.6
```
graingert, line 69:
```suggestion
                # Workaround for async generator shutdown on Python 3.6
                # https://bugs.python.org/issue38013
```

src/anyio/_backends/_asyncio.py
agronholm: If we want to do that on 3.7.5+, wouldn't we be missing such tasks on earlier versions with the proposed code?

graingert: missing?

agronholm: Not cancelling them even though we possibly should.

graingert: yes you have to wait an event loop turn to catch those missing tasks on 3.7.4-

agronholm:
Given the rapidly approaching EOL of Python 3.6, this is likely not a fix we will incorporate into AnyIO.

PR #371, opened 2021-09-19
Fixed level cancellation not resuming for all parent scopes after exiting a shielded scope

### LINKED ISSUE #370: CancelScope.cancel() does not work as expected if called within move_on_after() context
I ran  this: 
```python
from anyio import move_on_after, sleep, run, CancelScope

result1 = 0
result2 = 0

async def main1():
    global result1
    with CancelScope() as ascope:
        with move_on_after(1, shield=True) as scope:  # XXX
                print("Starting sleep")
                ascope.cancel()
                print("ascope cancelled")
                await sleep(2)

                print( "Cancel scopes=", scope.cancel_called, ascope.cancel_called)

        print("zzzz")
        await sleep(0)

        result1 = 1
        print("Exited cancel scopes=", scope.cancel_called, ascope.cancel_called)

async def main2():
    global result2
    with CancelScope() as ascope:
        with move_on_after(1) as scope:  # XXX
            with CancelScope(shield=True):  # XXX
                    print("Starting sleep")
                    ascope.cancel()
                    print("ascope cancelled")
                    await sleep(2)

                    print("Cancel scopes=", scope.cancel_called, ascope.cancel_called)

        print("zzzz")
        await sleep(0)

        result2 = 1
        print("Exited cancel scopes=", scope.cancel_called, ascope.cancel_called)

run(main1); print()
run(main2); print()
assert result1 == result2, "main1() and main2() should be equivalent"
```
And the outcome is:
```
Starting sleep
ascope cancelled
Cancel scopes= True True
zzzz

Starting sleep
ascope cancelled
Cancel scopes= True True
zzzz
Exited cancel scopes= True True

  File "C:\tmp\whichmove.py", line 45, in <module>
    assert result1 == result2, "main1() and main2() should be equivalent"
AssertionError: main1() and main2() should be equivalent
```
This came from `win10`, `python 3.9.6`, `asyncio`, `anyio-3.3.1`

From my reading of the documentation both main1 and main2 should be equivalent.  If that is the case, I would say there is an issue with main2.  Thank you.

agronholm:
We now find the highest level parent that has been directly cancelled (stopping at the first shielded scope) and restart the cancellation process there.

Fixes #370.

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -3,6 +3,12 @@ Version history
 
 This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
+**UNRELEASED**
+
+- Fixed cancellation problem on asyncio where level-triggered cancellation for **all** parent
+  cancel scopes would not resume after exiting a shielded nested scope
+  (`#370 <https://github.com/agronholm/anyio/issues/370>`_)
+
 **3.3.1**
 
 - Added missing documentation for the ``ExceptionGroup.exceptions`` attribute
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -305,7 +305,7 @@ class CancelScope(BaseCancelScope):
 
         host_task_state.cancel_scope = self._parent_scope
 
-        # Restart the cancellation effort in the nearest directly cancelled parent scope if this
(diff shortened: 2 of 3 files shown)


graingert (approved):

agronholm:
Thanks for reviewing!

PR #377, opened 2021-10-10
propagate native cancel during task group exit

### LINKED ISSUE #374: Task group is swallowing `CancelledError`.
If any of the tasks within the task group raise a `CancelledError`, then that does not propagate out of the task group. In my case, this happens because the code within the task group is cancelled from somewhere else.

```python
from anyio import create_task_group
import asyncio

async def main():
    async def in_task_group():
        raise asyncio.CancelledError

        tg.start_soon(in_task_group)

asyncio.run(main())
```

Would this be a bug? Or is this expected behavior?

graingert:
Fixes #374


agronholm:
Do you have a plan for this? Simply adding the `CancelledError` to the list of exceptions is going to break other things.

agronholm:
You should merge from master to get CI to pass.

PR #380, opened 2021-10-14
Fixed wrong exception raised on sudden TLS stream end on receive()

### LINKED ISSUE #379: When `standard_compatible=False` is set `.receive()` should `raise EndOfStream` for SSL EOF errors.
Prompted by https://github.com/encode/httpx/issues/1896 and https://github.com/encode/httpcore/issues/396.

Currently `standard_compatible=False` allows SSL streams to fail to unwrap properly without raising errors, which is great, and the expected behaviour.

It should *also* ensure that `receive()` operations that result in an `SSLEOFError` return `b""` rather than [raising a `BrokenResourceError`.](https://github.com/agronholm/anyio/blob/c0002bfbcbe83d69ea394cf7c8c9318c62426b26/src/anyio/streams/tls.py#L120-L121)

See [the equivalent in Trio](https://github.com/python-trio/trio/blob/0c2998f642d119f23a41e52d8b018fea1e1e1c28/trio/_ssl.py#L680-L690).

agronholm:
Fixes #379.

### DIFF
--- a/src/anyio/streams/tls.py
+++ b/src/anyio/streams/tls.py
@@ -118,11 +118,15 @@ class TLSStream(ByteStream):
             except ssl.SSLWantWriteError:
                 await self.transport_stream.send(self._write_bio.read())
             except ssl.SSLEOFError as exc:
+                self._read_bio.write_eof()
+                self._write_bio.write_eof()
                 if self.standard_compatible:
                     raise BrokenResourceError from exc
                 else:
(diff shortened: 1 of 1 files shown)


tomchristie:
I can't say that I understand *exactly* what effect that state change has there, but certainly the stream ought to end up in a no-longer-usable closed-state, yes.

PR #388, opened 2021-11-14
Fixed race condition when cancelling acquire() in Lock or Semaphore

### LINKED ISSUE #387: Potential race condition in `Lock`
Hello!

I'm trying to debug an issue we sometimes hit in production through httpx. We're not using anyio ourselves directly (it's regular asyncio), but httpx is based on it.

I'm not 100% sure the scenario I'm documenting here is the exact issue we're seeing, but there's a good chance. The symptoms we're seeing is that a lock instance gets stuck in the locked state with a finished task still set as the owner.

Here's the rough sketch, involving two tasks:

Task 1:
* acquires the lock

Task 2:
* tries acquiring the lock, fails, creates an event and starts waiting on it

Task 1:
* cancels Task 2, releases the lock setting the lock `_owner_task` to Task 2, gets Task 2's event and sets it

Task 2:
* wakes up with CancelledError on `event.wait()`, propagates it

The lock `_owner_task` is still set to Task 2, rendering the lock stuck.

In our production code it's not Task 1 cancelling Task 2, but aiohttp (but that's not relevant to the example).

Here's a Python snippet demonstrating the problem:
```
from asyncio import CancelledError, create_task, run, sleep
from contextlib import suppress

from anyio import Lock


async def main():
    lock = Lock()

    async def task_b():
        await sleep(0.1)  # Sleep to allow task_a to set up
        async with lock:
            pass

    t = create_task(task_b())

    async def task_a():
        async with lock:
            await sleep(0.1)  # Sleep to allow task_b to try locking
            t.cancel()
        await sleep(0.2)
        print(t.done())  # True
        print(lock._owner_task)  # TaskInfo(id=4337401360, name='Task-2')

    await task_a()
    with suppress(CancelledError):
        await t


run(main())
```
The lock should end up in a free state, but it gets stuck being owned by `task_b`.

agronholm:
Fixes #387.

### DIFF
--- a/src/anyio/_core/_synchronization.py
+++ b/src/anyio/_core/_synchronization.py
@@ -305,7 +305,7 @@ class Semaphore:
                 if not event.is_set():
                     self._waiters.remove(event)
                 else:
-                    self._value += 1
+                    self.release()
 
                 raise
         else:
--- a/src/anyio/_core/_synchronization.py
+++ b/src/anyio/_core/_synchronization.py
@@ -127,6 +127,8 @@ class Lock:
             except BaseException:
                 if not event.is_set():
                     self._waiters.remove(token)
+                elif self._owner_task == task:
+                    self.release()
 
                 raise
 
@@ -302,6 +304,8 @@ class Semaphore:
             except BaseException:
                 if not event.is_set():
                     self._waiters.remove(event)
+                else:
+                    self._value += 1
 
                 raise
         else:
--- a/src/anyio/from_thread.py
+++ b/src/anyio/from_thread.py
@@ -82,7 +82,6 @@ class _BlockingAsyncContextManager(AbstractContextManager):
             value = await self._async_cm.__aenter__()
         except BaseException as exc:
             self._enter_future.set_exception(exc)
-            raise
         else:
             self._enter_future.set_result(value)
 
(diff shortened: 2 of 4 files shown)


src/anyio/from_thread.py
```diff
@@ -82,7 +82,6 @@ async def run_async_cm(self) -> Optional[bool]:
             value = await self._async_cm.__aenter__()
         except BaseException as exc:
             self._enter_future.set_exception(exc)
-            raise
```
graingert, line 85: Is this related?

agronholm: Whoa, how did that get into the PR...must've been an artifact of my work on that other PR. Good job noticing this!

src/anyio/_core/_synchronization.py
```diff
@@ -302,6 +304,8 @@ async def acquire(self) -> None:
             except BaseException:
                 if not event.is_set():
                     self._waiters.remove(event)
+                else:
+                    self._value += 1
```
agronholm, line 308: I'm thinking, maybe this should call `self.release()` too, like the `Lock` version?

graingert (approved):

agronholm:
Thanks for the review!

Issue #363, opened 2021-08-22
to/from_thread and contextvars


Kyuuhachi:
The behavior of how contextvars are propagated through `to`/`from_thread` differs depending on which backend you use. With `asyncio`, they inherit the context of the calling thread; in `trio` they inherit the context of the system task group.
```py
import contextvars
import anyio

var = contextvars.ContextVar("var", default=None)

async def get():
	return var.get()

def anyio_from_thread():
	var.set("b")
	return anyio.from_thread.run(get)

async def anyio_to_from_thread():
	var.set("a")
	return await anyio.to_thread.run_sync(anyio_from_thread)

print("asyncio", anyio.run(anyio_to_from_thread, backend="asyncio"))
print("trio   ", anyio.run(anyio_to_from_thread, backend="trio"))
```

On anyio 3.3.0, this prints:
```
asyncio b
trio    None
```

This is easy (but inconvenient) to work around for `run_sync` by adding a `copy_context().run`, but I think any similar workaround for `run` would require special-casing for trio.

Also, `asyncio.to_thread` copies the context whereas `anyio.to_thread.run_sync` does not:
```py
import contextvars
import anyio
import asyncio

var = contextvars.ContextVar("var", default=None)

async def asyncio_to_thread():
	var.set("b")
	return await asyncio.to_thread(var.get)

async def anyio_to_thread():
	var.set("b")
	return await anyio.to_thread.run_sync(var.get)

print("asyncio", asyncio.run(asyncio_to_thread()))
print("anyio  ", anyio.run(anyio_to_thread, backend="asyncio"))
```
```
asyncio c
anyio   None
```

This is not as important since they are different functions, but it's still annoying. Same workaround as above works here, of course.

Kyuuhachi:
I think the most intuitive would be that both `to_thread` and `from_thread` inherit the context of the caller. I haven't investigated how contextvars are used in the wider ecosystem, but that would allow libraries like Eliot to be used without lots of extra cruft.

agronholm:
> I think the most intuitive would be that both to_thread and from_thread inherit the context of the caller.

Having thought about this some more, I agree. It seems though that I cannot force trio to do this when calling back to async-land from a worker thread. Asyncio, on the other hand, behaves this way naturally.

smurfix:
Then maybe we should add that feature to Trio.

agronholm:
[...]

PR #389, opened 2021-11-15
✨ Add support for copying `contextvars` to thread workers

tiangolo:
✨ Add support for copying `contextvars` to thread workers.

This adds the necessary logic to support copying the context in asyncio. There's an equivalent, supporting Trio PR here: https://github.com/python-trio/trio/pull/2160

This is related to https://github.com/agronholm/anyio/issues/363

This also includes the workaround for Trio for what I can do here on AnyIO (which wouldn't be necessary with https://github.com/python-trio/trio/pull/2160). But there's the particular case that I can't make work here without https://github.com/python-trio/trio/pull/2160, of calling `from_thread.run` (with an async function).

I don't know what's preferred, if having the extra workaround for Trio here or waiting for that PR in Trio and then removing the extra code here, but require a higher Trio version (once that's released). Let me know if I should remove all that.

Testing locally with that Trio branch the new tests seem to pass.

---

For completeness, here's the relevant bug in Python: https://bugs.python.org/issue34014 where the authors (asyncio, async, await) seem to agree that the context should be copied, but it was unfeasible to change in asyncio.

---

I want to get some feedback on this before updating docs, etc. 🤓

### DIFF
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -1,6 +1,7 @@
 import array
 import asyncio
 import concurrent.futures
+import contextvars
 import math
 import socket
 import sys
@@ -21,6 +22,8 @@ from typing import (
     Any, Awaitable, Callable, Collection, Coroutine, Deque, Dict, Generator, Iterable, List,
     Mapping, Optional, Sequence, Set, Tuple, Type, TypeVar, Union, cast)
 from weakref import WeakKeyDictionary
+
+from sniffio import current_async_library_cvar
 
 from .. import CapacityLimiterStatistics, EventStatistics, TaskInfo, abc
 from .._core._compat import DeprecatedAsyncContextManager, DeprecatedAwaitable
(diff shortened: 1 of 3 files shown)


agronholm:
I wish you had reached out to me first. I have a nearly complete PR of my own coming, just missing one `BlockingPortal` fix.

agronholm:
This PR is quite different from what I have locally. I will push my changes and make a PR tomorrow so we can compare notes.

tiangolo:
Sure thing!

And sorry, I did most of the work on a long flight a couple of weeks ago without internet (and I was mostly offline for the past two weeks). And the main part of my work was just porting my tests from the other Trio PR.

If we can check that the tests/use cases pass with your PR I'm perfectly fine with closing mine. 🙂

Also feel free to just copy paste anything you find useful here in this PR, no need to worry about "commit attributions" or anything like that. 🤓

agronholm:
Your PR seems to handle things better than mine. I had some ugly workarounds for trio but I'd rather have trio solve those problems on their end while I deal with asyncio here. I'll look into merging our efforts into one conclusive PR, with `BlockingPortal` also covered and without the trio workarounds.

agronholm:
I have my own PR now (#390) which incorporates tricks from this one. We still need trio to fix the problem on their end.

One unresolved issue is what `sniffio.current_async_library()` should give in a worker thread. In my opinion it should raise `AsyncLibraryNotFoundError` since it's not in an event loop thread.

smurfix:
I'm +1 on `AsyncLibraryNotFoundError` in principle.

On the other hand, the cases where that actually matters (i.e. you use a library within the thread that uses sniffio) are likely to be few and far between. Thus I'd be OK with requiring the user to explicitly clear `sniffio`'s state if that's required.

agronholm:
Looking at @tiangolo 's trio PR, I see that it is setting the current async library to `None`, but my test still fails on trio:
```python3
async def test_asynclib_detection() -> None:
    with pytest.raises(sniffio.AsyncLibraryNotFoundError):
        await to_thread.run_sync(sniffio.current_async_library)
```

agronholm:
What I think is going on here (correct me if I'm wrong) is: the cached name of the current async library is voided, but the trio token is still present in the context so sniffio's detection picks that up and declares that the worker thread is running on the trio event loop.

smurfix:
I suspect that `to_thread.run_sync` uses *some* context to run the thread … just not the one you're calling `to_thread.run_sync` from. As Trio's thread engine does thread caching AFAIR, that seems somewhat likely at least.

agronholm:
Ahh...it's anyio that's setting sniffio's contextvar in the worker thread. I think that's a no-no, and I need to use my own context for that instead.

agronholm:
Removed the lines that set the sniffio cvar; no ill effects noted in the test suite. The PR is now ready, once trio is. We should continue the discussion over at #390.

PR #391, opened 2021-11-19
Handle parametrized marks in pytest plugin

dlax:
We now detect 'anyio' marker when set within pytest.mark.parametrize,
for instance:
```python
    @pytest.mark.parametrize(
        "value",
        [
            pytest.param("1", marks=[pytest.mark.asyncio]),
            pytest.param("2", marks=[pytest.mark.anyio]),
        ]
    )
    async def test_param(value):
        ...
```
or included test case.

* * *

The code for detecting "inner" markers is pretty ugly; happy to rewrite if there's a better way.

### DIFF
--- a/src/anyio/pytest_plugin.py
+++ b/src/anyio/pytest_plugin.py
@@ -96,8 +96,18 @@ def pytest_pycollect_makeitem(collector: Any, name: Any, obj: Any) -> None:
         inner_func = obj.hypothesis.inner_test if hasattr(obj, 'hypothesis') else obj
         if iscoroutinefunction(inner_func):
             marker = collector.get_closest_marker('anyio')
-            own_markers = getattr(obj, 'pytestmark', ())
-            if marker or any(marker.name == 'anyio' for marker in own_markers):
+            own_markers = getattr(obj, 'pytestmark', [])
+            parametrized_markers = [
+                mark
+                for own_marker in own_markers
+                for paramset in (
+                    own_marker.args[1] if len(own_marker.args) >= 2 else ()
+                )
+                for mark in paramset.marks
+                if own_marker.name == 'parametrize'
+            ]
+            markers = own_markers + parametrized_markers
+            if marker or any(marker.name == 'anyio' for marker in markers):
                 pytest.mark.usefixtures('anyio_backend')(obj)
 
 
--- a/tests/test_pytest_plugin.py
+++ b/tests/test_pytest_plugin.py
@@ -58,11 +58,22 @@ def test_plugin(testdir: Testdir) -> None:
         async def test_skip_inline(some_feature):
             # Test for github #214
             pytest.skip("Test that skipping works")
+
+        @pytest.mark.parametrize(
+            "value",
+            [
+                pytest.param("1", marks=[pytest.mark.anyio]),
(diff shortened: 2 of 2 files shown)


agronholm:
What's the use case for this?

dlax:
I am introducing support for anyio in a library as an alternative to asyncio; I have two implementations of the same function and a single test that is currently written as:
```python
@pytest.mark.asyncio
async def test_func():
    # test 'func'
``` 
So I'd like to make it parametrized as:
```python
@pytest.fixture(
   "func",
    params=[
        pytest.param(func_asyncio, marks=pytest.mark.asyncio),
        pytest.param(func_anyio, marks=pytest.mark.anyio),
    ],
)
async def test_func(func):
    # test 'func'
    ...
```

agronholm:
But why not simply use the anyio pytest plugin for all the tests?

agronholm:
Are you also aware that `pytest-asyncio`, when present, runs any async fixtures, whether it was actually enabled or not?

dlax:
> But why not simply use the anyio pytest plugin for all the tests?

Can it be used as a drop-in replacement of asyncio plugin?
We merely use it to detect `async` test functions, nothing more fancy as far as I know.

agronholm:
> Can it be used as a drop-in replacement of asyncio plugin?

Mostly. There are some advanced cases that don't work with the anyio plugin, but in such cases you can just run the tests with the pytest-asyncio plugin alone.

dlax:
Ok; I'll try using the value of `anyio_backend` fixture to detect which implementation to use. This seems to be working.

agronholm:
I can probably offer more advice if you tell me what exactly you're doing. If you have two different code paths then you could parametrize the tests based on that. Do you need to run the any code path tests on trio too?

dlax:
I'm working on psycopg, where I defined an `AnyIOConnection` connection class  as an alternative to `AsyncIOConnection`. Then, in tests, we have a fixture that returns an opened connection and I would basically adjust it as:
```python
@pytest.fixture
async def aconn(dsn, anyio_backend):
    cls = {"asyncio":  AsyncIOConnection, "trio": AnyIOConnection}[anyio_backend]
    conn = await cls.connect(dsn)
    yield conn
    await conn.close()
```
So this runs the test suite with "asyncio" and "trio" backends. Unfortunately, this approach prevents `AnyIOConnection` from being used with the "asyncio" backend, but that's probably enough for us now.

Also, we need to force a particular event loop policy on windows. This is done currently with the `event_loop` fixture of `pytest-asyncio`. Is there a way to achieve this with the anyio plugin?

agronholm:
> Also, we need to force a particular event loop policy on windows. This is done currently with the event_loop fixture of pytest-asyncio. Is there a way to achieve this with the anyio plugin?

Something along the lines of:
```python3
asyncio_options = {'policy': asyncio.WindowsSelectorEventLoopPolicy()} if platform.system() == 'Windows' else {}

@pytest.fixture(params=[
    pytest.param(('asyncio', asyncio_options), id='asyncio'),
    pytest.param('trio')
])
def anyio_backend(request: SubRequest) -> Tuple[str, Dict[str, Any]]:
    return request.param
```

PR #397, opened 2021-11-28
Improved type annotations

### LINKED ISSUE #396: Type incompleteness
I am trying to use Pyright as type checker and getting errors like this:

```
error: Type of "start_soon" is partially unknown
    Type of "start_soon" is "(func: (*args: Any, **kwargs: Any) -> Coroutine[Unknown, Unknown, Unknown], *args: object, name: object | None = None) -> None" (reportUnknownMemberType)
```

I think, type-parameters for generic types must be explicitly stated (as `Any` or as `object` in some cases).

agronholm:
This gets us to 100% score with pyright. Fixes #396.

### DIFF
--- a/src/anyio/_core/_compat.py
+++ b/src/anyio/_core/_compat.py
@@ -2,8 +2,8 @@ from abc import ABCMeta, abstractmethod
 from contextlib import AbstractContextManager
 from types import TracebackType
 from typing import (
-    TYPE_CHECKING, AsyncContextManager, Callable, ContextManager, Generator, Generic, Iterable,
-    List, Optional, Tuple, Type, TypeVar, Union, overload)
+    TYPE_CHECKING, Any, AsyncContextManager, Callable, ContextManager, Generator, Generic,
+    Iterable, List, Optional, Tuple, Type, TypeVar, Union, overload)
 from warnings import warn
 
 if TYPE_CHECKING:
@@ -85,7 +85,7 @@ def maybe_async_cm(cm: Union[ContextManager[T], AsyncContextManager[T]]) -> Asyn
     return _ContextManagerWrapper(cm)
 
 
-def _warn_deprecation(awaitable: AnyDeprecatedAwaitable, stacklevel: int = 1) -> None:
+def _warn_deprecation(awaitable: 'AnyDeprecatedAwaitable[Any]', stacklevel: int = 1) -> None:
     warn(f'Awaiting on {awaitable._name}() is deprecated. Use "await '
          f'anyio.maybe_async({awaitable._name}(...)) if you have to support both AnyIO 2.x '
          f'and 3.x, or just remove the "await" if you are completely migrating to AnyIO 3+.',
--- a/src/anyio/_core/_compat.py
+++ b/src/anyio/_core/_compat.py
@@ -36,7 +36,7 @@ async def maybe_async(__obj: 'DeprecatedAwaitable') -> None:
     ...
 
 
-async def maybe_async(__obj: AnyDeprecatedAwaitable[T]) -> Union[TaskInfo, float, List[T], None]:
+async def maybe_async(__obj: 'AnyDeprecatedAwaitable[T]') -> Union[TaskInfo, float, List[T], None]:
     """
     Await on the given object if necessary.
 
--- a/src/anyio/_core/_compat.py
+++ b/src/anyio/_core/_compat.py
@@ -36,7 +36,7 @@ async def maybe_async(__obj: 'DeprecatedAwaitable') -> None:
     ...
 
 
-async def maybe_async(__obj: AnyDeprecatedAwaitable) -> Union[TaskInfo, float, List[T], None]:
+async def maybe_async(__obj: AnyDeprecatedAwaitable[T]) -> Union[TaskInfo, float, List[T], None]:
     """
     Await on the given object if necessary.
 
--- a/src/anyio/_core/_compat.py
+++ b/src/anyio/_core/_compat.py
@@ -13,7 +13,7 @@ else:
 
 T = TypeVar('T')
 AnyDeprecatedAwaitable = Union['DeprecatedAwaitable', 'DeprecatedAwaitableFloat',
-                               'DeprecatedAwaitableList', TaskInfo]
+                               'DeprecatedAwaitableList[T]', TaskInfo]
 
 
 @overload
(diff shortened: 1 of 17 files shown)


src/anyio/_core/_compat.py
```diff
@@ -13,7 +13,7 @@
 
 T = TypeVar('T')
 AnyDeprecatedAwaitable = Union['DeprecatedAwaitable', 'DeprecatedAwaitableFloat',
-                               'DeprecatedAwaitableList', TaskInfo]
+                               'DeprecatedAwaitableList[T]', TaskInfo]
```
tkukushkin, line 16: `AnyDeprecatedAwaitable` with `DeprecatedAwaitableList[T]` will become generic type alias. Are you sure this is what is needed?

See example:

```
In [3]: T = TypeVar('T')

In [4]: X = Union[int, List[T]]

In [5]: X
Out[5]: typing.Union[int, typing.List[~T]]

In [6]: X[str]
Out[6]: typing.Union[int, typing.List[str]]
```

tkukushkin: @erictraut, why pyright doesn't complain about usages of `AnyDeprecatedAwaitable` without specifying explicit type arguments (in `maybe_async` for example)?

erictraut: I would expect pyright to complain about that. If that's not what you're seeing, please file a bug with a repro description, and I'll investigate further. Thanks!

agronholm: > AnyDeprecatedAwaitable with DeprecatedAwaitableList[T] will become generic type alias. Are you sure this is what is needed?

The third overload was already using a type variable. The fix just added that to the union so it matches with the overload.

agronholm: I meant the third item in the return type union: `List[T]`.

tkukushkin: But now you must specify type argument for all usages of `AnyDeprecatedAwaitable`, because now it is genetic too. I think `DeprecatedAwaitableList[Any]` is what you need.

tkukushkin: Sorry, I haven't seen the latest changes...

src/anyio/_core/_compat.py
```diff
@@ -36,7 +36,7 @@ async def maybe_async(__obj: 'DeprecatedAwaitable') -> None:
     ...
 
 
-async def maybe_async(__obj: AnyDeprecatedAwaitable) -> Union[TaskInfo, float, list, None]:
+async def maybe_async(__obj: AnyDeprecatedAwaitable) -> Union[TaskInfo, float, List[T], None]:
```
tkukushkin, line 39: Did you mean:
```python
async def maybe_async(__obj: AnyDeprecatedAwaitable[T]) -> Union[TaskInfo, float, List[T], None]:
```
?

tkukushkin: Same with `_warn_deprecation`:

```python
def _warn_deprecation(awaitable: AnyDeprecatedAwaitable[Any], stacklevel: int = 1) -> None:
```

With exception, that type variable isn't needed here, return type doesn't depend on argument types, `Any` is enough.

agronholm: I've never seen or heard of a `Union` being further parametrized. That makes no sense. Most items in it don't have parameters, and only `DeprecatedAwaitableList` does but I already added the parameter to that one (which fixed the errors).

tkukushkin: See examples please:

```python
>>> from typing import Union, TypeVar, List
>>>
>>> X = TypeVar('X')
>>> Y = TypeVar('Y')
>>>
>>> U = Union[X, Y]
>>> U[int, str]
typing.Union[int, str]
>>>
>>> U = Union[int, str, List[X]]
>>> U[bool]
typing.Union[int, str, typing.List[bool]]
```

When you use `TypeVar` as type parameter for any generic type, you are only creating another **generic** type which must be further parametrized by concrete type to be _known type_.

Pyright doesn't complain about this because of bug (see https://github.com/agronholm/anyio/pull/397#discussion_r757828005, https://github.com/microsoft/pyright/issues/2625).

tkukushkin: I found a description of this behavior in [PEP-484 (Type aliases)](https://www.python.org/dev/peps/pep-0484/#type-aliases)

erictraut: Yes, PEP 484 provides for generic type aliases. A type alias is considered generic if it uses one or more type variables within its definition. A generic type alias must be specialized (i.e., type arguments must be supplied) when it is used in a type annotation.

agronholm: I've added the type parameter now, but now PyCharm complains:
image

agronholm: Not only that, but the code no longer runs on **any** version of Python:
`TypeError: typing.Union[_ForwardRef('DeprecatedAwaitable'), _ForwardRef('DeprecatedAwaitableFloat'), _ForwardRef('DeprecatedAwaitableList[T]'), object] is not a generic class`

tkukushkin: I think you should use forward references to `AnyDeprecatedAwaitable[T]` because of forward references in it's definition.

agronholm: I take it this is because of the forward references? If I make a union with an immediate reference to a parametrized type, it works fine.

agronholm: Okay, all the tests pass now. Do you have any further concerns about the changes or should I merge?

tkukushkin: Please fix `_warn_deprecation` too.

agronholm: Right, that one. Can I use `Any` there as I don't do anything with the argument?

tkukushkin: Of course. You can even use `object`.

`Any` allows to do whatever you want with value, `object`does not allow anything (as far as I understand).

agronholm: Doesn't look like `object` works though:
`src/anyio/_core/_compat.py:141: error: Argument 1 to "_warn_deprecation" has incompatible type "DeprecatedAwaitableList[T]"; expected "Union[DeprecatedAwaitable, DeprecatedAwaitableFloat, DeprecatedAwaitableList[object], TaskInfo]"
`

agronholm: I have to leave for a bit now, but I take it this was the last missing piece?

tkukushkin: Oh, I forgot that `List[T]` is invariant.

tkukushkin: > I have to leave for a bit now, but I take it this was the last missing piece?

Yeah, thank you!

agronholm: Thanks for the help! I'll merge this now.

PR #399, opened 2021-12-08
Fixed synchronization primitives not being released on native asyncio cancel during __aenter__()

### LINKED ISSUE #398: Deadlock in anyio.Lock
Related to this issue in httpcore: https://github.com/encode/httpcore/issues/454

TL;DR: With the right timing, the lock **acquire** step can be cancelled and the lock remain locked for other tasks because when using the lock context manager, if the task is cancelled during the `__aenter__` , `__aexit__` is not executed.

This snippet reproduces the issue:

```python
import asyncio
import anyio

class Holder:
    def __init__(self):
        # self.lock = asyncio.Lock()
        self.lock = anyio.Lock()

    async def work(self, index):
        print(f"Trying to acquire {index}")
        async with self.lock:
            print(f"Doing index {index}")
            await asyncio.sleep(5)


async def main_anyio():
    holder = Holder()
    t1 = asyncio.create_task(holder.work(1), name="holder 1")
    t2 = asyncio.create_task(holder.work(2), name="holder 2")

    print("Tasks created")
    await asyncio.sleep(0)

    print("Cancelling t1")
    t1.cancel()

    print("Waiting for t2")
    await asyncio.gather(t2)


if __name__ == "__main__":
    asyncio.run(main_anyio())
```

The issue goes away when using `asyncio.Lock` because that lock context manager never blocks during the acquire step. Even if the `__aenter__` method in `asyncio.Lock` is asynchronous, no asynchronous code is executed there.

agronholm:
Fixes #398.

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -3,6 +3,12 @@ Version history
 
 This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
+**UNRELEASED**
+
+- Fixed deadlock in synchronization primitives on asyncio which can happen if a task acquiring a
+  primitive is hit with a native (not AnyIO) cancellation with just the right timing, leaving the
+  next acquiring task waiting forever (`#398 <https://github.com/agronholm/anyio/issues/398>`_)
+
 **3.4.0**
 
 - Added context propagation to/from worker threads in ``to_thread.run_sync()``,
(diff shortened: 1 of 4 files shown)


mecampbellsoup:
When can we expect this to be released? Thanks.

agronholm:
I'm waiting for a review on another PR: https://github.com/agronholm/anyio/pull/400
When that's merged, I'm going to cut a patch release. If you have any insights on the contents on that PR, those would be most welcome and would expedite the process.

PR #400, opened 2021-12-12
Fixed compatibility with OpenSSL 3

agronholm:
There is probably a bug in CPython that necessitates this ugly workaround.

**EDIT**: https://bugs.python.org/issue46313


agronholm:
I've tested this manually with the `fedora:36` image from Dockerhub. I wonder if there is an easy way to run the tests on Github Actions.

PR #403, opened 2021-12-22
✨ Add alternative APIs that improve typing and tooling support (e.g. autocompletion)

tiangolo:
✨ Add alternative APIs that improve typing and tooling support (e.g. autocompletion in editors)

## Description

I want to propose adding a new API style for three functions (additional to the current API) that would take advantage of new type annotations ([PEP 612 - Parameter Specification Variables](https://www.python.org/dev/peps/pep-0612/)), to get better tooling and editor support.

In particular, this would allow/support **mypy** checks, **inline editor errors**, and ✨ **autocompletion** ✨ for positional and **keyword arguments** for:

* Sending async tasks in a task group (equivalent to `tg.start_soon()`)
* Sending blocking tasks to a worker thread (equivalent to `to_thread.run_sync()`)
* Sending async tasks to the main async loop from a worker thread (equivalent to `from_thread.run()`)

This would also allow mypy and editors to provide type checks, autocompletion, etc. for the **return values** received when sending tasks to a worker thread and when receiving them from a worker thread:

* Equivalent to `to_thread.run_sync()`
* Equivalent to `from_thread.run()`

## Usage

The main difference is that instead of having a function that takes the function to call plus:

* positional arguments (and that requires using partials for keyword arguments)
* configs (like task name)

```Python
anyio_function(worker_function, arg1, arg2)
```

...this new API only takes the function to call plus configs (like the task name), and then *returns* another function that takes the positional and keyword arguments. And when that function is called, it does the actual work. So it's used like this:

```Python
anyio_functionify(worker_function)(arg1, arg2)
```

And now it also supports keyword arguments, not only positional ones, so it's not necessary to use (and understand) partials to use keyword arguments:

```Python
anyio_functionify(worker_function)(arg1, arg2, kwarg1="a", kwarg2="b")
```

### Task Groups

```Python
from typing import Any
import anyio


async def program() -> Any:
    async def worker(i: int, message: str = "Default worker"):
        print(f"{message} - worker {i}")

        for i in range(3):
            tg.soonify(worker)(i=i, message="Hello ")


anyio.run(program)
```

* Autocompletion for task group:

Selection_009

* Inline errors for task group:

Selection_010

* Mypy error detection:

```
main.py:11: error: Too few arguments
Found 1 error in 1 file (checked 1 source file)
```

### To Thread (asyncify)

```Python
import time
import anyio


def run_slow(i: int, message: str = "Hello ") -> str:
    time.sleep(i)
    return message + str(i)


async def program() -> None:
    for i in range(3):
        result = await anyio.to_thread.asyncify(run_slow)(i=i)
        result + 3
        print(result)


anyio.run(program)
```

* Autocompletion for sending to worker thread:

Selection_019

* Inline errors for sending to worker thread:

Selection_020

* Autocompletion for result value:

Selection_021

* Inline errors for result value:

Selection_022

* Mypy error detection:

```
main.py:12: error: Too few arguments
main.py:13: error: Unsupported operand types for + ("str" and "int")
```

### From Thread (syncify)

```Python
import anyio


async def aio_run_slow(i: int, message: str = "Async Hello ") -> str:
    await anyio.sleep(i)
    return message + str(i)


def run_slow_embed(i: int) -> str:
    result = anyio.from_thread.syncify(aio_run_slow)(i=i, message="Hello sincify ")
    return result


async def program() -> None:
    for i in range(3):
        result = await anyio.to_thread.asyncify(run_slow_embed)(i=i)
        result + 3
        print(result)


anyio.run(program)
```

* Autocompletion for sending from worker thread:

Selection_027

* Inline errors for sending from worker thread:

Selection_028

* Autocompletion for result value:

Selection_030

* Inline errors for result value:

Selection_031

* Mypy error detection:

```
main.py:10: error: Too few arguments
main.py:17: error: Unsupported operand types for + ("str" and "int")
```

## Function names Bikeshedding

I wanted to have short function names that were descriptive enough, that's why the "somethingify".

I thought any more descriptive variant like `generate_async_function_that_calls_run_sync` would be too long. :sweat_smile: 

My rationale is that if a user wants to call a blocking function while inside async stuff, the user might want to have a way to "asyncify" that blocking function. And also the "somethingify" doesn't imply it is being executed/called right away, but that something is "converted" in some way. I would hope that would help with the intuition that it just returns another function that is then called, so it's necessary to add the extra chained parenthesis with any possible needed arguments.

But of course, I can change the names if there's another preference.

## Tests and Docs

I want to gather feedback first before writing tests and docs, but I'll add them if this or something like this is acceptable.

## Note on FastAPI

I want to have this type of interface for users and myself, I use autocompletion and tooling/types a lot. I thought of adding something like this to FastAPI itself, but I think this would really belong here, not in FastAPI.

I also intend to add it to the FastAPI docs to explain "advanced" use cases with concurrency and blocking code in threads run from async code, etc. I would like to **document** all those things **in FastAPI** referring to and explaining **AnyIO directly**, instead of writing wrappers in FastAPI or anything else. :nerd_face: 

**2022-01-08 Edit**

I see this (in particular taking kwargs) is quite debated, even more in Trio. And I know this is a somewhat drastic change, so I built a small package ([Asyncer](https://github.com/tiangolo/asyncer)) just with these functions and a couple of extra things. I documented it as temporary (in case these things land here) and very subjective (to my point of view).

Maybe that could help gauge how useful that API style is for Trio and AnyIO before making any commitments, in a safer way. I don't like the feel of "yet another library", but it might help separate what is more stable (Trio and AnyIO) and what is a Frankenstein temporary experiment (this idea(s) I'm proposing). I'm mentioning it here just for completeness, just in case.

### DIFF
--- a/src/anyio/abc/_tasks.py
+++ b/src/anyio/abc/_tasks.py
@@ -1,14 +1,21 @@
+import functools
 import typing
 from abc import ABCMeta, abstractmethod
 from types import TracebackType
 from typing import Any, Callable, Coroutine, Optional, Type, TypeVar
 from warnings import warn
+import sys
+
+if sys.version_info >= (3, 10):
+    from typing import ParamSpec
+else:
+    from typing_extensions import ParamSpec
 
 if typing.TYPE_CHECKING:
     from anyio._core._tasks import CancelScope
 
 T_Retval = TypeVar('T_Retval')
-
+T_ParamSpec = ParamSpec("T_ParamSpec")
 
 class TaskStatus(metaclass=ABCMeta):
     @abstractmethod
@@ -47,6 +54,36 @@ class TaskGroup(metaclass=ABCMeta):
         warn('spawn() is deprecated -- use start_soon() (without the "await") instead',
              DeprecationWarning)
         self.start_soon(func, *args, name=name)
+
+    def soonify(self, func: Callable[T_ParamSpec, Coroutine[Any, Any, Any]],
+                name: object = None) -> Callable[T_ParamSpec, None]:
+        """
+        Create and return a function that when called will start a new task in this
+        task group.
+
+        Internally it uses the same ``task_group.start_soon()`` method. But 
+        ``task_group.soonify()`` supports keyword arguments additional to positional
+        arguments and it adds better support for autocompletion and inline errors
+        for the arguments of the function called.
+
+        Use it like this:
+
+            async def do_work(arg1, arg2, kwarg1="", kwarg2=""):
+                # Do work
+
+            task_group.soonify(some_async_func)("spam", "ham", kwarg1="a", kwarg2="b")
+
+        :param func: a coroutine function
+        :param name: name of the task, for the purposes of introspection and debugging
+        :return: a function that takes positional and keyword arguments and when called
+            uses task_group.start_soon() to start the task in this task group.
+
+        .. versionadded:: 3.4.1
+        """
+        def wrapper(*args: T_ParamSpec.args, **kwargs: T_ParamSpec.kwargs) -> None:
+            partial_f = functools.partial(func, *args, **kwargs)
+            return self.start_soon(partial_f, name=name)
+        return wrapper
 
     @abstractmethod
     def start_soon(self, func: Callable[..., Coroutine[Any, Any, Any]],
--- a/src/anyio/abc/_tasks.py
+++ b/src/anyio/abc/_tasks.py
@@ -71,7 +71,7 @@ class TaskGroup(metaclass=ABCMeta):
             async def do_work(arg1, arg2, kwarg1="", kwarg2=""):
                 # Do work
 
-            task_group.soonify(some_async_func)("spam", "ham", kwarg1="a", kwarg2="b")
+            task_group.soonify(do_work)("spam", "ham", kwarg1="a", kwarg2="b")
 
         :param func: a coroutine function
         :param name: name of the task, for the purposes of introspection and debugging
--- a/src/anyio/abc/_tasks.py
+++ b/src/anyio/abc/_tasks.py
@@ -1,10 +1,10 @@
 import functools
+import sys
 import typing
 from abc import ABCMeta, abstractmethod
 from types import TracebackType
 from typing import Any, Callable, Coroutine, Optional, Type, TypeVar
 from warnings import warn
-import sys
 
 if sys.version_info >= (3, 10):
     from typing import ParamSpec
@@ -16,6 +16,7 @@ if typing.TYPE_CHECKING:
 
 T_Retval = TypeVar('T_Retval')
 T_ParamSpec = ParamSpec("T_ParamSpec")
+
 
 class TaskStatus(metaclass=ABCMeta):
     @abstractmethod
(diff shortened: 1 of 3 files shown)


graingert:
I'd like to see this API in trio before it's implemented in anyio.

Ultimately the best outcome here is for typing support for functions like asyncio.to_thread tg.start_soon etc via:

```
async def to_thread(fn: Callable[P, R], /, *args: P.args, **kwargs: P.kwargs) -> R: ...
```

uSpike:
FYI there's a long open issue in trio about this: https://github.com/python-trio/trio/issues/470

tiangolo:
Thanks for the feedback! And thanks for the link @uSpike, I hadn't seen that one.

@graingert, about the preference for:

```Python
P = ParamSpec("P")
R = TypeVar("R")
async def to_thread(fn: Callable[P, R], /, *args: P.args, **kwargs: P.kwargs) -> R: ...
```

As Nathaniel explains there, those functions take configs for them, so they can't take kwargs. I see that what I'm proposing here is one of the main proposals there: https://github.com/python-trio/trio/issues/470#issuecomment-473648115

---

And now with `ParamSpec` I think it would make even more sense, as there are obvious advantages.

Now, I'm not sure it would be easy to convince everyone one way or another in Trio given the huge thread and opinions there, but I'll try.

I also thought there was a chance it could make sense to solve/support it here even if Trio hasn't solved it yet, as was the case with https://github.com/agronholm/anyio/pull/390.

agronholm:
#390 was easier because it was fairly obvious what the correct behavior there was. This PR involves actual API design choices so it's not so clear cut.

tiangolo:
Get it, thanks @agronholm!

I'll comment in the Trio thread and see what happens.

agronholm:
So...rather than adding these elaborate functions, would it not work to just use a lambda? E.g. `tg.start_soon(lambda: my_async_func('foo', kwarg='bah'))`? Should play well with IDEs et al.

tiangolo:
That's a very good point! Very clever and simple. I don't know why that didn't occur to me. 🤦

I've been trying to see any reason why that would not work and I don't find any, even async functions would seem to work (they just can't be awaited inside the lambda, but if they return the awaitable thing then the lambda becomes an awaitable function 🤷 , so it should work).

That part, of calling an async function inside of a lambda but not awaiting it might be a bit mind-bending for newcomers, but it can probably be just documented.

I think that's a clever and simple approach that would work for most cases (in particular for the stuff that I'm interested in) and doesn't need any API change. I guess that would lose some introspection stuff as the wrapper and function name is not propagated, but I don't even know when that would be useful.

The only caveat I see is that it was not obvious to me, I would expect it to also not be obvious to others. In particular as the current API already receives positional arguments. But maybe just adding that to the docs and explaining it would be good enough for most cases.

Now I think it could be nice if none of the APIs received any arguments at all, so that there was only one way to do it, with lambdas, and it was obvious that it was the only way, but I know that's not possible now... anyway, maybe just documenting that approach with lambdas is good enough to achieve what this PR was intended to achieve. 🤔

agronholm:
Are you still pushing this instead of using lambdas?

tiangolo:
Ah, that last commit pushed was just because I had an uncommited typo fix in the same PR.

I think you would rather not have this here, right?

Would you like/accept a docs PR documenting lambdas and how that would help with types and stuff?

In any case, feel free to close this PR if you prefer.

agronholm:
> Would you like/accept a docs PR documenting lambdas and how that would help with types and stuff?

Yes, I think I'd like that!

agronholm:
Just remember that there is one ugly catch with lambdas: if you schedule more than one in the same function using the same variables, you may be surprised to find that they all bind to the latest values of said variable. `functools.partial()` does not have that problem. Also, correct me if I'm wrong, but if you return a lambda that references variables in an outer closure, that closure cannot be garbage collected before the lambda.

PR #408, opened 2022-01-05
Add ability to set `Semaphore.max_value` at runtime

Bluenix2:
## Summary

This pull requests adds a setter to `max_value` that will adjust the current- and max value.

## Considerations

Currently ``max_value`` can't be changed if it is set to None, because then the guarantee that all currently acquired tokens can be released will be broken. This *could* be fixed with an internal `_acquired` integer which counts how many tokens are acquired, would that be better than raising `RuntimeError`?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 *.pyc
 build
 dist
+venv
 docs/_build
 __pycache__
 .coverage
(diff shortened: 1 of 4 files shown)


agronholm:
What is the actual use case for this? Max value isn't a thing in the standard library, and trio does not allow mutating it either.

Bluenix2:
I have a ratelimiting system for the Discord API, this means that I don't know the maximum value until I have made at least one request and received one response.

This change would mean that I can initialize a semaphore with 1 as a max value and then increase it (and keep the tasks queued on it) - when I have received a response.

*Also I realize now that I don't handle the value increasing and waking up more waiters, there's no test for that either - I'll get to that.*

agronholm:
Wouldn't it be a lot simpler to just replace the semaphore?

agronholm:
Alternatively, would a `CapacityLimiter` not work? Its `total_tokens` can already be adjusted.

Bluenix2:
> Wouldn't it be a lot simpler to just replace the semaphore?

I could potentially already have a bunch of tasks waiting on this semaphore.

> Alternatively, would a `CapacityLimiter` not work? Its `total_tokens` can already be adjusted.

That's a good point, I didn't first consider it because I've seen more implementations with `Semaphore`s and was not sure I needed all the other functionality.

PR #409, opened 2022-01-06
Overhauled linting

agronholm:
This changeset should fix the new errors that appeared with mypy 0.930.

It also switches lint checks over to pre-commit.ci and enforces a specific pre-commit configuration. This also eliminates duplication between pre-commit, tox and Github Actions workflows. It also adds a number of useful checks, including pyupgrade and pygrep-hooks.


.pre-commit-config.yaml
```diff
@@ -0,0 +1,64 @@
...
+      - id: mypy
+        additional_dependencies: [ "pytest", "trio-typing", "packaging" ]
+
+  - repo: local
+    hooks:
+      - id: pyright
+        name: pyright
+        entry: pyright
+        language: node
+        pass_filenames: false
```
uSpike, line 51: Why have pyright check all files on every commit but not mypy?

agronholm: I'm confused. Mypy is right above pyright there.

agronholm: I've added pyright back to the GA test workflow and removed it from the pre-commit config.

PR #416, opened 2022-01-30
allow non-coroutine awaitbales in to_thread

adriangb:
For [#1453](https://github.com/encode/starlette/pull/1453), specifically allowing `starlette.endpoints.HTTPEndpoint` to be returned as the object to be `await`ed

### DIFF
--- a/src/anyio/from_thread.py
+++ b/src/anyio/from_thread.py
@@ -1,3 +1,4 @@
+import sys
 import threading
 from asyncio import iscoroutine
 from concurrent.futures import FIRST_COMPLETED, Future, ThreadPoolExecutor, wait
@@ -6,8 +7,13 @@ from inspect import isawaitable
 from types import TracebackType
 from typing import (
     Any, AsyncContextManager, Callable, ContextManager, Coroutine, Dict, Generator, Iterable,
-    Optional, Tuple, Type, TypeVar, Union, cast, overload)
+    Iterator, Optional, Tuple, Type, TypeVar, Union, cast, overload)
 from warnings import warn
+
+if sys.version_info < (3, 8):
+    from typing_extensions import Protocol
+else:
+    from typing import Protocol
 
 from ._core import _eventloop
 from ._core._compat import DeprecatedAwaitable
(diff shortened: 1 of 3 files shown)


src/anyio/from_thread.py
```diff
@@ -177,7 +179,9 @@ def callback(f: Future) -> None:
 
         try:
             retval = func(*args, **kwargs)
-            if iscoroutine(retval):
+            if iscoroutine(retval) or (
+                isawaitable(retval) and not isinstance(retval, DeprecatedAwaitable)
```
adriangb, line 183: I had to special case this to get the tests in anyio itself to pass, but there may be a better way

agronholm: Is the `iscoroutine()` check even needed anymore if `isawaitable()` is used?

agronholm:
A changelog entry and at least one new test would be required for merging.

adriangb:
@agronholm so I realized that we can just work around this in Starlette by wrapping our ASGI app in a coroutine, so I don't think we'll be needing this change

that said, I pushed a test, some type annotations and some doctoring changes in case this change is useful in general. I don't know if my wording in the docstrings is particularly helpful, the terminology around this stuff is quite obtuse / confusing.

src/anyio/from_thread.py
adriangb: not really, I checked and all tests still pass without it, I was just playing it safe. I removed it in 22c2ccb

tests/test_from_thread.py
```diff
@@ -187,6 +187,31 @@ def external_thread() -> None:
...
+        class AsyncGetThreadId:
+            def __init__(self) -> None:
+                self.thread_id = threading.get_ident()
+
+            def __await__(self) -> Iterator[int]:
+                assert self.thread_id == threading.get_ident()
+
+                async def coro() -> int:
+                    return self.thread_id
+                return coro().__await__()
```
adriangb, line 200: To be honest this is a bit lower-level into the asyncio world than I am familiar with, so this is the best I could come up with that made sense and worked, but I'm open to suggestions on how this can be made simpler w/o the closure

adriangb:
@agronholm is there anything I'm missing to move forward with this?

agronholm:
No, the ball is in my court. I need some time to ensure that the API is consistent everywhere regarding what kind of objects are accepted when starting tasks. That will determine whether this makes sense or not. I'm looking at `start_soon()` and `start()` in `TaskGroup` in particular.

agronholm:
Done in 538b739b.

PR #421, opened 2022-02-21
Support setting subprocess std* arguments to file objects or None

### LINKED ISSUE #419: Support passing file objects and None as std* arguments of `open_process()` and `run_process()`
Would it be possible for the `stdin`, `stdout`, and `stderr` arguments of `open_process()` and `run_process()` to be expanded to accept both `None` (to inherit the file descriptor from the parent) and (synchronous) file-like objects (to read from/write to the given file)?  Based on their documentation, both [asyncio](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.subprocess_exec) and [trio](https://trio.readthedocs.io/en/v0.19.0/reference-io.html#trio.open_process) seem to support such arguments.

jwodder:
Closes #419.

So it turns out that there's nothing currently in the code to stop users from pasing file objects or `None` to `open_process()`, so all this PR needed to do was update the type annotations & docs and add tests.

Regarding the type annotations for the `std*` arguments, [typeshed says][1] that asyncio takes `Union[int, IO[Any], None]`, while [trio-typing says][2] that trio takes `int`, `None`, or anything with a `fileno()` method.  As [`IO` comes with a `fileno()`][3], that makes it the stricter choice and thus the one supported by both backends.

As far as I'm aware, the only thing really missing from this PR is a test of `stdin=None`, but I'm not sure how to go about writing that.  Thoughts?

[1]: https://github.com/python/typeshed/blob/3afc0b1adeac3d74674c32605f112bdbfe24b530/stdlib/asyncio/subprocess.pyi
[2]: https://github.com/python-trio/trio-typing/blob/34fa626eaa971c50988eb54f71ddbe4384535b4a/trio-stubs/__init__.pyi
[3]: https://github.com/python/cpython/blob/c596ecbf821843de0e044f0d4da34c6b49a06907/Lib/typing.py#L2503


tests/test_subprocesses.py
```diff
@@ -99,3 +99,36 @@ async def test_process_new_session_sid() -> None:
...
+                'import sys; txt = sys.stdin.read().strip(); '
+                'print("stdin says", repr(txt), "but stderr says NO!", file=sys.stderr); '
+                'print("stdin says", repr(txt), "and stdout says YES!")'
+            ],
+            stdin=fin,
+            stdout=fout,
+            stderr=ferr,
+        ) as p:
+            assert await p.wait() == 0
+    assert stdoutfile.read_text() == "stdin says 'Hello, process!' and stdout says YES!\n"
```
agronholm, line 124: Add a blank line after a control block if the code continues on the same level.
```suggestion

    assert stdoutfile.read_text() == "stdin says 'Hello, process!' and stdout says YES!\n"
```

src/anyio/_core/_subprocesses.py
```diff
@@ -83,10 +88,12 @@ async def open_process(command: str | Sequence[str], *, stdin: int = PIPE,
...
-    :param stdin: either :data:`subprocess.PIPE` or :data:`subprocess.DEVNULL`
-    :param stdout: either :data:`subprocess.PIPE` or :data:`subprocess.DEVNULL`
-    :param stderr: one of :data:`subprocess.PIPE`, :data:`subprocess.DEVNULL` or
-        :data:`subprocess.STDOUT`
+    :param stdin: one of :data:`subprocess.PIPE`, :data:`subprocess.DEVNULL`, a
+        file-like object, or `None`
+    :param stdout: one of :data:`subprocess.PIPE`, :data:`subprocess.DEVNULL`,
+        a file-like object, or `None`
+    :param stderr: one of :data:`subprocess.PIPE`, :data:`subprocess.DEVNULL`,
+        :data:`subprocess.STDOUT`, a file-like object, or `None`
```
agronholm, line 96: Switched None to use two backticks as everywhere else in the codebase.
```suggestion
    :param stdin: one of :data:`subprocess.PIPE`, :data:`subprocess.DEVNULL`, a
        file-like object, or ``None``
    :param stdout: one of :data:`subprocess.PIPE`, :data:`subprocess.DEVNULL`,
        a file-like object, or ``None``
    :param stderr: one of :data:`subprocess.PIPE`, :data:`subprocess.DEVNULL`,
        :data:`subprocess.STDOUT`, a file-like object, or ``None``
```

agronholm (changes requested):
Just a few code style nits, otherwise LGTM.

agronholm:
Needs a few fixes still; I'm on it.

agronholm (approved):

agronholm:
Ok, seems to work now. The py3.11 failure is due to coverage incompatibility.

agronholm:
Thanks!

jmehnle:
@agronholm, for lack of a better forum (particularly, app.gitter.im seems to be down right now), is there a good reason `anyio.run_process()` can't accept a `stdin` argument with a file handle (such as `sys.stdin`)? Both [`subprocess.run()`](https://docs.python.org/3/library/subprocess.html#subprocess.run) and [`trio.run_process()`](https://trio.readthedocs.io/en/stable/reference-io.html#trio.run_process) support this, yet `anyio.run_process()` only accepts a static `input` string or bytes. (Why is this useful? I'm writing a shell wrapper that spawns a user-provided subprocess, which may or may not expect user interaction on `stdin`. This works with Trio but not with AnyIO.)

agronholm:
Does it work with asyncio's subprocess API too? If so, I could accept a PR to allow that.

jmehnle:
It seems to work on both `asyncio` and `trio`: https://github.com/agronholm/anyio/pull/859

Issue #430, opened 2022-03-29
Python 3.11: @asyncio.coroutine test failures


hroncok:
Hello. On Python 3.11.0a6, the tests fail with `AttributeError: module 'asyncio' has no attribute 'coroutine'`.

To reproduce, temporarily remove uvloop from test extra:

```diff
diff --git a/setup.cfg b/setup.cfg
index 63d5592..153cdbc 100644
--- a/setup.cfg
+++ b/setup.cfg
@@ -50,8 +50,6 @@ test =
     pytest >= 7.0
     pytest-mock >= 3.6.1
     trustme
-    uvloop < 0.15; python_version < '3.7' and (platform_python_implementation == 'CPython' and platform_system != 'Windows')
-    uvloop >= 0.15; python_version >= '3.7' and (platform_python_implementation == 'CPython' and platform_system != 'Windows')
 trio = trio >= 0.16
 doc =
     packaging
```

And run:

```
$ tox -re py311 -- -k 'test_wait_generator_based_task_blocked or test_cancel_generator_based_task'
py311 inst-nodeps: .../anyio/.tox/.tmp/package/1/anyio-3.5.0.post2+dirty.tar.gz
py311 installed: anyio @ file://.../anyio/.tox/.tmp/package/1/anyio-3.5.0.post2%2Bdirty.tar.gz,async-generator==1.10,attrs==21.4.0,cffi==1.15.0,coverage==6.3.2,cryptography==36.0.2,hypothesis==6.40.0,idna==3.3,iniconfig==1.1.1,outcome==1.1.0,packaging==21.3,pluggy==1.0.0,py==1.11.0,pycparser==2.21,pyparsing==3.0.7,pytest==7.1.1,pytest-mock==3.7.0,sniffio==1.2.0,sortedcontainers==2.4.0,tomli==2.0.1,trio==0.20.0,trustme==0.9.0
py311 run-test-pre: PYTHONHASHSEED='4277600544'
py311 run-test: commands[0] | coverage run -m pytest -k 'test_wait_generator_based_task_blocked or test_cancel_generator_based_task'
============================= test session starts ==============================
platform linux -- Python 3.11.0a6, pytest-7.1.1, pluggy-1.0.0
cachedir: .tox/py311/.pytest_cache
rootdir: .../anyio, configfile: pyproject.toml, testpaths: tests
plugins: anyio-3.5.0.post2+dirty, mock-3.7.0, hypothesis-6.40.0
[...]

PR #431, opened 2022-03-29
Python 3.11: @asyncio.coroutine decorator no longer available

hroncok:
Fixes https://github.com/agronholm/anyio/issues/430


hroncok:
I've amended this to be like 5fbe4effb7b7d3d7c8b80567347f0dd543698b89

agronholm:
Thanks.

PR #434, opened 2022-04-23
getaddrinfo does not take bytearray

hauntsaninja:
This matches the type in typeshed. asyncio just delegates directly to `socket.getaddrinfo`, which does not take bytearray.


JelleZijlstra:
Confirmed that getaddrinfo indeed doesn't accept bytearray: https://github.com/python/cpython/blob/b4e048411f4c62ad7343bca32c307f0bf5ef74b4/Modules/socketmodule.c#L6477

agronholm (approved):
Yup, indeed there is no bytearray in typeshed's `getaddrinfo()`.

PR #436, opened 2022-05-11
docs: run_sync_in_worker_thread -> run_sync

maartenbreddels:
Hi,

I think this is changed in version 3.0, but was missed in the docs.
Nice library, solves a lot of issues that I usually hand crafted.

Regards,

Maarten Breddels

### DIFF
--- a/docs/threads.rst
+++ b/docs/threads.rst
@@ -50,8 +50,8 @@ If you need to call a coroutine function from a worker thread, you can do this::
 
     run(main)
 
-.. note:: The worker thread must have been spawned using :func:`~run_sync`
-   for this to work.
+.. note:: The worker thread must have been spawned using :func:`~to_thread.run_sync` for this to
+   work.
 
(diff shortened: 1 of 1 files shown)


docs/threads.rst
```diff
@@ -50,7 +50,7 @@ If you need to call a coroutine function from a worker thread, you can do this::
 
     run(main)
 
-.. note:: The worker thread must have been spawned using :func:`~run_sync_in_worker_thread`
+.. note:: The worker thread must have been spawned using :func:`~run_sync`
    for this to work.
```
agronholm, line 54: This requires the `to_thread` prefix in order to work.
```suggestion
.. note:: The worker thread must have been spawned using :func:`~to_thread.run_sync` for this to
   work.
```

agronholm (approved):
Needed one little fix.

agronholm:
Thanks!

PR #446, opened 2022-06-22
Implemented UNIX datagram sockets. Closes #445

### LINKED ISSUE #445: Datagram streams on Unix sockets
I was trying to implement an asynchronous version of `SysLogHandler` using `anyio`, and I got stuck on the issue that the current `Stream` APIs do not handle datagram unix sockets.

I would propose adding the following APIs to `anyio`:

```python
class UnixDatagramSocket(anyio.abc.UnreliableObjectStream[Tuple[bytes, str]], anyio.abc._sockets._SocketProvider):
    async def sendto(data, path): ...

class ConnectedUnixDatagramSocket(anyio.abc.UnreliableObjectStream[bytes], anyio.abc._sockets._SocketProvider): ...

async def create_unix_datagram_socket(*, local_path: str) -> UnixDatagramSocket: ...

async def create_connected_unix_datagram_socket(remote_path: str, *, local_path: str) -> ConnectedDatagramSocket: ...
```

From my preliminary analysis, adding these APIs to `anyio` does not look like it would be too onerous, and I would be happy to submit a PR if you have any interest in such a feature.

I also wonder if you would find it useful/relevant to define `DatagramSocket[T]` and `ConnectedDatagramSocket` types in `anyio`, which would be common bases for `UnixDatagramSocket`/`UDPSocket` and `ConnectedUnixDatagramSocket`/`ConnectedUDPSocket`?

jhominal:
Closes #445.

As promised on #445, here is a PR that implements UNIX datagram sockets.

Here are a few notes on some implementation choices that were made on this PR which merit mention:

 * for the `asyncio` implementation, I chose not to rely on `AbstractEventLoop.create_datagram_endpoint` because support for UNIX datagram sockets was only added in Python 3.7, and using `socket.socket` objects directly means that a lot of the implementation can be shared with the existing `UNIXSocketStream`;
 * for the `asyncio` implementation, while refactoring a number of fields of `UNIXSocketStream` into a `_RawSocketMixin` class, the instance field `_loop` was dropped - given that most methods call `asyncio.get_running_loop` in order to get hold of the event loop instance, I did not see what remaining use the `_loop` instance served;
 * in the`[Connected]UNIXDatagramSocket.send`, the number of written bytes returned by `socket.send[to]` is not read - in the case where the datagram was only partially written on the socket, there is nothing that can be done for the application to handle that situation, as trying to send the remaining bytes would simply send a second incomplete datagram;
 * the size of the UNIX datagram socket recv buffer is hardcoded to 65536 - that size was chosen because it is already the limit configured for UDP sockets in `trio`;


agronholm:
I'll try to get this reviewed before Monday.

src/anyio/_backends/_asyncio.py
```diff
@@ -1741,6 +1814,32 @@ async def create_udp_socket(
...
+                f: asyncio.Future = asyncio.Future()
+                loop.add_writer(raw_socket, f.set_result, None)
+                f.add_done_callback(lambda _: loop.remove_writer(raw_socket))
+                await f
+            except BaseException:
+                raw_socket.close()
+                raise
+            else:
+                return ConnectedUNIXDatagramSocket(raw_socket)
+
```
agronholm, line 1838:
```suggestion
```

agronholm (approved):
I can't find any fault with this. You, sir, are a true professional.

PR #450, opened 2022-06-30
Fix type for max_buffer_size, it should be either int or math.inf

lalinsky:
You currently can't use simple int without upsetting mypy.

### DIFF
--- a/src/anyio/_core/_streams.py
+++ b/src/anyio/_core/_streams.py
@@ -1,5 +1,5 @@
 import math
-from typing import Any, Optional, Tuple, Type, TypeVar, overload
+from typing import Any, Literal, Optional, Tuple, Type, TypeVar, Union, overload
 
 from ..streams.memory import (
     MemoryObjectReceiveStream,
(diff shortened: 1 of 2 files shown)


lalinsky:
Hm, unfortunately we can't use `Literal[math.inf]`, so it will have to be `Union[int, float]`.

agronholm:
According to the [numeric tower](https://peps.python.org/pep-0484/#the-numeric-tower) in PEP 484, `float` is sufficient.

agronholm:
You say you got a mypy error passing `int` to a function that takes `float`? Can you show the code and the error please?

lalinsky:
```
audio_manager/server.py:272: error: No overload variant of "create_memory_object_stream" matches argument types "int", "Type[Playlist]"
audio_manager/server.py:272: note: Possible overload variants:
audio_manager/server.py:272: note:     def [T_Item] create_memory_object_stream(max_buffer_size: float, item_type: Type[T_Item]) -> Tuple[MemoryObjectSendStream[T_Item], MemoryObjectReceiveStream[T_Item]]
audio_manager/server.py:272: note:     def create_memory_object_stream(max_buffer_size: float = ...) -> Tuple[MemoryObjectSendStream[Any], MemoryObjectReceiveStream[Any]]
```

lalinsky:
I'll close this PR, as the changes are wrong and I can't get any version working with the overloading.

agronholm:
One more thing: did you misspell the second argument?

lalinsky:
Now that's really embarrassing. Yes, I did misspell the argument. It works fine. The mypy error message confused me and got me thinking the problem is something completely different. Thank you and sorry for wasting your time on this.

PR #451, opened 2022-06-30
Fix task reference count leak

amos402:
I found it as I'm using `httpx` which used `anyio` as its backend, there is a circle between the scope and parent scope, so the `asyncio.Task` won't be able to be released until the gc called.
Therefore we need to break the reference chain.

### DIFF
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -465,13 +465,13 @@ class CancelScope(BaseCancelScope):
     def _parent_cancelled(self) -> bool:
         # Check whether any parent has been cancelled
         cancel_scope = self._parent_scope
-        self._parent_scope = None
         while cancel_scope is not None and not cancel_scope._shield:
+            parent_scope = cancel_scope._parent_scope
+            cancel_scope._parent_scope = None
             if cancel_scope._cancel_called:
                 return True
             else:
-                cancel_scope = cancel_scope._parent_scope
-                cancel_scope._parent_scope = None
+                cancel_scope = parent_scope
         return False
 
     def cancel(self) -> DeprecatedAwaitable:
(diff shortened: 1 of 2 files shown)


agronholm:
How did the problem manifest itself in practice?

agronholm:
Could you elaborate on this a bit? I'm not understanding the problem, or your solution, or the test.

amos402:
The main problem is the Task object may still be referred to the CancelScope after the scope is exited.
```python
with TaskGroup() as tg:
    # host_task created in tg.__enter__
    tg.start_soon(...)
# After tg.__exit__, host_task still in tg's `cancel_scope`
```
Let's assume there're some circle references between the scopes, host_task would be collected until the GC is called. That said, if an exception occurred in the Task, you wouldn't know it in time if you don't add the done callback for the Task.

A simple example describes the above condition:
```python
async def do_raise():
    raise Exception

class A: pass


def case1():
    gc.disable()
    a = A()
    a.ref = a
    a.task = asyncio.ensure_future(do_raise())

def case2():
    asyncio.ensure_future(do_raise())

def case3():
    def on_done(fut):
        # catch the exception
    a = A()
    a.ref = a
    a.task = asyncio.ensure_future(do_raise())
    a.task.add_done_callback(on_done)
```

case1: You wouldn't know the exception(the warning of `exception never retrieved`), until the `a` object is collected as garbage.
case2: You would know that after the exception is raised.
case3: You owned the task by yourself, so it's your responsibility to catch the exception.

case1 may encounter in practice, as a library, we should keep our references saint and try to not depend on the GC.
Moreover, as the semantic, `__exit__` should clear the things which `__enter__` created.
The test is for ensured there was no other strong reference attached to the task after the group exited.

For the time being, I still need to fix one bug for this commit, a test enter dead-loop as I break the circle too early. It keeps call `CancelScope_deliver_cancellation`.

agronholm:
Would you mind fixing the conflicts? I would like to include this in the upcoming 4.0 release.

agronholm:
Closing due to lack of activity. I'm not convinced yet that this was a real problem. Let me know if you want to continue working on this. If there is a memory leak due to circular references, that would have to be proven.

PR #455, opened 2022-07-26
Order the waiting queue for asyncio ConcurrencyLimiter

ConorStevenson:
Previously we were using a LIFO, which is unfair and doesn't match `trio`'s implementation.

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -3,9 +3,10 @@ Version history
 
 This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
-**3.6.2**
+**UNRELEASED**
 
-- Fixed ``asyncio`` ``ConcurrencyLimiter`` to correctly order waiting tasks.
+- Fixed ``ConcurrencyLimiter`` on the asyncio backend to order waiting tasks in the FIFO
+  order (instead of LIFO) (PR by Conor Stevenson)
 
 **3.6.1**
 
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -3,6 +3,10 @@ Version history
 
 This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
+**3.6.2**
+
+- Fixed ``asyncio`` ``ConcurrencyLimiter`` to correctly order waiting tasks.
+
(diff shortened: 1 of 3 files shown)


tests/test_synchronization.py
```diff
@@ -524,3 +524,21 @@ async def acquire() -> None:
...
         task1.cancel()
         await asyncio.wait_for(task2, 1)
+
+    async def test_ordered_queue(self) -> None:
+        limiter = CapacityLimiter(1)
+        results = []
+        event = Event()
+
+        async def append(x: int, task_status: TaskStatus) -> None:
+            task_status.started()
```
agronholm, line 534: I think using `start()` instead of `start_soon()` is pointless here.

tests/test_synchronization.py
```diff
@@ -524,3 +524,21 @@ async def acquire() -> None:
...
+            async with limiter:
+                await event.wait()
+                results.append(x)
+
+        async with create_task_group() as tg:
+            for i in [0, 1, 2]:
+                await tg.start(append, i)
+            event.set()
```
agronholm, line 542: Blank line after a control block.
```suggestion

            event.set()
```

agronholm (changes requested):
Just minor changes needed. And would you mind adding a changelog entry too?

tests/test_synchronization.py
ConorStevenson: `start_soon()` didn't seem to guarantee the order of the tasks for `trio`, though the other backends worked fine. Is that normal `trio` behaviour?

docs/versionhistory.rst
```diff
@@ -3,6 +3,10 @@ Version history
 
 This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
+**3.6.2**
+
+- Fixed ``asyncio`` ``ConcurrencyLimiter`` to correctly order waiting tasks.
```
```suggestion
**UNRELEASED**

- Fixed ``ConcurrencyLimiter`` on the asyncio backend to order waiting tasks in the FIFO
  order (instead of LIFO) (PR by Conor Stevenson)
```

agronholm (approved):

agronholm:
Thanks! This was nice and clean, and useful too.

PR #460, opened 2022-08-23
Add note for UNIX datagram sockets

jhominal:
After checking back on anyio recent changes, I realized that I did not add a note for the new feature of UNIX datagram sockets to versionhistory.rst.


agronholm:
Thanks!

PR #464, opened 2022-09-19
Make Listener covariant in the stream type variable

mila:
This changes allows to properly annotate return type in a function like:

```python
import ssl
from typing import Optional

import anyio
from anyio.abc import ByteStream, Listener
from anyio.streams.tls import TLSListener


async def create_listener(
    local_host: str,
    local_port: int,
    ssl_context: Optional[ssl.SSLContext],
) -> Listener[ByteStream]:
    listener = await anyio.create_tcp_listener(
        local_host=local_host, local_port=local_port
    )
    if ssl_context is not None:
        return TLSListener(listener, ssl_context=ssl_context)
    else:
        return listener
```

Without this change, the function has to be annotated as returning `Union[Listener[SocketStream], TLSListener]`, which is long and it depends on specific implementation.

The stream type of the listeners acts [covariantly](https://peps.python.org/pep-0484/#covariance-and-contravariance). When I have a handler that expects `ByteStream`, I need `Listener[ByteStream]`. But the handler won't mind if it gets `TLSStream`, so I should not mind if I have `Listener[TLSStream]`. 

For comparison, [callables are covariant in the return type](https://peps.python.org/pep-0483/#covariance-and-contravariance). For the listeners, the socket acts similar as the return value of the callable - it is an output, just passed to the handler callback instead of returned directly.

Note that I did not propagate the change to `anyio.streams.stapled.MultiListener`. The reason is similar why mutable collections are not covariant. The following function should not accept `MultiListener[TLSStream]`:

```python
def replace_listeners(multi_listener: MultiListener[ByteStream]):
    multi_listener.listeners = [anyio.create_tcp_listener(...)]
```

I also did a small change to `anyio.abc.SocketListener`. Its use of `T_Stream` seems unnecessary, so I got rid of it instead of replacing it by `T_Stream_co`.

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -25,6 +25,7 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
   - The ``TaskStatus`` class is now generic, and should be parametrized to indicate the
     type of the value passed to ``task_status.started()``
+  - The ``Listener`` class is now covariant in its stream type
 - Fixed ``CapacityLimiter`` on the asyncio backend to order waiting tasks in the FIFO
   order (instead of LIFO) (PR by Conor Stevenson)
 - Fixed ``CancelScope.cancel()`` not working on asyncio if called before entering the
--- a/src/anyio/abc/_sockets.py
+++ b/src/anyio/abc/_sockets.py
@@ -14,7 +14,7 @@ from .._core._typedattr import (
     TypedAttributeSet,
     typed_attribute,
 )
-from ._streams import ByteStream, Listener, T_Stream, UnreliableObjectStream
+from ._streams import ByteStream, Listener, UnreliableObjectStream
 from ._tasks import TaskGroup
 
 IPAddressType = Union[str, IPv4Address, IPv6Address]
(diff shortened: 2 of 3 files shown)


mila:
The fail of `pre-commit.ci` is unrelated, so I addressed it in https://github.com/agronholm/anyio/pull/465.

agronholm:
Would you mind fixing the conflicts? I recently merged the 4.0-dev branch to master in preparation for a major release. I would like this change to be a part of that release.

docs/versionhistory.rst
```diff
@@ -25,6 +25,7 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
   - The ``TaskStatus`` class is now generic, and should be parametrized to indicate the
     type of the value passed to ``task_status.started()``
+  - The ``Listener`` class is now covariant in its stream type.
```
agronholm, line 28:
```suggestion
  - The ``Listener`` class is now covariant in its stream type
```

mila: I force-pushed a change that removes the dot.

agronholm (approved):
Checked with mypy, LGTM.

agronholm:
Thanks!

mila:
Thank you for merging.

BTW I am looking forward to the 4.0 release. I looked at the changelog today and found that at least three of the changes will help me somehow.

PR #472, opened 2022-10-08
Catch RuntimeError in finalization of start_blocking_portal()

### LINKED ISSUE #471: `BaseException` not returned though `blockingportal.start_task_soon(...).result()`
Take this code:
```python
async def test() -> None:
    raise KeyboardInterrupt

    # Instead of the above raise statement, we can use prompt_toolkit to read
    # input. Pressing control-c raises a `KeyboardInterrupt` here. (That's not
    # the default SIGINT handler, which would be handled in the main thread.)
    from prompt_toolkit import PromptSession  # pip install prompt-toolkit

    await PromptSession().prompt_async(">")

    # Instead of `KeyboardInterrupt`, raising `BaseException` yields similar
    # results.
    raise BaseException("test")


def main() -> None:
    with start_blocking_portal() as portal:
        portal.start_task_soon(test).result()

        # We never get here:
        # - Either the code just hangs,
        # - the portal gets killed
        print("------")
        portal.start_task_soon(test).result()


if __name__ == "__main__":
    main()
```

The outcome of this is very nondeterministic.

Sometimes it hangs. Sometimes we get a short traceback. Sometimes a very long traceback. Usually the portal gets killed and becomes unusable.

What I'd expect is that any exception raised by our `test` function is returned through the `concurrent.futures.Future`, not killing the portal. A `KeyboardInterrupt` is a valid outcome of an async function and should not kill the portal.

What worries me most is that it's nondeterministic, and that it sometimes hangs, as if there is a race condition.

agronholm:
Fixes #471.

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -30,6 +30,9 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
   order (instead of LIFO) (PR by Conor Stevenson)
 - Fixed ``CancelScope.cancel()`` not working on asyncio if called before entering the
   scope
+- Fixed ``start_blocking_portal()`` raising an unwarranted
+  ``RuntimeError: This portal is not running`` if a task raises an exception that causes
+  the event loop to be closed
 
 **3.6.1**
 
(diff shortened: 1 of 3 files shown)


smurfix (approved):
Looks sane at first glance. Unfortunately I have no time to do an in-depth analysis

agronholm:
Thanks for looking.

jonathanslenders:
Thank you!

PR #473, opened 2022-10-13
Allow passing only `item_type` to `create_memory_object_stream()`

gschaffner:
Hi! Calls with a signature like `create_memory_object_stream(item_type=int)` are supported by the implementation but not by the overloads.


agronholm:
If you rebase this on top of `master`, that should fix the pyright errors too.

agronholm (approved):

agronholm:
Thanks!

PR #477, opened 2022-10-17
Fixed to_thread.run_sync() hanging on asyncio on StopIteration

### LINKED ISSUE #475: `StopIteration` in a generator in a thread behaves differently in asyncio and Trio
`StopIteration` in a generator in a thread behaves differently in asyncio and Trio

## Asyncio

This code will show an error on the terminal and hang there forever:

```Python
import anyio

def hang():
    return next(item for item in [])

async def main():
    return await anyio.to_thread.run_sync(hang)

result = anyio.run(main)
```

The error shows:

```
Exception in callback WorkerThread._report_result(<Future pendi...ask_wakeup()]>, None, StopIteration())
handle: <Handle WorkerThread._report_result(<Future pendi...ask_wakeup()]>, None, StopIteration())>
  File "/Users/user/miniconda3/envs/python3.10/lib/python3.10/asyncio/events.py", line 80, in _run
    self._context.run(self._callback, *self._args)
  File "/Users/user/code/starlette/env3.10/lib/python3.10/site-packages/anyio/_backends/_asyncio.py", line 736, in _report_result
    future.set_exception(exc)
TypeError: StopIteration interacts badly with generators and cannot be raised into a Future
```

But I suspect the exception is being printed but not really raised, I'm not even sure. If I run it through the VS Code debugger the exception is not shown in the debugger, only printed in the terminal.

## Trio

Changing only the backend to Trio:

```Python
import anyio

def hang():
    return next(item for item in [])

async def main():
    return await anyio.to_thread.run_sync(hang)

result = anyio.run(main, backend="trio")
```

It raises a `RuntimeError` wrapping (`raise from`) the `StopIteration` error, and the program stops with that error instead of hanging:

```
  File "/Users/user/code/starlette/env3.10/lib/python3.10/site-packages/trio/_core/_traps.py", line 166, in wait_task_rescheduled
    return (await _async_yield(WaitTaskRescheduled(abort_func))).unwrap()
  File "/Users/user/code/starlette/env3.10/lib/python3.10/site-packages/outcome/_impl.py", line 138, in unwrap
    raise captured_error
  File "/Users/user/code/starlette/env3.10/lib/python3.10/site-packages/trio/_threads.py", line 157, in do_release_then_return_result
    return result.unwrap()
  File "/Users/user/code/starlette/env3.10/lib/python3.10/site-packages/outcome/_impl.py", line 138, in unwrap
    raise captured_error
  File "/Users/user/code/starlette/env3.10/lib/python3.10/site-packages/trio/_threads.py", line 170, in worker_fn
    ret = sync_fn(*args)
  File "/Users/user/code/starlette/env3.10/lib/python3.10/site-packages/anyio/_backends/_trio.py", line 175, in wrapper
    return func(*args)
  File "/Users/user/code/starlette/sandt_no_ex.py", line 4, in hang
    return next(item for item in [])
StopIteration

  File "/Users/user/code/starlette/sandt_no_ex.py", line 9, in <module>
    result = anyio.run(main, backend="trio")
  File "/Users/user/code/starlette/env3.10/lib/python3.10/site-packages/anyio/_core/_eventloop.py", line 56, in run
    return asynclib.run(func, *args, **backend_options)
  File "/Users/user/code/starlette/env3.10/lib/python3.10/site-packages/trio/_core/_run.py", line 1932, in run
    raise runner.main_task_outcome.error
  File "/Users/user/code/starlette/sandt_no_ex.py", line 7, in main
    return await anyio.to_thread.run_sync(hang)
  File "/Users/user/code/starlette/env3.10/lib/python3.10/site-packages/anyio/to_thread.py", line 28, in run_sync
    return await get_asynclib().run_sync_in_worker_thread(func, *args, cancellable=cancellable,
  File "/Users/user/code/starlette/env3.10/lib/python3.10/site-packages/anyio/_backends/_trio.py", line 180, in run_sync_in_worker_thread
    return await run_sync(context.run, wrapper, cancellable=cancellable, limiter=limiter)
  File "/Users/user/code/starlette/env3.10/lib/python3.10/site-packages/trio/_threads.py", line 207, in to_thread_run_sync
    return await trio.lowlevel.wait_task_rescheduled(abort)
RuntimeError: coroutine raised StopIteration
```

## Problem

This is a problem in FastAPI (via Starlette) because a sync (`def`) function that raises a `StopIteration` error hangs in there and never returns the response (it should return a 500 Server Error).

```Python
from fastapi import FastAPI


app = FastAPI()

@app.get("/hang")
def hang():
    return next(item for item in [])
```

## What to do

I see several places where this could be handled.

* On asyncio directly. Not sure how feasible that is, and it would still only work for some future Python, so the fix would still be needed in other places for current Python versions.

* On AnyIO: it could replicate the Trio behavior, even on asyncio. I guess that would mean AnyIO would have a more predictable behavior (in this specific corner case) independent of backend. But I understand if you would prefer to keep that backend-specific and not touch that.

* On Starlette: if you would rather not have it on AnyIO, the next place to put it would be on Starlette. Although I guess it's possible they would not want to do that and expect users to clean up their own code.

* On FastAPI: if none of the previous places would like to have that changed.

So, here's the question. Would you like a fix/change for that here on AnyIO?

agronholm:
Fixes #475.

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -33,6 +33,8 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
   scope
 - Fixed ``open_signal_receiver()`` inconsistently yielding integers instead of
   ``signal.Signals`` instances on the ``trio`` backend
+- Fixed ``to_thread.run_sync()`` hanging on asyncio if the target callable raises
+  ``StopIteration``
 
 **3.6.1**
 
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -674,6 +674,11 @@ class WorkerThread(Thread):
 
         if not future.cancelled():
             if exc is not None:
+                if isinstance(exc, StopIteration):
+                    new_exc = RuntimeError("coroutine raised StopIteration")
+                    new_exc.__cause__ = exc
+                    exc = new_exc
+
                 future.set_exception(exc)
             else:
                 future.set_result(result)
--- a/tests/test_to_thread.py
+++ b/tests/test_to_thread.py
@@ -265,3 +265,17 @@ def test_asyncio_no_recycle_stopping_worker(
     task1 = asyncio_event_loop.create_task(taskfunc1())
     task2 = asyncio_event_loop.create_task(taskfunc2())
     asyncio_event_loop.run_until_complete(asyncio.gather(task1, task2))
+
+
+async def test_stopiteration() -> None:
+    """
+    Test that raising StopIteration in a worker thread raises a RuntimeError on the
+    caller.
+
+    """
+
+    def raise_stopiteration() -> NoReturn:
+        raise StopIteration
+
+    with pytest.raises(RuntimeError, match="coroutine raised StopIteration"):
+        await to_thread.run_sync(raise_stopiteration)


tiangolo (approved):
Amazing, thank you! 🚀

PR #483, opened 2022-10-19
Made object streams co/contravariant

### LINKED ISSUE #466: `T_Item` for `MemoryObject{Send,Receive}Stream` should be {contra,co}variant.
Take this example snippet, and run ``mypy --strict`` over the code:

```python
from anyio.streams.memory import MemoryObjectReceiveStream, MemoryObjectSendStream

# * MemoryObjectReceiveStream issue *

stream: MemoryObjectReceiveStream[bytes]
stream2: MemoryObjectReceiveStream[bytes | BrokenPipeError]

# This should be fine, because when we're in a situation that can handle
# receiving both A and B, having something from which we receive only A, we're
# good. This fails, because `T_Item` in anyio.streams.memory is not covariant.
stream2 = stream

# * MemoryObjectSendStream issue *

stream: MemoryObjectSendStream[bytes | BrokenPipeError]
stream2: MemoryObjectSendStream[bytes]

# This should be fine, because when we're in a situation that can send both A
# and B, but we promise to only send A, we're good. This fails, because
# `T_Item` in anyio.streams.memory is not contravariant.
stream2 = stream
```

This would mean that the same type variable can't be used for both class definitions, because in one class it should be covariant, while in the other, it should be contravariant.

agronholm:
Fixes #466.

### DIFF
--- a/tests/streams/test_memory.py
+++ b/tests/streams/test_memory.py
@@ -13,6 +13,7 @@ from anyio import (
     fail_after,
     wait_all_tasks_blocked,
 )
+from anyio.abc import ObjectReceiveStream, ObjectSendStream
 from anyio.streams.memory import MemoryObjectReceiveStream, MemoryObjectSendStream
 
 pytestmark = pytest.mark.anyio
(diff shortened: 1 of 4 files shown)


tests/streams/test_memory.py
```diff
@@ -345,3 +346,18 @@ async def test_sync_close() -> None:
 
     with pytest.raises(ClosedResourceError):
         receive_stream.receive_nowait()
+
+
+async def test_type_variance() -> None:
```
jonathanslenders, line 351: <3

PR #487, opened 2022-10-25
Added the receive_nowait() method to all streams

agronholm:


agronholm:
I just realized that this PR requires backwards incompatible changes, as it introduces a new mandatory stream method that must raise either `WouldBlock` or `EndOfStream` depending on the state of the stream.

agronholm:
I've made the changes compatible with v4.x by only emitting a warning when `receive_nowait()` hasn't been implemented.

agronholm:
All that this still needs is added test coverage. The TLSStream implementation looks particularly tricky.

PR #491, opened 2022-11-04
typing: asyncio backend: fix `TaskGroup.start` mistyping and avoid `Any` returns

gschaffner:
This mistyping isn't user-facing, but it still helps with backend type checking coverage. A couple `asyncio.Future`s can also be parametrized now since `__future__` annotations was recently enabled.

### DIFF
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -622,7 +622,7 @@ class TaskGroup(abc.TaskGroup):
 
     async def start(
         self, func: Callable[..., Awaitable[Any]], *args: object, name: object = None
-    ) -> None:
+    ) -> object:
         future: asyncio.Future = asyncio.Future()
         task = self._spawn(func, args, name, future)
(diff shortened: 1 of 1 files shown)


agronholm:
Is there any evidence that these changes help with anything? Clearly `start()` returning `None` is wrong, but no checker seems to care.

agronholm:
So what do you suppose `start()` _should_ return then?

agronholm:
The trio-typing approach is unsustainable as it only works for a set number of positional arguments. I'd rather return `Any` instead until the Python typing system is fixed to cover this use case.

agronholm:
> What about `trio-typing`'s use of `mypy_extensions.NamedArg`? anyio could probably use `NamedArg` with an overload such that Mypy users get a proper return type and non-Mypy users get `-> Any` behavior still.

Do you know how to use it so that we don't have to specify an explicit number of positional arguments (and also not cause other type checkers to flip out)?

PR #494, opened 2022-11-04
🔥 Remove `start_blocking_portal` from `__all__,` as it seems it was removed

tiangolo:
🔥 Remove `start_blocking_portal` from `__all__,` as it seems it was removed


agronholm:
Yeah, it was moved to `anyio.from_thread`. Good catch.

agronholm (approved):

PR #495, opened 2022-11-06
Make `create_memory_object_stream` generic

gschaffner:
This supports annotating memory object streams with uninstantiable types that the current `item_type` hack currently does not work for, e.g.

```python
create_memory_object_stream(item_type=Callable)
# error: No overload variant of "create_memory_object_stream" matches argument type
# "_SpecialForm"  [call-overload]
```

(See also the discussion around https://matrix.to/#/!JfFIjeKHlqEVmAsxYP:gitter.im/$sGb5WyohiQVFuiPdfb0GZ54h-CqN76sxeKKiCWovCdY?via=gitter.im&via=matrix.org)

@agronholm it turns out I was mistaken—due to a bug(?) in mypy, what I thought was a way to do this in a backwards compatible way isn't quite backwards compatible. While c93f3ec6951982a360d1aad23e7f2bd68d93cbe4 (note: not in this PR) is backwards-compatible at runtime (emitting a deprecation warning currently, but v3 could skip this) and backwards compatible in the sense that mypy will correctly infer types for e.g.

```python
s0, r0 = create_memory_object_stream(item_type=float)
s1, r1 = create_memory_object_stream()
```

it would be a breaking change since it makes unannotated calls emit

```python
s1, r1 = create_memory_object_stream()
# error: Need type annotation for "s1"  [var-annotated]
# error: Need type annotation for "r1"  [var-annotated]
```

on mypy. Note that mypy *does* infer a default item type of `Any` for `s1` and `r1`, but it still wants a type annotation. This is not affected by `--allow-any-generic` or `--disallow-any-generic`. It appears to be a known problem, see https://github.com/python/mypy/issues/3737#issuecomment-322778284.

Given that c93f3ec6951982a360d1aad23e7f2bd68d93cbe4 would produce spurious `var-annotated` on anyio 4 too, I think you are right that this has to be a breaking change in v4.

### DIFF
--- a/docs/streams.rst
+++ b/docs/streams.rst
@@ -49,16 +49,20 @@ The loop exits when all clones of the send stream have been closed.
 Example::
 
     from anyio import create_task_group, create_memory_object_stream, run
+    from anyio.streams.memory import MemoryObjectReceiveStream
 
 
-    async def process_items(receive_stream):
+    async def process_items(receive_stream: MemoryObjectReceiveStream[str]) -> None:
         async with receive_stream:
             async for item in receive_stream:
                 print('received', item)
 
 
     async def main():
-        send_stream, receive_stream = create_memory_object_stream()
+        # The [str] specifies the type of the objects being passed through the
+        # memory object stream. This is a bit of trick, as create_memory_object_stream
+        # is actually a class masquerading as a function.
+        send_stream, receive_stream = create_memory_object_stream[str]()
         async with create_task_group() as tg:
             tg.start_soon(process_items, receive_stream)
             async with send_stream:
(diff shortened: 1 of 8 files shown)


agronholm (approved):
This looks really good! I took the liberty of making some additional changes. Would you mind checking them to make sure I didn't do anything stupid?

agronholm:
> if you don't mind, can you hold off merging this until I figure out why the trio folks did something different? I want to make sure I haven't done something wrong here.

Alright, there's no rush.

agronholm:
@gschaffner I'd like to get 4.0 out in a couple weeks. Have you made progress with your investigation?

agronholm:
I've given this another look, and I can't find any fault in it. Trio's way of doing this is more generic, but potentially also less compatible with type checkers. If there's a problem, we can sort it out during the release candidate phase.

agronholm:
Anyway, thank you for this!

Issue #432, opened 2022-03-30
Number of exceptions in `ExceptionGroup` differs between trio and asyncio


valgarf:
Example reproducing the issue:

```python
import anyio
import sys

async def raise_runtime():
    try:
        await anyio.sleep(999)
    except anyio.get_cancelled_exc_class():
        raise RuntimeError()

async def exc_group():
    with anyio.move_on_after(1):
        try:
            async with anyio.create_task_group() as tg:
                ### Variant 1
                tg.start_soon(anyio.sleep, 999)
                tg.start_soon(anyio.sleep, 999)
                await anyio.sleep(999)
                ### Variant 2
                # await raise_runtime()
                ### Variant 3
                # tg.start_soon(anyio.sleep, 999)
                # await raise_runtime()
        except anyio.get_cancelled_exc_class():
            print("Cancelled")
        except:
            print(sys.exc_info())


if __name__ == "__main__":
    print("--- trio ---")
    anyio.run(exc_group, backend="trio")
    print("--- asyncio ---")
    anyio.run(exc_group, backend="asyncio")
```

Running this yields

```
--- trio ---
(<class 'anyio._backends._trio.ExceptionGroup'>, <ExceptionGroup: Cancelled(), Cancelled(), Cancelled(), Cancelled()>, <traceback object at 0x7f1a0efc3180>)
--- asyncio ---
Cancelled
```
I expected the asyncio behavior - when cancelling a task group I would prefer to get a single `Cancelled` exception if no other exception was thrown. More importantly, I really expected the event loops to produce similar results.

There are also differences for the other variants in the example above.
Variant 2 prints:
```
--- trio ---
(<class 'anyio._backends._trio.ExceptionGroup'>, <ExceptionGroup: RuntimeError(), Cancelled()>, <traceback object at 0x7f00833dbbc0>)
--- asyncio ---
(<class 'RuntimeError'>, RuntimeError(), <traceback object at 0x7f008338ae00>)
```
Variant 3 prints:
 ```
--- trio ---
(<class 'anyio._backends._trio.ExceptionGroup'>, <ExceptionGroup: RuntimeError(), Cancelled(), Cancelled()>, <traceback object at 0x7f40e32b5e00>)
--- asyncio ---
(<class 'anyio._backends._asyncio.ExceptionGroup'>, <ExceptionGroup: RuntimeError(), CancelledError()>, <traceback object at 0x7f40e326d300>)
```

[...]

PR #496, opened 2022-11-06
Improved cancellation semantics

agronholm:
We now use `strict_exception_groups=True` on Trio. Any non-base exceptions raised out of task groups are now wrapped in an exception group. If only cancellation exceptions are raised from an excgroup, they are merged into a single exception. This also takes care of variant 2 at #432.


agronholm:
I'm getting a bit tired trying to figure out why mypy fails on CI but not locally.

agronholm:
Closing in favor of #586.

PR #502, opened 2022-11-16
Replace flake8 by ruff 🚀

JonathanPlasse:
[Ruff](https://github.com/charliermarsh/ruff) is equivalent to flake8 but much faster.
I recommend reading the [launch blog article](https://notes.crmarsh.com/python-tooling-could-be-much-much-faster)

- [x] Fix f-string used as docstring. This would have been interpreted by python as a joined string rather than a docstring.
- [x] Rename and remove, if possible, unused loop control variables
- [x] Fix `return` inside finally blocks cause exceptions to be silenced
- [x] Fix Do not call `getattr` with a constant attribute value. It is not any safer than normal property access.
- [x] Fix Do not `assert False` (`python -O` removes these calls), raise `AssertionError()`
- [x] Update configuration for the new release of `ruff`

### DIFF
--- a/src/anyio/from_thread.py
+++ b/src/anyio/from_thread.py
@@ -96,7 +96,7 @@ class _BlockingAsyncContextManager(Generic[T_co], AbstractContextManager):
             # `_BlockingAsyncContextManager.__exit__` is called, and an
             # `_exit_exc_info` has been set.
             result = await self._async_cm.__aexit__(*self._exit_exc_info)
-            return result
+        return result
 
     def __enter__(self) -> T_co:
         self._enter_future = Future()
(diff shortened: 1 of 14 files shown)


src/anyio/from_thread.py
```diff
@@ -96,7 +96,7 @@ async def run_async_cm(self) -> bool | None:
             # `_BlockingAsyncContextManager.__exit__` is called, and an
             # `_exit_exc_info` has been set.
             result = await self._async_cm.__aexit__(*self._exit_exc_info)
-            return result
+        return result
```
agronholm, line 99:
```suggestion

        return result
```

agronholm:
We're still debating this. It was said that ruff does not detect everything flake8 does yet, and also does not have a plugin system. I myself am leaning towards merging, so long as there is no significant degradation in coverage of code flaw detection.

JonathanPlasse:
[How does `ruff` compare to `flake8`](https://github.com/charliermarsh/ruff#how-does-ruff-compare-to-flake8).
It implements all rules from flake8 except those for Python 2 or Black incompatible.
It also implements the most popular flake8 plugins and is open to adding more.
It already caught some errors that I fixed.

I would wait before merging to add the latest version of `ruff`.

JonathanPlasse:
It is ready for review.

agronholm:
I've done this locally, with fewer changes, but thanks for getting the ball rolling on this.

PR #504, opened 2022-11-23
Shutdown the loop's default thread executor at the end of a test

### LINKED ISSUE #503: Thread leaks in async tests marked with @pytest.mark.anyio
This is because TestRunner doesn't call `loop.shutdown_default_executor()` at the end of a test. It calls [`loop.close()`](https://github.com/agronholm/anyio/blob/master/src/anyio/_backends/_asyncio.py#L1730-L1733) that [doesn't join threads (`wait=False`)](https://github.com/python/cpython/blob/3.11/Lib/asyncio/base_events.py#L684). 

A cleaner approach would call `loop.shutdown_default_executor()` right after [shutting down asyncgens](https://github.com/agronholm/anyio/blob/master/src/anyio/_backends/_asyncio.py#L1730). The approach is adopted in `IsolatedAsyncioTestCase` from the standard library: `asyncio.Runner.close` [called on tear down](https://github.com/python/cpython/blob/3.11/Lib/unittest/async_case.py#L124-L126) shuts down the loop's [default executor](https://github.com/python/cpython/blob/3.11/Lib/asyncio/runners.py#L73).

marcinbarczynski:
The same approach is adopted in IsolatedAsyncioTestCase. It prevents leaking asyncio threads as loop.close() doesn't join them.

Fixes #503

### DIFF
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -6,6 +6,7 @@ import concurrent.futures
 import math
 import socket
 import sys
+import threading
 from asyncio import (
     AbstractEventLoop,
     CancelledError,
@@ -95,7 +96,6 @@ T_Retval = TypeVar("T_Retval")
 
 # Check whether there is native support for task names in asyncio (3.8+)
 _native_task_names = hasattr(asyncio.Task, "get_name")
-
 
 _root_task: RunVar[asyncio.Task | None] = RunVar("_root_task")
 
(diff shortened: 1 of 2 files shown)


agronholm:
Surely there is a way to do this on Pythons earlier than 3.9, even if we have to use a private API?

marcinbarczynski:
> Surely there is a way to do this on Pythons earlier than 3.9, even if we have to use a private API?

That's true, but it requires copying [non-trivial code from Python 3.9+](https://github.com/python/cpython/blob/3.9/Lib/asyncio/base_events.py#L561-L579) to anyio and adapting it to Python 3.7 and 3.8 internals. Do you think it's worth the effort?

agronholm:
Yeah, those Python versions will stay supported for a while. Also please add a regression test.

marcinbarczynski:
> Yeah, those Python versions will stay supported for a while. Also please add a regression test.

Done.

agronholm (approved):

agronholm:
Thanks!

PR #510, opened 2022-12-22
Fix minor mistyping of `_trio._shutdown_process_pool`

gschaffner:
Mypy only finds this when using `trio_typing.plugin`, since without it the call

https://github.com/agronholm/anyio/blob/eaac4de83886c1487afa96aa83f1b4eacaa11dbf/src/anyio/_backends/_trio.py#L914

is not fully typed.


agronholm (approved):

agronholm:
Good catch.

PR #511, opened 2022-12-22
Add opt-in fuller typing for callables accepting `(func, *args, ...)`

gschaffner:
this follows up discussion in #491. this uses `trio_typing.plugin` to add opt-in fuller typing to callables accepting `(func, *args, unrelated_kwarg0, ...)`. see `versionhistory.rst` for more detailed info.

all of the errors that Mypy currently produces when checking AnyIO with this PR are due to either #510 or https://github.com/python/mypy/issues/14337, I believe.

even with this Mypy bug, this PR is quite useful for typed AnyIO-dependent works as-is, but the Mypy bug does limit its usefulness to (for the most part) only non-generic functions.

if there is interest in this work, I'd be happy to add some typing tests to this PR.

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -31,6 +31,25 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
   - ``create_memory_object_stream()`` now allows passing only ``item_type``
   - Object receive streams are now covariant and object send streams are correspondingly
     contravariant
+
+  - Several functions and methods that accept a function and positional arguments for it now have opt-in fuller typing for those using Mypy:
+
+    - ``anyio.run()``
+    - ``anyio.to_thread.run_sync()``
+    - ``anyio.from_thread.run()``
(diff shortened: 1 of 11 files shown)


killjoy1221:
I don't like how this is specific to mypy. Using `TypeVarTuple` from [PEP 646](https://peps.python.org/pep-0646/) would be a better alternative. [Support for it in mypy](https://github.com/python/mypy/issues/12280) requires additional flags, but it appears to be fully implemented in pyright and pyre.

A basic implementation would be like this.

```py
from typing import TypeVar, Awaitable, Callable, Any
from typing_extensions import TypeVarTuple, Unpack

T_Ret = TypeVar("T_Ret")
T_Args = TypeVarTuple("T_Args")


async def start(
    func: Callable[[Unpack[T_Args]], Awaitable[T_Ret]],
    *args: Unpack[T_Args],
    name: Any = None,
) -> Awaitable[T_Ret]:
    pass

```

agronholm:
Closing in favor of #652.

PR #512, opened 2022-12-22
Fix `TASK_STATUS_IGNORED`'s generic inheritance

gschaffner:
the following currently fails type checking on the 4.0 dev branch:

```python
from anyio import TASK_STATUS_IGNORED


async def foo(*, task_status: TaskStatus[None] = TASK_STATUS_IGNORED) -> None:
    # error: Incompatible default for argument "task_status" (default has type "_IgnoredTaskStatus", argument has type "TaskStatus[None]")  [assignment]
    raise NotImplementedError()
```

### DIFF
--- a/src/anyio/_core/_tasks.py
+++ b/src/anyio/_core/_tasks.py
@@ -2,12 +2,13 @@ from __future__ import annotations
 
 import math
 from types import TracebackType
+from typing import Any
 
 from ..abc._tasks import TaskGroup, TaskStatus
 from ._eventloop import get_async_backend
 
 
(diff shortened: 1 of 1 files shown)


agronholm:
With the latest changes on `master`, I get no error. `TaskStatus[object]` would be assignable to `TaskStatus[int]` due to the contravariant nature of `TaskStatus`. Do you feel that this is a mistake?

agronholm:
I added 66dbf1c2 to check that mypy is fine with `TASK_STATUS_IGNORED`.

PR #524, opened 2023-01-18
Fix `from_thread.run(_sync)?` not setting sniffio on asyncio

### LINKED ISSUE #523: `from_thread.run(_sync)?` does not set sniffio on asyncio
note that `from_thread.run(_sync)?` _does_ set sniffio on trio, however.

this causes a few bugs:

1.  `from_thread.run(_sync)?` fails when used to call sniffio-dependent functions when the `to_thread` worker is running a non-asyncio async backend known to sniffio (i.e. trio, curio, or (situationally) trio-asyncio) and the thread that started the worker is on asyncio.

    reproducer: bc3cc490714c6022243012b6c902c33f3c4b287a.

1.  because `BlockingPortal._spawn_task_from_thread` shares its underlying implementation with `from_thread.run_sync` on asyncio, using a `BlockingPortal` that was started by `start_blocking_portal("asyncio")` in a non-asyncio async backend known to sniffio deadlocks.

    reproducer: cd9deb1e20dcb3e84bf6adcd8d9d22266ce31d0e.

gschaffner:
closes #523.

### DIFF
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -14,9 +14,9 @@ from asyncio import (
     create_task,
     current_task,
     get_running_loop,
+    sleep,
 )
 from asyncio import run as native_run
-from asyncio import sleep
 from asyncio.base_events import _run_until_complete_cb  # type: ignore[attr-defined]
 from collections import OrderedDict, deque
 from collections.abc import AsyncIterator, Iterable
@@ -64,7 +64,7 @@ from .._core._exceptions import (
     EndOfStream,
     WouldBlock,
 )
-from .._core._sockets import GetAddrInfoReturnType, convert_ipv6_sockaddr
+from .._core._sockets import convert_ipv6_sockaddr
 from .._core._streams import create_memory_object_stream
 from .._core._synchronization import CapacityLimiter as BaseCapacityLimiter
 from .._core._synchronization import Event as BaseEvent
@@ -2214,11 +2214,18 @@ class AsyncIOBackend(AsyncBackend):
         type: int | SocketKind = 0,
         proto: int = 0,
         flags: int = 0,
-    ) -> GetAddrInfoReturnType:
-        result = await get_running_loop().getaddrinfo(
+    ) -> list[
+        tuple[
+            AddressFamily,
+            SocketKind,
+            int,
+            str,
+            tuple[str, int] | tuple[str, int, int, int],
+        ]
+    ]:
+        return await get_running_loop().getaddrinfo(
             host, port, family=family, type=type, proto=proto, flags=flags
         )
-        return cast(GetAddrInfoReturnType, result)
 
     @classmethod
     async def getnameinfo(
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -2035,8 +2035,10 @@ class AsyncIOBackend(AsyncBackend):
         token: object,
     ) -> T_Retval:
         loop = cast(AbstractEventLoop, token)
-        f: concurrent.futures.Future[T_Retval] = asyncio.run_coroutine_threadsafe(
-            func(*args), loop
+        context = copy_context()
+        context.run(sniffio.current_async_library_cvar.set, "asyncio")
+        f: concurrent.futures.Future[T_Retval] = context.run(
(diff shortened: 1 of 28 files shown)


src/anyio/_backends/_asyncio.py
```diff
@@ -2034,9 +2034,14 @@ def run_async_from_thread(
...
+        async def wrapper() -> T_Retval:
+            sniffio.current_async_library_cvar.set("asyncio")
+            return await func(*args)
+
         loop = cast(AbstractEventLoop, token)
         f: concurrent.futures.Future[T_Retval] = asyncio.run_coroutine_threadsafe(
-            func(*args), loop
+            wrapper(), loop
         )
         return f.result()
```
agronholm, line 2046: This is a rather unpleasant solution. Have you explored options that didn't involve a wrapper coroutine?

agronholm: This also passes the tests:
```python
    @classmethod
    def run_async_from_thread(
        cls,
        func: Callable[..., Awaitable[T_Retval]],
        args: tuple[Any, ...],
        token: object,
    ) -> T_Retval:
        loop = cast(AbstractEventLoop, token)
        ctx = copy_context()
        ctx.run(sniffio.current_async_library_cvar.set, "asyncio")
        f: concurrent.futures.Future[T_Retval] = ctx.run(
            asyncio.run_coroutine_threadsafe, 
            func(*args),
            loop
        )
        return f.result()
```

src/anyio/_backends/_asyncio.py
```diff
@@ -2046,6 +2048,7 @@ def run_sync_from_thread(
     ) -> T_Retval:
         @wraps(func)
         def wrapper() -> None:
+            sniffio.current_async_library_cvar.set("asyncio")
```
agronholm, line 2051: I would feel safer if this was inside the `try` block, just in case this call ever fails for whatever reason. At least that way the waiter of the future won't be left waiting forever.

tests/test_from_thread.py
```diff
@@ -63,6 +67,23 @@ def thread_worker_sync(func: Callable[..., T_Retval], *args: Any) -> T_Retval:
...
+@pytest.fixture
+def portal_backend_name(portal_backend: Any) -> str:
+    return anyio_backend_name.__wrapped__(portal_backend)  # type: ignore[attr-defined]
+
+
+@pytest.fixture
+def portal_backend_options(portal_backend: Any) -> dict[str, Any]:
+    return anyio_backend_options.__wrapped__(  # type: ignore[attr-defined]
+        portal_backend
+    )
```
agronholm, line 84: Do the tests specifically need to replicate the exact options for each backend? If not, you could just parametrize the tests based on `anyio.get_all_backends()`.

tests/test_from_thread.py
```diff
@@ -524,3 +562,20 @@ async def raise_baseexception() -> None:
...
+    ) -> None:
+        """Test that portals don't deadlock when started/used from async code."""
+
+        if anyio_backend_name == "trio" and portal_backend_name == "trio":
+            pytest.xfail("known bug (#525)")
+
+        with start_blocking_portal(
+            portal_backend_name, portal_backend_options
+        ) as portal:
+            portal.call(checkpoint)
```
agronholm, line 581: This is an annoying problem. Perhaps this could be removed from this PR, to be figured out later? I don't want `pytest-timeout` to block this from being merged.

agronholm (changes requested):
Let's get this done for v4.0!

tests/test_from_thread.py
agronholm: I think I'll just sidestep the issue of pytest-timeout for now and just rely on the timeout we have on GitHub. I want this merged for v3.7.0.

agronholm: That's just for GitHub Actions, right? Locally it would just hang forever, yes?

agronholm: Really? I haven't seen any such thing. All my searches point to `pytest-timeout` as the provider of any sort of a timeout effect.

agronholm: At any rate, if you could remove `pytest-timeout` for the time being, and make the other change I suggested, I could merge this and get it into v3.7.0. I think this is the last fix slated for that minor release.

agronholm (approved):

agronholm:
Thanks!

PR #528, opened 2023-02-04
Fix b'hello world' splitting example to include space.

gwk:
As currently written the example suggests that the space character is lost, which as I understand it is not the intent.


agronholm (approved):

agronholm:
Quite right, thanks!

PR #537, opened 2023-03-08
Fix `Event.wait()` cancellation race causing failure to notify all waiting tasks

### LINKED ISSUE #536: On asyncio, `Event.set()` sometimes fails to notify all waiting tasks
hi!

if:

1.  task A is awaiting `Event.wait()`
1.  `Event.set()` is called
1.  the `wait()`'s cancel scope is cancelled before the event loop schedules A

then, on the asyncio backend only, task A's `wait()` raises cancelled instead of getting notified of the event.

in contrast, on the trio backend (and on pure trio), `Event.set()` still notifies all waiters (including task A) in this situation. trio does this by implementing `Event.wait` via `wait_task_rescheduled`.

here is a small reproducer: `test_event_wait_before_set_before_cancel` in 65d74e483cbcb161223bc9216b020156104f09c0. `assert wait_woke` fails on asyncio only.

this problem is also the direct cause of another bug: `MemoryObjectSendStream.send(item)` can raise cancelled after `item` has been delivered to a receiver![^1] this `MemoryObjectSendStream.send` issue is just the `send_event.wait()` case to #146's `receive_event.wait()` case, i believe—the underlying issue for both seems to be this `Event.wait()` problem[^2].

a couple relevant comments from njsmith: https://github.com/agronholm/anyio/issues/146#issuecomment-673754182, https://github.com/agronholm/anyio/pull/147#issuecomment-674615765.

[^1]: raising cancelled after a receiver gets the item is a violation of trio's cancellation semantics for API in the `trio` namespace (https://trio.readthedocs.io/en/stable/reference-core.html#cancellation-and-primitive-operations). since `anyio` is documented as following `trio`'s cancellation model, i would guess that `anyio` is also intended to adhere to this (but in the `anyio` namespace)? if so, i think it would be good to document this at https://anyio.readthedocs.io/en/stable/cancellation.html and to document the couple rare exceptions to this rule (e.g. https://trio.readthedocs.io/en/stable/reference-io.html#trio.abc.SendStream.send_all) too.

[^2]: nit: #146 actually had two underlying issues: the problem that `MemoryObjectReceiveStream.receive()` didn't `checkpoint_if_cancelled()` before trying `receive_nowait()` (on all backends), as well as this `Event.wait()` issue (on asyncio only).

gschaffner:
fixes #536!

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -44,6 +44,11 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
   the event loop to be closed
 - Fixed ``current_effective_deadline()`` not returning ``-inf`` on asyncio when the
   currently active cancel scope has been cancelled (PR by Ganden Schaffner)
+- Fixed ``Event.set()`` failing to notify a waiter on asyncio if an ``Event.wait()``'s
+  scope was cancelled after ``Event.set()`` but before the scheduler resumed the waiting
+  task. This also fixed a race condition where ``MemoryObjectSendStream.send()`` could
+  raise a ``CancelledError`` on asyncio after successfully delivering an item to a
+  receiver (PR by Ganden Schaffner)
 
 **3.6.1**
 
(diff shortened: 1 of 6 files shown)


agronholm:
Closing in favor of #595.

PR #551, opened 2023-04-12
Uncancel tasks cancelled using CancelScope.cancel()

nikicat:
This patch is needed for correct task canceling when using asyncio backend on Python >= 3.11.
Related discussion https://github.com/encode/httpx/discussions/2506

### DIFF
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -271,26 +271,28 @@ class CancelScope(BaseCancelScope):
                 exc_val.exceptions if isinstance(exc_val, ExceptionGroup) else [exc_val]
             )
             if all(isinstance(exc, CancelledError) for exc in exceptions):
-
-                def uncancel() -> None:
-                    if sys.version_info >= (3, 11):
-                        # Uncancel all anyio cancellations
-                        for i in range(self._cancel_calls):
-                            self._host_task.uncancel()
-                        self._cancel_calls = 0
-
                 if self._timeout_expired:
-                    uncancel()
-                    return True
+                    return self._uncancel()
                 elif not self._cancel_called:
                     # Task was cancelled natively
                     return None
                 elif not self._parent_cancelled():
                     # This scope was directly cancelled
-                    uncancel()
-                    return True
+                    return self._uncancel()
 
         return None
+
+    def _uncancel(self) -> bool:
+        if sys.version_info < (3, 11) or self._host_task is None:
+            self._cancel_calls = 0
+            return True
+
+        # Uncancel all AnyIO cancellations
+        for i in range(self._cancel_calls):
+            self._host_task.uncancel()
+
+        self._cancel_calls = 0
+        return not self._host_task.cancelling()
 
     def _timeout(self) -> None:
         if self._deadline != math.inf:
@@ -594,7 +596,7 @@ class TaskGroup(abc.TaskGroup):
                 "This task group is not active; no new tasks can be started."
             )
 
-        options = {}
+        options: dict[str, Any] = {}
         name = get_callable_name(func) if name is None else str(name)
         if _native_task_names:
             options["name"] = name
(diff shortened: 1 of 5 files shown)


graingert:
uncancel needs to be called each time anyio calls cancel() not just the one time

tests/test_taskgroups.py
```diff
@@ -1092,3 +1092,15 @@ async def starter_task() -> None:
...
+
+@pytest.mark.skipif(
+    sys.version_info < (3, 11),
+    reason="Task uncancelling is only supported on Python 3.11",
+)
+@pytest.mark.parametrize("anyio_backend", ["asyncio"])
+async def test_uncancel_after_cancel() -> None:
+    with CancelScope() as scope:
+        scope.cancel()
+        await anyio.sleep(0)
```
graingert, line 1105:
```suggestion
        with pytest.raises(asyncio.CancelledError):
            await anyio.sleep(0)
        with pytest.raises(asyncio.CancelledError):
            await anyio.sleep(0)
```

src/anyio/_backends/_asyncio.py
```diff
@@ -270,13 +270,20 @@ def __exit__(
                 exc_val.exceptions if isinstance(exc_val, ExceptionGroup) else [exc_val]
             )
             if all(isinstance(exc, CancelledError) for exc in exceptions):
+
+                def uncancel() -> None:
```
graingert, line 274: can you copy the existing pattern for version compat:

https://github.com/agronholm/anyio/blob/e2ff31573aa14df5abfb96b3c56559886310fe36/src/anyio/_backends/_asyncio.py#L86-L94
```

**Comment by nikicat** [comment]
> uncancel needs to be called each time anyio calls cancel() not just the one time

I thought `.uncancel()` should be only called in `__exit__` and only when it suppressses `CancelledError` exception.

**Comment by nikicat** [comment]
> uncancel needs to be called each time anyio calls cancel() not just the one time

I thought `.uncancel()` should be only called in `__exit__` and only when it suppressses `CancelledError` exception.

**Comment by graingert** [comment]
yeah so you need something like:

```python
for i in range(cancel_calls):
    task.uncancel()
```

src/anyio/_backends/_asyncio.py
```diff
@@ -270,13 +270,20 @@ def __exit__(
...
+                    if hasattr(self._host_task, "uncancel"):
+                        self._host_task.uncancel()  # type: ignore[union-attr]
+
                 if self._timeout_expired:
+                    uncancel()
                     return True
                 elif not self._cancel_called:
                     # Task was cancelled natively
                     return None
```
graingert, line 283: native cancellation on 3.11 can now be detected by comparing the number of times anyio called .cance() and the state of `task.cancelling()`

agronholm:
A fix for uncancel() was included in my local WIP branch before I switched to other projects. Now that I've resumed working on AnyIO, I have to see which patches make the most sense.

nikicat:
I have updated this PR according to your suggestions. The only aspect I am uncertain about is allowing multiple context entries for a `CancelScope`. In the current implementation `CancelScope` remains in a canceled state after exiting.

agronholm:
I'm not sure that you should hurry with this. I did have code written to tackle this issue, so it's possible I might end up rejecting this PR. I will, however, take a look at it and see if I can pick up any good ideas.

src/anyio/_backends/_asyncio.py
agronholm: I've added detection of native cancellation, and a test to that end, on OP's behalf.

agronholm:
@nikicat I took the liberty of making some improvements to this PR. Would you like to do a sanity check on my changes so we can get this merged?

nikicat:
@agronholm looks good, LGTM! Thank you!

agronholm (approved):

agronholm:
Thanks!

PR #566, opened 2023-05-09
Fix `TaskStatus.started` type unsafety

### LINKED ISSUE #565: `TaskStatus.started` advertises type checking but it is not type-safe
### Things to check first

### AnyIO version

0414b4e705f21359c16240fc29006f3845cdaf06

### Python version

CPython 3.11.3

### What happened?

since c44c12fceeaa35bcd93009f66fb95cb32f0bffb5 (as well as its backport 75047d5594a9f3d5e122ed46d8ec9d819ca066e9 on 3.x), the following passes type-checking but is not type-safe:

```python
from anyio import TASK_STATUS_IGNORED


async def foo(*, task_status: TaskStatus[int]) -> None:
    task_status.started()
    # or
    # task_status.started(None)
```

see above.

gschaffner:
closes #565.

the bug this fixes predates #558, but for convenience of merging I've written this atop #558.

feel free to change the target branch if you prefer :)

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -25,8 +25,8 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
     - ``BlockingPortal.start_task_soon()``
     - ``BlockingPortal.start_task()``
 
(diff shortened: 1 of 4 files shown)


agronholm:
Thanks!

PR #568, opened 2023-05-13
Fixed error binding create_tcp_listener() to named link-local interfaces

### LINKED ISSUE #554: "OSError: [Errno 49] Can't assign requested address" for IPv6 address
### Things to check first

### AnyIO version

3.6.2

### Python version

3.10.6

### What happened?

`anyio.create_tcp_listener` was not able to bind to `fe80::1%lo0`.

### Details

It looks like the `create_tcp_listener` always resolves IPv6 addresses to a 2-element tuples:
* https://github.com/agronholm/anyio/blob/3.6.2/src/anyio/_core/_sockets.py#L500

Though 2-element tuples for IPv6 addresses are not always understandable by `socket.bind()`.

From https://docs.python.org/3/library/socket.html#socket-families:

> For [AF_INET6](https://docs.python.org/3/library/socket.html#socket.AF_INET6) address family, a four-tuple (host, port, flowinfo, scope_id) is used, where flowinfo and scope_id represent the sin6_flowinfo and sin6_scope_id members in struct sockaddr_in6 in C. For [socket](https://docs.python.org/3/library/socket.html#module-socket) module methods, flowinfo and scope_id can be omitted just for backward compatibility. Note, however, omission of scope_id can cause problems in manipulating scoped IPv6 addresses

`anyio.create_tcp_listener` under the hood uses `anyio.getaddrinfo`. 

It looks like the way how `anyio.getaddrinfo` works for IPv6 addresses is not fully compatible with the way how `socket.bind` is expecting the IPv6 addresses format to be.

`anyio.getaddrinfo` parses `fe80::1%lo0` (which is a link-local address) as `('fe80::1%1', 0)` (where `%1` is `scope_id`), whereas `socket.bind` is expecting `scope_id` as a separate parameter.

This is what `anyio.getaddrinfo` produces:
* `('fe80::1%1', 0)`

For comparison - this is the result of using `socket.getaddrinfo`:
```python
>>> import socket
>>> socket.getaddrinfo("fe80::1%lo0", 0, proto=socket.IPPROTO_TCP)
[(<AddressFamily.AF_INET6: 30>, <SocketKind.SOCK_STREAM: 1>, 6, '', ('fe80::1', 0, 0, 1))]
```

So we have `('fe80::1%1', 0)` vs. `('fe80::1', 0, 0, 1)`.

I'm not able to bind to `('fe80::1%1', 0)` using:
```python
import socket
raw_socket = socket.socket(socket.AddressFamily.AF_INET6)
raw_socket.bind(('fe80::1%1', 0))
```

This is what `socket.bind` expects (both work):
* `('fe80::1%1', 0, 0, 1)`
* `('fe80::1', 0, 0, 1)`

I am able to bind:
```python
import socket
raw_socket = socket.socket(socket.AddressFamily.AF_INET6)
raw_socket.bind(('fe80::1%1', 0, 0, 1))
# or
# raw_socket.bind(('fe80::1', 0, 0, 1))
```

1. Add this line to `/etc/hosts`:
    * `fe80::1%lo0     localhost`
2. Try to use `anyio.create_tcp_listener(host="localhost", port=0)`.
3. Which results in:
    * `OSError: [Errno 49] Can't assign requested address` 

Or even simpler:
1.  `anyio.create_tcp_listener(host="fe80::1%lo0", port=0)`

agronholm:
Fixes #554.


tests/test_sockets.py
```diff
@@ -692,6 +678,21 @@ async def handle(stream: SocketStream) -> None:
...
+    )
+    async def test_bind_link_local(self) -> None:
+        # Regression test for #554
+        for ifname, addresses in psutil.net_if_addrs().items():
+            for addr in addresses:
+                if addr.address.startswith("fe80::") and "%" in addr.address:
+                    async with await create_tcp_listener(local_host=addr.address):
+                        return
+
+        pytest.fail("Could not find a link-local IPv6 interface")
```
graingert, line 694: a for loop with early return makes this test a bit odd, maybe a genexpr makes it a bit more obvious what's going on?

also you don't need .items() you can use .values()

```suggestion
        link_local_ipv6_address = next(
            (
                addr.address
                for addresses in psutil.net_if_addrs().values()
                for addr in addresses
                if addr.address.startswith("fe80::") and "%" in addr.address
            ),
            None,
        )

        if link_local_ipv6_address is None:
            pytest.fail("Could not find a link-local IPv6 interface")
            
        async with await create_tcp_listener(local_host=addr.address):
            pass
```

docs/versionhistory.rst
```diff
@@ -50,6 +50,11 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
   ``TLSStream.wrap()`` being inadvertently set on Python 3.11.3 and 3.10.11
 - Fixed ``CancelScope`` to properly handle asyncio task uncancellation on Python 3.11
   (PR by Nikolay Bryskin)
+- Fixed ``OSError`` when trying to use ``create_tcp_listener()`` to bind to a link-local
+  IPv6 address (and worked around related bugs in `uvloop`)
```
webknjaz, line 54:
```suggestion
  IPv6 address (and worked around related bugs in ``uvloop``)
```

Issue #545, opened 2023-03-29
RuntimeError: dictionary changed size during iteration during import of anyio


ebw44:
# Context
The exception doesn't occur all the time and I have not been able to reproduce the problem with unit tests.  
It has happened only when running my code in vscode debugger, but once it happens it seems to reoccur more.  
In a regular terminal with the same python conda environment I haven't been able to trigger the problem.  
After a restart of vscode, I'm not able to reproduce the problem. I forgot to try to close the vscode 'Python debug console', before I did the restart of vscode, so I don't know if the state of the console was involved.

## Version
anyio=3.6.1=pyhd8ed1ab_1
python=3.11.0=hcf16a7b_0_cpython

# Error
I'm getting a `RuntimeError: dictionary changed size during iteration` from the `for key, value in list(locals().items())` code at the end of anyio/__init__.py
The initial code that triggers the exception is a simple get request using httpx
```python
async with httpx.AsyncClient(timeout=5.0, transport=self.transport) as client:
      res = await client.get(url, headers=self._with_token())
```
with `self.transport = httpx.AsyncHTTPTransport(retries=3)` and `self._with_token()` just returning a dict with authorization token.

Here is the full trace
```
Traceback (most recent call last):
  File "C:\Users\username\Documents\devel\Other\automation\rules\msgraphService.py", line 36, in _get
    res = await client.get(url, headers=self._with_token())
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\username\mambaforge\envs\msgraph\Lib\site-packages\httpx\_client.py", line 1757, in get
    return await self.request(
           ^^^^^^^^^^^^^^^^^^^
  File "C:\Users\username\mambaforge\envs\msgraph\Lib\site-packages\httpx\_client.py", line 1533, in request
    return await self.send(request, auth=auth, follow_redirects=follow_redirects)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\username\mambaforge\envs\msgraph\Lib\site-packages\httpx\_client.py", line 1620, in send
    response = await self._send_handling_auth(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #570, opened 2023-05-17
Solid init of `anyio` and `anyio.abc`

T-256:


agronholm:
Would you like to explain this? The `... as X` imports are so mypy recognizes them as re-exports.

agronholm:
Did you run the test suite, or `pre-commit` locally?

T-256:
sorry, didn't aware of pre commit hook.
No I didn't test locally, this try was for https://github.com/agronholm/anyio/issues/545
let me fix it

agronholm:
If there's a problem, how do I reproduce it? And how does this PR fix it?

T-256:
now is ok
> If there's a problem, how do I reproduce it? And how does this PR fix it?

discussed in https://github.com/agronholm/anyio/issues/545 and https://github.com/encode/httpcore/issues/663 and it seems anyio issue on lazy imports.

agronholm:
Why have I never seen this myself? And how does this PR propose to fix it?

Can you write me a trivial script to reproduce the issue?

T-256:
I can't repro currently, but I had this problem before by iterating over `locals()`. I thought this would help.
Maybe @ebw44 can help us here, bcz I also didn't able to repro via his httpx example.

Currently I'm going to close this PR until we found an example to reproduce it.

smurfix:
Huh. Iterating `locals` will do that. You need to create the loop variable before starting the iterator.

agronholm:
But that's why I make a copy of locals _before_ I start iterating.

T-256:
@agronholm do you have new idea on why this error happened?
```
File "C:\Users\username\mambaforge\envs\msgraph\Lib\site-packages\anyio\__init__.py", line 165, in <module>
    for key, value in list(locals().items()):
                      ^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: dictionary changed size during iteration
```

Maybe `list` didn't do the job to shallow copy. document suggest to use `dict.copy` or `copy.deepcopy`

agronholm:
But that is what `list()` does – it makes a shallow copy, just like `dict.copy()`. A deep copy on the other hand is slow and problematic, so we don't want that.

Under what circumstances have you been able to reproduce the problem?

agronholm:
Let me put this in another way – can you prove that `locals().copy()` would solve the problem you're seeing?

T-256:
> But that is what `list()` does – it makes a shallow copy, just like `dict.copy()`.

so perhaps the error is from `items` method, that is maybe iterating over locals

> Under what circumstances have you been able to reproduce the problem?

As I said above I couldn't even repro it. But i think this is coming from multi-threading problem.

T-256:
> Let me put this in another way – can you prove that `locals().copy()` would solve the problem you're seeing?

@ebw44 should test this.

ebw44:
I have tried to reproduce the problem I had without success. Mainly because my code has changed and I cannot go back to the problematic commit because the infrastructure my code depends on has evolved and is not compatible with that commit anymore.
At the time the problem was happening it was not consistent and seemed to be only when running under vscode debugger. Back then I was unable to create a unit test that triggered the problem. Maybe #692 gives some clue of why I got that problem and a way to reproduce it.

Issue #575, opened 2023-06-05
Futures from BlockingPortal.spawn_task don't wake up concurrent.futures.wait when cancelled


tapetersen:
### Things to check first

- [X] I have searched the existing issues and didn't find my bug already reported there

- [X] I have checked that my bug is still present in the latest release

### AnyIO version

3.7.0

### Python version

3.10.6

### What happened?

Not sure if this is supposed to be supported and digging into the internals of futures.Future and wait I can see why it's a headache todo so but for what it's worth and what I've found.

When creating a blocking portal and spawning tasks in it from a worker thread I can wait on the `concurrent.futures.Future` returned from `BlockingPortal.spawn_task()`  using `Future.result()` and if I stop the portal with `portal.stop(cancel_remaining=True)` the waiting thread will exit with a `CancelledError`

If I try to wait on the result with `concurrent.futures.wait([future], return_when=futures.FIRST_COMPLETED)` however it is never notified of the cancellation and hangs the thread.

Tested on asyncio backend

### How can we reproduce the bug?

```python
import anyio

from concurrent import futures

import anyio.abc
from anyio import from_thread, to_thread

def sync_thread(portal: anyio.abc.BlockingPortal):
    # Normal result() method of future works as expected
    portal.spawn_task(anyio.sleep_forever).result()
    # wait will never return
    # futures.wait([portal.spawn_task(anyio.sleep_forever)], return_when=futures.FIRST_COMPLETED)


async def main():
    async with (
        from_thread.BlockingPortal() as portal,
        anyio.create_task_group() as tg,
    ):
        tg.start_soon(to_thread.run_sync, sync_thread, portal)
        # Ensure thread has time to start the task
        await anyio.sleep(0.1)
        await portal.stop(cancel_remaining=True)

if __name__ == "__main__":
    anyio.run(main)
```

tapetersen:
Ok after digging a whole lot deeper the problem is basically that `Future.cancel()` nofifies the condition variable which will wake up direct callers of `Future.result()` but will not take care of `Future._waiters` where `future.wait` adds itself.

This is since it waits for the "Executor" to call `Future.set_running_or_notify_cancel` to tell that this is really cancelled and won't be run.

Adding a `future.set_running_or_notify_cancel()` to the except block after https://github.com/agronholm/anyio/blob/f7a880ffac4766efb39e6fb60fc28d944f5d2f65/src/anyio/from_thread.py#L219 solves it.

[...]

PR #576, opened 2023-06-05
Notify waiters when cancelling Future in BlockingPortal

tapetersen:
This solves #575
Will look into creating a test for it as well if needed


agronholm:
I fixed this via a6ab281bca6574166bc76ae9604c8f5a97341301.

tapetersen:
> I fixed this via [a6ab281](https://github.com/agronholm/anyio/commit/a6ab281bca6574166bc76ae9604c8f5a97341301).

Thanks. I understand explicit use of Portals probably isn't that common but it really fits what we're trying to do and need.

PR #586, opened 2023-07-12
Updated cancellation semantics to better match the new trio and asyncio ones

### LINKED ISSUE #374: Task group is swallowing `CancelledError`.

agronholm:
This is a lighter version of #496 which I could never get right.

Fixes #374.

### DIFF
--- a/docs/tasks.rst
+++ b/docs/tasks.rst
@@ -139,3 +139,21 @@ host task that will be copied, but the context of the task that calls
 :meth:`TaskGroup.start_soon() <.abc.TaskGroup.start_soon>`.
 
 .. _context: https://docs.python.org/3/library/contextvars.html
+
+Differences with asyncio.TaskGroup
+----------------------------------
+
+The :class:`asyncio.TaskGroup` class, added in Python 3.11, is very similar in design to
+the AnyIO :class:`~TaskGroup` class. The asyncio counterpart has some important
+differences in its semantics, however:
+
+* Tasks are spawned solely through :meth:`~asyncio.TaskGroup.create_task`; there is no
+  ``start()`` or ``start_soon()`` method
+* The :meth:`~asyncio.TaskGroup.create_task` method returns a task object which can be
+  awaited on (or cancelled)
+* Tasks can only be cancelled individually (there is no ``cancel()`` method or similar
+  in the task group)
+* When a task is cancelled before its coroutine has started running, it will not get a
+  chance to handle the cancellation exception
+* New tasks cannot be started after an exception in one of the tasks has triggered a
+  shutdown
--- a/docs/tasks.rst
+++ b/docs/tasks.rst
@@ -151,9 +151,10 @@ differences in its semantics, however:
   ``start()`` or ``start_soon()`` method
 * The :meth:`~asyncio.TaskGroup.create_task` method returns a task object which can be
   awaited on (or cancelled)
-* Tasks can only be cancelled individually (there is no ``cancel()`` method or similar
-  in the task group)
-* When a task is cancelled before its coroutine has started running, it will not get a
-  chance to handle the cancellation exception
-* New tasks cannot be started after an exception in one of the tasks has triggered a
-  shutdown
+* Tasks spawned via :meth:`~asyncio.TaskGroup.create_task` can only be cancelled
+  individually (there is no ``cancel()`` method or similar in the task group)
+* When a task spawned via :meth:`~asyncio.TaskGroup.create_task` is cancelled before its
+  coroutine has started running, it will not get a chance to handle the cancellation
+  exception
+* :class:`asyncio.TaskGroup` does not allow starting new tasks after an exception in
+  one of the tasks has triggered a shutdown of the task group
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -7,6 +7,13 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
 - **BACKWARDS INCOMPATIBLE** Replaced AnyIO's own ``ExceptionGroup`` class with the PEP
   654 ``BaseExceptionGroup`` and ``ExceptionGroup``
+- **BACKWARDS INCOMPATIBLE** Changes to cancellation semantics:
+
+  - Any exceptions raising out of a task groups are now nested inside an
+    ``ExceptionGroup`` (or ``BaseExceptionGroup`` if one or more ``BaseException`` were
+    included)
+  - Fixed task group not raising a cancellation exception on asyncio at exit if no child
+    tasks were spawned and an outer cancellation scope had been cancelled before
 - **BACKWARDS INCOMPATIBLE** Changes the pytest plugin to run all tests and fixtures in
   the same task, allowing fixtures to set context variables for tests and other fixtures
 - **BACKWARDS INCOMPATIBLE** Changed ``anyio.Path.relative_to()`` and
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -14,6 +14,8 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
     included)
   - Fixed task group not raising a cancellation exception on asyncio at exit if no child
     tasks were spawned and an outer cancellation scope had been cancelled before
+  - Ensured that exiting a ``TaskGroup`` always hits a yield point, regardless of
+    whether there are running child tasks to be waited on
 - **BACKWARDS INCOMPATIBLE** Changes the pytest plugin to run all tests and fixtures in
   the same task, allowing fixtures to set context variables for tests and other fixtures
 - **BACKWARDS INCOMPATIBLE** Changed ``anyio.Path.relative_to()`` and
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -499,28 +499,23 @@ class TaskGroup(abc.TaskGroup):
         while self.cancel_scope._tasks:
             try:
                 await asyncio.wait(self.cancel_scope._tasks)
-            except asyncio.CancelledError:
+            except asyncio.CancelledError as exc:
+                # This task was cancelled natively; reraise the CancelledError later
+                # unless this task was already interrupted by another exception
                 self.cancel_scope.cancel()
+                if exc_val is None:
+                    exc_val = exc
+                    ignore_exception = False
+                    self._exceptions.append(exc)
 
         self._active = False
         if self._exceptions:
-            exc: BaseException | None
             group = BaseExceptionGroup("multiple tasks failed", self._exceptions)
-            if not self.cancel_scope._parent_cancelled():
-                # If any exceptions other than AnyIO cancellation exceptions have been
-                # received, raise those
-                _, exc = group.split(is_anyio_cancelled_exc)
-            elif all(is_anyio_cancelled_exc(e) for e in walk_exception_group(group)):
-                # All tasks were cancelled by AnyIO
-                exc = CancelledError()
-            else:
-                exc = group
-
-            if isinstance(exc, BaseExceptionGroup):
-                exc = collapse_exception_group(exc)
-
-            if exc is not None and exc is not exc_val:
-                raise exc
+            _, errors = group.split(CancelledError)
+            if errors is not None:
+                raise errors
+            elif exc_val is not None and not ignore_exception:
+                raise exc_val
 
         return ignore_exception
 
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -510,7 +510,9 @@ class TaskGroup(abc.TaskGroup):
 
         self._active = False
         if self._exceptions:
-            group = BaseExceptionGroup("multiple tasks failed", self._exceptions)
+            group = BaseExceptionGroup(
+                "unhandled errors in a TaskGroup", self._exceptions
+            )
             _, errors = group.split(CancelledError)
             if errors is not None:
                 raise errors
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -499,11 +499,15 @@ class TaskGroup(abc.TaskGroup):
             )
 
         # Raise the CancelledError received while waiting for child tasks to exit,
-        # unless the context manager itself was previously exited with another exception
+        # unless the context manager itself was previously exited with another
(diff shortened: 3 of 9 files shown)


docs/tasks.rst
```diff
@@ -139,3 +139,21 @@ host task that will be copied, but the context of the task that calls
 :meth:`TaskGroup.start_soon() <.abc.TaskGroup.start_soon>`.
 
 .. _context: https://docs.python.org/3/library/contextvars.html
+
+Differences with asyncio.TaskGroup
+----------------------------------
```
zanieb, line 144: Thank you! This is a great summary.

src/anyio/_backends/_asyncio.py
```diff
@@ -499,28 +499,25 @@ async def __aexit__(
...
-
-            if isinstance(exc, BaseExceptionGroup):
-                exc = collapse_exception_group(exc)
-
-            if exc is not None and exc is not exc_val:
-                raise exc
+            group = BaseExceptionGroup(
+                "unhandled errors in a TaskGroup", self._exceptions
+            )
+            _, errors = group.split(CancelledError)
```
zanieb, line 516: Was not clear to me at first (without an IDE to hint the order of the items returned by `split`)
```suggestion
            # Remove cancellation errors from the group
            _, errors = group.split(CancelledError)
```

src/anyio/_backends/_asyncio.py
```diff
@@ -499,28 +499,25 @@ async def __aexit__(
...
+            if errors is not None:
+                raise errors
+            elif exc_val is not None and not ignore_exception:
+                raise exc_val
```
zanieb, line 520: This is the follow-up for "reraise the CancelledError later unless this task was already interrupted by another exception" right? As well as if `exc_val` is not null on context exit? Perhaps a comment about these cases (particularly when the second case would occur) would be helpful.

src/anyio/_backends/_asyncio.py
```diff
@@ -499,28 +499,25 @@ async def __aexit__(
         while self.cancel_scope._tasks:
             try:
                 await asyncio.wait(self.cancel_scope._tasks)
-            except asyncio.CancelledError:
+            except asyncio.CancelledError as exc:
```
zanieb, line 502: `CancelledError` is already imported and using it differently here made me think it was a different type

```suggestion
            except CancelledError as exc:
```

src/anyio/_backends/_asyncio.py
```diff
@@ -499,28 +499,25 @@ async def __aexit__(
...
+                # This task was cancelled natively; reraise the CancelledError later
+                # unless this task was already interrupted by another exception
                 self.cancel_scope.cancel()
+                if exc_val is None:
+                    exc_val = exc
+                    ignore_exception = False
+                    self._exceptions.append(exc)
```
zanieb, line 509: Why do you push the `CancelledError` onto `self._exceptions` if it will be dropped on the `group.split` call?

zanieb (approved):
This looks great! I have some minor comments.

Does this also address cases from https://github.com/agronholm/anyio/issues/432?

src/anyio/_backends/_asyncio.py
agronholm: Because without that, the `if self._exceptions` would be `False`, so we never end up re-raising the `CancelledError` unless other exceptions crop up too.

zanieb: Ah I missed that conditional — just outside the diff :) thanks!

src/anyio/_backends/_asyncio.py
agronholm: This was just about the hardest part to figure out, as AnyIO is treading into uncharted territory as far as cancellation semantics go. I need to double check that I got this right, or at least that I've made it behave consistently.

agronholm: IIRC the idea was to raise a single cancellation exception if the host task was cancelled and there were no other errors.

agronholm: At least `asyncio.TaskGroup` behaves this way.

oremanj: In the case where `exc_val` is the exception originally pased to `__aexit__`, you will get a nicer traceback (no extraneous frame) if you `return False` rather than explicitly reraising it.

docs/tasks.rst
```diff
@@ -139,3 +139,21 @@ host task that will be copied, but the context of the task that calls
...
+
+The :class:`asyncio.TaskGroup` class, added in Python 3.11, is very similar in design to
+the AnyIO :class:`~TaskGroup` class. The asyncio counterpart has some important
+differences in its semantics, however:
+
+* Tasks are spawned solely through :meth:`~asyncio.TaskGroup.create_task`; there is no
```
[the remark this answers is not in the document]
zanieb: I considered saying this as well but at least they consistently refer to the `asyncio` implementation. If two people thought it, might be worth doing anyway.

docs/versionhistory.rst
```diff
@@ -7,6 +7,13 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
 - **BACKWARDS INCOMPATIBLE** Replaced AnyIO's own ``ExceptionGroup`` class with the PEP
   654 ``BaseExceptionGroup`` and ``ExceptionGroup``
+- **BACKWARDS INCOMPATIBLE** Changes to cancellation semantics:
+
+  - Any exceptions raising out of a task groups are now nested inside an
+    ``ExceptionGroup`` (or ``BaseExceptionGroup`` if one or more ``BaseException`` were
+    included)
+  - Fixed task group not raising a cancellation exception on asyncio at exit if no child
+    tasks were spawned and an outer cancellation scope had been cancelled before
```
oremanj, line 16: Trio would say that a Cancelled exception means some work was interrupted. In this case, it sounds like no work was interrupted, so I'm not sure it's actually important to raise the exception. Trio does raise the exception, because of its rule that every async operation that can block (including `NurseryManager.__aexit__`) is always both a schedule point and a cancel point, but this has caused problems occasionally and there has been recent discussion of changing it; IIRC consensus was that it would be fine to change but no one has done it yet. Discussion in python-trio/trio#1457.

agronholm: The issue that this refers to was that an AnyIO `TaskGroup` would swallow a `CancelledError` that was meant to cancel the entire task.

src/anyio/_backends/_asyncio.py
agronholm: In that case the test here fails and the execution proceeds to the end of the method where we do exactly that.

docs/versionhistory.rst
oremanj: Edge-triggered cancellation rears its ugly head again!

My intuition is that for fully consistent behavior here you need to treat each asyncio task as implicitly enclosed in a CancelScope that would be cancelled by `task.cancel()`, and include that implicit cancel scope in your logic of whether each cancel scope should swallow a cancellation or forward it outward. You can indirectly detect whether a task-level cancellation has occurred, at least on 3.11+, by counting anyio-originated calls to `task.cancel()` that are not yet uncancelled, and comparing that against `task.cancelling()`.

Making task group exit a cancel point is defensible though. Even if Trio stops doing that, you can easily add it back by putting in an explicit `checkpoint_if_cancelled`.

src/anyio/_backends/_asyncio.py
oremanj: I disagree. Imagine that there are no child tasks, and the nested child is cancelled by `Task.cancel()` rather than `CancelScope.cancel()`. Then you wind up with `_exceptions` containing one CancelledError, no `errors`, `exc_val is not None`, and `not ignore_exception` (because the cancellation wasn't for this task group's cancel scope). You will explicitly reraise the CancelledError when you should be implicitly reraising it by returning False.

agronholm: I don't think I understand the scenario you're describing here. Could you illustrate with some sample code?

oremanj:
```
async def main():
    asyncio.get_running_loop().call_later(1, asyncio.current_task().cancel)
    await anyio.sleep(5)
asyncio.run(main())
```
I contend that the resulting traceback will have a stack frame from `TaskGroup.__aexit__`. It doesn't need to, and on Trio it wouldn't. It's a minor polish thing, not a big deal, but under the context manager protocol, if you want to have the same exception raised out of the context manager as reached it, then you should return False rather than raising anything.

agronholm: I think I remembered a different implementation of `__aexit__()` in which I didn't add `exc_val` into `self._exceptions`. But IIRC that had different issues. Now the issue is that the code in the `if self._exceptions:` block has no way to determine whether it should reraise `exc_val` (if it was caught inside `__aexit__()`) or to return `False` (if the context manager was exited with it).

docs/tasks.rst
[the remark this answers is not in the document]
agronholm: Under what circumstances could I possibly be talking about `anyio.TaskGroup`, after leading into the list with:
> The asyncio counterpart has some important differences in its semantics, however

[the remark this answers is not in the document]
oremanj: I think that is a lot less obvious than you think it is. I generally like to write docs in a way that eliminates all possible doubt. You have feedback from two different people that the wording you've used here created doubt. What you do with that is up to you.

src/anyio/_backends/_asyncio.py
agronholm: I didn't mean this wasn't fixable. What I left unsaid was that I would need to add an extra local variable just for distinguishing between these two cases.

zanieb: It might be clearer not to reassign `exc_val` anyway. That'd make the case I had questions about at https://github.com/agronholm/anyio/pull/586#discussion_r1263933896 more obvious.

agronholm: Maybe you're right. I'll poke the code and see what I come up with.

docs/tasks.rst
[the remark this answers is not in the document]
agronholm: I've changed the wording. Hopefully it is less ambiguous now.

src/anyio/_backends/_asyncio.py
agronholm: I opted not to store any CancellationErrors which then simplifies the code somewhat.

agronholm:
> This looks great! I have some minor comments.
> 
> Does this also address cases from #432?

The differences in what comes out when cancellation happens is not really something I can reasonably control. Trio does its own thing and asyncio needs a single `CancelledError` to pop out in order for asyncio to work with it.

docs/versionhistory.rst
agronholm: > and include that implicit cancel scope in your logic of whether each cancel scope should swallow a cancellation or forward it outward.

This is actually something I was going to propose for asyncio! I'll be sending that  post to Async-SIG shortly. Meanwhile, I can again look for a workaround for Python < 3.11, but that should not block this PR.

agronholm:
So, what's the verdict? Are more changes needed here?

src/anyio/_backends/_asyncio.py
```diff
@@ -490,33 +477,35 @@ async def __aexit__(
...
+        # Raise the CancelledError received while waiting for child tasks to exit,
+        # unless the context manager itself was previously exited with another exception
+        if cancelled_exc_while_waiting_tasks:
+            if exc_val is None or ignore_exception:
+                raise cancelled_exc_while_waiting_tasks
 
-            if exc is not None and exc is not exc_val:
-                raise exc
+        if not waited_for_tasks_to_finish:
+            await AsyncIOBackend.checkpoint()
```
zanieb, line 508: Worth documenting why we checkpoint here?

src/anyio/_backends/_asyncio.py
```diff
@@ -490,33 +477,35 @@ async def __aexit__(
...
-            else:
-                exc = group
+            raise BaseExceptionGroup(
+                "unhandled errors in a TaskGroup", self._exceptions
+            )
 
-            if isinstance(exc, BaseExceptionGroup):
-                exc = collapse_exception_group(exc)
+        # Raise the CancelledError received while waiting for child tasks to exit,
+        # unless the context manager itself was previously exited with another exception
```
zanieb, line 502: This will also be skipped if an exception was raised in the the group (i.e. a value is in `self._exceptions`) from the clause above. Perhaps worth noting for clarity.

src/anyio/_backends/_asyncio.py
```diff
@@ -490,33 +477,35 @@ async def __aexit__(
...
+        if cancelled_exc_while_waiting_tasks:
+            if exc_val is None or ignore_exception:
```
zanieb, line 504: Nit: This could be one statement

zanieb:
Overall I think this is clearer than before! I poked at it a little bit but couldn't come up with a better way to express the logic.

src/anyio/_backends/_asyncio.py
agronholm: It would not increase readability. I've seen something similar used in CPython code, and my initial thoughts were the same as yours until I saw the wisdom in doing it like this.

src/anyio/_backends/_asyncio.py
agronholm: Clarified.

src/anyio/_backends/_asyncio.py
agronholm: Clarified.

agronholm:
I'm fairly happy with this, so I'm merging. If anything comes up, we can sort it out during the RC cycle.

agronholm:
> > This looks great! I have some minor comments.
> > Does this also address cases from #432?
> 
> The differences in what comes out when cancellation happens is not really something I can reasonably control. Trio does its own thing and asyncio needs a single `CancelledError` to pop out in order for asyncio to work with it.

So I just talked to Joshua Oreman and it seems that Trio wouldn't have an issue with a single `Cancelled` coming out of a task group; it was just my misconception. Expect another PR to address that.

PR #591, opened 2023-07-17
Improved detection of timeout based vs direct cancellation of cancel scopes

### LINKED ISSUE #589: Timeout error after success
### Things to check first

### AnyIO version

3.7 or master, Trio 0.22.2 or asyncio

### Python version

3.11

### What happened?

Consider this code snippet:
``` python3
     try:
         with anyio.fail_after(10):
             msg = await self._wqueue_r.receive()
             logger.debug("We got %s", msg)  ## A
     except TimeoutError:
         msg = NOPMsg()
         logger.debug("We timed out and use %s", msg)  ## B
     ... # do something with the message
     logger.debug("We processed %s", msg)  ## C
```

This seems straightforward, no? We log either A or B, depending on whether something's in the queue or not, then C.

… except that once in a blue moon I get A+B+C.

The problem occurs with Trio or asyncio back-ends. It does not occur when using Trio directly.

On my (fast) development VM, this code exhibits the problem pretty much immediately. On a Raspberry Pi 3, T=0.1 causes it to show up after a few seconds. If it doesn't, change the `0.9` in the third-last line so that the reported numbers increase roughly at the same speed.

```python3

import sys
import anyio
import trio
import random

T = 0.0001
N=int(2/T)

async def eat(r):
    a=0
    b=0
    while True:
        msg = None
        try:
            with anyio.fail_after(T):
            # with trio.fail_after(T):  # bug disappears
                msg = await r.receive()
        except (TimeoutError, trio.TooSlowError):
            if msg is not None:
                raise RuntimeError("Owch {a} {b}")
            b += 1
        else:
            a += 1
        if a%N == 0 or b%N == 0:
            sys.stderr.write(f"  {a} {b}\r"); sys.stderr.flush()

async def main():
    s,r = anyio.create_memory_object_stream(10)
        tg.start_soon(eat,r)
        while True:
            await anyio.sleep(T*(random.random() * 0.7 + 0.9)) ## might need to adapt
            await s.send(True)

anyio.run(main)
```

agronholm:
Fixes #589.

### DIFF
--- a/docs/cancellation.rst
+++ b/docs/cancellation.rst
@@ -7,7 +7,7 @@ The ability to cancel tasks is the foremost advantage of the asynchronous progra
 model. Threads, on the other hand, cannot be forcibly killed and shutting them down will
 require perfect cooperation from the code running in them.
 
-Cancellation in AnyIO follows the model established by the trio_ framework. This means
+Cancellation in AnyIO follows the model established by the Trio_ framework. This means
 that cancellation of tasks is done via so called *cancel scopes*. Cancel scopes are used
 as context managers and can be nested. Cancelling a cancel scope cancels all cancel
 scopes nested within it. If a task is waiting on something, it is cancelled immediately.
@@ -17,7 +17,7 @@ requiring waiting, such as :func:`~sleep`.
 A task group contains its own cancel scope. The entire task group can be cancelled by
 cancelling this scope.
 
-.. _trio: https://trio.readthedocs.io/en/latest/reference-core.html#cancellation-and-timeouts
+.. _Trio: https://trio.readthedocs.io/en/latest/reference-core.html#cancellation-and-timeouts
 
 Timeouts
 --------
@@ -50,6 +50,13 @@ Here's how you typically use timeouts::
             print('Exited cancel scope, deadline reached =', scope.deadline_reached)
 
     run(main)
+
+.. warning:: On Trio, there is a known issue where explicitly cancelling a cancel scope
+    with a deadline can lead to :exc:`TimeoutError` being erronously raised from
+    the :func:`fail_after` context manager if the manual cancellation happens before the
+    deadline is reached, but exiting from the context block is delayed long enough for'
+    the deadline timer to go off. See the relevant
+    `Trio issue <https://github.com/python-trio/trio/issues/698>`_ for details.
 
 Shielding
 ---------
--- a/docs/cancellation.rst
+++ b/docs/cancellation.rst
@@ -46,17 +46,14 @@ Here's how you typically use timeouts::
                 await sleep(2)
                 print('This should never be printed')
 
-            # The deadline_reached property will be True if a timeout occurred
-            print('Exited cancel scope, deadline reached =', scope.deadline_reached)
+            # The cancelled_caught property will be True if timeout was reached
+            print('Exited cancel scope, cancelled =', scope.cancelled_caught)
 
     run(main)
 
-.. warning:: On Trio, there is a known issue where explicitly cancelling a cancel scope
-    with a deadline can lead to :exc:`TimeoutError` being erronously raised from
-    the :func:`fail_after` context manager if the manual cancellation happens before the
-    deadline is reached, but exiting from the context block is delayed long enough for'
-    the deadline timer to go off. See the relevant
-    `Trio issue <https://github.com/python-trio/trio/issues/698>`_ for details.
+.. note:: It's recommended not to directly cancel a scope from :func:`~fail_after`, as
+    that may currently result in :exc:`TimeoutError` being erroneously raised if exiting
+    the scope is delayed long enough for the deadline to be exceeded.
 
 Shielding
 ---------
--- a/docs/cancellation.rst
+++ b/docs/cancellation.rst
@@ -46,8 +46,8 @@ Here's how you typically use timeouts::
                 await sleep(2)
                 print('This should never be printed')
 
-            # The cancel_called property will be True if timeout was reached
-            print('Exited cancel scope, cancelled =', scope.cancel_called)
+            # The deadline_reached property will be True if a timeout occurred
+            print('Exited cancel scope, deadline reached =', scope.deadline_reached)
 
     run(main)
 
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -33,8 +33,7 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 - Added the ``anyio.Path.is_junction()`` and ``anyio.Path.walk()`` methods
 - Added ``create_unix_datagram_socket`` and ``create_connected_unix_datagram_socket`` to
   create UNIX datagram sockets (PR by Jean Hominal)
-- Added the ``deadline_reached`` property to ``CancelScope`` to determine whether the
-  scope was cancelled due to the deadline being reached, or for some other reason
+- Added the ``cancelled_caught`` property to ``CancelScope``
 - Fixed ``fail_after()`` raising an unwarranted ``TimeoutError`` when the cancel scope
   was cancelled before reaching its deadline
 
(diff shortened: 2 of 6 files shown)


smurfix:
Unfortunately this patch doesn't fix anything on my system. Tested against anyio/fix-timeout-after-success and Trio/v0.22.2.

Test run:
```
Crash TRIO=False Owch 33 34
Crash TRIO=None Owch 7015 7016
Crash TRIO=True Owch 76 77
OK with TRIO=2
```

using this code:

```
import sys
import anyio
import trio
import random

T = 0.0001
N = int(0.1/T)

async def eat(r):
    a=0
    b=0
    while True:
        msg = None
        try:
            with (trio if TRIO==2 else anyio).fail_after(T):
                msg = await r.receive()
        except (trio.TooSlowError,TimeoutError):
            if msg is not None:
                raise RuntimeError(f"Owch {a} {b}")
            b += 1
        else:
            a += 1
        if a%N == 0 or b%N == 0:
            sys.stderr.write(f"  {a} {b}\r"); sys.stderr.flush()

async def main():
    s,r = (trio.open_memory_channel if TRIO else anyio.create_memory_object_stream)(10)
    async with (trio.open_nursery if TRIO else anyio.create_task_group)() as tg:
        tg.start_soon(eat,r)
        for _ in range(100*N):
            await (trio if TRIO else anyio).sleep(T*(random.random() * 0.7 + 0.9))
            await s.send(True)
        tg.cancel_scope.cancel()

for TRIO in (False,None,True,2):
    try:
        if TRIO:
            trio.run(main)
        else:
            anyio.run(main, backend="asyncio" if TRIO is None else "trio")
    except Exception as exc:
        if hasattr(exc,"exceptions"):
            exc = exc.exceptions[0]
        print(f"Crash TRIO={TRIO} {exc}")
    else:
        print(f"OK with TRIO={TRIO}")

```

smurfix:
The test program does not crash (or at least not reliably enough …) if I increase T to 0.01 on my test machine, which suggests that if you tried my previous version and didn't see the problem you need to decrease T somewhat; after doing that, you might have to adjust the randomization so that both values increase at roughly the same speed. If the first value stays at zero then T is too low.

agronholm:
I was not able to 100% fix this on trio because it's unable to tell the difference between timeout based cancellation vs manual cancellation + deadline exceeded later, unlike the asyncio backend where the timer expiring event is explicitly recorded and checked against. Are you able to make the test program fail with asyncio though? If so, that then would indicate a problem.

agronholm:
With the changes I pushed, I can no longer repro the problem with your script. Can you confirm?

smurfix:
With the current changes there is no longer a problem with the Trio backend.

There is however now a *different* problem with asyncio. The inner test loop terminates *way* too early.

To reproduce:
* replace the outer loop with `for TRIO in (None,):`
* play with T values

The `sleep` call in the `for _ in range(10*N)` loop runs a few times, then its cancellation escapes and silently terminates the taskgroup.

Thus the code, changed to
```
    async with (trio.open_nursery if TRIO else anyio.create_task_group)() as tg:
        tg.start_soon(eat,r)
        for _ in range(10*N):
            print("a", end="")
            await (trio if TRIO else anyio).sleep(T*(random.random() * 0.7 + 0.9))
            print("b", end="")
            await s.send(True)
            print("c", end="")
        tg.cancel_scope.cancel()

```
ends up emitting `abcabcabcabcabcabcaOK with TRIO=None` which is kindof an unsatisfactory result. :-/

agronholm:
Back to the drawing board I guess :frowning_face: 
I'll let you know once everything works again.

agronholm:
I have played around with the code without being able to repro the problem. Would you mind pasting your entire script here with which you repro'd the new issue, so I can be sure I'm doing it right?

smurfix:
Sure.
```python3
import sys
import anyio
import asyncio
import trio
import random
TRIO=None

T = 0.00005
N = int(0.1/T)

async def eat(r):
    a=0
    b=0
    while True:
        msg = None
        try:
            with (trio if TRIO==2 else anyio).fail_after(T):
                msg = await r.receive()
        except (trio.TooSlowError,TimeoutError):
            if msg is not None:
                raise RuntimeError(f"Owch {a} {b}")
            b += 1
        else:
            a += 1
        if a%N == 0 or b%N == 0:
            sys.stderr.write(f"  {a} {b}\r"); sys.stderr.flush()

async def main():
    s,r = (trio.open_memory_channel if TRIO else anyio.create_memory_object_stream)(10)
    async with (trio.open_nursery if TRIO else anyio.create_task_group)() as tg:
        tg.start_soon(eat,r)
        for _ in range(10*N):
            try:
                await (trio if TRIO else asyncio if TRIO is None else anyio).sleep(T*(random.random() * 0.7 + 0.9))
            except BaseException as exc:
                raise RuntimeError("What?") from exc
            await s.send(True)
        tg.cancel_scope.cancel()


for TRIO in (None,):
#for TRIO in (False,None,True,2):
    try:
        if TRIO:
            trio.run(main)
        else:
            anyio.run(main, backend="asyncio" if TRIO is None else "trio")
    except Exception as exc:
        if hasattr(exc,"exceptions"):
            exc = exc.exceptions[0]
        raise exc
    else:
        print(f"OK with TRIO={TRIO}")

$ PYTHONPATH=/src/anyio/src python3 /tmp/t.py 
  File "/tmp/t.py", line 34, in main
    await (trio if TRIO else asyncio if TRIO is None else anyio).sleep(T*(random.random() * 0.7 + 0.9))
  File "/usr/lib/python3.11/asyncio/tasks.py", line 639, in sleep
    return await future
           ^^^^^^^^^^^^
asyncio.exceptions.CancelledError

  File "/tmp/t.py", line 51, in <module>
    raise exc
  File "/tmp/t.py", line 36, in main
    raise RuntimeError("What?") from exc
RuntimeError: What?
```

agronholm:
Thanks, I was now able to repro the problem locally.

agronholm:
What seems to be happening is that the task is cancelled by a cancel scope, but the cancellation is only delivered to the task _after_ a new scope has been entered. The new scope is exited with `CancelledError`, and sees that it didn't cancel the task, so it lets the exception escape.

agronholm:
I checked and it seems that the stdlib `async with asyncio.timeout(...):` suffers from the exact same issue, and for the exactly same reason.

agronholm:
I went to talk to Łukasz Langa here at the EP2023 sprints and agreed that I should report this as an issue.

But given that this is a fundamental issue in asyncio, I don't think this should delay the 4.0 release any longer. Would you agree that this PR substantially improves the semantics even though it doesn't fully resolve all the issues? We can work on improving things in later releases.

smurfix:
Ah. Frankly I already suspected that.

Sure, go ahead with 4.0. This issue is not a problem for me, I'm using Trio. ;-) and even if it was it's not anyio's fault and even if you can work around it in anyio, given that it affects non-anyio code, doing so would not really help much.

agronholm:
One issue I noticed on py3.11 is that when this happens, the cancel scope doesn't uncancel the host task since that is currently only triggered when a `CancelledError` is caught by the CM. I think I'll at least fix that before merging.

agronholm:
The tricky part is constructing a test to consistently repro this situation.

agronholm:
@graingert and I both came to the conclusion that AnyIO's memory object streams are the fundamental problem here. Neither of us could reproduce the problem with asyncio queues.

smurfix:
Are they? AFAICT the only async thing memory streams do is wait on (and set, and (probably most relevant here) cancel tasks waiting on) perfectly normal `Event`s … no idea why that should be a "fundamental problem".

Thus the next step I'd consider would be to copy the relevant parts into your test program, maybe inline some of the code in question, and then boil that down to a somewhat-minimal reproducer.

agronholm:
Memory object streams cheat a little bit as far as cancellation goes, see [here](https://github.com/agronholm/anyio/blob/master/src/anyio/streams/memory.py#L107-L111). We suspected that this is the cause of your troubles. I'm working on an alternative solution.

smurfix:
Owch. Yeah, I seem to have missed that one.

I agree that you really shouldn't do that.

agronholm:
That code was added to fix #146.

agronholm:
After a lengthy discussion, I decided to back out of `deadline_reached` and instead implement `CancelScope.cancelled_caught` from Trio. We discussed ways to improve the detection of cancellation reason/source, but those improvements will have to be deferred to a later feature release. The memory object stream receive bug still needs to be addressed before v4.0 final, but I may do that in a separate PR.

agronholm:
As I made some significant tweaks before merging, it might be prudent to test that `master` still works for you. I did add more tests to ensure that there are no regressions, but...you never know :stuck_out_tongue:

PR #595, opened 2023-07-24
Added an alternate fix for `MemoryObjectReceiveStream.receive()`` on asyncio

### LINKED ISSUE #146: Object stream randomly drops items

agronholm:
This fixes #146 properly.

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -16,6 +16,8 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
     tasks were spawned and an outer cancellation scope had been cancelled before
   - Ensured that exiting a ``TaskGroup`` always hits a yield point, regardless of
     whether there are running child tasks to be waited on
+  - On asyncio, cancel scopes will defer cancelling tasks that are scheduled to resume
+    with a finished future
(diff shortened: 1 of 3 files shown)


agronholm:
Asyncio's cancellation, as implemented today, is not trivially fixable, and requires some hard decisions upstream. I've had a lengthy post about this in the works for a while which I will post to Async-SIG.

PR #599, opened 2023-08-03
Added v4 migration documentation

agronholm:


zanieb (approved):
LGTM!

goodboy (approved):
Looks good to me 😎 

Glad to see this tech finally land in 3.11 and also think we're going to try to adopt the 

```python
send, receive = create_memory_object_stream[int](100)
```

style annotation for `tractor`'s `MsgStream` constructor 🥳

agronholm:
Thanks. By the way, this is just a trick to get PEP 695 parametrized functions on older Pythons: https://docs.python.org/3.12/whatsnew/3.12.html#pep-695-type-parameter-syntax

PR #600, opened 2023-08-10
Fix changelog merges

gschaffner:
i think there were some issues with merges into the main branch changelog. this should make it so that the 3.7.0 section of the changelog on the main branch matches the 3.7.0 section of the changelog on the 3.x branch.

i've placed the entries in the 4.0.0rc1 changelog where they would have been merged to if each of the incorrect commits (see commit messages) had been correct. i'm not sure if this is the preferred order of entries though.


agronholm (approved):

agronholm:
Thanks!

PR #606, opened 2023-08-26
Also consider cancel counts for swallowing CancelledError on asyncio

agronholm:
This fixes cases on Python 3.11+ where a third party library catches a `CancelledError` and raises another one without including the original message. At least asyncpg is [known to do this](https://github.com/MagicStack/asyncpg/blob/7cb4e70d88d165273997d914280c6d109fbbc8f6/asyncpg/compat.py#L62-L63).

### DIFF
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -428,10 +428,12 @@ class CancelScope(BaseCancelScope):
             self._cancel_calls = 0
             return True
 
-        # Uncancel all AnyIO cancellations
+        # Undo all cancellations done by this scope
         if sys.version_info >= (3, 11):
-            for i in range(self._cancel_calls):
-                self._host_task.uncancel()
+            while self._cancel_calls:
+                self._cancel_calls -= 1
+                if not self._host_task.uncancel() and not self._cancel_calls:
+                    return True
 
         self._cancel_calls = 0
         return f"Cancelled by cancel scope {id(self):x}" in cancelled_exc.args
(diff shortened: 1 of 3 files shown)


src/anyio/_backends/_asyncio.py
```diff
@@ -428,10 +428,12 @@ def _uncancel(self, cancelled_exc: CancelledError) -> bool:
...
             return True
 
-        # Uncancel all AnyIO cancellations
+        # Undo all cancellations done by this scope
         if sys.version_info >= (3, 11):
-            for i in range(self._cancel_calls):
-                self._host_task.uncancel()
+            while self._cancel_calls:
+                self._cancel_calls -= 1
+                if not self._host_task.uncancel() and not self._cancel_calls:
```
graingert, line 435: I think cancel scopes need to store the current task.cancelling() value and compare to that rather than 0 like asyncio.timeout does in 1.11.4

graingert:
```python
try:
    asyncio.current_task.cancel()
    await anyio.sleep(1)
except CancelledError:
        try:
            with CancelScope() as scope:
                scope.cancel()
                try:
                    await anyio.sleep(0)
                    raise asyncio.CancelledError
            pytest.fail("Should have swallowed the CancelledError")
```

graingert (changes requested):

src/anyio/_backends/_asyncio.py
graingert: https://github.com/python/cpython/pull/102815/files#diff-0a6da8818df8b29c3c649167aa37c906eb27283c5fe1cacb40e36a6ce0458e7eR87

agronholm:
Updated the code and the test.

graingert (approved):

Issue #614, opened 2023-09-08
contextvars setup when mixing sync and async pytest fixtures in async test


heqile:
### Things to check first

- [X] I have searched the existing issues and didn't find my bug already reported there

- [X] I have checked that my bug is still present in the latest release

### AnyIO version

4.0.0

### Python version

3.8

### What happened?

In the tests, we need to set the `ContextVar` to some special values by using pytest fixtures.  After updating to `v4.0.0`, some of our tests broke, as if the `sync` fixture who setup `ContextVar` does not work.

In fact, the `sync` `ContextVar` setup fixture works or not, depends on the `sync` fixture runs before `async` fixture or after. (see the code below to reproduce).

I do notice the [changes](https://anyio.readthedocs.io/en/stable/testing.html#context-variable-propagation) in `v4.0.0`:
> The asynchronous test runner runs all async fixtures and tests in the same task, so context variables set in async fixtures or tests, within an async test runner, will affect other async fixtures and tests within the same runner. However, these context variables are not carried over to synchronous tests and fixtures, or to other async test runners.

But the case that a `ContextVar` setup in `sync` fixture and runs after certain `async` fixture should also be a common use case, (especially when you want to re-use the setup fixutre in other `non-async` project). And it does work in `v3.7.1`, so I'd like to consider it as a bug.

(correct me, if I'm wrong :) )

### How can we reproduce the bug?

```python
import pytest
import contextvars

pytestmark = pytest.mark.anyio

some_data_var = contextvars.ContextVar("some_data", default=1)


@pytest.fixture
def anyio_backend():
    return "asyncio"


@pytest.fixture
async def async_setup():
    some_data_var.set(2)


@pytest.fixture
def sync_setup():
    some_data_var.set(3)


async def async_main_func():
    return some_data_var.get()


async def test_one(
    async_setup, sync_setup  # run "async_setup" first, test passes in anyio v3.7.1, but not in v4.0.0
):
    assert 3 == await async_main_func()  # the function returns 2, but expected to be 3
    assert 3 == some_data_var.get()  # the contextvar's value is 2, but expected to be 3


async def test_two(
    sync_setup, async_setup  # run "sync_setup" first, test does not pass in v3.7.1, but fixed in v4.0.0
):
    assert 2 == await async_main_func()
    assert 2 == some_data_var.get()

```

jhominal:
Full disclosure: I work in the same team as @heqile, on the affected projects

[...]

PR #616, opened 2023-09-24
Implemented contextvars isolation in test runner using distinct tasks for each scope

jhominal:
This is an implementation of contextvars isolation between tests, using distinct tasks and queues for each pytest scope.

I am sure that there is a lot to review (e.g. the `_TaskManager` name is not very inspired ;)), but it seems that the basic functionality works. (The documentation should also be updated to document this new behaviour of the test runner)

I am going to implement the solution to issue #614 on top of the implementation in this branch, when I have a bit more time, but in the meantime I welcome any and all feedback.

### DIFF
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -1808,9 +1808,45 @@ def _create_task_info(task: asyncio.Task) -> TaskInfo:
     return TaskInfo(id(task), parent_id, task.get_name(), task.get_coro())
 
 
-class TestRunner(abc.TestRunner):
+class _TaskManager:
     _send_stream: MemoryObjectSendStream[tuple[Awaitable[Any], asyncio.Future[Any]]]
+    _task: asyncio.Task
+    _loop: asyncio.AbstractEventLoop
 
+    @staticmethod
+    async def _run_coroutines(
+        receive_stream: MemoryObjectReceiveStream[
+            tuple[Awaitable[Any], asyncio.Future[Any]]
+        ],
+    ) -> None:
+        with receive_stream:
+            async for coro, future in receive_stream:
+                try:
+                    retval = await coro
+                except BaseException as exc:
+                    if not future.cancelled():
+                        future.set_exception(exc)
+                else:
+                    if not future.cancelled():
+                        future.set_result(retval)
+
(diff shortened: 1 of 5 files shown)


agronholm:
Nice! I'll try to find time for a review in the coming week.

agronholm:
Looking at this, I have just one concern: do contextvars still propagate from higher scoped fixtures to lower scoped fixtures (e.g. module -> function)?

jhominal:
Yes, higher scoped fixtures still propagate to lower scoped fixtures - that is thanks to pytest ordering guarantees, that higher scoped fixtures run before lower scoped ones. However, this propagation is implemented by making lower scoped fixtures run on a copy of the higher level context, meaning that lower scoped fixtures cannot modify the context of the higher level context. I guess I will add another test to confirm that.

However, trying to fix issue #614 is something of a pain with this system of queuing coroutines to background tasks, because suddenly I have to try and run even synchronous fixtures and tests in these background tasks and I have a hard time identifying when that is possible/desirable. I also suspect that reentrancy related to calling `getfixturevalue` would be difficult. Ideally I would like to run asynchronous fixtures in the "current context" without spinning a context copy (meaning, without spawning a new task), but I do not know if that is even possible.

agronholm:
Hm, I suspected this might not be so easy. I've only taken a fairly cursory look, but I'll delve deeper when I get the chance. Maybe I can think of something then.

jhominal:
So, after thinking about the issue during the night:

 * I am beginning to think that context isolation is too much of a deviation from "regular pytest" to be something that should be introduced as a matter of course in the anyio test runner;
   * Isolation between tests can actually be handled via a decorator on the test function, same as what people would be able to do with a sync or async test;
 * I think I have identified two possible, not too invasive, ways of running all fixtures and tests, including the synchronous ones, in the same context;
   * One which already promises to be extremely hacky if I can make it work at all but would allow me to avoid touching to the synchronous fixtures and tests at all;
   * One which would be slightly more invasive but could probably work by using an explicit, pytest session-scoped `contextvars.Context` object;

As a result of that, I am going to mark this PR as a draft, and I will submit PRs related to these two options in the coming days.

jhominal:
Given how #618 has turned out, I am going to close this PR - I do not think anymore that this PR should be integrated in `anyio`.

PR #639, opened 2023-11-22
Fixed asyncio `CancelScope` not recognizing its own cancellation exception

### LINKED ISSUE #634: Re-raising a CancelledError after catching it in an ExceptionGroup confuses asyncio backend
### Things to check first

### AnyIO version

4.0.0

### Python version

3.11

### What happened?

If you catch a CancelledError as part of an ExceptionGroup and then re-raise it when using the asyncio backend, the surrounding CacnelScope won't recognize it as a cancellation exception anymore. The end result, is that instead of cancellation working as expected and the exception being swallowed by the CancelScope, its instead re-raised.

This issue only occurs when using the asyncio backend. It does not occur on the trio backend.

I _think_ the issue is that trio internally will unwrap cancellation exception inside of Exception Groups: https://github.com/python-trio/trio/blob/d5c0fb83bba6b56c35956c65092411cc9a2be8c3/src/trio/_core/_run.py#L595-L604. However, I don't _think_ the asyncio backend in anyio does the same thing.

Run the following code, which exits with an exception. Change the backend to "trio" and it exits without an exception.

```python
import anyio


async def main():
        tg.cancel_scope.cancel()
        try:
            await anyio.sleep_forever()
        except* anyio.get_cancelled_exc_class():
            raise


if __name__ == "__main__":
```

agronholm:
...if it comes wrapped in an exception group.
Fixes #634.

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -12,6 +12,8 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
   - Removed a checkpoint when exiting a task group
 - Bumped minimum version of trio to v0.23
 - Exposed the ``ResourceGuard`` class in the public API
+- Fixed discrepancy between ``asyncio`` and ``trio`` where reraising a cancellation
+  exception in an ``except*`` block would incorrectly bubble out of its cancel scope
 
 **4.0.0**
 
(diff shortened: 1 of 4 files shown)


agronholm:
@dagenix I'm only unwrapping one layer of exception groups; do you think it's necessary to unwrap them all?

agronholm:
I replaced some dead code with a helper function that should cover all our bases.

PR #640, opened 2023-11-23
Accept bytes for Unix socket paths.

Fuyukai:
*nix paths are an arbitrary sequence of bytes.

### DIFF
--- a/src/anyio/_core/_sockets.py
+++ b/src/anyio/_core/_sockets.py
@@ -3,11 +3,11 @@ from __future__ import annotations
 import os
 import socket
 import ssl
+import stat
 import sys
 from collections.abc import Awaitable
 from ipaddress import IPv6Address, ip_address
 from os import PathLike, chmod
-from pathlib import Path
 from socket import AddressFamily, SocketKind
 from typing import Literal, cast, overload
 
@@ -343,7 +343,10 @@ async def create_tcp_listener(
 
 
 async def create_unix_listener(
-    path: str | PathLike[str], *, mode: int | None = None, backlog: int = 65536
+    path: str | bytes | PathLike[str] | PathLike[bytes],
+    *,
+    mode: int | None = None,
+    backlog: int = 65536,
 ) -> SocketListener:
     """
     Create a UNIX socket listener.
@@ -469,7 +472,7 @@ async def create_connected_udp_socket(
 
 async def create_unix_datagram_socket(
     *,
-    local_path: None | str | PathLike[str] = None,
+    local_path: None | str | bytes | PathLike[str] = None,
     local_mode: int | None = None,
 ) -> UNIXDatagramSocket:
     """
@@ -496,9 +499,9 @@ async def create_unix_datagram_socket(
 
 
 async def create_connected_unix_datagram_socket(
-    remote_path: str | PathLike[str],
+    remote_path: str | bytes | PathLike[str],
     *,
-    local_path: None | str | PathLike[str] = None,
+    local_path: None | str | bytes | PathLike[str] = None,
     local_mode: int | None = None,
 ) -> ConnectedUNIXDatagramSocket:
     """
@@ -519,7 +522,7 @@ async def create_connected_unix_datagram_socket(
     :return: a connected UNIX datagram socket
 
     """
-    remote_path = str(Path(remote_path))
+    remote_path = os.fspath(remote_path)
     raw_socket = await setup_unix_local_socket(
         local_path, local_mode, socket.SOCK_DGRAM
     )
@@ -668,7 +671,7 @@ def convert_ipv6_sockaddr(
 
 
 async def setup_unix_local_socket(
-    path: None | str | PathLike[str],
+    path: None | str | bytes | PathLike[str] | PathLike[bytes],
     mode: int | None,
     socktype: int,
 ) -> socket.socket:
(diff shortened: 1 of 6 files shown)


agronholm:
While this is *almost* an annotations only patch, it does replace the use of `str()`. Perhaps there should be a test that creates a socket from a bytes-based path that doesn't cleanly translate to ASCII?

Fuyukai:
What would that actually be testing? It's just passed straight through to the underlying socket.

agronholm:
It would test that a bytes-based socket path would not result in `UnicodeDecodeError`, as it could without this patch.

agronholm:
It's even clearer now that I tried:
```
>>> Path(b"/tmp/foo2")
  File "<stdin>", line 1, in <module>
  File "/usr/lib64/python3.12/pathlib.py", line 1163, in __init__
    super().__init__(*args)
  File "/usr/lib64/python3.12/pathlib.py", line 373, in __init__
    raise TypeError(
TypeError: argument should be a str or an os.PathLike object where __fspath__ returns a str, not 'bytes'
```

Fuyukai:
Also added bytes supports for the other methods and added tests for each distinct one.

src/anyio/_core/_sockets.py
```diff
@@ -245,7 +247,9 @@ async def try_connect(remote_host: str, event: Event) -> None:
     return connected_stream
 
 
-async def connect_unix(path: str | PathLike[str]) -> UNIXSocketStream:
+async def connect_unix(
+    path: str | bytes | PathLike[str] | PathLike[bytes]
```
agronholm, line 251: Hm, wouldn't `PathLike[Any]` work here, given that the variable is bound to `str` and `bytes` anyway?

src/anyio/_core/_sockets.py
```diff
@@ -340,7 +344,10 @@ async def create_tcp_listener(
 
 
 async def create_unix_listener(
-    path: str | PathLike[str], *, mode: int | None = None, backlog: int = 65536
+    path: str | bytes | PathLike[str] | PathLike[bytes],
```
agronholm, line 347: Same as above.

src/anyio/_core/_sockets.py
```diff
@@ -466,7 +473,7 @@ async def create_connected_udp_socket(
 
 async def create_unix_datagram_socket(
     *,
-    local_path: None | str | PathLike[str] = None,
+    local_path: None | str | bytes | PathLike[str] = None,
```
agronholm, line 476: Shouldn't this also accept `PathLike[bytes]`?

tests/test_sockets.py
```diff
@@ -1002,6 +1002,24 @@ async def test_cannot_connect(self, socket_path: Path) -> None:
...
             await connect_unix(socket_path)
 
+    async def test_connecting_using_bytes(
+        self, server_sock: socket.socket, socket_path: Path
+    ) -> None:
+        async with await connect_unix(str(socket_path).encode()):
+            pass
+
+    @pytest.mark.skipif(
+        sys.platform == "darwin", reason="OSX requires valid UTF-8 paths"
```
agronholm, line 1012: Other AnyIO skip markers use `platform.system()`; this should remain consistent.
```suggestion
        platform.system() == "Darwin", reason="macOS requires valid UTF-8 paths"
```

tests/test_sockets.py
```diff
@@ -1101,6 +1119,18 @@ async def test_bind_twice(self, socket_path: Path) -> None:
             async with await create_unix_listener(socket_path):
                 pass
 
+    async def test_listening_bytes_path(self, socket_path: Path) -> None:
+        async with await create_unix_listener(str(socket_path).encode()):
+            pass
+
+    @pytest.mark.skipif(
+        sys.platform == "darwin", reason="OSX requires valid UTF-8 paths"
```
agronholm, line 1127: As above.
```suggestion
        platform.system() == "Darwin", reason="macOS requires valid UTF-8 paths"
```

tests/test_sockets.py
```diff
@@ -1495,6 +1525,20 @@ async def test_send_after_close(self, socket_path: Path) -> None:
...
             await unix_dg.sendto(b"foo", path)
 
+    async def test_local_path_bytes(self, socket_path: Path) -> None:
+        async with await create_unix_datagram_socket(
+            local_path=str(socket_path).encode()
+        ):
+            pass
+
+    @pytest.mark.skipif(
+        sys.platform == "darwin", reason="OSX requires valid UTF-8 paths"
```
agronholm, line 1535: Ditto.
```suggestion
        platform.system() == "Darwin", reason="macOS requires valid UTF-8 paths"
```

agronholm (changes requested):

agronholm:
Also please add a changelog entry in `docs/versionhistory.rst`.

src/anyio/_core/_sockets.py
Fuyukai: Pylance gave me an error when I first tried, but Pyright doesn't. Not sure what's going on there.. I'll trust the CLI and change it.

agronholm: If the CI passes, then that's good enough for me.

tests/test_sockets.py
Fuyukai: For what it's worth, I copied it from the one immediately below which uses ``sys.platform`` :)

agronholm: Oh, that's on me then :)
All the newer skip marks use `platform.system()`.

agronholm (approved):
LGTM now, thanks!

PR #646, opened 2023-12-07
Fix asyncio didn't drop waiters on increase token

EgorBlagov:
I've found several corner cases when on increase of  `CapacityLimiter.total_tokens` (for asyncio backend) new tasks cannot acquire lock, and release mechanism didn't release correct waiters as well.
I've written couple of tests
1. When waiters queue is not empty and new task couldn't acquire the limiter
2. When there is one waiter, and the task that relases the lock, releases the wrong lock.

Unfortunately one of new tests is not stable for `trio`, maybe you can help me to finish that fix?

Thanks

### DIFF
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -1651,17 +1651,11 @@ class CapacityLimiter(BaseCapacityLimiter):
 
         old_value = self._total_tokens
         self._total_tokens = value
-        events = []
-        for event in self._wait_queue.values():
-            if value <= old_value:
-                break
-
-            if not event.is_set():
-                events.append(event)
-                old_value += 1
-
-        for event in events:
+        to_unblock_count = min(len(self._wait_queue), value - old_value)
+        while to_unblock_count > 0:
+            event = self._wait_queue.popitem(last=False)[1]
             event.set()
+            to_unblock_count -= 1
 
     @property
     def borrowed_tokens(self) -> int:
@@ -1669,7 +1663,7 @@ class CapacityLimiter(BaseCapacityLimiter):
 
     @property
     def available_tokens(self) -> float:
-        return self._total_tokens - len(self._borrowers)
+        return self._total_tokens - self.borrowed_tokens
 
     def acquire_nowait(self) -> None:
         self.acquire_on_behalf_of_nowait(current_task())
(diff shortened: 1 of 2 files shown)


agronholm:
I think your tests rely on a specific scheduling order, something that `trio` doesn't guarantee.

agronholm:
The core problem here seems to be that the setter for `total_tokens` doesn't pop the items from the waiter queue like `release_on_behalf_of()` does, and then `acquire_on_behalf_of()` sees the waiter queue and won't let the caller acquire a token even though it should. I'm working on my own set of changes to fix this.

EgorBlagov:
Ok, maybe I can check how trio behaves, I initially encountered it when tried to increase capacity limiter for FastAPI handle dynamically, and it didn't work reliably. Let me try couple of experiments with trio, and see if test can be not that strict.

Though I'm not sure if the blinking test (`test_increase_tokens_lets_others_acquire_after_release `) fails on trio due to inconsistent ordering. (1st is joined, then we increase, 2nd is joined, then 1st releases, and there must be a room for the 3rd worker). I'm guessing that trio misses something like `await checkpoint`. Though I'm not familiar with trio at all.

Thanks for your reply)

EgorBlagov:
here is how I checked it:
```python
import anyio
from anyio import CapacityLimiter
from datetime import datetime


async def do_stuff(x, limiter: CapacityLimiter):
    print(datetime.now(), 'wait limiter', x, limiter.borrowed_tokens)
    
    async with limiter:
        print(datetime.now(), 'entered', 'do_stuf_' + str(x), limiter.borrowed_tokens)
        await anyio.sleep(1)

    print(datetime.now(), 'left', 'do_stuf_' + str(x), limiter.borrowed_tokens)

async def main():

    limiter = CapacityLimiter(1)
        for x in range(100):
            tg.start_soon(do_stuff, x, limiter, name='do_stuff_'+str(x))

        await anyio.sleep(3.5)

        limiter.total_tokens = 10
        print('limiter increased')

anyio.run(main, backend='trio')
```

Indeed trio works fine (and scheduling order is quite random). And if you set backend to asyncio, you'll notice that after that capacity is not increased (at first 10 workers join, and then only one by one they release).
Yeah, test should be somehow simlified.

agronholm:
I've pushed the change and credited you in the changelog.

EgorBlagov:
tnx, glad to help!

PR #647, opened 2023-12-09
Fixed asyncio backend options not being used

### LINKED ISSUE #643: Backend specific options don't appear to be used
### Things to check first

### AnyIO version

4.1.0

### Python version

3.11

### What happened?

As best as I can tell, the backend specific options for both the asyncio and trio options aren't being used

https://github.com/agronholm/anyio/blob/cc18942a384dcb881e78a81c898d7e89d4e1ce0b/src/anyio/_backends/_asyncio.py#L1939-L1960

https://github.com/agronholm/anyio/blob/cc18942a384dcb881e78a81c898d7e89d4e1ce0b/src/anyio/_backends/_trio.py#L807-L814

I'm sorry if I'm missing something, but the backend options don't appear to be used, except for `debug` for asyncio. I don't have a test case, however, so, I apologize again if I'm not seeing how the options are getting applied.

agronholm:
Fixes #643.


DaGenix (approved):
Looks good to me!

Issue #500, opened 2022-11-14
Creating Event when in from_thread.run_sync() context


tapetersen:
When using `to_thread.run_sync` and then calling back again to the event loop with `from_thread.run_sync` I can't create an `anyio.Event` since sniffio fails to detect the running loop (no current_task is set in asyncio which of course is kindof true).

Reproduced in a gist here https://gist.github.com/tapetersen/e85a808f0bba7dd396fa68eafb41012d

tapetersen:
This is not critical for us and can be worked around by creating `asyncio.Event` objects directly (not ideal to tie to backend but ok  in our case) or by making the function async which I would want to avoid since it doesn't really yield to the event loop or await anything.

`from_thread` here is really just used to ensure thread safety since the method modifies some other state that isn't explicitly locked but relies on only being modified by the main thread running the loop.

I don't know about `trio`but at least in asyncio it has been ok to create Event objects outside async contexts and also in different event loops as they don't actually bind to a loop until waited upon since 3.9

PR #651, opened 2023-12-12
Enabled Event and CapacityLimiter to be instantiated outside an event loop

agronholm:
Relates to #500.

### DIFF
--- a/src/anyio/_core/_synchronization.py
+++ b/src/anyio/_core/_synchronization.py
@@ -399,7 +399,7 @@ class Semaphore:
 
 
 class CapacityLimiter:
-    def __new__(cls, total_tokens: int) -> CapacityLimiter:
+    def __new__(cls, total_tokens: float) -> CapacityLimiter:
         return CapacityLimiterAdapter(total_tokens)
 
     async def __aenter__(self) -> None:
@@ -512,10 +512,10 @@ class CapacityLimiter:
 class CapacityLimiterAdapter(CapacityLimiter):
     _internal_limiter: CapacityLimiter | None = None
 
-    def __new__(cls, total_tokens: int) -> CapacityLimiterAdapter:
+    def __new__(cls, total_tokens: float) -> CapacityLimiterAdapter:
         return object.__new__(cls)
 
-    def __init__(self, total_tokens: int) -> None:
+    def __init__(self, total_tokens: float) -> None:
         if not isinstance(total_tokens, int) and total_tokens is not math.inf:
             raise TypeError("total_tokens must be an int or math.inf")
         elif total_tokens < 1:
--- a/src/anyio/_core/_synchronization.py
+++ b/src/anyio/_core/_synchronization.py
@@ -130,7 +130,10 @@ class EventAdapter(Event):
         await self._event.wait()
 
     def statistics(self) -> EventStatistics:
-        return self._event.statistics()
+        if self._internal_event is None:
+            return EventStatistics(tasks_waiting=0)
+
+        return self._internal_event.statistics()
 
 
 class Lock:
@@ -553,7 +556,10 @@ class CapacityLimiterAdapter(CapacityLimiter):
 
     @property
     def total_tokens(self) -> float:
-        return self._limiter.total_tokens
+        if self._internal_limiter is None:
+            return self._total_tokens
+
+        return self._internal_limiter.total_tokens
 
     @total_tokens.setter
     def total_tokens(self, value: float) -> None:
@@ -561,11 +567,17 @@ class CapacityLimiterAdapter(CapacityLimiter):
 
     @property
     def borrowed_tokens(self) -> int:
-        return self._limiter.borrowed_tokens
+        if self._internal_limiter is None:
+            return 0
+
+        return self._internal_limiter.borrowed_tokens
 
     @property
     def available_tokens(self) -> float:
-        return self._limiter.available_tokens
+        if self._internal_limiter is None:
+            return self._total_tokens
+
+        return self._internal_limiter.available_tokens
 
     def acquire_nowait(self) -> None:
         self._limiter.acquire_nowait()
@@ -586,7 +598,15 @@ class CapacityLimiterAdapter(CapacityLimiter):
         self._limiter.release_on_behalf_of(borrower)
 
     def statistics(self) -> CapacityLimiterStatistics:
-        return self._limiter.statistics()
+        if self._internal_limiter is None:
+            return CapacityLimiterStatistics(
+                borrowed_tokens=0,
+                total_tokens=self.total_tokens,
+                borrowers=(),
+                tasks_waiting=0,
+            )
+
+        return self._internal_limiter.statistics()
 
 
 class ResourceGuard:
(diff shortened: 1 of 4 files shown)


graingert (approved):

agronholm:
The motivation is to allow users to create events and capacity limiters on the module level, as import side effects, and in the REPL too. You can already do this with `asyncio.Event()` and `trio.Event()`.

src/anyio/_core/_synchronization.py
```diff
@@ -101,6 +102,32 @@ def statistics(self) -> EventStatistics:
...
+        if self._internal_event is None:
+            self._internal_event = get_async_backend().create_event()
+
+        return self._internal_event
+
+    def set(self) -> None:
+        self._event.set()
+
+    def is_set(self) -> bool:
+        return self._internal_event is not None and self._internal_event.is_set()
```
davidbrochart, line 122: I understand that if the event is not created then it cannot be set, but should `is_set` actually create the event?
```suggestion
        return self._event.is_set()
```
`statistics` also doesn't need the event to be created (it could return empty statistics if it's not), yet it creates it.

src/anyio/_core/_synchronization.py
```diff
@@ -482,6 +509,78 @@ def statistics(self) -> CapacityLimiterStatistics:
         raise NotImplementedError
 
 
+class CapacityLimiterAdapter(CapacityLimiter):
+    _internal_limiter: CapacityLimiter | None = None
+
+    def __new__(cls, total_tokens: int) -> CapacityLimiterAdapter:
+        return object.__new__(cls)
+
+    def __init__(self, total_tokens: int) -> None:
```
davidbrochart, line 518: Line 519 checks if `total_tokens` is not `math.inf` which is a `float`, so should this be reflected in its type?
```suggestion
    def __init__(self, total_tokens: int | float) -> None:
```

agronholm: In that case, the annotation should be `float`, as that already includes integers according to the PEP 484 numeric tower.

smurfix: Well, on the other hand it shouldn't be **any** float. Just `math.inf`.
Shouldn't `int | math.inf` Just Work?

agronholm: No, because `math.inf` is not a type.

smurfix: Neither is `None`, but that works just fine.

agronholm: `None` is explicitly special cased in PEP 484.

src/anyio/_core/_synchronization.py
```diff
@@ -76,7 +77,7 @@ class SemaphoreStatistics:
 
 class Event:
     def __new__(cls) -> Event:
-        return get_async_backend().create_event()
+        return EventAdapter()
```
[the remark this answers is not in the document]
agronholm: Good idea.

src/anyio/_core/_synchronization.py
smurfix: Ugh. They should just have accepted arbitrary singletons. (I do wonder whether any type checkers do …)

src/anyio/_core/_synchronization.py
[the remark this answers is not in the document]
agronholm: Done.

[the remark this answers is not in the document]
agronholm: Done.

src/anyio/_core/_synchronization.py
agronholm: Added.

src/anyio/_core/_synchronization.py
davidbrochart: I think it should be `int | Literal[math.inf]`, but mypy doesn't like it. Maybe it's a bug in mypy?

agronholm: No, it's not. Floats are not valid literals according to [PEP 586](https://peps.python.org/pep-0586/#illegal-parameters-for-literal-at-type-check-time).

davidbrochart: Thanks.
BTW, why not using `None` to mean an infinite number of tokens? `math.inf` carries a direct meaning, but `None` also means "I don't want to specify a finite number of tokens", which easily translates to a infinite number of tokens IMO.

agronholm: It couldn't been done, but I was following trio's API design which uses `math.inf` instead.

tests/test_synchronization.py
```diff
@@ -208,6 +210,18 @@ async def waiter() -> None:
 
         assert event.statistics().tasks_waiting == 0
 
+    def test_instantiate_outside_event_loop(
```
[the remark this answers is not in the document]
agronholm: Done.

src/anyio/_core/_synchronization.py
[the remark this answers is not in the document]
agronholm: With my latest change, they won't be getting the adapters, will they?

src/anyio/_core/_synchronization.py
```diff
@@ -482,6 +520,95 @@ def statistics(self) -> CapacityLimiterStatistics:
...
+    @property
+    def total_tokens(self) -> float:
+        if self._internal_limiter is None:
+            return self._total_tokens
+
+        return self._internal_limiter.total_tokens
+
+    @total_tokens.setter
+    def total_tokens(self, value: float) -> None:
+        self._limiter.total_tokens = value
```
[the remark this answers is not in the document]
agronholm: Frankly, the idea that someone would change the adapter's total tokens from a worker thread had not even occurred to me. The whole idea seems pretty bizarre to me.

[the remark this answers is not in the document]
agronholm: I'll revert my latest commit then?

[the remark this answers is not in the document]
agronholm: I have some mixed feelings about this. It might be reasonable to allow the tokens to be set from outside the event loop, just like `trio.CapacityLimiter` allows.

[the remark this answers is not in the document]
agronholm: The more I think about it, the more I'm leaning towards allowing users to mutate `total_tokens` before the initialization.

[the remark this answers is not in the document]
agronholm: Yeah, generally the API should be considered thread-unsafe, save for that parts that were explicitly designed to allow access from worker threads (the `anyio.from_thread` module).

agronholm:
Thank you for the reviews!

tapetersen:
> is the motivation for this [#500 (comment)](https://github.com/agronholm/anyio/issues/500#issue-1448408878) / [#501 (comment)](https://github.com/agronholm/anyio/discussions/501#discussion-4572066)? if so, @tapetersen's example code has been working as intended since #524 was merged, so a delayed instantiation trick may not be necessary. see also [#523 (comment)](https://github.com/agronholm/anyio/issues/523#issuecomment-1429653302) / [python-trio/sniffio#35 (comment)](https://github.com/python-trio/sniffio/issues/35#issuecomment-1635306783).

Oh i unfortunately hadn't noticed before. Like mentioned it wasn't really hard to workaround in our case. Nice to see it getting solved outside a loop at all as well as that was the case of asyncio and why I was surprised it didn't work.

PR #655, opened 2023-12-18
Path.relative_to() walk_up support

cjntaylor:
Adding support for the `walk_up` keyword argument introduced in `Path.relative_to()` in python 3.12

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -5,7 +5,7 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
 **UNRELEASED**
 
-- Added support for the python 3.12 ``walk_up`` keyword argument in
+- Added support for the Python 3.12 ``walk_up`` keyword argument in
   ``anyio.Path.relative_to()`` (PR by Colin Taylor)
 
 **4.2.0**
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -3,6 +3,11 @@ Version history
 
 This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
+**UNRELEASED**
+
+- Added support for the python 3.12 ``walk_up`` keyword argument in
+  ``anyio.Path.relative_to()`` (PR by Colin Taylor)
+
 **4.2.0**
 
 - Add support for ``byte``-based paths in ``connect_unix``, ``create_unix_listeners``,
--- a/src/anyio/_core/_fileio.py
+++ b/src/anyio/_core/_fileio.py
@@ -514,8 +514,17 @@ class Path:
     ) -> str:
         return await to_thread.run_sync(self._path.read_text, encoding, errors)
 
-    def relative_to(self, *other: str | PathLike[str]) -> Path:
-        return Path(self._path.relative_to(*other))
+    if sys.version_info >= (3, 12):
+
+        def relative_to(
+            self, *other: str | PathLike[str], walk_up: bool = False
+        ) -> Path:
+            return Path(self._path.relative_to(*other, walk_up=walk_up))
+
+    else:
+
+        def relative_to(self, *other: str | PathLike[str]) -> Path:
+            return Path(self._path.relative_to(*other))
 
     async def readlink(self) -> Path:
         target = await to_thread.run_sync(os.readlink, self._path)
--- a/tests/test_fileio.py
+++ b/tests/test_fileio.py
@@ -1,18 +1,16 @@
 from __future__ import annotations
 
-import contextlib
 import os
 import pathlib
 import platform
 import socket
 import stat
 import sys
-from types import TracebackType
 
 import pytest
 from _pytest.tmpdir import TempPathFactory
 
-from anyio import AsyncFile, Path, open_file, to_thread, wrap_file
+from anyio import AsyncFile, Path, open_file, wrap_file
 
 pytestmark = pytest.mark.anyio
 
@@ -66,27 +64,6 @@ class TestAsyncFile:
 
 
 class TestPath:
-    class ChangeDirectory(contextlib.AbstractAsyncContextManager):
-        def __init__(self, cwd: Path) -> None:
-            super().__init__()
-            self.__cwd: Path = cwd
-            self.__pwd: Path | None = None
-
-        async def __aenter__(self) -> Path:
-            self.__pwd = await Path.cwd()
-            await to_thread.run_sync(os.chdir, self.__cwd)
-            return self.__cwd
-
-        async def __aexit__(
-            self,
-            exc_type: type[BaseException] | None,
-            exc_value: BaseException | None,
-            traceback: TracebackType | None,
-        ) -> bool | None:
-            if self.__pwd is not None:
-                await to_thread.run_sync(os.chdir, self.__pwd)
-            return await super().__aexit__(exc_type, exc_value, traceback)
-
     @pytest.fixture
     def populated_tmpdir(self, tmp_path: pathlib.Path) -> pathlib.Path:
         tmp_path.joinpath("testfile").touch()
@@ -467,16 +444,20 @@ class TestPath:
         sys.version_info < (3, 12),
         reason="Path.relative_to(walk_up=<bool>) is only available on Python 3.12+",
     )
-    async def test_relative_to_sibling(self, populated_tmpdir: pathlib.Path) -> None:
+    async def test_relative_to_sibling(
+        self,
+        populated_tmpdir: pathlib.Path,
+        monkeypatch: pytest.MonkeyPatch,
+    ) -> None:
         subdir = Path(populated_tmpdir / "subdir")
         sibdir = Path(populated_tmpdir / "sibdir")
 
         with pytest.raises(ValueError):
             subdir.relative_to(sibdir, walk_up=False)
 
-        async with TestPath.ChangeDirectory(sibdir) as path:
-            relpath = subdir.relative_to(path, walk_up=True) / "dummyfile1.txt"
-            assert os.access(relpath, os.R_OK)
+        monkeypatch.chdir(sibdir)
+        relpath = subdir.relative_to(sibdir, walk_up=True) / "dummyfile1.txt"
+        assert os.access(relpath, os.R_OK)
 
     async def test_rename(self, tmp_path: pathlib.Path) -> None:
         path = tmp_path / "somefile"
--- a/tests/test_fileio.py
+++ b/tests/test_fileio.py
@@ -1,16 +1,18 @@
 from __future__ import annotations
 
+import contextlib
 import os
 import pathlib
 import platform
 import socket
 import stat
 import sys
+from types import TracebackType
 
 import pytest
 from _pytest.tmpdir import TempPathFactory
 
-from anyio import AsyncFile, Path, open_file, wrap_file
+from anyio import AsyncFile, Path, open_file, to_thread, wrap_file
 
 pytestmark = pytest.mark.anyio
 
@@ -64,14 +66,40 @@ class TestAsyncFile:
 
 
 class TestPath:
+    class ChangeDirectory(contextlib.AbstractAsyncContextManager):
+        def __init__(self, cwd: Path) -> None:
+            super().__init__()
+            self.__cwd: Path = cwd
+            self.__pwd: Path | None = None
+
+        async def __aenter__(self) -> Path:
+            self.__pwd = await Path.cwd()
+            await to_thread.run_sync(os.chdir, self.__cwd)
+            return self.__cwd
+
+        async def __aexit__(
+            self,
+            exc_type: type[BaseException] | None,
+            exc_value: BaseException | None,
+            traceback: TracebackType | None,
+        ) -> bool | None:
+            if self.__pwd is not None:
+                await to_thread.run_sync(os.chdir, self.__pwd)
+            return await super().__aexit__(exc_type, exc_value, traceback)
+
     @pytest.fixture
     def populated_tmpdir(self, tmp_path: pathlib.Path) -> pathlib.Path:
         tmp_path.joinpath("testfile").touch()
         tmp_path.joinpath("testfile2").touch()
+
         subdir = tmp_path / "subdir"
-        subdir.mkdir()
-        subdir.joinpath("dummyfile1.txt").touch()
-        subdir.joinpath("dummyfile2.txt").touch()
+        sibdir = tmp_path / "sibdir"
+
+        for subpath in (subdir, sibdir):
+            subpath.mkdir()
+            subpath.joinpath("dummyfile1.txt").touch()
+            subpath.joinpath("dummyfile2.txt").touch()
+
         return tmp_path
 
     async def test_properties(self) -> None:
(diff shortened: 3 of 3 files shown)


docs/versionhistory.rst
```diff
@@ -40,6 +40,8 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
   * ``anyio.Path.with_segments()``
 
   (PR by Ganden Schaffner)
+- Added support for the python 3.12 ``walk_up`` keyword argument in
+  ``anyio.Path.relative_to()`` (PR by Colin Taylor)
```
agronholm, line 44: Don't try the amend the release notes of an already published release. Add a new section for **UNRELEASED**.

src/anyio/_core/_fileio.py
```diff
@@ -514,8 +514,17 @@ async def read_text(
...
         return await to_thread.run_sync(self._path.read_text, encoding, errors)
 
-    def relative_to(self, *other: str | PathLike[str]) -> Path:
-        return Path(self._path.relative_to(*other))
+    if sys.version_info >= (3, 12):
+
+        def relative_to(
+            self, *other: str | PathLike[str], walk_up: bool = False
+        ) -> Path:
+            return Path(self._path.relative_to(*other, walk_up=walk_up))
```
agronholm, line 522: Add a test for this please.

agronholm (changes requested):

docs/versionhistory.rst
cjntaylor: Apologies, wasn't intended - I transposed into the wrong section. I'll shift to the correct place

src/anyio/_core/_fileio.py
cjntaylor: Added. Required a little bit of modification to the test environment to provide a sibling folder to try to reference. Let me know if this isn't acceptable - I can iterate on another approach

agronholm:
What's all the socket stuff, and why did you reformat `.pre-commit-config.yaml`?

cjntaylor:
> What's all the socket stuff, and why did you reformat `.pre-commit-config.yaml`?

Did you read the review sections in the PR that explain WHY the socket stuff was modified? 

To summarize, the full set of tests would not pass on my system without the change. On many modern unices (and every modern MacOS), creating socket files in mktemp(d) created directories (including python's implementation) creates socket paths that are too long. This causes socket operations to fail with `OSError: AF_UNIX path too long`. 

I needed to alter the behaviour to get the unix tests to pass, since I was adding tests for `walk_up` as requested. While my tests only affect a single test file, it's considered polite/good practice to make sure you're "all green" when modifying tests in a PR. My guess is that either the MacOS CI you're using has an altered `/tmp` mount, or this issue doesn't affect older versions of MacOS. I'm developing and testing on MacOS Ventura 13.6.1, and AFAIK this would be an issue pretty far back. The mktemp implementation hasn't changed in MacOS in a long time, as soon as the tempdir implementation uses `/private/var/folders` the path will end up too long. See the referenced cpython issue here: python/cpython#93852. Max path length is 107 bytes on MacOS, and 108 on Linux.

I included them in the PR because while they're a separate concern, they do block this PR from passing tests on updated systems. If you'd prefer, I can split this change out into a separate PR - but then that means this PR has the potential to break tests without this change.

The `--no-warn-unused-ignores` flag passed to mypy hook is to get that phase to pass with minimal changes. I'm assuming the comments are there/needed for a reason. Without either removing the comments or passing the flags, I get the following:

```
mypy.....................................................................Failed
- hook id: mypy
- exit code: 1

tests/test_taskgroups.py:263: error: Unused "type: ignore" comment  [unused-ignore]
tests/test_taskgroups.py:1357: error: Unused "type: ignore" comment  [unused-ignore]
tests/test_taskgroups.py:1358: error: Unused "type: ignore" comment  [unused-ignore]
tests/test_taskgroups.py:1367: error: Unused "type: ignore" comment  [unused-ignore]
tests/test_synchronization.py:674: error: Unused "type: ignore" comment  [unused-ignore]
tests/test_sockets.py:871: error: Unused "type: ignore" comment  [unused-ignore]
tests/test_sockets.py:877: error: Unused "type: ignore" comment  [unused-ignore]
tests/test_sockets.py:1223: error: Unused "type: ignore" comment  [unused-ignore]
tests/test_sockets.py:1242: error: Unused "type: ignore" comment  [unused-ignore]
tests/test_sockets.py:1310: error: Unused "type: ignore" comment  [unused-ignore]
tests/test_sockets.py:1346: error: Unused "type: ignore" comment  [unused-ignore]
tests/test_sockets.py:1351: error: Unused "type: ignore" comment  [unused-ignore]
tests/test_sockets.py:1369: error: Unused "type: ignore" comment  [unused-ignore]
tests/test_sockets.py:1371: error: Unused "type: ignore" comment  [unused-ignore]
Found 14 errors in 3 files (checked 15 source files)
```

I've done nothing special for pre-commit, literally just `pre-commit install` using the exact unmodified file in the repository. All of the installed package versions look identical, so I'm unsure why this happens for me and not on the CI.

The whitespace formatting shifts are an unintentional side-effect of my local dev setup; I use `prettier` by default to minimize whitespace thrashing in my files. The file as modified is actually formatted per the YAML spec since that's well defined, but, if you'd prefer, I can revert the whitespace changes. My apologies - I wasn't intending to modify formatting, only the arg.

cjntaylor:
Apologies, I realized I messed up my change to split the socket test modifications out into a separate commit. That's now fixed, so you can see just the changes that were made to that test file

cjntaylor:
**Edited to reflect recent changes**

> What's all the socket stuff, and why did you reformat `.pre-commit-config.yaml`?

The full set of unit tests would not pass on my system; Unix socket tests were failing with `OSError: AF_UNIX path too long`. Paths generated by `tmp_path_factory` exceeding the max unix socket path length on macOS. My new tests for `walk_up` are limited to `test_fileio.py`, but I thought it polite/good practice to make sure the test suite was "all green" in the PR.

On macOS, the unix socket path length limit is 107 bytes; the base temp directory of `/private/var/folders` combined with the deep hierarchy of subdirectories pytest creates regularly exceeds this limit by 10+ bytes. Linux also has this limit (108 bytes), but because it's almost always rooted at `/tmp` or `/var/tmp` it's unlikely to trigger this error. These limits are discussed more here: python/cpython#93852

A simple workaround was to use `tempdir.TemporaryDirectory` for socket paths. Even on macOS, these paths are short enough that they shouldn't trigger this error, while still using standard tempdir conventions. As you can see, the implementation can easily be substituted in place for the original one. I've updated the commits with this in mind.

I'm unsure why these path lengths aren't an issue on the macOS CI instances; it's very odd they're not. The only major difference I can find is that I'm running Ventura (macOS 13) vs the CI's Monterey (macOS 12); otherwise the pytest version and all other packages match up exactly. Even if they somehow still didn't, I don't see anything in the pytest commit history indicating a change to how `tmp_path_factory` locations are generated. Maybe the CI skirt's the limit using a shorter username (mine's only 8 characters) or somehow different temp path generation?

I included these changes in this PR because while they're a separate concern, they do block this one from passing all tests. I figure this is relatively important so unit tests can be checked locally, without requiring a `push` or putting load on the CI. However, if you prefer, I can create a second PR for them specifically, and then re-cut this PR to be based on that one (since I would want the new PR to precede this one so I can run local tests).

The `--no-warn-unused-ignores` flag passed to mypy hook is for a similar reason. On my system, using the unmodified `.pre-commit-config.yaml` and `pre-commit install`, the mypy hook fails as follows:

```
mypy.....................................................................Failed
- hook id: mypy
- exit code: 1

tests/test_taskgroups.py:263: error: Unused "type: ignore" comment  [unused-ignore]
tests/test_taskgroups.py:1357: error: Unused "type: ignore" comment  [unused-ignore]
tests/test_taskgroups.py:1358: error: Unused "type: ignore" comment  [unused-ignore]
tests/test_taskgroups.py:1367: error: Unused "type: ignore" comment  [unused-ignore]
tests/test_synchronization.py:674: error: Unused "type: ignore" comment  [unused-ignore]
tests/test_sockets.py:871: error: Unused "type: ignore" comment  [unused-ignore]
tests/test_sockets.py:877: error: Unused "type: ignore" comment  [unused-ignore]
tests/test_sockets.py:1223: error: Unused "type: ignore" comment  [unused-ignore]
tests/test_sockets.py:1242: error: Unused "type: ignore" comment  [unused-ignore]
tests/test_sockets.py:1310: error: Unused "type: ignore" comment  [unused-ignore]
tests/test_sockets.py:1346: error: Unused "type: ignore" comment  [unused-ignore]
tests/test_sockets.py:1351: error: Unused "type: ignore" comment  [unused-ignore]
tests/test_sockets.py:1369: error: Unused "type: ignore" comment  [unused-ignore]
tests/test_sockets.py:1371: error: Unused "type: ignore" comment  [unused-ignore]
Found 14 errors in 3 files (checked 15 source files)
```

I'm assuming the referenced `#ignore` comments are there/needed for a reason. The added arg allows this hook to pass without removing the comments. If the comments shouldn't be there instead, I can revert this change and remove the comments instead. If this hook is going to work locally, I think it needs to be one or the other? It's also odd to me this is doesn't affect the CI - the package versions used by my local hook are identical. I can't find any differences between the two environments. Since I can't commit without this passing, this either must be part of this PR, or another preceding PR. Let me know if you'd prefer the latter.

My default dev environment uses `prettier` to minimize whitespace thrashing / bikeshedding in supported files - I didn't catch that it had altered the formatting when I added the mypy arg. However, it did make the formatting more consistent across file and closer to "standard" YAML formatting. If you prefer, I can revert these changes - just let me know.

agronholm:
> However, it did make the formatting more consistent across file and closer to "standard" YAML formatting. If you prefer, I can revert these changes - just let me know.

Could you elaborate on where this "standard" YAML formatting comes from? As far as I can tell, my formatting is consistent with the Y[AML 1.2](https://yaml.org/spec/1.2.2/) standard.

cjntaylor:
From the section that talks about [indentation](https://yaml.org/spec/1.2.2/#61-indentation-spaces):

> Each [node](https://yaml.org/spec/1.2.2/#nodes) must be indented further than its parent [node](https://yaml.org/spec/1.2.2/#nodes). All sibling [nodes](https://yaml.org/spec/1.2.2/#nodes) must use the exact same indentation level. However the [content](https://yaml.org/spec/1.2.2/#nodes) of each sibling [node](https://yaml.org/spec/1.2.2/#nodes) may be further indented independently.

I put "standard" in quotes because that definition is ambiguous ("may be further indented independently" - should you? how much?). Prettier chose an unambiguous format based on developer feedback, to eliminate this exact bikeshedding: prettier/prettier#4723. 

I happen to agree with the issue argument for indentation - it's both valid per the spec AND clearer to read. However, I'd also accept standardized/consistent non-indentation even if I don't care for it. A standardized format makes all files consistent, regardless of developer, editor, etc. It _eliminates_ entire classes of bugs, not the least of which is repo thrashing. This the exact same reason black/ruff python formatting isn't configurable. It's also why I run `prettier` by default.

FWIW: [pre-commit's docs](https://pre-commit.com/#2-add-a-pre-commit-configuration) use an absolutely wild format I've never seen before. I don't recommend it at all; I've never seen any other yaml formatted that way. However, you could make the argument this is what the tool developer _intends_ for their configuration file, and I'd accept that. The entire file would need reformatted to match though 😉 

Regardless of the general argument, though, the [file](https://github.com/agronholm/anyio/blob/master/.pre-commit-config.yaml#L37-L39) isn't consistently indented to begin with. That's why prettier picked up on and made the change. In other parts of the file, the `hooks` sequence and other child sequences are indented, but the modified one is not:

(I've substituted "·" to make the leading whitespace visible)

```yaml
# This is the configuration file for pre-commit (https://pre-commit.com/).
# To use:
# * Install pre-commit (https://pre-commit.com/#installation)
# * Copy this file as ".pre-commit-config.yaml"
# * Run "pre-commit install".
repos:
··- repo: https://github.com/pre-commit/pre-commit-hooks
····rev: v4.5.0
····hooks: # Children indented here (6 spaces)
······- id: check-toml
······- id: check-yaml
······- id: debug-statements
······- id: end-of-file-fixer
······- id: mixed-line-ending
········args: ["--fix=lf"]
······- id: trailing-whitespace

··- repo: https://github.com/astral-sh/ruff-pre-commit
····rev: v0.1.8
····hooks: # Children indented here (6 spaces)
······- id: ruff
········args: [--fix, --show-fixes]
······- id: ruff-format

··- repo: https://github.com/pre-commit/mirrors-mypy
····rev: v1.7.1
····hooks: # Children indented here (6 spaces)
······- id: mypy
········args: [--no-warn-unused-ignores]
········additional_dependencies: # Children indented here (10 spaces)
··········- pytest
··········- trio >= 0.23
··········- packaging

··- repo: https://github.com/pre-commit/pygrep-hooks
····rev: v1.10.0
····hooks: # !! Children not indented here (4 spaces)
····- id: rst-backticks
····- id: rst-directive-colons
····- id: rst-inline-touching-normal
```

At the end of the day, this is bikeshedding. I can revert the whitespace change if you'd like; it's water under the bridge to me. While the original change was accidental, I'd argue retroactively I want to include the change because I already need to modify the file to add the `--no-warn-unused-ignores` flag and it makes the format more consistent/easier to read.

uSpike:
Hi, I just wanted to add that I think it'd be good to open a separate issue/PR for changes unrelated to the title of this PR.  I think the original changes are great and I think it's a lot less cognitive load on reviewers and agronholm to review when done separately.

Cheers!

cjntaylor:
> Hi, I just wanted to add that I think it'd be good to open a separate issue/PR for changes unrelated to the title of this PR

The only reason I haven't is because they're indirectly related; the full test suite (which includes new tests relevant to this PR that were specifically requested) won't pass without the other test changes. At least locally. For some inexplicable reason it doesn't break on the macOS CI instances, so maybe that's good enough?

I'm happy to do so though. I'll have to wrap my head around making this PR contingent on the PR you're requesting be added (it'll need to be merged first), but, other than that, I think it's fine

Sorry for the confusion. From the beginning I've been indicating that I was unsure of how to handle the other changes w.r.t. this PR so I'm glad someone has weighed in.

Thanks!

cjntaylor:
@uSpike Should I split out the `.pre-commit-config.yaml` changes into a third PR? That one's trickier because if I do I'll have to disable pre-commit checking locally; I won't be able to commit otherwise (the `mypy` hook fails with `Unused "type: ignore" comment`). 

This is the issue with splitting things out - it becomes a bit of a rabbit hole and it's very easy to end up in broken states.

agronholm:
The changes to `.pre-commit-config.yaml` are largely unncessary. There seems to be a mypy bug that causes it to behave differently when only a portion of the files are being scanned, as opposed to the whole code base. When that happens, I check with `pre-commit run -a` to see if it's a real problem, and if it doesn't show up in that scan, I just disable the git hooks for that commit. The workaround would be to force mypy to always check all the files with the following config:
```yaml
        args: [src, tests]
        pass_filenames: false
```

cjntaylor:
@uSpike - PR's have been split into #659, #660 and #661, with 659's pre-commit changes determined as irrelevant and removed.

@agronholm - Thanks for your patience in all of this, and apologies for over-complicating things. I've completely isolated the changes into separate PRs so you can consider them independently, and made sure that each one is as minimal a change as possible. 

Thank you for the workaround. I've closed #659 accordingly.

Please note that this PR will now fail to run the full test suite on macOS >= 13. Running `test_fileio.py` in isolation appears to work correctly, and the CI does pass

tests/test_fileio.py
```diff
@@ -64,14 +66,40 @@ async def test_wrap_file(self, tmp_path: pathlib.Path) -> None:
...
+
+        async def __aexit__(
+            self,
+            exc_type: type[BaseException] | None,
+            exc_value: BaseException | None,
+            traceback: TracebackType | None,
+        ) -> bool | None:
+            if self.__pwd is not None:
+                await to_thread.run_sync(os.chdir, self.__pwd)
+            return await super().__aexit__(exc_type, exc_value, traceback)
```
agronholm, line 88: This seems a bit overkill, why not use `monkeypatch.chdir()` instead?

cjntaylor: First time I've ever heard of it 😅

That's much simpler. I've scoped it using `monkeypatch.context()` to try to minimize error cases

agronholm: Explain – how does that help?

cjntaylor: The pytest docs have a warning:

> Mind that patching stdlib functions and some third-party libraries used by pytest might break pytest itself, therefore in those cases it is recommended to use [MonkeyPatch.context()](https://docs.pytest.org/en/latest/reference/reference.html#pytest.MonkeyPatch.context) to limit the patching to the block you want tested

The note is [here](https://docs.pytest.org/en/latest/how-to/monkeypatch.html#global-patch-example-preventing-requests-from-remote-operations). Since pytest is highly dependent on locating and working with tests relative to file paths, I figured `os.chdir` was semi-dangerous to patch. Scoping it like this seems like a good precaution

agronholm: You're not patching `os.chdir()`, you're just switching the working directory.

agronholm: Docs: https://docs.pytest.org/en/6.2.x/reference.html#pytest.MonkeyPatch.chdir

cjntaylor: My mistake. Fixed. I greatly dislike the naming of that fixture / chdir belongs somewhere else - but that's for me to take up with the pytest folks.

agronholm (approved):
LGTM, thanks!

PR #659, opened 2023-12-28
Adding mypy argument to ignore comment error

cjntaylor:
The `mypy` hook in pre-commit fails with `error: Unused "type: ignore" comment` on several test files. Assuming that these comments are necessary, this flag configures mypy to ignore these errors and allows the hook to pass. 

Alternatively, the offending comments could be removed if that solution is acceptable/preferable; doing so would also allow the hook to pass. Please let me know if this would be the case.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -26,6 +26,7 @@ repos:
     rev: v1.8.0
     hooks:
       - id: mypy
+        args: [--no-warn-unused-ignores]
         additional_dependencies:
           - pytest
           - trio >= 0.23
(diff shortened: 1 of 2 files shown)


agronholm:
I'm hoping that you saw my [comment](https://github.com/agronholm/anyio/pull/655#issuecomment-1871464714) on the other PR. I'm not favorable to ignoring the comment errors, and I provided an alternative.

cjntaylor:
I just did - I'll drop this PR. The workaround is fine but feels unnecessary as it'll slow down the commit check pretty considerably. If it's acceptable to you to temporarily disable pre-commit in that context, I'll just do that.

agronholm:
The checks are not really skipped if you do a manual run when this happens. _Some_ workaround will be necessary until this is sorted on the mypy side.

PR #660, opened 2023-12-28
Adding prettier formatting

cjntaylor:
*Lets handle the yaml formatting changes properly* 😉

Adding prettier for consistent formatting of non-python files. Ruff format only applies to python files, however, other files in the repo still have similar concerns (consistency, whitespace thrashing, etc). Prettier addresses this concern in the same way as ruff; minimal configurability to prevent bikeshedding.

Prettier is written in javascript but is a generalized tool, handling several other source types (`markdown`, `yaml`, `json`, etc). While a tool in the python ecosystem would be preferrable, AFAIK no such tool exists. This appears to be why `pre-commit` itself directly hosts a [prettier integration](https://github.com/pre-commit/mirrors-prettier).

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -14,20 +14,20 @@ jobs:
     runs-on: ubuntu-latest
     environment: release
     steps:
-    - uses: actions/checkout@v4
-    - name: Set up Python
-      uses: actions/setup-python@v5
-      with:
-        python-version: 3.x
-    - name: Install dependencies
-      run: pip install build
-    - name: Create packages
-      run: python -m build
-    - name: Archive packages
-      uses: actions/upload-artifact@v3
-      with:
-        name: dist
-        path: dist
+      - uses: actions/checkout@v4
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: 3.x
+      - name: Install dependencies
+        run: pip install build
+      - name: Create packages
+        run: python -m build
+      - name: Archive packages
+        uses: actions/upload-artifact@v3
+        with:
+          name: dist
+          path: dist
 
   publish:
     name: Publish build artifacts to the PyPI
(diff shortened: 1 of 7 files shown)


cjntaylor:
I'll leave this up for posterity, on the off-chance you're interested. It's water under the bridge to me. I've stripped the unnecessary mypy changes so it merges cleanly from master.

Feel free to close

agronholm:
Doesn't prettier support reading its configuration from `pyproject.toml`?

cjntaylor:
> Doesn't prettier support reading its configuration from `pyproject.toml`?

I feel like I've seen this before, but I can't find any reference to it. Have you come across something; I would also much prefer that (I'm a big fan of single config file).

agronholm:
No, I haven't actually come across prettier itself before this.

cjntaylor:
Ah, no worries. 

Unfortunately this is the only tool I've found that works consistently for these other file formats, but it comes from the javascript/node world, so it's not particularly designed with python in mind. From my basic searching, I haven't yet found anything that would put the config internal to pyproject.toml; it supports a number of different files but not pyproject: https://prettier.io/docs/en/configuration.html

I've always used the TOML file variants when working with python, since the language standardized it's main config format on TOML (at least as far as I'm concerned now that PEP621 is final). Don't merge for now - I'll see if there's a way to accomplish this without too much effort.

cjntaylor:
I looked into it more. It's possible to point prettier at pyproject.toml with a "--config" argument, but it spews warnings about unrecognized keys. The config settings would also have to be top level (not in a tool section), so I don't think it's a good solution.

It is possible to pass the few configuration parameters as arguments themselves instead of using a config. However, this only works in the context of pre-commit and would be inconsistent with any other tool that also uses prettier. I gave this an attempt and the results were also not stellar - you just end up with different configuration files.

So for now, I think these additional config files are the best general solution that seems to work everywhere, including within pre-commit. It has given me an idea to see about fixing this more generically for python, including `pyproject.toml` support - but I wouldn't wait if you think this is useful (that'll take a bit)

agronholm:
YAML formatting is now handled via `yamllint`.

PR #661, opened 2023-12-28
Tests fail with "OSError: AF_UNIX path too long" because socket paths exceed max length (107 bytes)

cjntaylor:
*Splitting apart into a separate PR*

Unix socket tests were failing with OSError: AF_UNIX path too long. The paths generated by pytest's `tmp_path_factory` fixture on macOS exceed the system max socket path length.

On macOS, the unix socket path length limit is 107 bytes; the base temp directory of `/private/var/folders` combined with the subdirectories added by `tmp_path_factory` generates paths regularly exceeding this limit by 10+ bytes. Linux also has this limit (108 bytes), but because it's almost always rooted at `/tmp` or `/var/tmp` it's unlikely to trigger this error. These limits are discussed more here: python/cpython#93852

A simple workaround was to use `tempdir.TemporaryDirectory` for socket paths. Even on macOS, these paths are short enough that they shouldn't trigger this error, while still using standard tempdir conventions/locations.

I'm unsure why these path lengths aren't an issue on the macOS CI instances; it's very odd they're not. The only major difference I can find is that I'm running Ventura (macOS 13) vs the CI's Monterey (macOS 12); otherwise my pytest version matches up exactly. I also don't see anything in the pytest commit history indicating a change to how tmp_path_factory locations are generated.

### DIFF
--- a/tests/test_sockets.py
+++ b/tests/test_sockets.py
@@ -7,6 +7,7 @@ import os
 import platform
 import socket
 import sys
+import tempfile
 import threading
 import time
 from contextlib import suppress
@@ -14,7 +15,7 @@ from pathlib import Path
 from socket import AddressFamily
 from ssl import SSLContext, SSLError
 from threading import Thread
-from typing import Any, Iterable, Iterator, NoReturn, TypeVar, cast
+from typing import Any, Generator, Iterable, Iterator, NoReturn, TypeVar, cast
 
 import psutil
 import pytest
@@ -707,8 +708,11 @@ class TestTCPListener:
 )
 class TestUNIXStream:
     @pytest.fixture
-    def socket_path(self, tmp_path_factory: TempPathFactory) -> Path:
-        return tmp_path_factory.mktemp("unix").joinpath("socket")
+    def socket_path(self) -> Generator[Path, None, None]:
+        # Use stdlib tempdir generation
+        # Fixes `OSError: AF_UNIX path too long` from pytest generated temp_path
+        with tempfile.TemporaryDirectory() as path:
+            yield Path(path) / "socket"
 
     @pytest.fixture(params=[False, True], ids=["str", "path"])
     def socket_path_or_str(self, request: SubRequest, socket_path: Path) -> Path | str:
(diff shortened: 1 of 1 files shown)


agronholm:
Out of curiosity, what kinds of paths are you getting from `tmp_path_factory`?

docs/versionhistory.rst
```diff
@@ -1,6 +1,12 @@
 Version history
 ===============
 
+**UNRELEASED**
+
+- Modified socket tests to generate unix socket paths of lengths less than 107 bytes.
+  Modern unices enforce this limitation and throw an OSError when the path is too long
+  (PR by Colin Taylor)
```
agronholm, line 8: There's no need for a changelog entry as this is just an internal change.
```suggestion
```

docs/versionhistory.rst
```diff
@@ -1,6 +1,12 @@
...
+
```
agronholm, line 9: There's no need for a changelog entry as this is just an internal change.
```suggestion
```

cjntaylor: Just being thorough - removed.

agronholm (approved):

cjntaylor:
> Out of curiosity, what kinds of paths are you getting from `tmp_path_factory`?

They look like this:

```
/private/var/folders/0w/hxs339xd1jx78kt6gpy0b8_w0000gp/T/pytest-of-cjntaylor/pytest-80/unix50/socket
/private/var/folders/0w/hxs339xd1jx78kt6gpy0b8_w0000gp/T/pytest-of-cjntaylor/pytest-80/unix51/socket
/private/var/folders/0w/hxs339xd1jx78kt6gpy0b8_w0000gp/T/pytest-of-cjntaylor/pytest-80/unix52/socket
/private/var/folders/0w/hxs339xd1jx78kt6gpy0b8_w0000gp/T/pytest-of-cjntaylor/pytest-80/unix53/socket
```

These are short enough to not throw AF_UNIX errors. I must have run the tests enough to increase the counts enough to push the path past 107 bytes. Explains why the CI works - it's tearing down it's temp folder each run so those counts will never get that high

Up to you if you still think this is an issue; it's a little close for comfort for me but the average person would probably not run into it 🤷‍♂️

agronholm:
Looks uncomfortably close to the limit still.

PR #663, opened 2024-01-05
Emit ResourceWarnings for unclosed `MemoryObject*Stream`

heckad:


agronholm:
The best one can safely do in `__del__()` is to raise a `ResourceWarning`. Closing is not safe, as the destructor could be called in any thread.

agronholm:
Do you intend to replace the close() into raising a ResourceWarning? If not, then I'll close this.

agronholm (changes requested):
Aside from the suggested changes, this needs tests, and a note in the changelog. Note that those tests will probably need to call `gc.collect()` to get the desired result on PyPy.

src/anyio/_backends/_asyncio.py
```diff
@@ -1865,6 +1866,9 @@ def __exit__(
     ) -> None:
         self._runner.__exit__(exc_type, exc_val, exc_tb)
 
+        if (send_stream := self._send_stream) is not None:
+            send_stream.close()
```
agronholm, line 1870: It would be better to set up `.close()` as a "done callback" of the runner task, in `_call_in_runner_task()`.

agronholm (changes requested):
Much better, but please also drop the storing of tracebacks. Storing tracebacks is problematic from the memory management standpoint, and although I can see the objective, I think the `tracemalloc` system can be used to retrieve the same information. Nevertheless, I don't want the tracebacks stored.

heckad:
Unfortunately it doesn't work.
```python
import tracemalloc
import warnings


class Foo:
    pass


def main():
    tracemalloc.start(65535)

    foo = Foo()

    warnings.warn("Foo", ResourceWarning, source=foo)

    bar = object()

    warnings.warn("Bar", ResourceWarning, source=bar)


if __name__ == "__main__":
    main()
```
Outputs
```
D:\projects\python\test\test16.py:14: ResourceWarning: Foo
  warnings.warn("Foo", ResourceWarning, source=foo)
D:\projects\python\test\test16.py:18: ResourceWarning: Bar
  warnings.warn("Bar", ResourceWarning, source=bar)
Object allocated at (most recent call last):
  File "D:\projects\python\test\test16.py", lineno 22
    main()
  File "D:\projects\python\test\test16.py", lineno 16
    bar = object()
```
Test on `Python 3.11.4 (tags/v3.11.4:d2340ef, Jun  7 2023, 05:45:37) [MSC v.1934 64 bit (AMD64)] on win32`
The same problem with `MemoryObject*Stream`. 
Additionally, the maximum number of stack traces is 65535. This means that if our streams live long, we won't know where we created them. I can place stacktraces inside 
```python
if sys.flags.dev_mode:
    ....
```
What do you think?

agronholm:
I think I would feel a lot more at ease if the tracebacks were at least formatted, so as to remove any circular references. And it would still have to be hidden behind that dev mode flag.

Is any other project storing tracebacks like this? I'd like to check what others are doing and if there are alternatives.

heckad:
> I think I would feel a lot more at ease if the tracebacks were at least formatted, so as to remove any circular references.

Sorry, could you clarify what cyclic dependencies you are talking about? The function `traceback.extract_stack` returns StackSummary type which contains info about the stack, not the stack like in `Exception.__traceback__`
image

> Is any other project storing tracebacks like this? I'd like to check what others are doing and if there are alternatives.

For example [`aiohttp`](https://github.com/aio-libs/aiohttp/blob/58a9190ed092e5748e80e16082592b886d105975/aiohttp/client.py#L263C1-L268C42)

smurfix:
While the `StackSummary` contains a `locals` attribute, it defaults to being empty unless you specifically tell the `TracebackException` that `capture_locals=True`.

That being said, building and storing a stack traceback in every stream is prohibitively expensive even if there are no reference cycles. Don't do that.

If/when it's really necessary to save the entire call chain, well, people should be perfectly capable of monkeypatching `create_m_o_stream`.

agronholm:
> While the `StackSummary` contains a `locals` attribute, it defaults to being empty unless you specifically tell the `TracebackException` that `capture_locals=True`.
> 
> That being said, building and storing a stack traceback in every stream is prohibitively expensive even if there are no reference cycles. Don't do that.
> 
> If/when it's really necessary to save the entire call chain, well, people should be perfectly capable of monkeypatching `create_m_o_stream`.

OP's idea was to only do that when the dev mode is enabled, though.

agronholm:
If indeed this is hidden behind dev mode, then I am warming up to the idea.

agronholm:
Regarding tracemalloc, why are you saying it's not working? This is what I see:
```
bug.py:14: ResourceWarning: Foo
  warnings.warn("Foo", ResourceWarning, source=foo)
Object allocated at (most recent call last):
  File "bug.py", lineno 22
    main()
  File "bug.py", lineno 12
    foo = Foo()
bug.py:18: ResourceWarning: Bar
  warnings.warn("Bar", ResourceWarning, source=bar)
Object allocated at (most recent call last):
  File "bug.py", lineno 22
    main()
  File "bug.py", lineno 16
    bar = object()
```
Seems to work fine?

agronholm:
@heckad I'd like to get this into the new release. We just need to resolve the issue with stack summaries. You haven't replied to my latest comment. If the code is going to be merged with stack summaries, then that needs to be opt-in using the `dev_mode` flag.

src/anyio/streams/memory.py
```diff
@@ -59,10 +62,17 @@ def statistics(self) -> MemoryObjectStreamStatistics:
...
+    _source_traceback: traceback.StackSummary | None = field(init=False, default=None)
 
     def __post_init__(self) -> None:
         self._state.open_receive_channels += 1
 
+        if self.is_source_traceback_capturing_enabled():
+            self._source_traceback = traceback.extract_stack(sys._getframe(1))
+
+    def is_source_traceback_capturing_enabled(self) -> bool:
+        return sys.flags.dev_mode
```
agronholm, line 74: I'd rather not add a new public method for this.
```suggestion
        if sys.flags.dev_mode:
            self._source_traceback = traceback.extract_stack(sys._getframe(1))
```

src/anyio/streams/memory.py
```diff
@@ -164,15 +174,36 @@ def __exit__(
...
+    _source_traceback: traceback.StackSummary | None = field(init=False, default=None)
 
     def __post_init__(self) -> None:
         self._state.open_send_channels += 1
 
+        if self.is_source_traceback_capturing_enabled():
+            self._source_traceback = traceback.extract_stack(sys._getframe(1))
+
+    def is_source_traceback_capturing_enabled(self) -> bool:
+        return sys.flags.dev_mode
```
agronholm, line 205: As above.
```suggestion
        if sys.flags.dev_mode:
            self._source_traceback = traceback.extract_stack(sys._getframe(1))
```

agronholm (changes requested):
Almost perfect now, just two small changes I want.

heckad:
> Why'd you still keep that extra method?

For the ability to override detecting dev mode logic in child classes.

agronholm:
> > Why'd you still keep that extra method?
> 
> For the ability to override detecting dev mode logic in child classes.

Why not just look at `sys.flags.dev_mode` then?

agronholm:
At any rate, just using `sys.flags.dev_mode` is as far as I'm willing to take this. Please remove the method and I'll merge.

agronholm:
Why'd you close this?

PR #672, opened 2024-01-17
Fix `run_process()` and `open_process().__aexit__` leaking an orphan process when cancelled

### LINKED ISSUE #669: `run_process` with `fail_after` causes RuntimeError on timeout
### Things to check first

### AnyIO version

4.2.0

### Python version

3.12.1

### What happened?

I am using `fail_after` to timeout if some task group is taking too long, usually due to being stuck in some subprocess call longer than expected. I was able to reproduce this in a small script (posted in the other section) which results in the following exception after hitting the timeout handler in this case.

```
Exception ignored in: <function BaseSubprocessTransport.__del__ at 0x1011d8360>
  File "/Users/jmayeres/.pyenv/versions/3.12.1/lib/python3.12/asyncio/base_subprocess.py", line 126, in __del__
    self.close()
  File "/Users/jmayeres/.pyenv/versions/3.12.1/lib/python3.12/asyncio/base_subprocess.py", line 104, in close
    proto.pipe.close()
  File "/Users/jmayeres/.pyenv/versions/3.12.1/lib/python3.12/asyncio/unix_events.py", line 568, in close
    self._close(None)
  File "/Users/jmayeres/.pyenv/versions/3.12.1/lib/python3.12/asyncio/unix_events.py", line 592, in _close
    self._loop.call_soon(self._call_connection_lost, exc)
  File "/Users/jmayeres/.pyenv/versions/3.12.1/lib/python3.12/asyncio/base_events.py", line 792, in call_soon
    self._check_closed()
  File "/Users/jmayeres/.pyenv/versions/3.12.1/lib/python3.12/asyncio/base_events.py", line 539, in _check_closed
    raise RuntimeError('Event loop is closed')
RuntimeError: Event loop is closed
```

The expectation would be that no RuntimeError is raised. I noticed that I don't have this issue if I use `open_process` and handle the output via text streams, but if I use `run_process` like in this example I get errors when the timeout is raised. I believe there's also some other weird stuff happening with these exceptions but I wanted to keep the example simple (in our working code there's some nested task groups and it seems like we don't get the timeout exception handler reached but the exception gets absorbed somewhere).

Invoking the following in python results in an exception after printing "Timed out!". Note that I'm running this on macOS 14.2.1

```python
from anyio import create_task_group, fail_after, run, run_process


async def subprocess_output_no_stream():
    result = await run_process("while true; do ps aux; sleep 1; done")
    stdout = result.stdout
    print(f"Subprocess done: got {len(stdout)} bytes")


async def main():
    try:
        with fail_after(1):
                print("Starting")
                tg.start_soon(subprocess_output_no_stream)
            print("Task group done")
    except TimeoutError:
        print("Timed out!")


if __name__ == "__main__":
    run(main)
```

gschaffner:
fixes #669.

## underlying AnyIO bug

currently, cancellation of `async with await open_process()` (i.e. `Process.__aexit__()` in a cancelled scope) can cause the `AsyncResource` to not get fully closed, leaking an orphan process. this is because a cancelled `Process.__aexit__()` is not performing the minimum level of cleanup: Trio's cancellation model states that

> Async cleanup operations–like `__aexit__` methods or async close methods–are cancellable just like anything else except that if they are cancelled, they still perform a minimum level of cleanup before raising `Cancelled`. (https://trio.readthedocs.io/en/stable/reference-core.html#cancellation-and-primitive-operations)

## how it caused #669

#669 is a special case of this bug: in #669, even though `run_process` does call `process.kill` if cancelled, it still (very briefly) leaks an orphan process. this is because when `process.kill()` returns, the process is not necessarily dead yet: `process.kill` (`kill(2)`/`TerminateProcess`) is like `call_soon` and only schedules the kill. so after `process.kill()` we need to wait briefly for the death to occur and for the event loop to learn about it.

so the problem seen in #669 was: an orphan process was leaked briefly. this started a race between
*   the event loop closing and
*   the OS killing the process and the event loop learning about it

note that this race is present on both backends. if the event loop closing happened first and won the race, then:
*   on Trio it caused `Popen.__del__` to emit a `ResourceWarning`
*   on asyncio it _should_ have caused `BaseSubprocessTransport.__del__` to emit a `ResouceWarning`, but due to an asyncio bug (https://github.com/python/cpython/issues/114177) it caused an unraisable error instead of a `ResourceWarning`.

the tests currently added in this PR should cover the underlying AnyIO bug that caused #669 while being deterministic (not flaky). however, if you also want a test that looks more like the original example in #669 (testing asyncio/Trio `close` & `__del__` behavior), here's another test that could be added to this PR:

```python
def test_process_aexit_cancellation_doesnt_orphan_del(
    anyio_backend_name: str, anyio_backend_options: dict[str, Any]
) -> None:
    """
    Regression test for #669.

    Ensure that, when cancelled, open_process.__aexit__() doesn't leave behind any
    __del__() finalizers that emit ResourceWarning (or fail due to
    https://github.com/python/cpython/issues/114177) if there is a race where the event
    loop is closed too soon.

    N.B.: This is a test of asyncio and Trio, not AnyIO: it should pass if
    test_process_aexit_cancellation_doesnt_orphan_process() passes and asyncio/Trio are
    not bugged. This test can also have false negatives (passes when it should fail)
    because it relies on run() finishing either before the OS kills the process or
    before the event loop learns the process died.

    """
    # Don't del process until after run() finishes
    process: Process

    async def main() -> None:
        nonlocal process
        with CancelScope() as scope:
                [sys.executable, "-c", "import time; time.sleep(1)"]
            ) as process:
                scope.cancel()

    run(main, backend=anyio_backend_name, backend_options=anyio_backend_options)
```

### DIFF
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -940,14 +940,21 @@ class Process(abc.Process):
     _stderr: StreamReaderWrapper | None
 
     async def aclose(self) -> None:
-        if self._stdin:
-            await self._stdin.aclose()
-        if self._stdout:
-            await self._stdout.aclose()
-        if self._stderr:
-            await self._stderr.aclose()
+        with CancelScope(shield=True):
+            if self._stdin:
+                await self._stdin.aclose()
+            if self._stdout:
+                await self._stdout.aclose()
+            if self._stderr:
+                await self._stderr.aclose()
 
-        await self.wait()
+        try:
+            await self.wait()
+        except BaseException:
+            self.kill()
+            with CancelScope(shield=True):
+                await self.wait()
+            raise
 
     async def wait(self) -> int:
         return await self._process.wait()
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -918,6 +918,7 @@ class StreamReaderWrapper(abc.ByteReceiveStream):
 
     async def aclose(self) -> None:
         self._stream.feed_eof()
+        await AsyncIOBackend.checkpoint()
 
 
 @dataclass(eq=False)
@@ -930,6 +931,7 @@ class StreamWriterWrapper(abc.ByteSendStream):
 
     async def aclose(self) -> None:
         self._stream.close()
+        await AsyncIOBackend.checkpoint()
 
 
 @dataclass(eq=False)
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -956,6 +956,7 @@ class Process(abc.Process):
             self.kill()
             with CancelScope(shield=True):
                 await self.wait()
+
             raise
 
     async def wait(self) -> int:
--- a/src/anyio/_core/_subprocesses.py
+++ b/src/anyio/_core/_subprocesses.py
@@ -68,8 +68,10 @@ async def run_process(
         async with create_task_group() as tg:
             if process.stdout:
                 tg.start_soon(drain_stream, process.stdout, 0)
+
             if process.stderr:
                 tg.start_soon(drain_stream, process.stderr, 1)
+
             if process.stdin and input:
                 await process.stdin.send(input)
                 await process.stdin.aclose()
--- a/src/anyio/_core/_subprocesses.py
+++ b/src/anyio/_core/_subprocesses.py
@@ -65,20 +65,16 @@ async def run_process(
         start_new_session=start_new_session,
     ) as process:
         stream_contents: list[bytes | None] = [None, None]
-        try:
-            async with create_task_group() as tg:
-                if process.stdout:
-                    tg.start_soon(drain_stream, process.stdout, 0)
-                if process.stderr:
-                    tg.start_soon(drain_stream, process.stderr, 1)
-                if process.stdin and input:
-                    await process.stdin.send(input)
-                    await process.stdin.aclose()
+        async with create_task_group() as tg:
+            if process.stdout:
+                tg.start_soon(drain_stream, process.stdout, 0)
+            if process.stderr:
+                tg.start_soon(drain_stream, process.stderr, 1)
+            if process.stdin and input:
+                await process.stdin.send(input)
+                await process.stdin.aclose()
 
-                await process.wait()
-        except BaseException:
-            process.kill()
-            raise
+            await process.wait()
 
     output, errors = stream_contents
     if check and process.returncode != 0:
--- a/tests/test_subprocesses.py
+++ b/tests/test_subprocesses.py
@@ -9,7 +9,7 @@ from textwrap import dedent
 
 import pytest
 
-from anyio import CancelScope, open_process, run_process
+from anyio import CancelScope, ClosedResourceError, open_process, run_process
 from anyio.streams.buffered import BufferedByteReceiveStream
 
 pytestmark = pytest.mark.anyio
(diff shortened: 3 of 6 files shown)


src/anyio/_core/_subprocesses.py
```diff
@@ -65,20 +65,16 @@ async def drain_stream(stream: AsyncIterable[bytes], index: int) -> None:
...
-            process.kill()
-            raise
+        async with create_task_group() as tg:
+            if process.stdout:
+                tg.start_soon(drain_stream, process.stdout, 0)
+            if process.stderr:
+                tg.start_soon(drain_stream, process.stderr, 1)
+            if process.stdin and input:
+                await process.stdin.send(input)
+                await process.stdin.aclose()
```
agronholm, line 75: Since we're modifying this code, let's add blank lines between control blocks.
```suggestion
            if process.stdout:
                tg.start_soon(drain_stream, process.stdout, 0)

            if process.stderr:
                tg.start_soon(drain_stream, process.stderr, 1)

            if process.stdin and input:
                await process.stdin.send(input)
                await process.stdin.aclose()
```

src/anyio/_backends/_asyncio.py
```diff
@@ -940,14 +942,21 @@ class Process(abc.Process):
...
+            if self._stderr:
+                await self._stderr.aclose()
+
+        try:
+            await self.wait()
+        except BaseException:
+            self.kill()
+            with CancelScope(shield=True):
+                await self.wait()
+            raise
```
agronholm, line 959:
```suggestion

            raise
```

tests/test_subprocesses.py
```diff
@@ -176,3 +176,49 @@ async def test_run_process_inherit_stdout(capfd: pytest.CaptureFixture[str]) ->
...
+    """
+    Regression test for #669.
+
+    Ensures that open_process.__aexit__() closes standard streams when cancelled. Also
+    ensures that process.std{in.send,{out,err}.receive}() raise ClosedResourceError on a
+    closed stream.
+
+    """
+    pytest.xfail("#671 needs to be resolved first")
+
```
agronholm, line 209:
```suggestion
@pytest.mark.xfail("#671 needs to be resolved first")
async def test_process_aexit_cancellation_closes_standard_streams() -> None:
    """
    Regression test for #669.

    Ensures that open_process.__aexit__() closes standard streams when cancelled. Also
    ensures that process.std{in.send,{out,err}.receive}() raise ClosedResourceError on a
    closed stream.

    """
```

tests/test_subprocesses.py
```diff
@@ -176,3 +176,49 @@ async def test_run_process_inherit_stdout(capfd: pytest.CaptureFixture[str]) ->
...
+    assert process.stdin is not None
+    with pytest.raises(ClosedResourceError):
+        await process.stdin.send(b"foo")
+    assert process.stdout is not None
+    with pytest.raises(ClosedResourceError):
+        await process.stdout.receive(1)
+    assert process.stderr is not None
+    with pytest.raises(ClosedResourceError):
+        await process.stderr.receive(1)
```
agronholm, line 224:
```suggestion
    assert process.stdin is not None

    with pytest.raises(ClosedResourceError):
        await process.stdin.send(b"foo")

    assert process.stdout is not None

    with pytest.raises(ClosedResourceError):
        await process.stdout.receive(1)

    assert process.stderr is not None

    with pytest.raises(ClosedResourceError):
        await process.stderr.receive(1)
```

agronholm (changes requested):
Just some formatting changes.

tests/test_subprocesses.py
```diff
@@ -176,3 +176,53 @@ async def test_run_process_inherit_stdout(capfd: pytest.CaptureFixture[str]) ->
...
+        async with await open_process(
+            [sys.executable, "-c", "import time; time.sleep(1)"]
+        ) as process:
+            scope.cancel()
+
+    assert process.returncode is not None
+    assert process.returncode != 0
+
+
+@pytest.mark.xfail("#671 needs to be resolved first")
```
agronholm, line 199:
```suggestion
@pytest.mark.xfail(reason="#671 needs to be resolved first")
```

agronholm:
How is it that the xfail test passes? Explain please?

agronholm (approved):

agronholm:
Thanks!

Issue #678, opened 2024-01-25
Stricter typing for `TaskGroup.start()`


jonathanslenders:
### Things to check first

- [X] I have searched the existing issues and didn't find my feature already requested there

### Feature description

Looking at how we type `TaskGroup.start_soon`, we can do the same approach to have stricter type checking on `TaskGroup.start`.

We can do the following:

```python
from typing_extensions import Protocol
from typing_extensions import TypeVarTuple, Unpack, TypeVar, Generic


T = TypeVar("T")


class TaskStatus(Generic[T]):
    def ready(self, result: T) -> None:
        ...


# ---

PosArgsT = TypeVarTuple("PosArgsT")
TaskStatusResult = TypeVar("TaskStatusResult")


class _StartFunc(Protocol[Unpack[PosArgsT], TaskStatusResult]):
    async def __call__(
        self, *args: Unpack[PosArgsT], task_status: TaskStatus[TaskStatusResult]
    ) -> None:
        ...


class TaskGroup:
    async def start(
        self,
        func: _StartFunc[Unpack[PosArgsT], TaskStatusResult],
        *args: Unpack[PosArgsT],
    ) -> TaskStatusResult:
        ...
```

Example usage::

```python
async def task(a: int, b: str, task_status: TaskStatus[bool]) -> None:
    task_status.ready(True)

async def main() -> None:
    tg = TaskGroup()
    r: float = await tg.start(task, 6, "t")
```

On the last line, typing the result as `float` is important. Otherwise it will be `Any`. But if we would type it as `str`, then mypy would say:

```python
error: Argument 1 to "start" of "TaskGroup" has incompatible type "Callable[[int, str, TaskStatus[bool]], Coroutine[Any, Any, None]]"; expected "_StartFunc[int, str, str]"  [arg-type]
```

That is significantly better than `object` what we currently have here:
https://github.com/agronholm/anyio/blob/master/src/anyio/abc/_tasks.py#L70

### Use case

(better type checking.)

jonathanslenders:
I just see previous attempts have been made to supply types for `start()`.

See: https://github.com/agronholm/anyio/pull/511
And most recent mypy issue: https://github.com/python/mypy/issues/16522

When trying to finalize a PR, I was hitting exactly those issues, so I guess we have to wait for mypy...

rmorshea:
Another consideration, in my own code I've been doing the following:

```python
from collections.abc import AsyncIterator, Callable
from typing import Unpack

[...]

PR #679, opened 2024-01-26
Add stricter types for TaskGroup.start()

jonathanslenders:
See: https://github.com/agronholm/anyio/issues/678

There are still some typing issues here. I'm trying to figure out some edge cases.


jonathanslenders:
Let's close this for now. We need this mypy issue to be resolved: https://github.com/python/mypy/issues/16522

PR #688, opened 2024-02-08
Remove conditionals and workarounds for Python interpreter versions 3.7 and older

musicinmybrain:
Since `anyio` already requires Python 3.8 or later, these conditionals and workarounds are no longer needed.


agronholm (approved):
Very nice, thanks!

PR #694, opened 2024-03-02
Fix silent in asyncio if python dev-mode enabled

heckad:
Example
```python
import time
import anyio

async def main():
    time.sleep(10)

anyio.run(main)
```
If you run this script with dev mode enabled using `PYTHONDEVMODE=1` env var then you don't got any warning because anyio call `asyncio.run` with `debug=False`


agronholm:
Please add a test and a changelog entry. You can use the `monkeypatch` fixture to set the env variable in the test, like `monkeypatch.setenv("PYTHONDEVMODE", "1")`.

PR #699, opened 2024-03-13
Fixed handling of KeyError raised by a typed attribute getter

agronholm:
`KeyError`s coming out of the getter call shouldn't be caught by `extra()`.


graingert (approved):

agronholm:
Thanks :+1:

PR #703, opened 2024-03-18
Fixed EOF detection on asyncio when there is also data in the buffer

### LINKED ISSUE #701: TCP listener handler does not disconnect on EOF (netcat -N)
### Things to check first

### AnyIO version

4.3.0

### Python version

3.11.8

### What happened?

While trying to implement a Python TCP receiver for a line based protocol using anyio, I have issues closing the stream with an EOF on the client side with netcat. This works with a pure Python synchronous implementation. The code uses the anyio backend. 

```python
import anyio
from anyio.streams.text import TextReceiveStream

async def read_lines(port: int):
    """Read lines from TCP socket and output."""
    
    # FIXME: GNU netcat does not deterministically shutdown after sending EOF (-N option), one has to use -w 0 to close on idle
    async def handle(client):
        async with client:
            receive_stream = TextReceiveStream(client)
            async for chunk in receive_stream:
                for line in chunk.splitlines():
                    print(f"Line: {line.rstrip()}")
    
    listener = await anyio.create_tcp_listener(local_port=port)
    await listener.serve(handle)


anyio.run(read_lines, 1234)
```

Now, the following command hangs on most, but not all executions (it might have to do with buffering):
`echo 'hello world' | netcat -N localhost 1234`

The following works, but uses an idle workaround
`echo 'hello world' | netcat -N -w 0 localhost 1234`

agronholm:
I also made `receive()` more efficient by not having two consecutive checkpoints in the case where it needs to wait for data, and avoided unnecessary task switching if the task hasn't been cancelled.

Closes #701.

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -7,6 +7,8 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
 - Fixed erroneous ``TypedAttributeLookupError`` if a typed attribute getter raises
   ``KeyError``
+- Fixed ``SocketStream.receive()`` not detecting EOF on asyncio if there is also data in
+  the read buffer (`#701 <https://github.com/agronholm/anyio/issues/701>`_)
 
 **4.3.0**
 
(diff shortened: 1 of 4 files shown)


fungs (approved):
I couldn't go deep, but the changes in the functions are small and the test looks good!

graingert (approved):

PR #707, opened 2024-03-27
Fix shield future to prevent erroneous RuntimeError

### LINKED ISSUE #706: Raising an error inside a task of a task group can produce: `RuntimeError: called 'started' twice on the same task status`
### Things to check first

### AnyIO version

4.3.0

### Python version

3.12.2

### What happened?

Calling task_status.started() can raise `RuntimeError: called 'started' twice on the same task status` under certain circumstances (see examples), even though it was not called twice.
The reason for this is that the underlying future gets awaited in the start() method. When the start() method gets canceled, the underlying future gets also canceled which makes calling set_result on the future later fail and raise the erroneous RuntimeError.
This can be fixed by using asyncio.shield when awaiting the future.
https://github.com/agronholm/anyio/blob/05a3d02ed383f996ad93e8a7a4f6dcce153a06b6/src/anyio/_backends/_asyncio.py#L782

```python
from anyio import create_task_group, run


async def taskfunc(*, task_status: TaskStatus):
    task_status.started()
    raise ValueError


async def main():
        await tg.start(taskfunc)
        await tg.start(taskfunc)


run(main)
```
or
```python
from anyio import create_task_group, run


async def taskfunc(*, task_status: TaskStatus):
    task_status.started()


async def main():
        coro = tg.start(taskfunc)
        tg.cancel_scope.cancel()
        await coro


run(main)
```

dominik-schwabe:
Fixes #706.

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -6,7 +6,8 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 **UNRELEASED**
 
 - Fixed erroneous ``RuntimeError: called 'started' twice on the same task status``
-  when cancelling a task in a TaskGroup created with the ``start`` method
+  when cancelling a task in a TaskGroup created with the ``start()`` method before
+  the first checkpoint is reached after calling ``task_status.started()``
   (`#706 <https://github.com/agronholm/anyio/issues/706>`_; PR by Dominik Schwabe)
 - Fixed erroneous ``TypedAttributeLookupError`` if a typed attribute getter raises
   ``KeyError``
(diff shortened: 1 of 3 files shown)


agronholm:
I'm reviewing this now. Would you mind adding a changelog entry?

dominik-schwabe:
I updated the changelog, but I wasn't happy with it and changed it with a force push and I think this removed you commit https://github.com/agronholm/anyio/commit/6704b9ed0bfd2f651d04a8c7871073d268cfcfa2 .

agronholm:
Yeah, just don't force push your changes :) I think my revisions made the entry more accurate, yes?

dominik-schwabe:
Sorry for this, I didn't expect you to be so fast.

agronholm (approved):

agronholm:
Thanks!

PR #708, opened 2024-03-29
add `wrapped` property to `Path`

valq7711:


agronholm:
Why is `pathlib.Path(anyio_path)` not enough?

valq7711:
> Why is `pathlib.Path(anyio_path)` not enough?

 it requires to import `pathlib.Path`

agronholm:
So?

valq7711:
It is redundant code (importing) and an extra `pathlib.Path` instance created every time we need a sync path

agronholm:
I'd like to know more about your use case. This hasn't come up before.

valq7711:
```python
class MockData:
    mock_data_json = anyio.Path('/mock_data/data.json')
    mock_data_json.parent.wrapped.mkdir(exist_ok=True)
```

agronholm:
Doesn't seem like you need this for performance, just to save you a little bit of typing? I'll think this over.

valq7711:
There is `wrapped` property for async file, why not to have the same for async path?

valq7711:
There are sync-libraries (e.g. csv) that work with sync path and usually you need to do some async path checking (like a `.is_file()`) and then pass sync-path to these libraries `to_thread.run_sync()`, so I think it is a fairly common case when it needs to get sync-path from async-path.

agronholm:
Do these APIs work solely with `Path`, or do they accept `os.PathLike`, like virtually all stdlib functions dealing with paths do?

agronholm:
In the case of `AsyncFile`, it's stateful so you can't just convert it into a sync file descriptor without repercussions. That's why it has the `wrapped` property. Another reason for not having a `wrapped` property on the async `Path` is that, as alluded above, most APIs that take paths take any path-like objects, so an explicit conversion tends not be necessary anyway.

agronholm:
Closing due to stalled discussion and a negative consensus on Gitter regarding this.

PR #711, opened 2024-04-05
Added the BlockingPortalProvider class

agronholm:
This feature helps make efficient sync counterparts to async APIs by providing a single portal on an on-demand basis.


graingert:
You should test non overlapping uses of the portal, and potentially uses from more than one thread

agronholm:
> You should test non overlapping uses of the portal, and potentially uses from more than one thread

Alright, will do.

agronholm:
> You should test non overlapping uses of the portal, and potentially uses from more than one thread

Done.

graingert (approved):

PR #722, opened 2024-04-16
fix: `is_junction` only for python>=3.12

YXL76:


agronholm:
What's this for? If you're not on 3.12, you will get an error in any event.

Issue #728, opened 2024-05-09
MemoryObjectStream can drop items when the receiving end is cancelled


agronholm:
### Things to check first

- [X] I have searched the existing issues and didn't find my bug already reported there

- [X] I have checked that my bug is still present in the latest release

### AnyIO version

4.3.0

### Python version

3.8

### What happened?

If a task (A) sends an item to another task (B) via a memory object stream, and task B is in a state of waiting for an item, and has a pending cancellation, the item is still sent to B but as cancellation is then delivered to B, that item is essentially dropped on the floor.

A similar issue was reported in #146, but it seems that it wasn't fixed as thoroughly as I had hoped.

### How can we reproduce the bug?

```py
import anyio
from anyio import (
    CancelScope, create_memory_object_stream, create_task_group,
    wait_all_tasks_blocked
)

async def receiver(receive, task_status):
    with CancelScope() as cancel_scope:
        task_status.started(cancel_scope)
        await receive.receive()

async def main():
    send, receive = create_memory_object_stream(1)
    with send, receive:
        async with create_task_group() as tg:
            cancel_scope = await tg.start(receiver, receive)
            await wait_all_tasks_blocked()
            cancel_scope.cancel()
            send.send_nowait("item")

        assert receive.receive_nowait() == "item"

anyio.run(main)
```
The above snippet reproduces the problem (`WouldBlock` is raised) on both asyncio and Trio. On Trio, if `create_memory_object_stream()` is replaced with `trio.open_memory_channel()`, the snippet no longer raises an exception.

agronholm:
Ah, my intention was to create a memory object stream with a buffer. Then the `assert` makes sense, yes?

agronholm:
I've adjusted the snippet accordingly.

agronholm:
Compare to the use of `open_memory_channel()`:
```py
import anyio
from anyio import (
    CancelScope, create_task_group,
    wait_all_tasks_blocked,
)
from trio import open_memory_channel

async def receiver(receive, task_status):
    with CancelScope() as cancel_scope:
        task_status.started(cancel_scope)
        await receive.receive()

async def main():
    send, receive = open_memory_channel(1)
    with send, receive:
        async with create_task_group() as tg:
            cancel_scope = await tg.start(receiver, receive)
            await wait_all_tasks_blocked()
            cancel_scope.cancel()
            send.send_nowait("item")

        assert receive.receive_nowait() == "item"

anyio.run(main, backend="trio")
```

PR #729, opened 2024-05-11
Fix memory streams incorrectly raising cancelled when `*_nowait()` is called immediately after cancelling `send()`/`receive()`

gschaffner:
### Changes

Fixes:

*   #728 (in sense (1) as defined in https://github.com/agronholm/anyio/issues/728#issuecomment-2105664591). cancellation of `receive()` could drop items, meaning that _neither_ the attempted sender nor the attempted receiver think they are responsible for the item.

*   the analogous issue about cancellation of `send()`. cancellation of `send()` could result in `send()` raising cancelled despite succeeding, meaning that _both_ the attempted sender and the attempted receiver think they are responsible for the item.

### Checklist

the you've fulfilled the following conditions (where applicable):

features)

these instructions.

### DIFF
--- a/src/anyio/streams/memory.py
+++ b/src/anyio/streams/memory.py
@@ -11,6 +11,7 @@ from .. import (
     ClosedResourceError,
     EndOfStream,
     WouldBlock,
+    get_cancelled_exc_class,
 )
 from ..abc import Event, ObjectReceiveStream, ObjectSendStream
 from ..lowlevel import checkpoint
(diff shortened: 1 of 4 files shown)


agronholm:
This was the first solution that came to my mind, but then I realized it would interact poorly with asyncio's native cancellation.

src/anyio/streams/memory.py
```diff
@@ -230,6 +236,12 @@ async def send(self, item: T_contra) -> None:
             self._state.waiting_senders[send_event] = item
             try:
                 await send_event.wait()
+            except get_cancelled_exc_class():
+                # Ignore the immediate cancellation if we already sent the item, so as
+                # to not indicate failure despite success
+                if send_event in self._state.waiting_senders:
+                    del self._state.waiting_senders[send_event]
+                    raise
```
agronholm, line 244: This would be fine when the cancellation originates from a cancel scope, but not when a task was natively cancelled (`Task.cancel()`, because the originator of that cancellation expects the cancellation to be acted on and not ignored). Doing this will lead to hangs.

agronholm:
Closing in favor of #735, as I've come to the conclusion that avoiding the item delivery to a task with a pending cancellation is the only viable solution.

PR #733, opened 2024-05-12
Add `__slots__` to `AsyncResource`

injust:
### Changes
`AsyncResource` should have `__slots__` so child classes can use `__slots__`.

Similar to https://github.com/python/cpython/issues/105726 and https://github.com/python/cpython/pull/106771 for `contextlib.AbstractAsyncContextManager`.

### Checklist

features)

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -29,7 +29,7 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
   ``MemoryObjectSendStream`` that were garbage collected without being closed (PR by
   Andrey Kazantcev)
 - Added ``__slots__`` to ``AsyncResource`` so that child classes can use ``__slots__``
-  (PR by Justin Su)
+  (`#733 <https://github.com/agronholm/anyio/pull/733>`_; PR by Justin Su)
 
 **4.3.0**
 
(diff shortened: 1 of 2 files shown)


agronholm:
Do you have a use case for this? None of the subclasses in AnyIO itself use `__slots__`.

injust:
I'm creating my own child classes of `AsyncResource` with `attrs`, and was surprised that they weren't slotted classes. I'm assuming that's a supported use case, similar to how one might use `contextlib.AbstractAsyncContextManager`.

agronholm:
Attrs doesn't require the superclasses to have `__slots__` in order to create a slotted class:
```pyconsole
>>> import attrs
>>> class Foo:
...   pass
... 
>>> @attrs.define(slots=True)
... class Bar:
...   x: int
... 
>>> b = Bar(1)
>>> b.y = 5
  File "<stdin>", line 1, in <module>
AttributeError: 'Bar' object has no attribute 'y'
```

injust:
You forgot to subclass `Bar` from `Foo` in your example.

From https://threeofwands.com/attrs-ii-slots/:
> Slot classes lose a bunch of their advantages if the inheritance chain contains even a single dict class, because there **will** be an instance dictionary associated with every instance. Slot classes all the way!

agronholm:
Right, I forgot to do that. It seems that `attrs` has changed the way it operates, because it _used_ to work this way (although I will say it was the wrong thing to do). The only question that remains then, is whether we want to include the `__weakref__` slot here.

injust:
> It seems that `attrs` has changed the way it operates, because it _used_ to work this way

I haven't been using `attrs` for too long so I can't comment on that. Only thing that comes to mind is that slotted classes are default for `attrs.define` and `attrs.frozen`,

> whether we want to include the `__weakref__` slot here

I'm -0 on this. If someone wanted a weak referenceable-`AsyncResource`, they could create their own subclass (and have it be slotted or not).

agronholm (approved):

agronholm:
Alright, you convinced me.

Issue #123, opened 2020-07-09
max_size bs. max_bytes


smurfix:
```
$ git grep def.receive_until
src/anyio/_networking.py:    async def receive_until(self, delimiter: bytes, max_size: int) -> bytes:
src/anyio/abc.py:    async def receive_until(self, delimiter: bytes, max_bytes: int) -> bytes:
```
pylint complains about that. Which to use?

agronholm:
I noticed this myself just today. I will fix it shortly. Is this a blocker issue for you?

smurfix:
No, not particularly. I'll just patch it in my local copy / Debian package.

PR #735, opened 2024-05-14
Fixed memory object stream sometimes dropping sent items

### LINKED ISSUE #728: MemoryObjectStream can drop items when the receiving end is cancelled

agronholm:
## Changes

Fixes #728.

## Checklist

features)

these instructions.

### Updating the changelog

<https://github.com/agronholm/anyio/issues/123>_; PR by Yourname)`

If possible, use your real name in the changelog entry. If not, use your GitHub
username.

### DIFF
--- a/src/anyio/_core/_testing.py
+++ b/src/anyio/_core/_testing.py
@@ -67,7 +67,7 @@ def get_running_tasks() -> list[TaskInfo]:
     """
     Return a list of running tasks in the current event loop.
 
-    :return: a sequence of task info objects
+    :return: a list of task info objects
 
     """
     return cast("list[TaskInfo]", get_async_backend().get_running_tasks())
--- a/src/anyio/_core/_testing.py
+++ b/src/anyio/_core/_testing.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from collections.abc import Awaitable, Generator
-from typing import Any
+from typing import Any, cast
 
 from ._eventloop import get_async_backend
 
@@ -45,8 +45,8 @@ class TaskInfo:
     def __repr__(self) -> str:
         return f"{self.__class__.__name__}(id={self.id!r}, name={self.name!r})"
 
-    def _unwrap(self) -> TaskInfo:
-        return self
+    def has_pending_cancellation(self) -> bool:
+        return False
 
 
 def get_current_task() -> TaskInfo:
@@ -63,10 +63,10 @@ def get_running_tasks() -> list[TaskInfo]:
     """
     Return a list of running tasks in the current event loop.
 
-    :return: a list of task info objects
+    :return: a sequence of task info objects
 
     """
-    return get_async_backend().get_running_tasks()
+    return cast("list[TaskInfo]", get_async_backend().get_running_tasks())
 
 
 async def wait_all_tasks_blocked() -> None:
--- a/src/anyio/_core/_testing.py
+++ b/src/anyio/_core/_testing.py
@@ -46,6 +46,10 @@ class TaskInfo:
         return f"{self.__class__.__name__}(id={self.id!r}, name={self.name!r})"
 
     def has_pending_cancellation(self) -> bool:
+        """
+        Return ``True`` if the task has a cancellation pending, ``False`` otherwise.
+
+        """
         return False
 
 
--- a/src/anyio/streams/memory.py
+++ b/src/anyio/streams/memory.py
@@ -45,7 +45,7 @@ class MemoryObjectStreamState(Generic[T_Item]):
     buffer: deque[T_Item] = field(init=False, default_factory=deque)
     open_send_channels: int = field(init=False, default=0)
     open_receive_channels: int = field(init=False, default=0)
-    waiting_receivers: OrderedDict[Event, MemoryObjectItemReceiver] = field(
+    waiting_receivers: OrderedDict[Event, MemoryObjectItemReceiver[T_Item]] = field(
         init=False, default_factory=OrderedDict
     )
     waiting_senders: OrderedDict[Event, T_Item] = field(
--- a/src/anyio/streams/memory.py
+++ b/src/anyio/streams/memory.py
@@ -12,6 +12,7 @@ from .. import (
     EndOfStream,
     WouldBlock,
 )
+from .._core._testing import TaskInfo, get_current_task
 from ..abc import Event, ObjectReceiveStream, ObjectSendStream
 from ..lowlevel import checkpoint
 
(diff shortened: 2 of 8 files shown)


src/anyio/streams/memory.py
```diff
@@ -199,11 +206,14 @@ def send_nowait(self, item: T_contra) -> None:
         if not self._state.open_receive_channels:
             raise BrokenResourceError
 
-        if self._state.waiting_receivers:
-            receive_event, container = self._state.waiting_receivers.popitem(last=False)
-            container.append(item)
-            receive_event.set()
-        elif len(self._state.buffer) < self._state.max_buffer_size:
+        while self._state.waiting_receivers:
+            receive_event, receiver = self._state.waiting_receivers.popitem(last=False)
```
Zac-HD, line 210: I suspect that this can deadlock if the receiving task is cancelled, then send_nowait invoked (removing it from the waiting_receivers), and then _uncancelled_ before the event loop actually kicks the receiving task out of `await receive`.

Of course this is a pretty nasty use of the already brittle uncancel semantics, but we probably want a comment here to make the possibility salient to future readers?  Or maybe I'm wrong about this.

tests/streams/test_memory.py
```diff
@@ -305,28 +305,49 @@ async def test_cancel_during_receive() -> None:
...
+            await receive.receive()
+
+    send, receive = create_memory_object_stream[str](1)
+    with send, receive:
+        async with create_task_group() as tg:
+            cancel_scope = await tg.start(scoped_receiver, receive)
+            await wait_all_tasks_blocked()
+            cancel_scope.cancel()
+            send.send_nowait("item")
+
```
Zac-HD, line 348: from comment above: uncancelling that scope is my maybe-deadlock.

Zac-HD (approved):
This looks good to me, modulo the single concern about a possible uncancel edge case.  

I'm not sure whether that even _can_ happen, but it seems worth checking.  If it can, I'd probably just add a code comment + note in the docs and merge anyway.

src/anyio/streams/memory.py
agronholm: I was not able to create a situation where an uncancel task would get its foot in the door to prevent the delivery of the cancellation. Here's my best attempt so far:
```py
import asyncio
from asyncio import CancelledError, current_task, get_running_loop

from anyio import create_memory_object_stream, create_task_group


async def receiver(receive, task_status):
    with receive:
        task_status.started(current_task())
        try:
            item = await receive.receive()
        except CancelledError:
            print("cancelled")
            raise

        print("got", item)


async def uncanceller(task):
    task.uncancel()
    print("uncancelled")


async def main():
    send, receive = create_memory_object_stream()
        task = await tg.start(receiver, receive)
        get_running_loop().call_soon(task.uncancel)
        task.cancel()
        await send.send(6)


asyncio.run(main())
```

agronholm: Looks like I was trying hard enough. I was able to repro the situation using just asyncio's task APIs:
```python
import asyncio
from asyncio import CancelledError, create_task

from anyio import create_memory_object_stream


async def receiver(receive):
    with receive:
        try:
            item = await receive.receive()
        except CancelledError:
            print("cancelled")
            raise

        print("got", item)


async def uncanceller(task):
    task.uncancel()
    print("uncancelled, cancelling =", task.cancelling())


async def main():
    send, receive = create_memory_object_stream()
    task = create_task(receiver(receive))
    uncancel_task = create_task(uncanceller(task))
    task.cancel()
    await send.send(6)


asyncio.run(main())
```

agronholm: Scratch that too, because the task hadn't even started and that's why the sender was blocking.

agronholm: Ok, here's my third attempt which seems to confirm that the cancellation is scheduled ahead of everything else:
```py
import asyncio
from asyncio import CancelledError, create_task, sleep

from anyio import create_memory_object_stream


async def receiver(receive):
    with receive:
        print("waiting for item")
        try:
            item = await receive.receive()
        except CancelledError:
            print("cancelled")
        else:
            print("got", item)


async def uncanceller(task):
    task.uncancel()
    print("uncancelled, cancelling =", task.cancelling())


async def main():
    send, receive = create_memory_object_stream(1)
    receive_task = create_task(receiver(receive))
    await sleep(0)
    uncancel_task = create_task(uncanceller(receive_task))
    receive_task.cancel()
    await send.send(6)
    await uncancel_task
    await receive_task


asyncio.run(main())
```
Output:
```
waiting for item
cancelled
uncancelled, cancelling = 0
...
```
Interestingly, if I cancel the receiver task and uncancel it right after without yielding to another task, I get different results based on the Python version. On 3.13.0b1, the task is scheduled normally and receives the item:
```py
import asyncio
from asyncio import CancelledError, create_task, sleep

from anyio import create_memory_object_stream


async def receiver(receive):
    with receive:
        print("waiting for item")
        try:
            item = await receive.receive()
        except CancelledError:
            print("cancelled")
        else:
            print("got", item)


async def main():
    send, receive = create_memory_object_stream()
    receive_task = create_task(receiver(receive))
    await sleep(0)
    receive_task.cancel()
    receive_task.uncancel()
    await send.send(6)
    await receive_task


asyncio.run(main())
```

agronholm: I suppose this is the effect of Guido's recent cancellation PR.

src/anyio/_core/_testing.py
```diff
@@ -63,10 +67,10 @@ def get_running_tasks() -> list[TaskInfo]:
     """
     Return a list of running tasks in the current event loop.
 
-    :return: a list of task info objects
+    :return: a sequence of task info objects
```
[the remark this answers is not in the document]
agronholm: That was my first thought too, but:

1. it is factually returning a list
2. a Sequence (or even a MutableSequence) would narrow the return type, thus breaking backwards compatibility, albeit only from the type checker standpoint.

I see no ideal solution here, so I chose the least disruptive way forward. Thoughts?

agronholm:
I've now added an asyncio-only test that ensures that a native cancellation won't cause an item to be dropped. I've verified that the test fails on current `master`.

agronholm:
Huh, now it's failing on py < 3.11. I need to investigate.

src/anyio/_core/_testing.py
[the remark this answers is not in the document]
Zac-HD: Let's just say "list" instead of "sequence" here in the docstring.

As a broader comment, I tend to treat type annotations as documentation rather than a stable API, and thus don't mind updating them in minor versions if that otherwise makes sense.  But continuing to use `list` here seems best anyway.

Zac-HD (approved):
Native-cancellation handling looks good.  I'm looking forward to using this!

src/anyio/_core/_testing.py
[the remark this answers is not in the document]
agronholm: I've reverted the wording.

PR #736, opened 2024-05-26
Added opt-in locking of first detected async backend

agronholm:
## Changes

Prompted by https://github.com/agronholm/anyio/discussions/543, this adds an optimization, enabled by the `ANYIO_BACKEND` environment variable, for skipping the costly sniffio automatic async backend detection after the first call. Whatever backend is detected is assumed to be running on every call since. This avoids the `getpid()` system calls in this extremely often called function.

## Checklist

features)

these instructions.

### Updating the changelog

<https://github.com/agronholm/anyio/issues/123>_; PR by Yourname)`

If possible, use your real name in the changelog entry. If not, use your GitHub
username.

### DIFF
--- a/src/anyio/_core/_eventloop.py
+++ b/src/anyio/_core/_eventloop.py
@@ -28,7 +28,7 @@ PosArgsT = TypeVarTuple("PosArgsT")
 threadlocals = threading.local()
 loaded_backends: dict[str, type[AsyncBackend]] = {}
 forced_backend_name = os.getenv("ANYIO_BACKEND")
-forced_backend: type[AsyncBackend]
+forced_backend: type[AsyncBackend] | None = None
 
 
 def run(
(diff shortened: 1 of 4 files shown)


src/anyio/_core/_eventloop.py
```diff
@@ -152,8 +170,16 @@ def claim_worker_thread(
 
 
 def get_async_backend(asynclib_name: str | None = None) -> type[AsyncBackend]:
+    global forced_backend
+
     if asynclib_name is None:
-        asynclib_name = sniffio.current_async_library()
+        if os.getenv("PYTEST_CURRENT_TEST"):
```
rafalkrupinski, line 176: not `ANYIO_BACKEND`?

agronholm: Ah...this must be a leftover from my alternate solutions. I'll see if I can improve this further.

rafalkrupinski: Actually I don't think it should be `ANYIO_BACKEND` either. You'll figure it out :)

agronholm: I think I got it right this time around.

provinzkraut:
@agronholm are there plans to get this one merged or are you no longer considering this a viable option?

agronholm:
> @agronholm are there plans to get this one merged or do you no longer consider this a viable option?

Benchmarks showed that the performance difference was negligible, so I'm not likely going to work on this further.

provinzkraut:
> Benchmarks showed that the performance difference was negligible, so I'm not likely going to work on this further.

Ah, didn't manage to extract that information from the existing discourse about this topic. Would you suggest addressing performance differences between anyio/trio/asyncio separately on a case-by-case basis then?

agronholm:
> > Benchmarks showed that the performance difference was negligible, so I'm not likely going to work on this further.
> 
> Ah, didn't manage to extract that information from the existing discourse about this topic. Would you suggest addressing performance differences between anyio/trio/asyncio separately on a case-by-case basis then?

The case where I was focused on was locking operations in `httpcore`, and there it turned out that simply optimizing `anyio.Lock` was the real bottleneck, not `sniffio`. So I would suggest the same approach: benchmark to find out where the bottlenecks are in your use case, and fix those.

Issue #737, opened 2024-05-27
Test failures with Python 3.13.0b1 in 4.4.0


musicinmybrain:
### Things to check first

- [X] I have searched the existing issues and didn't find my bug already reported there

- [X] I have checked that my bug is still present in the latest release

### AnyIO version

4.4.0

### Python version

3.13.0b1

### What happened?

```
______________________________________________________________________________________ TestPath.test_properties[asyncio] _______________________________________________________________________________________
tests/test_fileio.py:103: in test_properties
    assert not missing
E   AssertionError: assert not {'from_uri', 'full_match', 'parser'}
___________________________________________________________________________________ TestPath.test_properties[asyncio+uvloop] ___________________________________________________________________________________
tests/test_fileio.py:103: in test_properties
    assert not missing
E   AssertionError: assert not {'from_uri', 'full_match', 'parser'}
________________________________________________________________________________________ TestPath.test_properties[trio] ________________________________________________________________________________________
tests/test_fileio.py:103: in test_properties
    assert not missing
E   AssertionError: assert not {'from_uri', 'full_match', 'parser'}
__________________________________________________________________________________________ TestPath.test_is_reserved ___________________________________________________________________________________________
tests/test_fileio.py:274: in test_is_reserved
    assert Path("nul").is_reserved() == expected_result
src/anyio/_core/_fileio.py:445: in is_reserved
    return self._path.is_reserved()  
/usr/lib64/python3.13/pathlib/_local.py:423: in is_reserved
    warnings.warn(msg, DeprecationWarning, stacklevel=2)
E   DeprecationWarning: pathlib.PurePath.is_reserved() is deprecated and scheduled for removal in Python 3.15. Use os.path.isreserved() to detect reserved paths on Windows.
```

I expected these four tests to pass, as they do on Python 3.12.

### How can we reproduce the bug?

```
$ gh repo clone agroholm/anyio
$ cd anyio
$ python3.13 --version
Python 3.13.0b1
$ python3.13 -m venv _e
$ . _e/bin/activate
```

[...]

PR #739, opened 2024-05-27
Python 3.13: expect DeprecationWarning for pathlib.PurePath.is_reserved()

musicinmybrain:
## Changes

This is a *partial* fix for https://github.com/agronholm/anyio/issues/737.

On Python 3.13 and later, `test_is_reserved` now expects that [`pathlib.PurePath.is_reserved()` produces a `DeprecationWarning`](https://docs.python.org/3.13/library/pathlib.html#pathlib.PurePath.is_reserved).

## Checklist

- [ ] You've added tests (in `tests/`) added which would fail without your patch **N/A**
features) **N/A**

these instructions.


agronholm:
Sorry, but I already had local changes ready that tackle not only this issue, but also the issue of missing 3.13 Path features.

PR #750, opened 2024-06-23
Switched start_blocking_portal() to use daemon threads

agronholm:
## Changes

ThreadPoolExecutor used daemon threads in Python 3.8 and non-daemon threads in 3.9 onwards. But daemon threads are essential for the use case where we want to have a "loitering" blocking portal which will only be shut down via an atexit hook.

Related discussion: https://github.com/agronholm/anyio/discussions/743

## Checklist

features)

these instructions.

### Updating the changelog

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -10,6 +10,8 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
   (`#737 <https://github.com/agronholm/anyio/issues/737>`_)
 - Changed the ``ResourceWarning`` from an unclosed memory object stream to include its
   address for easier identification
+- Changed ``start_blocking_portal()`` to always use daemonic threads, to accommodate the
+  "loitering event loop" use case
 - Fixed ``to_process.run_sync()`` failing to initialize if ``__main__.__file__`` pointed
   to a file in a nonexistent directory
   (`#696 <https://github.com/agronholm/anyio/issues/696>`_)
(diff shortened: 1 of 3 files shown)


dhirschfeld:
I can confirm this appears to fix the hang I was observing :+1:

agronholm:
Thanks for confirming!

dhirschfeld:
Possibly a random error which might be fixed by retrying the CI
```
FAILED tests/streams/test_tls.py::TestTLSStream::test_extra_attributes[trio] - pytest.PytestUnhandledThreadExceptionWarning: Exception in thread Thread-4
<snip>
ConnectionResetError: [WinError 10054] An existing connection was forcibly closed by the remote host
```

agronholm:
> Possibly a random error which might be fixed by retrying the CI?
> 
> ```
> FAILED tests/streams/test_tls.py::TestTLSStream::test_extra_attributes[trio] - pytest.PytestUnhandledThreadExceptionWarning: Exception in thread Thread-4
> <snip>
> ConnectionResetError: [WinError 10054] An existing connection was forcibly closed by the remote host
> ```

Yup, randomly occurring Windows socket errors are among the hardest to figure out.

dhirschfeld (approved):
FWIW it seems to work for me and AFAICT the changes look fine.

agronholm:
> One possibility to avoid breaking subinterpreters here could be to make this opt-in: `def start_blocking_portal(daemon: bool = False)`.

I'm not too concerned about subinterpreters. I would've loved to add support for them, but they still don't have a stable Python API, and furthermore, once `nogil` becomes usable, they might become completely obsolete. I think we'll cross that bridge when we come to it, if ever.

PR #762, opened 2024-07-19
Removed check for coroutine in asyncio backend

davidbrochart:
## Changes

See https://github.com/agronholm/anyio/discussions/759#discussioncomment-10096103.
In the asyncio backend, there is no need to check if the object is a coroutine before launching it as a task:
- asyncio already does it,
- it prevents launching a class instance that looks like a coroutine, for instance:
```py
import typing

import anyio

class Foo(typing.Coroutine):
    def __await__(self):
        return self.run().__await__()

    def send(self, value=None):
        return self.__await__().send(value)

    def throw(self, *_unused):
        raise RuntimeError("Shouldn't happen")

    async def run(self):
        print("run")

async def main():
        tg.start_soon(Foo)

anyio.run(main, backend="trio")  # works fine
anyio.run(main, backend="asyncio")  # doesn't work
```
This already works with the Trio backend, but currently not with the asyncio backend.

## Checklist

features)

these instructions.

### Updating the changelog

### DIFF
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -747,7 +747,7 @@ class TaskGroup(abc.TaskGroup):
         else:
             parent_id = id(self.cancel_scope._host_task)
 
-        coro = func(*args, **kwargs)
+        coro = cast(Coroutine, func(*args, **kwargs))
         name = get_callable_name(func) if name is None else str(name)
         task = create_task(coro, name=name)
         task.add_done_callback(task_done)
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -29,7 +29,6 @@ from inspect import (
     CORO_RUNNING,
     CORO_SUSPENDED,
     getcoroutinestate,
-    iscoroutine,
 )
 from io import IOBase
 from os import PathLike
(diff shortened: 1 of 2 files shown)


davidbrochart:
@agronholm What do you think about this?

agronholm:
Frankly I considered the inability to run generator-based coroutine functions a *feature* rather than a limitation. The main problem with this, however, is that I don't know what ends up a the `coro` of a task, and we can't accept just anything there. Maybe it works, maybe it won't, but you haven't added any tests to that end so it's hard to say. The problem with the `coro` of a task is outlined in a review comment of my own PR (#675).

Furthermore, if we go through with this and flat out disable such checks, we may get inconsistent or misleading error messages across backends when a user tries to pass an incompatible object as the callable to `anyio.run()`.

smurfix:
Well. We could fix this by replacing anyio's `iscoroutine` test with `isinstance(…, typing.Coroutine)`.

davidbrochart:
Thanks @agronholm, I didn't see #675.
In particular, the following argument makes sense: "the problem is that anything that AnyIO allows past this check needs to support `inspect.getcoroutinestate` (because of `_task_started`)". But then, since the [code](https://github.com/agronholm/anyio/pull/762#issue-2419003800) at the top comment works with the Trio backend, I'm wondering if it should?

agronholm:
> Thanks @agronholm, I didn't see #675. In particular, the following argument makes sense: "the problem is that anything that AnyIO allows past this check needs to support `inspect.getcoroutinestate` (because of `_task_started`)". But then, since the [code](https://github.com/agronholm/anyio/pull/762#issue-2419003800) at the top comment works with the Trio backend, I'm wondering if it should?

The question is, where do we draw the line, and how? I don't know just what `isinstance(coro, Coroutine)` will pass through. Does it guarantee that the object works with `inspect.getcoroutinestate()`? I have a faint recollection that I've experimented with alternate checks in the past, but without much success.

davidbrochart:
> Does it guarantee that the object works with `inspect.getcoroutinestate()`?

No it doesn't. But then how can Trio accept such coroutine-like objects? I must have some other way of handling their state.
Anyway, this seems to be a limitation of the asyncio backend and indeed such objects shouldn't be accepted.

graingert:
> No it doesn't. But then how can Trio accept such coroutine-like objects? It must have some other way of handling their state.

trio works by checking `frame = getattr(coro, "cr_frame", None)` and adding a wrapper coroutine if `frame` is missing

PR #769, opened 2024-08-23
Avoid expose extra vars (key, value)

waketzheng:
## Changes
```py
import anyio
print(anyio.key)
# typed_attribute
print(anyio.value)
# <function typed_attribute at 0x1033a5e10>
```

It is unnecessary to expose anyio.key and anyio.value

## Checklist

these instructions.

### DIFF
--- a/src/anyio/__init__.py
+++ b/src/anyio/__init__.py
@@ -1,7 +1,5 @@
 from __future__ import annotations
 
-from typing import Any
-
 from ._core._eventloop import current_time as current_time
 from ._core._eventloop import get_all_backends as get_all_backends
 from ._core._eventloop import get_cancelled_exc_class as get_cancelled_exc_class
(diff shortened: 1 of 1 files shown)


src/anyio/__init__.py
```diff
@@ -69,8 +67,6 @@
...
 
 # Re-export imports so they look like they live directly in this package
-key: str
-value: Any
-for key, value in list(locals().items()):
-    if getattr(value, "__module__", "").startswith("anyio."):
-        value.__module__ = __name__
+for __value in list(locals().values()):
+    if getattr(__value, "__module__", "").startswith("anyio."):
+        __value.__module__ = __name__
```
agronholm, line 72: If you wanted to be thorough, why not add `del __value` at the end?

agronholm (approved):
Looks good, thanks!

PR #777, opened 2024-08-30
Convert bytearray to bytes in `StreamProtocol.receive()`

### LINKED ISSUE #776: `SocketStream.receive` sometimes returns `bytearray` objects on Windows
### Things to check first

### AnyIO version

4.4.0

### Python version

3.11

### What happened?

`StreamProtocol`, here: https://github.com/agronholm/anyio/blob/master/src/anyio/_backends/_asyncio.py#L1075
incorrectly assumes that it will only get `bytes` objects as an argument. Unfortunately, on Windows, that's not the case. For the `ProactorEventLoop`, we'll get `bytearray` objects here.

See:
- https://github.com/python/typeshed/issues/12606
- https://github.com/python/cpython/issues/123496

I think the quickest fix for anyio is to accept `bytearray` as well, and convert it into `bytes`, regardless of what typeshed/cpython do.

(Sorry, I don't have access to a Windows machine right now to create an example, but please let me know if an example script is needed.)

agronholm:
## Changes

Checks if `Protocol.data_received()` was passed a `bytearray` instead of `bytes`, and converts it accordingly.

Fixes #776. 

## Checklist

features)

these instructions.

### Updating the changelog

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -20,6 +20,9 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
   (`#696 <https://github.com/agronholm/anyio/issues/696>`_)
 - Fixed ``TaskInfo.has_pending_cancellation()`` on asyncio not respecting shielded
   scopes (`#771 <https://github.com/agronholm/anyio/issues/771>`_; PR by @gschaffner)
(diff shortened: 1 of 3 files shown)


graingert (approved):

agronholm:
Thanks!

PR #778, opened 2024-08-31
Fixed debugger quit not working when inside a task group

agronholm:
## Changes

If you set a breakpoint inside a task group, quitting the debugger won't exit the pytest test session, but instead fails the current test. These changes find the exit exception in an exception group, and raise just that, ignoring the rest of the exceptions (if any).

## Checklist

features)

these instructions.

### Updating the changelog

### DIFF
--- a/src/anyio/pytest_plugin.py
+++ b/src/anyio/pytest_plugin.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import sys
 from collections.abc import Iterator
 from contextlib import ExitStack, contextmanager
 from inspect import isasyncgenfunction, iscoroutinefunction
@@ -8,11 +9,13 @@ from typing import Any, Dict, Tuple, cast
 import pytest
 import sniffio
 from _pytest.outcomes import Exit
-from exceptiongroup import ExceptionGroup
 
 from ._core._eventloop import get_all_backends, get_async_backend
 from ._core._exceptions import iterate_exceptions
 from .abc import TestRunner
+
+if sys.version_info < (3, 11):
+    from exceptiongroup import ExceptionGroup
 
 _current_runner: TestRunner | None = None
 _runner_stack: ExitStack | None = None
--- a/src/anyio/pytest_plugin.py
+++ b/src/anyio/pytest_plugin.py
@@ -7,8 +7,11 @@ from typing import Any, Dict, Tuple, cast
 
 import pytest
 import sniffio
+from _pytest.outcomes import Exit
+from exceptiongroup import ExceptionGroup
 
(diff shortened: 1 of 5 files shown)


src/anyio/pytest_plugin.py
```diff
@@ -1,16 +1,22 @@
...
 
+import sys
 from collections.abc import Iterator
 from contextlib import ExitStack, contextmanager
 from inspect import isasyncgenfunction, iscoroutinefunction
 from typing import Any, Dict, Tuple, cast
 
 import pytest
 import sniffio
+from _pytest.outcomes import Exit
```
graingert, line 11: It's publicly available at pytest.exit.Exception

agronholm: `pytest.exit` is a function, not a module.

agronholm: Oh, I misunderstood. The exception is somehow attached to the function?

agronholm: I know it would work like that, but PyCharm currently sees this as an error:
image
I would prefer to not make this change so long as this is the case. I know it's just one of the myriad of PyCharm's bugs, but still...

graingert: The type annotation is off the charts https://github.com/pytest-dev/pytest/blob/9a444d113658be6ccb8dc9f57ed118a1ef17c94c/src/_pytest/outcomes.py#L105

src/anyio/pytest_plugin.py
```diff
@@ -121,7 +127,12 @@ def run_with_hypothesis(**kwargs: Any) -> None:
             funcargs = pyfuncitem.funcargs
             testargs = {arg: funcargs[arg] for arg in pyfuncitem._fixtureinfo.argnames}
             with get_runner(backend_name, backend_options) as runner:
-                runner.run_test(pyfuncitem.obj, testargs)
+                try:
+                    runner.run_test(pyfuncitem.obj, testargs)
+                except ExceptionGroup as excgrp:
+                    for exc in iterate_exceptions(excgrp):
+                        if isinstance(exc, (Exit, KeyboardInterrupt, SystemExit)):
```
graingert, line 134: Technically I feel this is a pytest bug and should be fixed there

agronholm: Eventually, yes, but for the time being I feel this is a big quality-of-life improvement to the AnyIO testing experience.

graingert: Yeah definitely

graingert (approved):

agronholm:
Thanks!

PR #782, opened 2024-09-04
Accept abstract namespace paths for unix domain sockets

### LINKED ISSUE #781: create_unix_listener doesn't accept abstract namespace sockets
### Things to check first

### AnyIO version

4.4.0

### Python version

3.10

### What happened?

On anyio version 3.7 at least it's possible to pass a path starting with a null byte to `anyio.create_unix_listener` to create an abstract unix domain socket that isn't backed by a real file and is cleaned up automatically on program close.

In version 4.4.0 (at least probably earlier) this gives a `ValueError: embedded null byte` when the code tries to call `os.stat` on the path.

```
import anyio
anyio.run(anyio.create_unix_listener, '\0abstract-path')
```

I havn't tested on trio backend or more versions but may do so later.
Assuming this is something that is desired to fix I may try creating a PR for it as well.

tapetersen:
## Changes

Accept paths starting with a null byte in create_unix_listener and connect_unix_socket to allow creating abstract unix sockets. Fixes #781

## Checklist

features)

### Updating the changelog

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -37,6 +37,8 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 - Fixed quitting the debugger in a pytest test session while in an active task group
   failing the test instead of exiting the test session (because the exit exception
   arrives in an exception group)
+- Re-add support for linux abstract namespace sockets (#781
+  <https://github.com/agronholm/anyio/issues/781>_; PR by @tapetersen)
 
 **4.4.0**
 
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -37,8 +37,8 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 - Fixed quitting the debugger in a pytest test session while in an active task group
   failing the test instead of exiting the test session (because the exit exception
   arrives in an exception group)
-- Re-add support for linux abstract namespace sockets (#781
-  <https://github.com/agronholm/anyio/issues/781>_; PR by @tapetersen)
+- Fixed support for Linux abstract namespaces in UNIX sockets that was broken in v4.2
+  (#781 <https://github.com/agronholm/anyio/issues/781>_; PR by @tapetersen)
 
 **4.4.0**
 
--- a/src/anyio/_core/_sockets.py
+++ b/src/anyio/_core/_sockets.py
@@ -683,16 +683,30 @@ async def setup_unix_local_socket(
     path_str: str | bytes | None
     if path is not None:
         path_str = os.fspath(path)
+        is_abstract = (
+            path_str.startswith(b"\0")
+            if isinstance(path_str, bytes)
+            else path_str.startswith("\0")
+        )
 
-        # Copied from pathlib...
-        try:
-            stat_result = os.stat(path)
-        except OSError as e:
-            if e.errno not in (errno.ENOENT, errno.ENOTDIR, errno.EBADF, errno.ELOOP):
-                raise
+        if is_abstract:
+            # Unix abstract namespace socket. No file backing so skip stat call
+            pass
         else:
-            if stat.S_ISSOCK(stat_result.st_mode):
-                os.unlink(path)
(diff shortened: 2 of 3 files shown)


tests/test_sockets.py
```diff
@@ -735,12 +735,17 @@ async def test_bind_link_local(self) -> None:
...
+    def socket_path(
+        self, request: SubRequest, tmp_path_factory: TempPathFactory
+    ) -> Generator[Path, None, None]:
         # Use stdlib tempdir generation
         # Fixes `OSError: AF_UNIX path too long` from pytest generated temp_path
         with tempfile.TemporaryDirectory() as path:
-            yield Path(path) / "socket"
+            if request.param == "path":
+                yield Path(path) / "socket"
+            elif request.param == "abstract":
```
agronholm, line 747: Redudant `elif` – what else could `request.param be?

tests/test_sockets.py
```diff
@@ -735,12 +735,17 @@ async def test_bind_link_local(self) -> None:
     sys.platform == "win32", reason="UNIX sockets are not available on Windows"
 )
 class TestUNIXStream:
-    @pytest.fixture
-    def socket_path(self) -> Generator[Path, None, None]:
+    @pytest.fixture(params=["path", "abstract"])
```
agronholm, line 738: The latter parameter needs a `skipif` marker when not on Linux, as this is a Linux specific feature.

agronholm (changes requested):

tests/test_sockets.py
tapetersen: Just a remnant from me normally being paranoid with this pattern and a guarding else with assert but I'll just change to an else here.

tests/test_sockets.py
tapetersen: Fixed

tapetersen:
@agronholm  I'm not sure about what if any updates to the documentation should be done apart from the changelog entry.

docs/versionhistory.rst
```diff
@@ -37,6 +37,8 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 - Fixed quitting the debugger in a pytest test session while in an active task group
   failing the test instead of exiting the test session (because the exit exception
   arrives in an exception group)
+- Re-add support for linux abstract namespace sockets (#781
```
agronholm, line 40: What did you mean by "Re-add" (also notice the tense which differs from all the rest of the changelog entries)? Did AnyIO used to have this support before?

src/anyio/_core/_sockets.py
```diff
@@ -683,16 +683,30 @@ async def setup_unix_local_socket(
...
+            else path_str.startswith("\0")
+        )
 
-        # Copied from pathlib...
-        try:
-            stat_result = os.stat(path)
-        except OSError as e:
-            if e.errno not in (errno.ENOENT, errno.ENOTDIR, errno.EBADF, errno.ELOOP):
-                raise
+        if is_abstract:
```
agronholm, line 692: I really dislike no-op `if` branches. Please refactor the stat checking branch to be the only one. Move the comment about abstract namespaces to whatever place you deem most appropriate.

agronholm (changes requested):
Finishing touches.

docs/versionhistory.rst
tapetersen: Indeed. It worked/works in 3.7.1 which we use currently to delay the ExceptionGroup change. I havn't bisected or tried to find where it broke more closely yet though.

agronholm: If this used to work before, then I guess then the correct phrasing would be "fixed" because I never intentionally removed support for it.

agronholm, line 40:
```suggestion
- Fixed support for Linux abstract namespaces in UNIX sockets (#781
```

src/anyio/_core/_sockets.py
tapetersen: Updated and removed extra var since the check is now more concise

PR #790, opened 2024-09-22
Fix `asyncio.Task.cancelling` issues

gschaffner:
## Changes

This is a small patch that fixes the more obscure case that we skipped yesterday (https://github.com/agronholm/anyio/pull/774#discussion_r1766604109).

After a night's rest I woke up this morning with the realization that this case shouldn't involve whack-a-mole as we feared. IIUC it seems to be relatively straightforward to fix.

d342ae144083993f671bfe702af3ec6f25449ea6 is the core patch here in its simplest form. The subsequent two commits are some minor follow-up refactoring for performance optimization:

*   Avoid doing repeated computations of `_effectively_cancelled` & `_parent_cancellation_is_visible_to_us` in `CancelScope.__exit__` (I missed this during review).

*   Move the counter back from `TaskState` to `CancelScope`. This makes the impl. a little harder to understand IMO (as this makes the impl. diverge a bit further from Trio's model of delaying deciding which scope a level cancellation exception belongs to), but it may(?) save a little memory (as the counter is only needed for host tasks).

## Checklist

features)

### DIFF
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -377,7 +377,7 @@ def is_anyio_cancellation(exc: CancelledError) -> bool:
     # matching cancel message
     while True:
         if (
-            bool(exc.args)
+            exc.args
             and isinstance(exc.args[0], str)
             and exc.args[0].startswith("Cancelled by cancel scope ")
         ):
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -372,11 +372,22 @@ def _task_started(task: asyncio.Task) -> bool:
 
 
 def is_anyio_cancellation(exc: CancelledError) -> bool:
-    return (
-        bool(exc.args)
-        and isinstance(exc.args[0], str)
-        and exc.args[0].startswith("Cancelled by cancel scope ")
-    )
+    # Sometimes third party frameworks catch a CancelledError and raise a new one, so as
+    # a workaround we have to look at the previous ones in __context__ too for a
+    # matching cancel message
+    while True:
+        if (
+            bool(exc.args)
+            and isinstance(exc.args[0], str)
+            and exc.args[0].startswith("Cancelled by cancel scope ")
+        ):
+            return True
+
+        if isinstance(exc.__context__, CancelledError):
+            exc = exc.__context__
+            continue
+
+        return False
 
 
 class CancelScope(BaseCancelScope):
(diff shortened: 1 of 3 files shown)


src/anyio/_backends/_asyncio.py
```diff
@@ -385,8 +396,11 @@ def __init__(self, deadline: float = math.inf, shield: bool = False):
         self._cancel_handle: asyncio.Handle | None = None
         self._tasks: set[asyncio.Task] = set()
         self._host_task: asyncio.Task | None = None
-        self._cancel_calls: int = 0
-        self._cancelling: int | None = None
+        self._pending_uncancellations: int | None
+        if sys.version_info >= (3, 11):
+            self._pending_uncancellations = 0
+        else:
+            self._pending_uncancellations = None
```
agronholm, line 403: Are you sure we're okay with not looking at the task cancellation counter when entering the scope? I think there was a reason why `asyncio.timeout()` does that.

agronholm: Potentially related? https://github.com/python/cpython/pull/102815/files#diff-0a6da8818df8b29c3c649167aa37c906eb27283c5fe1cacb40e36a6ce0458e7eR87

agronholm: It's *possible* that we don't need this, as we rely on the cancel messages rather than the cancel counter, as I believe you meant by that previous comment of yours.

agronholm:
I'm reviewing this now for inclusion in the next release. Would you mind resolving the merge conflicts? I'm not 100% confident about doing that myself in this case.

src/anyio/_backends/_asyncio.py
```diff
@@ -372,11 +372,22 @@ def _task_started(task: asyncio.Task) -> bool:
 
 
 def is_anyio_cancellation(exc: CancelledError) -> bool:
-    return (
-        bool(exc.args)
-        and isinstance(exc.args[0], str)
-        and exc.args[0].startswith("Cancelled by cancel scope ")
-    )
+    # Sometimes third party frameworks catch a CancelledError and raise a new one, so as
```
graingert, line 375: Can you include this in the change log?

graingert: Oh I see oh well

graingert: Let's do (or don't) the warning on another PR. I'll investigate asyncpg

graingert: looks like it's fixed in asyncpg: https://github.com/MagicStack/asyncpg/blob/e9bb695413b2ee923310f9106277198cc19c47b3/asyncpg/_asyncio_compat.py#L25-L73

src/anyio/_backends/_asyncio.py
```diff
@@ -397,8 +408,11 @@ def __init__(self, deadline: float = math.inf, shield: bool = False):
         self._cancel_handle: asyncio.Handle | None = None
         self._tasks: set[asyncio.Task] = set()
         self._host_task: asyncio.Task | None = None
-        self._cancel_calls: int = 0
-        self._cancelling: int | None = None
+        self._pending_uncancellations: int | None
+        if sys.version_info >= (3, 11):
+            self._pending_uncancellations = 0
```
agronholm, line 413:
```suggestion
        if sys.version_info >= (3, 11):
            self._pending_uncancellations: int | None = 0
```

agronholm (approved):

PR #799, opened 2024-10-02
Fixed missing or inconsistent error when acquiring already owned Lock

### LINKED ISSUE #798: Lock behavior differs between asyncio and trio after the lock changes in 4.5.0
### Things to check first

### AnyIO version

AnyIO 4.6.0

### Python version

3.12.5

### What happened?

Prior to AnyIO 4.5.0, an exception would be raised when a caller attempted to acquire a lock that was already claimed on both the asyncio and trio backends.
Exception output: `RuntimeError: attempt to re-acquire an already held Lock`

Now, in both 4.5.0 and 4.6.0, asyncio will hang indefinitely. In the context manager example, `RuntimeError: The current task is not holding this lock` is output when the process is interrupted with ctrl-c . Trio still behaves the same as it did prior to 4.5.0, so it appears this issue is only with the asyncio changes.

Simplest case to demonstrate the error:

```python3
async def main() -> None:
    _lock = anyio.Lock()

    await _lock.acquire()


    await _lock.acquire()

if __name__ == "__main__":
    anyio.run(main)#, backend="trio")
```

Similar behavior with context managers:

```python3
async def main() -> None:
    _lock = anyio.Lock()

    async def get_lock() -> None:
        async with _lock:
            print("Error should be raised here")

    async with _lock:
        await get_lock()

if __name__ == "__main__":
    anyio.run(main)#, backend="trio")
```

agronholm:
## Changes

Reintroduces a check to see if the lock is already owned by this task, on asyncio. On Trio, harmonize the `RuntimeError` message to match that on asyncio.

Fixes #798. 

## Checklist

features)

these instructions.

### Updating the changelog

### DIFF
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -1731,9 +1731,10 @@ class Lock(BaseLock):
         self._waiters: deque[tuple[asyncio.Task, asyncio.Future]] = deque()
 
     async def acquire(self) -> None:
+        task = cast(asyncio.Task, current_task())
         if self._owner_task is None and not self._waiters:
             await AsyncIOBackend.checkpoint_if_cancelled()
-            self._owner_task = current_task()
+            self._owner_task = task
 
             # Unless on the "fast path", yield control of the event loop so that other
             # tasks can run too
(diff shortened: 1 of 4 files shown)


src/anyio/_backends/_asyncio.py
```diff
@@ -1762,10 +1765,14 @@ async def acquire(self) -> None:
         self._waiters.remove(item)
 
     def acquire_nowait(self) -> None:
+        task = cast(asyncio.Task, current_task())
         if self._owner_task is None and not self._waiters:
-            self._owner_task = current_task()
+            self._owner_task = task
             return
 
+        if self._owner_task == task:
```
graingert, line 1773: Why not use is?

```suggestion
        if self._owner_task is task:
```

Should slightly faster?

graingert (approved):

agronholm:
Thanks for the review!

PR #800, opened 2024-10-02
Updated Path documentation and made is_junction() conditional

### LINKED ISSUE #794: AttributeError: 'Path' object has no attribute 'walk'
### Things to check first

### AnyIO version

4.6.0

### Python version

3.11

### What happened?

Anyio.Path implements the `walk` method which is available in pathlib.Parth from python 3.12 onwards. For lower versions of python, this causes anyio.Path.walk to through an attribute error. 

I expected anyio to implement an alternative using os.walk on lower python versions, or at least to through an informative error instructing the user. 

Try using anyio.Path.walk in a lower Python version.

agronholm:
## Changes

Updated the `anyio.Path` docs to fix several omissions, and to mention limitations on earlier Python versions.
Also, the `is_junction()` method is now hidden unless on Python 3.12 or later, to match its stdlib counterpart (as it wouldn't work anyway).

Fixes #794. 

## Checklist

features)

these instructions.

### Updating the changelog

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -3,6 +3,12 @@ Version history
 
 This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
+**UNRELEASED**
+
+- Corrected documentation on ``anyio.Path`` regarding the limitations imposed by the
+  current Python version on several of its methods, and made the ``is_junction`` method
+  unavailable on Python versions earlier than 3.12
+
 **4.6.0**
 
 - Dropped support for Python 3.8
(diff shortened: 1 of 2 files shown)


docs/versionhistory.rst
```diff
@@ -13,6 +13,9 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 - Fixed TaskGroup and CancelScope producing cyclic references in tracebacks
   when raising exceptions (`#806 <https://github.com/agronholm/anyio/pull/806>`_)
   (PR by @graingert)
+- Corrected documentation on ``anyio.Path`` regarding the limitations imposed by the
+  current Python version on several of its methods, and made the ``is_junction`` method
+  unavailable on Python versions earlier than 3.12
```
graingert, line 18: Link to the issue?

graingert (changes requested):

docs/versionhistory.rst
agronholm: Added.

graingert (approved):

PR #804, opened 2024-10-11
gh-633: fix fixture methods run with different self than the test method

### LINKED ISSUE #633: Pytest async method fixture is run with different `self` than the test method
### Things to check first

### AnyIO version

4.0.0

### Python version

3.11.4

### What happened?

I'm trying to use an fixture class method to initialize some data for tests and store it in attributes. With synchronous fixture everything works as expected, the filled attributes are available in the test methods. But attributes filled in async function are not available in test methods. This happens because different objects are passed as "self" for the fixture and the test method. 

I found [similar issue for pytest-asyncio](https://github.com/pytest-dev/pytest-asyncio/issues/197) which was fixed

```
import pytest


class TestSomething:
    @pytest.fixture()
    async def do_setup(self):
        self.some_attr = 'value'

        print()
        print('in fixture')
        print(self)
        print(self.__dict__)

    async def test_something(self, do_setup):
        print()
        print('in test')
        print(self)          # different object
        print(self.__dict__) # some_attr missing
```

Output:

```
in fixture
<tests.test_anyio.TestSomething object at 0x107b0ca50>
{'some_attr': 'value'}

in test
<tests.test_anyio.TestSomething object at 0x107f06590>
{}
```

graingert:
## Changes

Fixes #633 . 

## Checklist

features)

these instructions.

### Updating the changelog

### DIFF
--- a/src/anyio/pytest_plugin.py
+++ b/src/anyio/pytest_plugin.py
@@ -125,12 +125,12 @@ def pytest_fixture_setup(fixturedef: Any, request: Any) -> Generator[Any, None, 
                 has_backend_arg = "anyio_backend" in fixturedef.argnames
                 original_func = fixturedef.func
                 fixturedef.func = wrapper
-                stack.callback(setattr, fixturedef, 'func', original_func)
+                stack.callback(setattr, fixturedef, "func", original_func)
 
                 if not has_backend_arg:
                     original_argnames = fixturedef.argnames
                     fixturedef.argnames = original_argnames + ("anyio_backend",)
-                    stack.callback(setattr, fixturedef, 'argnames', original_argnames)
+                    stack.callback(setattr, fixturedef, "argnames", original_argnames)
 
         return (yield)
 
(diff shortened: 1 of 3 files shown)


src/anyio/pytest_plugin.py
```diff
@@ -82,15 +116,23 @@ def wrapper(*args, anyio_backend, **kwargs):  # type: ignore[no-untyped-def]
...
+    with ExitStack() as stack:
+        # Only apply this to coroutine functions and async generator functions in requests
+        # that involve the anyio_backend fixture
+        func = _resolve_fixture_function(fixturedef, request)
+        if isasyncgenfunction(func) or iscoroutinefunction(func):
+            if "anyio_backend" in request.fixturenames:
+                has_backend_arg = "anyio_backend" in fixturedef.argnames
+                original_func = fixturedef.func
+                fixturedef.func = wrapper
+                stack.callback(setattr, fixturedef, "func", original_func)
```
graingert, line 128: this should fix a case where anyio adopts a fixture and then subsequent tests use it

agronholm:
Closing in favor of #807.

PR #806, opened 2024-10-12
fix cyclic garbage that keeps traceback frames alive in taskgroup exceptions

graingert:
## Changes
fix cyclic garbage that keeps traceback frames alive in taskgroup exceptions

This reproduces the behaviour in trio (the trio versions of the test pass with minor modifications of the anyio wrappers). Producing cyclic garbage is not usually a problem, however when an exception traceback references that exception again the entire traceback and therefore all frames in the stack remain alive until the garbage collector runs.

## Checklist

features)

these instructions.

### Updating the changelog

### DIFF
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -426,15 +426,16 @@ class CancelScope(BaseCancelScope):
         exc_tb: TracebackType | None,
     ) -> bool | None:
         del exc_tb
-        try:
-            if not self._active:
-                raise RuntimeError("This cancel scope is not active")
-            if current_task() is not self._host_task:
-                raise RuntimeError(
-                    "Attempted to exit cancel scope in a different task than it was "
-                    "entered in"
-                )
 
+        if not self._active:
+            raise RuntimeError("This cancel scope is not active")
+        if current_task() is not self._host_task:
+            raise RuntimeError(
+                "Attempted to exit cancel scope in a different task than it was "
+                "entered in"
+            )
+
+        try:
             assert self._host_task is not None
             host_task_state = _task_states.get(self._host_task)
             if host_task_state is None or host_task_state.cancel_scope is not self:
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -481,7 +481,8 @@ class CancelScope(BaseCancelScope):
             self._restart_cancellation_in_parent()
             return swallow_exception and not not_swallowed_exceptions
         finally:
-            del self._host_task, exc_val
+            self._host_task = None
+            del exc_val
 
     @property
     def _effectively_cancelled(self) -> bool:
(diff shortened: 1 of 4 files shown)


graingert:
depends on https://github.com/python-trio/trio/pull/3107

graingert:
Needs to be skipped on PyPy

src/anyio/_backends/_asyncio.py
```diff
@@ -683,6 +687,27 @@ def started(self, value: T_contra | None = None) -> None:
...
+            waiter.set_result(None)
+
+    for task in tasks:
+        task.add_done_callback(on_completion)
+        del task
+
+    try:
+        await waiter
+    finally:
+        while tasks:
```
graingert, line 707: asyncio.wait leaves cyclic garbage here https://github.com/python/cpython/blob/eb2d268ac7480b5e2b4ffb9a644cad7ac75ae954/Lib/asyncio/tasks.py#L526

graingert:
> depends on [python-trio/trio#3107](https://github.com/python-trio/trio/pull/3107)

Ok maybe this isn't needed, but I couldn't get the tests passing without it

graingert:
I'll add the changelog etc and shape this into a proper PR shortly

tests/test_taskgroups.py
```diff
@@ -1548,6 +1549,111 @@ async def in_task_group(task_status: TaskStatus[None]) -> None:
...
+            raise _Done
+    except ExceptionGroup as excs:
+        exc = excs.exceptions[0]
+
+    assert isinstance(exc, _Done)
+    assert gc.get_referrers(exc) == no_other_refs()
+
+
+async def test_exception_refcycles_parent_task() -> None:
+    """Test that TaskGroup deletes self._parent_task"""
```
graingert, line 1599: These are the docstrings from the cpython versions of these tests

graingert:
This PR is a lot easier to review with whitespace changes hidden: https://github.com/agronholm/anyio/pull/806/files?w=1

src/anyio/_backends/_trio.py
```diff
@@ -183,16 +183,17 @@ async def __aexit__(
         exc_val: BaseException | None,
         exc_tb: TracebackType | None,
     ) -> bool | None:
+        rest = None
         try:
             return await self._nursery_manager.__aexit__(exc_type, exc_val, exc_tb)
         except BaseExceptionGroup as exc:
-            _, rest = exc.split(trio.Cancelled)
+            rest = exc.split(trio.Cancelled)[1]
             if not rest:
```
agronholm, line 191: Since `rest` isn't referenced outside of the `if` statement, we could just do this:
```suggestion
            # If the exception group only contains Cancelled exceptions, raise a single one to
            # cover them all
            if not exc.split(trio.Cancelled)[1]:
```

agronholm (changes requested):
Just a cosmetic change in a suggestion.

agronholm (approved):

agronholm:
Thanks!

PR #807, opened 2024-10-12
Rebind instance method fixtures to the same instance as the test

### LINKED ISSUE #633: Pytest async method fixture is run with different `self` than the test method

agronholm:
## Changes

Fixes instance method fixtures to run with the same `self` as the test.

Fixes #633. 

## Checklist

features)

these instructions.

### Updating the changelog

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -8,7 +8,7 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 - Fixed acquring a lock twice in the same task on asyncio hanging instead of raising a
   ``RuntimeError`` (`#798 <https://github.com/agronholm/anyio/issues/798>`_)
 - Fixed an async fixture's ``self`` being different than the test's ``self`` in
-  unittest-style tests (`#633 <https://github.com/agronholm/anyio/issues/633>`_)
+  class-based tests (`#633 <https://github.com/agronholm/anyio/issues/633>`_)
   (PR by @agronholm and @graingert)
 
 **4.6.0**
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -9,6 +9,7 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
   ``RuntimeError`` (`#798 <https://github.com/agronholm/anyio/issues/798>`_)
 - Fixed an async fixture's ``self`` being different than the test's ``self`` in
   unittest-style tests (`#633 <https://github.com/agronholm/anyio/issues/633>`_)
+  (PR by @agronholm and @graingert)
 
 **4.6.0**
 
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -7,6 +7,9 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
 - Fixed acquring a lock twice in the same task on asyncio hanging instead of raising a
   ``RuntimeError`` (`#798 <https://github.com/agronholm/anyio/issues/798>`_)
+- Fixed ``self`` being different than the test's ``self`` for async instance method
+  fixtures in unittest-style tests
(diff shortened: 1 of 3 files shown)


docs/versionhistory.rst
```diff
@@ -7,6 +7,9 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
 - Fixed acquring a lock twice in the same task on asyncio hanging instead of raising a
   ``RuntimeError`` (`#798 <https://github.com/agronholm/anyio/issues/798>`_)
+- Fixed ``self`` being different than the test's ``self`` for async instance method
+  fixtures in unittest-style tests
```
graingert, line 11: It's not just unittest style tests, it also applies to class based pytest style tests

src/anyio/pytest_plugin.py
```diff
@@ -70,28 +71,56 @@ def pytest_configure(config: Any) -> None:
...
+            if not (has_request_arg := "request" in fixturedef.argnames):
+                fixturedef.argnames += ("request",)
+
+            try:
+                return (yield)
+            finally:
+                fixturedef.func = func
+                fixturedef.argnames = original_argname
+
+    yield
```
graingert, line 122:
```suggestion
```

graingert (changes requested):

src/anyio/pytest_plugin.py
graingert, line 122:
```suggestion
    return (yield)
```

agronholm: Why?

graingert (approved):

agronholm:
Thanks for the review!

src/anyio/pytest_plugin.py
```diff
@@ -70,28 +71,56 @@ def pytest_configure(config: Any) -> None:
...
+            local_func = func.__func__.__get__(request.instance)
+        else:
+            local_func = func
+
         backend_name, backend_options = extract_backend_and_options(anyio_backend)
         if has_backend_arg:
             kwargs["anyio_backend"] = anyio_backend
 
+        if has_request_arg:
+            kwargs["request"] = anyio_backend
```
layday, line 94: This overwrites the `request` param of all async fixtures so that you are no longer able to parameterise async fixtures at all. The following example with throw an `AttributeError: 'str' object has no attribute 'param'`:

```py
@pytest.fixture(params=['foo', 'bar'])
async def dummy_fixture(
    request: pytest.FixtureRequest,
):
    if request.param == 'foo':
        do_this()
    else:
        do_that()
```

agronholm: Seems like this was missed in the review. I'll make a fix promptly.

agronholm: I've released a fix now.

layday: Thank you!

PR #810, opened 2024-10-14
Switch backend detection to use entrypoints.

Fuyukai:
## Changes

This allows for external libraries to provide their own custom async backends, without needing to hack anyio internals.

## Checklist

features)

these instructions.

### Updating the changelog

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -3,6 +3,11 @@ Version history
 
 This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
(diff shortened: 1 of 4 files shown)


agronholm:
I have no plans to integrate this, since it would require making the `AsyncBackend` class public. Even recently I've made backwards incompatible changes to its interface - something that would've required a major version bump if it was part of the public API. On top of that, I've been burned by entry points before. PyInstaller and the likes require special consideration to not cause this feature to mysteriously break when packaged by them.

PR #820, opened 2024-11-09
Use ThreadSelectorEventLoop on Windows with ProactorEventLoop

davidbrochart:
## Changes

This PR allows `anyio.wait_socket_readable(sock)` to work on Windows when using the asyncio backend with a proactor event loop, by using Tornado's [SelectorThread](https://github.com/tornadoweb/tornado/blob/65a9e48f8ce645f104e3e0aa772222e70b0376d9/tornado/platform/asyncio.py#L438), as discussed with @minrk in https://github.com/zeromq/pyzmq/issues/1827.

## Checklist

features)

### DIFF
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -99,6 +99,44 @@ from ..abc._eventloop import StrOrBytesPath
 from ..lowlevel import RunVar
 from ..streams.memory import MemoryObjectReceiveStream, MemoryObjectSendStream
 
+# registry of asyncio loop : selector thread
+_selectors: WeakKeyDictionary = WeakKeyDictionary()
+
+
+def _get_selector_windows(
+    asyncio_loop: AbstractEventLoop,
+) -> AbstractEventLoop:
+    """Get selector-compatible loop.
+
+    Sets ``add_reader`` family of methods on the asyncio loop.
+
+    Workaround Windows proactor removal of *reader methods.
+    """
+
+    if asyncio_loop in _selectors:
+        return _selectors[asyncio_loop]
+
+    from ._selector_thread import AddThreadSelectorEventLoop
+
+    selector_loop = _selectors[asyncio_loop] = AddThreadSelectorEventLoop(  # type: ignore[abstract]
+        asyncio_loop
+    )
+
+    # patch loop.close to also close the selector thread
+    loop_close = asyncio_loop.close
+
+    def _close_selector_and_loop() -> None:
+        # restore original before calling selector.close,
+        # which in turn calls eventloop.close!
+        asyncio_loop.close = loop_close  # type: ignore[method-assign]
+        _selectors.pop(asyncio_loop, None)
+        selector_loop.close()
+
+    asyncio_loop.close = _close_selector_and_loop  # type: ignore[method-assign]
+
+    return selector_loop
+
+
 if sys.version_info >= (3, 10):
     from typing import ParamSpec
 else:
@@ -2683,6 +2721,12 @@ class AsyncIOBackend(AsyncBackend):
             raise BusyResourceError("reading from") from None
 
         loop = get_running_loop()
+        if (
+            sys.platform == "win32"
+            and asyncio.get_event_loop_policy().__class__.__name__
+            == "WindowsProactorEventLoopPolicy"
+        ):
+            loop = _get_selector_windows(loop)
         event = read_events[sock] = asyncio.Event()
         loop.add_reader(sock, event.set)
         try:
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -98,44 +98,7 @@ from ..abc import (
 from ..abc._eventloop import StrOrBytesPath
 from ..lowlevel import RunVar
 from ..streams.memory import MemoryObjectReceiveStream, MemoryObjectSendStream
-
-# registry of asyncio loop : selector thread
-_selectors: WeakKeyDictionary = WeakKeyDictionary()
-
-
-def _get_selector_windows(
-    asyncio_loop: AbstractEventLoop,
-) -> AbstractEventLoop:
-    """Get selector-compatible loop.
-
-    Sets ``add_reader`` family of methods on the asyncio loop.
-
-    Workaround Windows proactor removal of *reader methods.
-    """
-
-    if asyncio_loop in _selectors:
-        return _selectors[asyncio_loop]
-
-    from ._selector_thread import AddThreadSelectorEventLoop
-
-    selector_loop = _selectors[asyncio_loop] = AddThreadSelectorEventLoop(  # type: ignore[abstract]
-        asyncio_loop
-    )
-
-    # patch loop.close to also close the selector thread
-    loop_close = asyncio_loop.close
-
-    def _close_selector_and_loop() -> None:
-        # restore original before calling selector.close,
-        # which in turn calls eventloop.close!
-        asyncio_loop.close = loop_close  # type: ignore[method-assign]
-        _selectors.pop(asyncio_loop, None)
-        selector_loop.close()
-
-    asyncio_loop.close = _close_selector_and_loop  # type: ignore[method-assign]
-
-    return selector_loop
-
+from ._selector_thread import _get_selector_windows
 
 if sys.version_info >= (3, 10):
     from typing import ParamSpec
@@ -2726,14 +2689,19 @@ class AsyncIOBackend(AsyncBackend):
             and asyncio.get_event_loop_policy().__class__.__name__
             == "WindowsProactorEventLoopPolicy"
         ):
-            loop = _get_selector_windows(loop)
+            selector = _get_selector_windows(loop)
+            add_reader = selector.add_reader
+            remove_reader = selector.remove_reader
+        else:
+            add_reader = loop.add_reader
+            remove_reader = loop.remove_reader
         event = read_events[sock] = asyncio.Event()
-        loop.add_reader(sock, event.set)
+        add_reader(sock, event.set)
         try:
             await event.wait()
         finally:
             if read_events.pop(sock, None) is not None:
-                loop.remove_reader(sock)
+                remove_reader(sock)
                 readable = True
             else:
                 readable = False
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -98,7 +98,6 @@ from ..abc import (
 from ..abc._eventloop import StrOrBytesPath
 from ..lowlevel import RunVar
 from ..streams.memory import MemoryObjectReceiveStream, MemoryObjectSendStream
-from ._selector_thread import _get_selector_windows
 
 if sys.version_info >= (3, 10):
     from typing import ParamSpec
@@ -2684,19 +2683,19 @@ class AsyncIOBackend(AsyncBackend):
             raise BusyResourceError("reading from") from None
 
         loop = get_running_loop()
-        if (
-            sys.platform == "win32"
-            and asyncio.get_event_loop_policy().__class__.__name__
-            == "WindowsProactorEventLoopPolicy"
-        ):
+        add_reader = loop.add_reader
+        event = read_events[sock] = asyncio.Event()
+        try:
+            add_reader(sock, event.set)
+        except NotImplementedError:
+            # Proactor on Windows does not yet implement add/remove reader
+            from ._selector_thread import _get_selector_windows
+
             selector = _get_selector_windows(loop)
-            add_reader = selector.add_reader
+            selector.add_reader(sock, event.set)
             remove_reader = selector.remove_reader
         else:
-            add_reader = loop.add_reader
             remove_reader = loop.remove_reader
-        event = read_events[sock] = asyncio.Event()
-        add_reader(sock, event.set)
         try:
             await event.wait()
         finally:
@@ -2722,13 +2721,24 @@ class AsyncIOBackend(AsyncBackend):
             raise BusyResourceError("writing to") from None
 
         loop = get_running_loop()
+        add_writer = loop.add_writer
         event = write_events[sock] = asyncio.Event()
-        loop.add_writer(sock.fileno(), event.set)
+        try:
+            add_writer(sock.fileno(), event.set)
+        except NotImplementedError:
+            # Proactor on Windows does not yet implement add/remove writer
+            from ._selector_thread import _get_selector_windows
+
+            selector = _get_selector_windows(loop)
+            selector.add_writer(sock, event.set)
+            remove_writer = selector.remove_writer
+        else:
+            remove_writer = loop.remove_writer
         try:
             await event.wait()
         finally:
             if write_events.pop(sock, None) is not None:
-                loop.remove_writer(sock)
+                remove_writer(sock)
                 writable = True
             else:
                 writable = False
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -2690,15 +2690,15 @@ class AsyncIOBackend(AsyncBackend):
             raise BusyResourceError("reading from") from None
 
         loop = get_running_loop()
-        event = read_events[sock] = asyncio.Event()
+        event = read_events[obj] = asyncio.Event()
         try:
-            loop.add_reader(sock, event.set)
+            loop.add_reader(obj, event.set)
         except NotImplementedError:
             # Proactor on Windows does not yet implement add/remove reader
             from ._selector_thread import _get_selector_windows
 
             selector = _get_selector_windows(loop)
-            selector.add_reader(sock, event.set)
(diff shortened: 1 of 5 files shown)


agronholm:
What I'm saying is that a simpler solution would be better, where we just spawn a thread on demand that runs `select(...)` on all the sockets that need to wait for read/write. Or am I missing something?

davidbrochart:
But that would mean one thread per socket, right? While with this solution there is only one thread per event loop.

agronholm:
> But that would mean one thread per socket, right? While with this solution there is only one thread per event loop.

No, I never said that. There would be just one thread which would `select()` on multiple sockets.

davidbrochart:
OK, well there might be a better way than this PR, which is trying to bring Tornado's implementation almost untouched.

agronholm:
I think Tornado had different needs, hence their more elaborate solution.

minrk:
> Why would we have to shoehorn these into the async event loop?

I don't really understand what you mean by shoehorning into the loop. The selector is not being added to the loop itself, it is a single Thread running `select` which implements the `add_reader` interface of the EventLoop API in a threadsafe way, to call `select` in a thread and run callbacks in the event loop when `select` returns. It is an object with a reference to the event loop and a weakref registry to store one of these threads per event loop (only once `add_reader` is called).

However you want to do it, I think it is appropriate to implement it optimistically _as if_ `add_reader` can be assumed, because asyncio maintainers agree that it should be, even on Proactor, even though nobody has had the time to implement it yet (I still think it was a mistake to allow Proactor to become the default without this fundamental API). That way you can remove the workaround when you drop support for Pythons that didn't support it without changing ~anything. That is I think the main thing the tornado approach accomplishes because you only need to replace `loop.add_reader` with `get_selector(loop).add_reader`, which does nothing but return `loop.add_reader` everywhere but Windows + Proactor.

> where we just spawn a thread on demand that runs select(...) on all the sockets that need to wait for read/write. Or am I missing something?

It is indeed very simple if you have a single thread per socket, but if you want to use a shared thread instead of thread-per-socket, you need to handle the fact that `add_reader` and friends may be called when `select` is already blocking in a background thread, which means needing:

1. a threadsafe way to update the sockets for `select`, and
2. a mechanism to wake the thread when the socket list is updated

which is where most of the logic in the tornado implementation resides. You could reimplement it, but it seems to me like it's going to need to be quite similar to tornado's implementation if it's going to work. But maybe you have an idea I'm missing about how to handle waking the select thread in a simpler way?

src/anyio/_backends/_selector_thread.py
```diff
@@ -0,0 +1,341 @@
...
+        try:
+            del self._writers[fd]
+        except KeyError:
+            return False
+        self._wake_selector()
+        return True
+
+
+# AddThreadSelectorEventLoop: unmodified from tornado 6.4.0
+class AddThreadSelectorEventLoop(asyncio.AbstractEventLoop):
```
minrk, line 285: Maybe this class is part of the objection. It's not actually needed, since you can instantiate `selector = SelectorThread(asyncio_loop)` directly.

agronholm: > It is indeed very simple if you have a single thread per socket, but if you want to use a shared thread instead of thread-per-socket, you need to handle the fact that add_reader and friends may be called when select is already blocking in a background thread, which means needing:

I'm not sure where everybody gets the idea that I'm advocating for a one-thread-per-socket solution. That would be terribly inefficient.

> Maybe this class is part of the objection. It's not actually needed, since you can instantiate selector = SelectorThread(asyncio_loop) directly.

Yes, this class was the main point of my objection: I figured the code could be significantly trimmed down.

davidbrochart:
I don't understand that `test_wait_socket_readable` fails on trio now.

davidbrochart:
And now I don't understand that the test won't be skipped for Windows/Trio :)

agronholm:
> And now I don't understand that the test won't be skipped for Windows/Trio :)

Have you tried using `pytest.mark.skipif` to skip the test?

davidbrochart:
It seems that `pytest.mark.skipif` doesn't support fixtures, and we need the `anyio_backend_name` fixture. I don't see any other way than using `pytest.skip` inside the test.

tests/test_sockets.py
```diff
@@ -1849,3 +1850,28 @@ async def test_connect_tcp_getaddrinfo_context() -> None:
...
+async def test_wait_socket_readable(anyio_backend_name: str) -> None:
+    if anyio_backend_name == "trio" and platform.system() == "Windows":
+        pytest.skip("Internal error in Trio")
+
+    def client(port: int) -> None:
+        with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as sock:
+            sock.connect(("127.0.0.1", port))
+            sock.sendall(b"Hello, world")
+
+    with move_on_after(0.1):
```
agronholm, line 1864: Why a timeout this strict? It's bound to cause flakiness on slow/busy systems.
Also, wouldn't `fail_after` eliminate the need to set a flag?
```suggestion
    with fail_after(5):
```
On another note, I think this should be on the same line as `with conn:`, given that it's the only place where the timeout can work.

tests/test_sockets.py
```diff
@@ -1849,3 +1850,28 @@ async def test_connect_tcp_getaddrinfo_context() -> None:
...
+        with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as sock:
+            sock.bind(("127.0.0.1", 0))
+            port = sock.getsockname()[1]
+            sock.listen()
+            thread = Thread(target=client, args=(port,), daemon=True)
```
agronholm, line 1869: I don't think this needs to be daemon, given how it will either succeed or fail instantly. Given the continuous effort on CPython's side to get rid of daemon threads, I'm also looking for alternatives.

Also, the thread needs to be joined uncoditionally, even if the test fails. With this code, it won't be joined if the test fails.

agronholm (changes requested):

agronholm:
> It seems that `pytest.mark.skipif` doesn't support fixtures, and we need the `anyio_backend_name` fixture. I don't see any other way than using `pytest.skip` inside the test.

Do you have a Windows machine to run this locally on? Setting a breakpoint on the first line would probably shed some light on this.

davidbrochart:
> Do you have a Windows machine to run this locally on?

No, but with `platform.system() == "Linux"` I see that the test is skipped on my Linux machine for trio, so I really have no clue.

agronholm:
> > Do you have a Windows machine to run this locally on?
> 
> No, but with `platform.system() == "Linux"` I see that the test is skipped on my Linux machine for trio, so I really have no clue.

I have a Windows machine here, maybe I could take a look.

agronholm:
Something weird is going on here. If I only run the test on Trio, it's skipped as expected. I would say that the exception is some sort of fallout from the asyncio version of the test.

agronholm:
Specifically, it has to do with the thread that's spawned when `wait_socket_readable()` is called, not so much the thread that's spawned in the test code directly – that looks totally fine.

agronholm:
I checked that the `ResourceWarning` about the unclosed socket had nothing to do with any of the three sockets created during the test.

davidbrochart:
Thanks @agronholm, I had forgotten to close the loop when removing `AddThreadSelectorEventLoop`.
I think the [failure](https://github.com/agronholm/anyio/actions/runs/11793619202/job/32849543595?pr=820) in PyPy is unrelated?

src/anyio/_backends/_asyncio.py
```diff
@@ -2683,13 +2684,24 @@ async def wait_socket_readable(cls, sock: socket.socket) -> None:
             raise BusyResourceError("reading from") from None
 
         loop = get_running_loop()
+        if (
+            sys.platform == "win32"
```
graingert, line 2688: I think it would be better to try to use add_reader, and catch the NotImplementedError - raising if we're not on sys.platform == "win32"?

src/anyio/_backends/_selector_thread.py
```diff
@@ -0,0 +1,316 @@
...
+        self._closed = True
+
+    async def _thread_manager(self) -> typing.AsyncGenerator[None, None]:
+        # Create a thread to run the select system call. We manage this thread
+        # manually so we can trigger a clean shutdown from an atexit hook. Note
+        # that due to the order of operations at shutdown, only daemon threads
+        # can be shut down in this way (non-daemon threads would require the
+        # introduction of a new hook: https://bugs.python.org/issue41962)
+        self._thread = threading.Thread(
+            name="Tornado selector",
```
graingert, line 134: probably will need a different name

src/anyio/_backends/_asyncio.py
minrk: Yeah, I think if you catch the error you can register the selector independent of the platform in case someone else decides to make another incomplete EventLoop implementation like Proactor. The selector thread should work everywhere.

src/anyio/_backends/_asyncio.py
```diff
@@ -2683,13 +2684,24 @@ async def wait_socket_readable(cls, sock: socket.socket) -> None:
...
+        ):
+            selector = _get_selector_windows(loop)
+            add_reader = selector.add_reader
+            remove_reader = selector.remove_reader
+        else:
+            add_reader = loop.add_reader
+            remove_reader = loop.remove_reader
         event = read_events[sock] = asyncio.Event()
-        loop.add_reader(sock, event.set)
+        add_reader(sock, event.set)
```
minrk, line 2699: perhaps something like:

```suggestion
        try:
            add_reader(sock, event.set)
        except NotImplementedError:
            # Proactor on Windows does not yet implement add/remove reader
            selector = _get_selector_windows(loop)
            selector.add_reader(sock, event.set)
            remove_reader = selector.remove_reader
        else:
            remove_reader = loop.remove_reader
```

src/anyio/_backends/_selector_thread.py
```diff
@@ -0,0 +1,316 @@
...
+        # If we don't join our (daemon) thread here, we may get a deadlock
+        # during interpreter shutdown. I don't really understand why. This
+        # deadlock happens every time in CI (both travis and appveyor) but
+        # I've never been able to reproduce locally.
+        assert loop._thread is not None
+        loop._thread.join()
+    _selector_loops.clear()
+
+
+atexit.register(_atexit_callback)
```
graingert, line 59: This causes an issue for dask/distributed where they can't use tornado on the proactor because they send a shutdown message on a socket then close the event loop in atexit

the socket add_reader/add_writer then stops working while it's trying to send the shutdown message

anyio already has a thread pool implementation that shuts down when run_forever completes I think that should be used here

src/anyio/_backends/_selector_thread.py
```diff
@@ -0,0 +1,316 @@
...
+        self._closing_selector = False
+        self._thread: threading.Thread | None = None
+        self._thread_manager_handle = self._thread_manager()
+
+        async def thread_manager_anext() -> None:
+            # the anext builtin wasn't added until 3.10. We just need to iterate
+            # this generator one step.
+            await self._thread_manager_handle.__anext__()
+
+        # When the loop starts, start the thread. Not too soon because we can't
```
graingert, line 94: I don't think we need this, because we know the loop has started when we create the Selector thread. We should probably also stash the task somewhere

src/anyio/_backends/_selector_thread.py
graingert: Also the thread here is closed explicitly using the async generator hook *and* patching EventLoop.close so I think it's ok to leak the demon thread atexit if nobody explicitly closes the loop

src/anyio/_backends/_selector_thread.py
```diff
@@ -0,0 +1,290 @@
+"""Ensure asyncio selector methods (add_reader, etc.) are available.
+Running select in a thread and defining these methods on the running event loop.
+Originally in tornado.platform.asyncio.
+Redistributed under license Apache-2.0
```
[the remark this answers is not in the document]
graingert: eg from here https://github.com/tornadoweb/tornado/blob/e22eaeff0e60079cbdd0310f8f65fa8b677c81a9/tornado/__init__.py#L4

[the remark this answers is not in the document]
agronholm: Yeah, if we end up going with their implementation.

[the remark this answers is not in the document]
davidbrochart: I don't know how this plays with AnyIO's MIT license? Would it be an issue to vendor this code?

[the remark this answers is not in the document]
graingert: No, the Apache 2 license supports re-distribution under a different license

graingert:
I think what @agronholm is looking for is something that uses the existing thread shutdown approach

https://github.com/agronholm/anyio/blob/master/src/anyio/_backends/_asyncio.py#L2417

davidbrochart:
> I think what @agronholm is looking for is something that uses the existing thread shutdown approach

It's already here:
https://github.com/agronholm/anyio/blob/69800620adb78da8daccfee08a41949c67a4311a/src/anyio/_backends/_selector_thread.py#L269

But maybe you'd like to get rid of the loop patch?
https://github.com/agronholm/anyio/blob/69800620adb78da8daccfee08a41949c67a4311a/src/anyio/_backends/_selector_thread.py#L272-L283

agronholm:
> > I think what @agronholm is looking for is something that uses the existing thread shutdown approach
> 
> It's already here:
> 
> https://github.com/agronholm/anyio/blob/69800620adb78da8daccfee08a41949c67a4311a/src/anyio/_backends/_selector_thread.py#L269
> 
> But maybe you'd like to get rid of the loop patch?
> 
> https://github.com/agronholm/anyio/blob/69800620adb78da8daccfee08a41949c67a4311a/src/anyio/_backends/_selector_thread.py#L272-L283

If we could hook it into the life cycle of the root task, that would the ideal solution from the AnyIO PoV I think.

graingert:
Do we also want to close the selector thread after 5 seconds of not listening to sockets? Might not be worth bothering in this PR but maybe tomorrow

agronholm:
I think I'd like to take a stab at using the `selectors` API for this to potentially simplify the implementation (and to not have to think about the licensing stuff).

graingert:
Perhaps the way to go is to create an absolutely paired down copy of trio guest mode that only supports wait_readable/wait_writable?

agronholm:
> Perhaps the way to go is to create an absolutely paired down copy of trio guest mode that only supports wait_readable/wait_writable, then we can use the CFFI IOCP trick instead of selectors, and only run the thread when waiting for IO

Isn't the entire point of using the IOCP trick to not have to spawn threads?

graingert:
I think yes it does help with that, but it also lets you wait on more sockets because select on windows is limited

agronholm:
> I think yes it does help with that, but it also lets you wait on more sockets because select on windows is limited (and less efficient in other ways)

Ah, yeah, I totally forgot about that. OTOH, the most common cases where these low-level functions are needed are libraries like libzmq and libpq5 (postgresql), and it's unlikely that users will run into the limitations of Windows `select()` then.
 
> What's particularly helpful here is you can register/unregister an FD in the guest loop's IOCP without using a threading.Condition etc

Does the `selectors` API use that?

agronholm:
> > > What's particularly helpful here is you can register/unregister an FD in the guest loop's IOCP without using a threading.Condition etc
> > 
> > 
> > Does the `selectors` API use that?
> 
> I don't know what you're asking? I'm assuming it's not possible to change what select is waiting on from outside the call to select, you need to wake it up and give it a new set of fds

Ah, that's what you were getting at. And you're saying that the IOCP trick does not require the running system call to be interrupted in order to add/remove fds to it (I haven't looked at it at all, just asking)?

graingert:
> > > > What's particularly helpful here is you can register/unregister an FD in the guest loop's IOCP without using a threading.Condition etc
> > > 
> > > 
> > > Does the `selectors` API use that?
> > 
> > 
> > I don't know what you're asking? I'm assuming it's not possible to change what select is waiting on from outside the call to select, you need to wake it up and give it a new set of fds
> 
> Ah, that's what you were getting at. And you're saying that the IOCP trick does not require the running system call to be interrupted in order to add/remove fds to it (I haven't looked at it at all, just asking)?

Yep! Thanks for bearing with me

davidbrochart:
Also, guest mode could be useful to integrate AnyIO with other event loops, right?

agronholm:
> Also, guest mode could be useful to integrate AnyIO with other event loops, right?

I don't think I'm following.

davidbrochart:
Just in general, like [in Trio](https://trio.readthedocs.io/en/stable/reference-lowlevel.html#using-guest-mode-to-run-trio-on-top-of-other-event-loops).

graingert:
> Also, guest mode could be useful to integrate AnyIO with other event loops, right?

This is the wrong sort of guest mode - it's making a pared down trio IO manager the guest to asyncio, it's not going to let asyncio operate as a guest on anything else

PR #822, opened 2024-11-10
Made asyncio TaskGroup work with eager task factories

### LINKED ISSUE #764: `get_coro` is missing a `None` check in `_task_started`
### Things to check first

### AnyIO version

4.4.0

### Python version

3.12

### What happened?

Hello!

This was discovered upstream in CPython, see [this issue](https://github.com/python/cpython/issues/122332). In AnyIO's `_task_started` function (see [here](https://github.com/agronholm/anyio/blob/master/src/anyio/_backends/_asyncio.py#L335)), `get_coro()` can return `None` as of 3.12, with asyncio's new `eager_task_factory`. 

Unfortunately, there's not all that much you can do about this at this very moment -- a bug in C caused this code to segfault on current versions (this is [fixed now](https://github.com/python/cpython/pull/122338), but not released).

This was the reproducer used in the CPython issue, which eventually fails at `_task_started`. It's not all that minimal, sorry! I haven't used AnyIO enough to come up with my own reproducer.

```py
import asyncio

import httpx
import uvicorn
from fastapi import FastAPI
from starlette.responses import StreamingResponse


async def main():
    loop.set_task_factory(asyncio.eager_task_factory)

    app = FastAPI()

    @app.get("/")
    async def get():
        async def _():
            yield "1"

        return StreamingResponse(
            _(),
        )

    server = uvicorn.Server(
        uvicorn.Config(
            app,
            host="0.0.0.0",
            port=8080,
        )
    )
    asyncio.create_task(server.serve())

    client = httpx.AsyncClient()
    await client.get("http://localhost:8080")


if __name__ == "__main__":
    asyncio.run(main())
```

Again, this code segfaults on 3.12.4, you'll need to wait for the next patch release for the fix. In the meantime, it might be a good idea to throw an exception if using AnyIO with `eager_task_factory`.

agronholm:
## Changes

Fixes #764.

This updates the task group code and type annotations to accommodate eager task factories.

Additionally, this updates pre-commit modules. The updated `mypy` catches the annotation issues due to its updated `typeshed`.

## Checklist

features)

these instructions.

### Updating the changelog

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -22,14 +22,14 @@ repos:
       - id: trailing-whitespace
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.6.9
+    rev: v0.7.1
     hooks:
       - id: ruff
         args: [--fix, --show-fixes]
       - id: ruff-format
 
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v1.11.2
+    rev: v1.13.0
     hooks:
       - id: mypy
         additional_dependencies:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -22,7 +22,7 @@ repos:
       - id: trailing-whitespace
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.7.1
+    rev: v0.8.1
     hooks:
       - id: ruff
         args: [--fix, --show-fixes]
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -5,6 +5,12 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
 **UNRELEASED**
 
+- Added support for asyncio's eager task factories:
+
+  * Updated the annotation of ``TaskInfo.coro`` to allow it to be ``None``
+  * Updated ``TaskGroup`` to work with asyncio eager task factories
+
+  (`#764 <https://github.com/agronholm/anyio/issues/764>`_)
 - Fixed a misleading ``ValueError`` in the context of DNS failures
   (`#815 <https://github.com/agronholm/anyio/issues/815>`_; PR by @graingert)
 
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -5,7 +5,7 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
 **UNRELEASED**
 
-- Added support for asyncio's eager task factories:
+- Improved compatibility with asyncio's eager task factories:
 
   * Updated the annotation of ``TaskInfo.coro`` to allow it to be ``None``
   * Updated ``TaskGroup`` to work with asyncio eager task factories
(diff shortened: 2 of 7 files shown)


src/anyio/_backends/_asyncio.py
```diff
@@ -844,16 +888,22 @@ def task_done(_task: asyncio.Task) -> None:
...
-
         # Make the spawned task inherit the task group's cancel scope
-        _task_states[task] = TaskState(
+        _task_states[coro] = task_state = TaskState(
             parent_id=parent_id, cancel_scope=self.cancel_scope
         )
+        name = get_callable_name(func) if name is None else str(name)
+        try:
+            task = create_task(coro, name=name)
+        except BaseException:
```
graingert, line 898: This should just be a finally

agronholm:
Sigh...I improved my test and it's failing now. Back to the drawing board.

graingert:
> Sigh...I improved my test and it's failing now. Back to the drawing board.

CI looks green to me, what's failing?

agronholm:
> > Sigh...I improved my test and it's failing now. Back to the drawing board.
> 
> CI looks green to me, what's failing?

Fetching task state didn't work, as my amended test showed. I've pushed it now.

.pre-commit-config.yaml
```diff
@@ -22,14 +22,14 @@ repos:
       - id: trailing-whitespace
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.6.9
+    rev: v0.7.1
```
graingert-coef, line 25: I think there's 0.8 out now?

graingert: I think there's 0.8 out now?

docs/versionhistory.rst
```diff
@@ -5,6 +5,12 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
 **UNRELEASED**
 
+- Improved compatibility with asyncio's eager task factories:
+
+  * Updated the annotation of ``TaskInfo.coro`` to allow it to be ``None``
```
graingert, line 10: Is it possible for TaskInfo.coro to be None? Can't you only access it from a running task?

agronholm: This stemmed from a `mypy` complaint, but I think you're right – `TaskInfo` objects should never be created from completed tasks under any circumstances.

docs/versionhistory.rst
```diff
@@ -5,6 +5,8 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
 **UNRELEASED**
 
+- Updated ``TaskGroup`` to work with asyncio's eager task factories
```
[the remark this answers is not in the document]
agronholm: I'm not very inclined to apply this wording. I at least wouldn't know what that second sentence means, and I think it's too early to say anything about the future of eager task factories.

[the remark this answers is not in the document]
graingert: I think we should at least say experimental until the test suite is passing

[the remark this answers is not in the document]
agronholm: The `TaskGroup` test for this is passing though, and it's the only change we claim was made. Perhaps we could add a note in the task group's actual docs about this?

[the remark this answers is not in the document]
graingert: Yeah something to say it's experimental and different from trio behavior. If we find out it's too hard to support both behaviors from user code we can add a workaround without making a major release

graingert (approved):

PR #824, opened 2024-11-11
Allowed wait_socket_readable/writable to accept a file descriptor

### LINKED ISSUE #821: Allow `anyio.wait_socket_readable` to accept a file descriptor
### Things to check first

### Feature description

It seems that `anyio.wait_socket_readable(sock)` can not only accept a `socket.socket` but also a file descriptor (`int`).
- On asyncio, [add_reader](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.add_reader) accepts a file descriptor.
- On trio, [wait_readable](https://trio.readthedocs.io/en/stable/reference-lowlevel.html#trio.lowlevel.wait_readable) accepts a file descriptor or an object with a `.fileno()` method returning a file descriptor.

So I think there is no need to call `socket.fromfd()` as mentioned in https://github.com/agronholm/anyio/issues/386, but we just need to change the typing of the argument to be `socket.socket | int`?

### Use case

I think `socket.fromfd` doubles the number of open file descriptors. Also, it might not be easy to auto-detect the file descriptor's socket family, type and protocol, which are required for `socket.fromfd`.

davidbrochart:
## Changes

Fixes #821.

It seems that `anyio.wait_socket_readable(sock)` and `anyio.wait_socket_writable(sock)` can not only accept a `socket.socket` but also the socket's file descriptor (`int`):
- On asyncio, [add_reader](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.add_reader) and [add_writer](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.add_writer) accept a file descriptor.
- On trio, [wait_readable](https://trio.readthedocs.io/en/stable/reference-lowlevel.html#trio.lowlevel.wait_readable) and [wait_writable](https://trio.readthedocs.io/en/stable/reference-lowlevel.html#trio.lowlevel.wait_writable) accept a file descriptor or an object with a `.fileno()` method returning a file descriptor.

## Checklist

features)

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -7,9 +7,10 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
 - Fixed a misleading ``ValueError`` in the context of DNS failures
   (`#815 <https://github.com/agronholm/anyio/issues/815>`_; PR by @graingert)
-- Allowed ``wait_socket_readable`` and ``wait_socket_writable`` to accept a socket
-  file descriptor (`#824 <https://github.com/agronholm/anyio/pull/824>`_)
-  (PR by @davidbrochart)
+- Added ``wait_readable`` and ``wait_writable`` functions that accept an object with a
+  ``.fileno()`` method or an integer handle, and deprecated ``wait_socket_readable``
+  and ``wait_socket_writable``.
+  (`#824 <https://github.com/agronholm/anyio/pull/824>`_) (PR by @davidbrochart)
 
 **4.6.2**
 
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -7,10 +7,10 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
 - Fixed a misleading ``ValueError`` in the context of DNS failures
   (`#815 <https://github.com/agronholm/anyio/issues/815>`_; PR by @graingert)
-- Added ``wait_readable`` and ``wait_writable`` functions that accept an object with a
-  ``.fileno()`` method or an integer handle, and deprecated ``wait_socket_readable``
-  and ``wait_socket_writable``.
-  (`#824 <https://github.com/agronholm/anyio/pull/824>`_; PR by @davidbrochart)
+- Added the ``wait_readable()`` and ``wait_writable()`` functions which will accept
+  an object with a ``.fileno()`` method or an integer handle, and deprecated
+  their now obsolete versions (``wait_socket_readable()`` and
+  ``wait_socket_writable()`` (PR by @davidbrochart)
 
 **4.6.2**
 
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -7,6 +7,8 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
 - Fixed a misleading ``ValueError`` in the context of DNS failures
   (`#815 <https://github.com/agronholm/anyio/issues/815>`_; PR by @graingert)
+- Allowed ``wait_socket_readable`` and ``wait_socket_writable`` to accept a socket
+  file descriptor
 
 **4.6.2**
 
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -8,7 +8,8 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 - Fixed a misleading ``ValueError`` in the context of DNS failures
   (`#815 <https://github.com/agronholm/anyio/issues/815>`_; PR by @graingert)
 - Allowed ``wait_socket_readable`` and ``wait_socket_writable`` to accept a socket
-  file descriptor
+  file descriptor (`#824 <https://github.com/agronholm/anyio/pull/824>`_)
+  (PR by @davidbrochart)
 
 **4.6.2**
 
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -10,7 +10,7 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 - Added ``wait_readable`` and ``wait_writable`` functions that accept an object with a
   ``.fileno()`` method or an integer handle, and deprecated ``wait_socket_readable``
   and ``wait_socket_writable``.
-  (`#824 <https://github.com/agronholm/anyio/pull/824>`_) (PR by @davidbrochart)
+  (`#824 <https://github.com/agronholm/anyio/pull/824>`_; PR by @davidbrochart)
 
 **4.6.2**
 
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -2671,7 +2671,7 @@ class AsyncIOBackend(AsyncBackend):
         return await get_running_loop().getnameinfo(sockaddr, flags)
 
     @classmethod
-    async def wait_socket_readable(cls, sock: socket.socket) -> None:
+    async def wait_socket_readable(cls, sock: socket.socket | int) -> None:
         await cls.checkpoint()
         try:
             read_events = _read_events.get()
@@ -2698,7 +2698,7 @@ class AsyncIOBackend(AsyncBackend):
             raise ClosedResourceError
 
     @classmethod
-    async def wait_socket_writable(cls, sock: socket.socket) -> None:
+    async def wait_socket_writable(cls, sock: socket.socket | int) -> None:
         await cls.checkpoint()
         try:
             write_events = _write_events.get()
@@ -2711,7 +2711,7 @@ class AsyncIOBackend(AsyncBackend):
 
         loop = get_running_loop()
         event = write_events[sock] = asyncio.Event()
-        loop.add_writer(sock.fileno(), event.set)
+        loop.add_writer(sock, event.set)
         try:
             await event.wait()
         finally:
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -1722,8 +1722,8 @@ class ConnectedUNIXDatagramSocket(_RawSocketMixin, abc.ConnectedUNIXDatagramSock
                     return
 
 
-_read_events: RunVar[dict[int, asyncio.Event]] = RunVar("read_events")
-_write_events: RunVar[dict[int, asyncio.Event]] = RunVar("write_events")
+_read_events: RunVar[dict[socket.socket | int, asyncio.Event]] = RunVar("read_events")
+_write_events: RunVar[dict[socket.socket | int, asyncio.Event]] = RunVar("write_events")
 
 
 #
@@ -2675,16 +2675,13 @@ class AsyncIOBackend(AsyncBackend):
         return await get_running_loop().getnameinfo(sockaddr, flags)
 
     @classmethod
-    async def wait_socket_readable(cls, sock: HasFileno | int) -> None:
+    async def wait_socket_readable(cls, sock: socket.socket) -> None:
         await cls.checkpoint()
         try:
             read_events = _read_events.get()
         except LookupError:
             read_events = {}
             _read_events.set(read_events)
-
-        if not isinstance(sock, int):
-            sock = sock.fileno()
 
         if read_events.get(sock):
             raise BusyResourceError("reading from") from None
@@ -2705,7 +2702,7 @@ class AsyncIOBackend(AsyncBackend):
             raise ClosedResourceError
 
     @classmethod
-    async def wait_socket_writable(cls, sock: HasFileno | int) -> None:
+    async def wait_socket_writable(cls, sock: socket.socket) -> None:
         await cls.checkpoint()
         try:
             write_events = _write_events.get()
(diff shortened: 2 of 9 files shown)


davidbrochart:
Not sure about the [mypy error](https://results.pre-commit.ci/run/github/145337588/1731315406.qdZt6u-4R_2PduzVmTtobQ):
```
tests/test_sockets.py:1884: error: Argument 1 has incompatible type "object"; expected "socket | int"  [arg-type]
```

tests/test_sockets.py
```diff
@@ -1849,3 +1851,38 @@ async def test_connect_tcp_getaddrinfo_context() -> None:
...
+    with move_on_after(0.1):
+        with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as sock:
+            sock.bind(("127.0.0.1", 0))
+            port = sock.getsockname()[1]
+            sock.listen()
+            thread = Thread(target=client, args=(port,), daemon=True)
+            thread.start()
+            conn, addr = sock.accept()
+            with conn:
+                sock_or_fd = conn.fileno() if use_fd else conn
```
graingert, line 1882: I think this might be cleaner as two different tests, should keep mypy happy also.

Although I think you can fix mypy with isinstance instead of `type(x) is`

src/anyio/_backends/_asyncio.py
```diff
@@ -2711,7 +2711,7 @@ async def wait_socket_writable(cls, sock: socket.socket) -> None:
 
         loop = get_running_loop()
         event = write_events[sock] = asyncio.Event()
```
graingert, line 2713: I think write events now needs to be a RunVar of a dict of int to Event, rather than Any to Event. Otherwise someone could pass a socket and then later pass that socket's fd

agronholm: I concur – the key needs to be the `fileno()` of the socket.

tests/test_sockets.py
```diff
@@ -1849,3 +1851,38 @@ async def test_connect_tcp_getaddrinfo_context() -> None:
...
+                assert type(sock_or_fd) in (int, socket.socket)
```
agronholm, line 1883: This fixes the mypy error.
```suggestion
                sock_or_fd: socket.socket | int = conn.fileno() if use_fd else conn
```

src/anyio/_backends/_asyncio.py
```diff
@@ -2698,7 +2698,7 @@ async def wait_socket_readable(cls, sock: socket.socket) -> None:
             raise ClosedResourceError
 
     @classmethod
-    async def wait_socket_writable(cls, sock: socket.socket) -> None:
+    async def wait_socket_writable(cls, sock: socket.socket | int) -> None:
```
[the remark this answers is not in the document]
davidbrochart: But the method name has "socket" in it, so I think AnyIO expects a socket.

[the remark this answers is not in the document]
graingert: Probably the methods should be renamed to remove the socket from them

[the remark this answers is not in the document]
minrk: makes sense to me. I believe trio did the same rename to `wait_readable`  in 0.12 (2019), so it makes sense to follow suit.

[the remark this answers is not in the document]
agronholm: They were named that way because on Windows, you can't wait for readability on a non-socket FD. On Linux, you can.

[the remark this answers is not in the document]
agronholm: The right type annotation would likely be `_typeshed.HasFileno`.

agronholm:
You really should install `pre-commit` locally.

agronholm:
The docs build is now failing due to the typeshed annotation. I don't remember off hand how I've dealt with this previously.

davidbrochart:
I'm not sure why [docs](https://github.com/agronholm/anyio/actions/runs/11801705380/job/32875677930?pr=824) fails.

agronholm:
> The docs build is now failing due to the typeshed annotation. I don't remember off hand how I've dealt with this previously.

Ah, I remember now. Check out `anyio._core_fileio` to find out what I did about that.

docs/versionhistory.rst
```diff
@@ -7,6 +7,8 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
 - Fixed a misleading ``ValueError`` in the context of DNS failures
   (`#815 <https://github.com/agronholm/anyio/issues/815>`_; PR by @graingert)
+- Allowed ``wait_socket_readable`` and ``wait_socket_writable`` to accept a socket
+  file descriptor
```
agronholm, line 11:
```suggestion
  file descriptor (PR by @davidbrochart)
```

tests/test_sockets.py
```diff
@@ -1849,3 +1852,41 @@ async def test_connect_tcp_getaddrinfo_context() -> None:
...
+            pytest.skip("Does not work on asyncio/Windows/ProactorEventLoop")
+
+    wait_socket = wait_socket_readable if event == "readable" else wait_socket_writable
+
+    def client(port: int) -> None:
+        with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as sock:
+            sock.connect(("127.0.0.1", port))
+            sock.sendall(b"Hello, world")
+
+    with move_on_after(0.1):
```
agronholm, line 1876: This makes the test flaky; let's give it more breathing room.
```suggestion
    with move_on_after(10):
```
Also, move it closer to `await wait_socket(...)` as it's the only place where it has any effect. And maybe change to `fail_after()` so that you don't need to set those flags.

tests/test_sockets.py
```diff
@@ -1849,3 +1852,41 @@ async def test_connect_tcp_getaddrinfo_context() -> None:
...
+            conn, addr = sock.accept()
+            with conn:
+                sock_or_fd: HasFileno | int = (
+                    conn.fileno() if socket_type == "fd" else conn
+                )
+                await wait_socket(sock_or_fd)
+                socket_ready = True
+
+    assert socket_ready
+    thread.join()
```
agronholm, line 1892: The thread needs to joined uncondtionally before the test finishes.

src/anyio/_backends/_asyncio.py
```diff
@@ -1718,8 +1722,8 @@ async def send(self, item: bytes) -> None:
                     return
 
 
-_read_events: RunVar[dict[Any, asyncio.Event]] = RunVar("read_events")
-_write_events: RunVar[dict[Any, asyncio.Event]] = RunVar("write_events")
+_read_events: RunVar[dict[socket.socket | int, asyncio.Event]] = RunVar("read_events")
```
graingert, line 1725:
```suggestion
_read_events: RunVar[dict[int, asyncio.Event]] = RunVar("read_events")
```

src/anyio/_backends/_asyncio.py
```diff
@@ -1718,8 +1722,8 @@ async def send(self, item: bytes) -> None:
...
+_write_events: RunVar[dict[socket.socket | int, asyncio.Event]] = RunVar("write_events")
```
graingert, line 1726:
```suggestion
_write_events: RunVar[dict[int, asyncio.Event]] = RunVar("write_events")
```

src/anyio/_backends/_asyncio.py
graingert: It shouldn't be possible to put an int and a socket in here, otherwise a socket with the same fileno as an int could be added

agronholm:
Please stop force pushing your changes. It makes the review process much harder.

src/anyio/_backends/_asyncio.py
```diff
@@ -2679,17 +2683,18 @@ async def wait_socket_readable(cls, sock: socket.socket) -> None:
             read_events = {}
             _read_events.set(read_events)
 
-        if read_events.get(sock):
+        sock_fileno = sock.fileno()
```
graingert, line 2686: We should be able to remove the wait_socket_writable and wait_socket_readable from the backend as they're not public

graingert (changes requested):

src/anyio/_core/_sockets.py
```diff
@@ -591,8 +601,11 @@ def getnameinfo(sockaddr: IPSockAddrType, flags: int = 0) -> Awaitable[tuple[str
     return get_async_backend().getnameinfo(sockaddr, flags)
 
 
+@deprecated("This function is deprecated; use `wait_readable` instead", stacklevel=2)
```
graingert, line 604: Wait you need to set the stack level here? That can't be right

davidbrochart: Default value is `1`, would you prefer that?

src/anyio/_backends/_asyncio.py
davidbrochart: You mean using `wait_readable` and `wait_writable` instead?

src/anyio/_core/_sockets.py
graingert: Ah no you don't, the stack level is already set to +1 https://github.com/python/cpython/blob/3.13/Lib/warnings.py

```suggestion
@deprecated("This function is deprecated; use `wait_readable` instead")
```

src/anyio/_core/_sockets.py
```diff
@@ -611,8 +624,11 @@ def wait_socket_readable(sock: socket.socket) -> Awaitable[None]:
     return get_async_backend().wait_socket_readable(sock)
 
 
+@deprecated("This function is deprecated; use `wait_writable` instead", stacklevel=2)
```
graingert, line 627:
```suggestion
@deprecated("This function is deprecated; use `wait_writable` instead")
```

graingert (changes requested):

src/anyio/_core/_sockets.py
graingert: The default is 1, but that then calls warn with 2. If we set 2 then warn gets called with 3

src/anyio/_backends/_asyncio.py
graingert: Yep

src/anyio/_core/_sockets.py
```diff
@@ -591,8 +601,11 @@ def getnameinfo(sockaddr: IPSockAddrType, flags: int = 0) -> Awaitable[tuple[str
     return get_async_backend().getnameinfo(sockaddr, flags)
 
 
+@deprecated("This function is deprecated; use `wait_readable` instead")
 def wait_socket_readable(sock: socket.socket) -> Awaitable[None]:
     """
+    Deprecated, use `wait_readable` instead.
```
agronholm, line 607:
```suggestion
    .. deprecated:: 4.7.0
       Use :func:`wait_readable` instead.
```

src/anyio/_core/_sockets.py
```diff
@@ -608,11 +621,14 @@ def wait_socket_readable(sock: socket.socket) -> Awaitable[None]:
...
 
     """
-    return get_async_backend().wait_socket_readable(sock)
+    return get_async_backend().wait_readable(sock.fileno())
 
 
+@deprecated("This function is deprecated; use `wait_writable` instead")
 def wait_socket_writable(sock: socket.socket) -> Awaitable[None]:
     """
+    Deprecated, use `wait_writable` instead.
```
agronholm, line 630:
```suggestion
    .. deprecated:: 4.7.0
       Use :func:`wait_writable` instead.
```

agronholm (changes requested):
Let's use the proper Sphinx annotation for deprecations.

src/anyio/_core/_sockets.py
```diff
@@ -628,7 +646,59 @@ def wait_socket_writable(sock: socket.socket) -> Awaitable[None]:
...
+    descriptors aren't supported, and neither are handles that refer to anything besides
+    a ``SOCKET``.
+
+    This does **NOT** work on Windows when using the asyncio backend with a proactor
+    event loop (default on py3.8+).
+
+    .. warning:: Only use this on raw sockets that have not been wrapped by any higher
+        level constructs like socket streams!
+
+    :param obj: an object with a ``.fileno()`` method or an integer handle.
```
agronholm, line 672: No stop mark at the end, as other parameter descriptions don't have that either.
```suggestion
    :param obj: an object with a ``.fileno()`` method or an integer handle
```

src/anyio/_core/_sockets.py
```diff
@@ -628,7 +646,59 @@ def wait_socket_writable(sock: socket.socket) -> Awaitable[None]:
...
+
+    See `wait_readable` for the definition of ``obj``.
+
+    This does **NOT** work on Windows when using the asyncio backend with a proactor
+    event loop (default on py3.8+).
+
+    .. warning:: Only use this on raw sockets that have not been wrapped by any higher
+        level constructs like socket streams!
+
+    :param obj: an object with a ``.fileno()`` method or an integer handle.
```
agronholm, line 694: No stop mark at the end, as other parameter descriptions don't have that either.
```suggestion
    :param obj: an object with a ``.fileno()`` method or an integer handle
```

src/anyio/_core/_sockets.py
```diff
@@ -628,7 +646,59 @@ def wait_socket_writable(sock: socket.socket) -> Awaitable[None]:
...
+
+
+def wait_writable(obj: HasFileno | int) -> Awaitable[None]:
+    """
+    Wait until the given object can be written to.
+
+    See `wait_readable` for the definition of ``obj``.
+
+    This does **NOT** work on Windows when using the asyncio backend with a proactor
+    event loop (default on py3.8+).
```
agronholm, line 689:
```suggestion
    This does **NOT** work on Windows when using the asyncio backend with a proactor
    event loop (default on py3.8+).

    .. seealso:: See the documentation of :func:`wait_readable` for the definition of ``obj``.
```

tests/test_sockets.py
```diff
@@ -1849,3 +1854,60 @@ async def test_connect_tcp_getaddrinfo_context() -> None:
...
+            sock.connect(("127.0.0.1", port))
+            sock.sendall(b"Hello, world")
+
+    with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as sock:
+        sock.bind(("127.0.0.1", 0))
+        port = sock.getsockname()[1]
+        sock.listen()
+        thread = Thread(target=client, args=(port,), daemon=True)
+        thread.start()
+        conn, addr = sock.accept()
```
agronholm, line 1884: Maybe drop the thread entirely since it's apparently not needed here? Otherwise this should make sure that the thread is joined even if the test fails, which currently isn't the case.

tests/test_sockets.py
```diff
@@ -1849,3 +1854,60 @@ async def test_connect_tcp_getaddrinfo_context() -> None:
...
+        with conn:
+            sock_or_fd: HasFileno | int = conn.fileno() if socket_type == "fd" else conn
+            with fail_after(10):
+                await wait(sock_or_fd)
+                assert conn.recv(1024) == b"Hello, world"
+        thread.join()
+
+
+async def test_deprecated_wait_socket(anyio_backend_name: str) -> None:
+    if anyio_backend_name == "asyncio" and sys.platform == "win32":
```
agronholm, line 1894: I prefer the use of `platform.system` in this code base.
```suggestion
    if anyio_backend_name == "asyncio" and platform.system() == "Windows"
```

tests/test_sockets.py
```diff
@@ -1849,3 +1854,60 @@ async def test_connect_tcp_getaddrinfo_context() -> None:
...
+            pytest.skip("Does not work on asyncio/Windows/ProactorEventLoop")
+
+    with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as sock:
+        with pytest.warns(
+            DeprecationWarning,
+            match="This function is deprecated; use `wait_readable` instead",
+        ):
+            with move_on_after(0.1):
+                await wait_socket_readable(sock)
+        with pytest.warns(
```
agronholm, line 1908: Add a blank line after decreasing indentation.
```suggestion

        with pytest.warns(
```

agronholm (changes requested):

agronholm:
One more thing: the new functions need to be added to the API documentation too.

tests/test_sockets.py
davidbrochart: I dropped the thread :+1:

agronholm (approved):

graingert (approved):

PR #836, opened 2024-12-08
use settings from trio for Selector waker socketpair

graingert:
## Changes

Lifted from https://github.com/python-trio/trio/blob/da89ae850658fb365064cb12a47659c8a0ac44b5/src/trio/_core/_wakeup_socketpair.py#L12

## Checklist

features)

these instructions.

### Updating the changelog

```
```


graingert:
not sure how to test this, also needs a changelog entry

PR #839, opened 2024-12-09
Fixed shielded subtask cancel scope not protecting host from cancellation during start()

### LINKED ISSUE #837: CancelScope shielding fails to protect against cancellation until task_status.started is called on asyncio
### Things to check first

### AnyIO version

4.7.0

### Python version

3.10

### What happened?

Running some quite involved code that needs to do shielded cleanup if it fails sometime exits a shielded scope due to external cancellation if it was cancelled before it had called `task_status.started()` 

I know these are edge cases and may also be really hard to solve so feel free to resolve as such.

I've managed to reduce it to the following test-case.

```python

async def test_anyio_cancel_shielding_start():
    entered_inner_scope = anyio.Event()
    async with (
        anyio.create_task_group() as tg,
        anyio.create_task_group() as inner_tg,
    ):
        async def inner_task(*, task_status: anyio.abc.TaskStatus = anyio.TASK_STATUS_IGNORED):
            # task_status.started() # Uncommenting this lets the task hang forever instead

                entered_inner_scope.set()
                try:
                    await anyio.sleep_forever()
                    assert False, "Shouldn't be cancelled in a shielded scope"

        async def start_inner_task():
            await inner_tg.start(inner_task)

        tg.start_soon(start_inner_task)
        await entered_inner_scope.wait()
        tg.cancel_scope.cancel()



if __name__ == "__main__":
    anyio.run(test_anyio_cancel_shielding_start)
```

agronholm:
## Changes

Fixes #837. 

## Checklist

features)

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -3,6 +3,13 @@ Version history
 
 This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
+**UNRELEASED**
+
+- Fixed cancellation edge case on asyncio where a task spawning another with
+  ``TaskGroup.start()`` is not protected from external cancellation even when the
+  subtask has not yet called ``task_status.started()`` and is in a shielded cancel scope
+  (`#837 <https://github.com/agronholm/anyio/issues/837>`_)
+
 **4.7.0**
 
 - Updated ``TaskGroup`` to work with asyncio's eager task factories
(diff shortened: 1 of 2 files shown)


tapetersen:
@agronholm I was just about to check on this again and realised your TaskHandle rewrite added an explicit `CancelScope` around the part before `TaskStatus.started()` is called and replaced the raw `asyncio.Task.cancel()` by that which completely solves the original bug :) .

Not sure if rebasing and merging this to have a regression test for it is worth it or if it's better to just close the original issue directly.

tapetersen:
Closing as it's fixed and the test was merged in #1178

PR #841, opened 2024-12-15
Fixed AssertionError when using nest_asyncio

### LINKED ISSUE #840: Assertion error when `current_task()` returns an instance of `_asyncio.Task`
### Things to check first

### AnyIO version

4.7.0

### Python version

3.12.8

### What happened?

It seems that sometimes when `asyncio.current_task()` is called in

https://github.com/agronholm/anyio/blob/97f064295b65dc1d7a2ded6fd24ba5be46d4c4f4/src/anyio/_backends/_asyncio.py#L2367

an instance of `_asyncio.Task` is returned. This then causes an assertion to be raised in

https://github.com/agronholm/anyio/blob/97f064295b65dc1d7a2ded6fd24ba5be46d4c4f4/src/anyio/_backends/_asyncio.py#L685-L694

This is triggered in our test suite (fastapi, httpx, pytest-asyncio) after upgrading to anyio 4.7.0, but I haven't been able to come up with a MRE so hopefully this is enough to point maintainers to the potential fix.

agronholm:
## Changes

Refactored `TaskStateStore` to not check explicitly against `asyncio.Task`, as `nest-asyncio` replaces the asyncio modules.

Fixes #840. 

## Checklist

features)

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -3,6 +3,11 @@ Version history
 
 This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
+**UNRELEASED**
+
+- Fixed ``AssertionError`` when using ``nest-asyncio``
+  (`#840 <https://github.com/agronholm/anyio/issues/840>`_)
+
 **4.7.0**
 
 - Updated ``TaskGroup`` to work with asyncio's eager task factories
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -677,32 +677,33 @@ class TaskState:
         self.cancel_scope = cancel_scope
 
 
-class TaskStateStore(MutableMapping["Awaitable[Any] | asyncio.Task", TaskState]):
+class TaskStateStore(
+    MutableMapping["Coroutine[Any, Any, Any] | asyncio.Task", TaskState]
+):
     def __init__(self) -> None:
-        self._task_states = WeakKeyDictionary[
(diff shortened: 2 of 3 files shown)


agronholm:
I don't know how to write a test for this, as `nest_asyncio.apply()` is too invasive to be used in the test suite.

graingert:
you could try with:

```
with mock.patch("asyncio.Task", asyncio.tasks._PyTask):
    ...
```

agronholm:
I've added a test and verified that it fails without this patch.

edgarrmondragon:
I can confirm that this patch fixes the problem for us

sigridjineth:
@agronholm I just have came across this, but would be better implementing the logic for accomodating some varities of task-like objects I think. what's your opinion?

agronholm:
> @agronholm I just have came across this, but would be better implementing the logic for accomodating some varities of task-like objects I think. what's your opinion?

If you find a problem not fixed by this PR, let me know. Otherwise I'm not inclined to add any more code just because.

graingert (approved):

sandangel:
Hi, thank you so much for providing the fix for this issue. I wonder if we can create a release for this so I can download?

sandangel:
Please take your time. I pin to 4.6.2 at the moment.

agronholm:
> Please take your time. I pin to 4.6.2 at the moment.

All done now; v4.8.0 is out.

PR #843, opened 2024-12-17
[Fix] Add support for task-like objects without strict asyncio.

### LINKED ISSUE #840: Assertion error when `current_task()` returns an instance of `_asyncio.Task`

sigridjineth:
## Changes
Instead of asserting isinstance(key, asyncio.Task), the store now relies on the presence of get_coro() and asyncio.iscoroutine() checks. This ensures that objects behaving like tasks (e.g., those provided by nest_asyncio or PyCharm’s debugger) are handled gracefully.

Additioanlly, by removing the strict type checks, this PR improves interoperability with non-standard or monkey-patched asyncio environments, as well as newer Python features like asyncio.eager_task_factory.

## Fixes:
* Fixes #840
* credits to @seedspirit

## Checklist

features)

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -56,7 +56,7 @@ from typing import (
     Any,
     Optional,
     TypeVar,
-    cast,
+    cast, Union,
 )
 from weakref import WeakKeyDictionary
 
@@ -307,6 +307,8 @@ PosArgsT = TypeVarTuple("PosArgsT")
 P = ParamSpec("P")
 
 _root_task: RunVar[asyncio.Task | None] = RunVar("_root_task")
+
+TaskLike = Union[asyncio.Task, Coroutine[Any, Any, Any]]
 
 
 def find_root_task() -> asyncio.Task:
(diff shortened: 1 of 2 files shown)


agronholm:
This is a duplicate of #841, isn't it?

sigridjineth:
@agronholm aha, thanks for noticing it. it is fine to close this but I want to clarify that rather than only adjusting the type check, the implementation would be better by introducing some accomodity in the variety of task-like objects (testing for nest_asyncio, custom task factories, debugger level monkey patching)

that my intention behind it.

agronholm:
I'm not going to merge this because:

1. You've made a lot of unnecessary changes without any explanation for them
2. You've used `nest_asyncio.apply()` which will affect all subsequent asyncio tests
3. You've added redundant tests
4. You checked the checkboxes for having updated the docs and added a changelog entry, neither of which you've actually done
5. The test suite is failing and even pre-commit checks failed

All in all, #841 is a much leaner and more straightforward fix for the issue.

PR #850, opened 2024-12-31
Added support for subinterpreter workers

agronholm:
## Changes

Adds experimental support for subinterpreter workers

## Checklist

features)

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/docs/subinterpreters.rst
+++ b/docs/subinterpreters.rst
@@ -21,7 +21,7 @@ Running functions in a worker interpreter makes sense when:
 * The code is either pure Python code, or extension code that does not release the
   Global Interpreter Lock (GIL)
 
-If the code you'try trying to run only does blocking network I/O, or file I/O, then
+If the code you're trying to run only does blocking network I/O, or file I/O, then
 you're better off using :doc:`worker thread <threads>` instead.
 
 This is done by using :func:`.interpreter.run_sync`::
--- a/docs/subinterpreters.rst
+++ b/docs/subinterpreters.rst
@@ -1,0 +1,51 @@
+Working with subinterpreters
+============================
+
+.. py:currentmodule:: anyio
+
+Subinterpreters offer a middle ground between worker threads and worker processes. They
+allow you to utilize multiple CPU cores to run Python code while avoiding the overhead
+and complexities of spawning subprocesses.
+
+.. warning:: Subinterpreter support is considered **experimental**. The underlying
+   Python API for managing subinterpreters has not been finalized yet, and has had
+   little real-world testing. As such, it is not recommended to use this feature for
+   anything important yet.
+
+Running a function in a worker interpreter
+------------------------------------------
+
+Running functions in a worker interpreter makes sense when:
+
+* The code you want to run in parallel is CPU intensive
+* The code is either pure Python code, or extension code that does not release the
+  Global Interpreter Lock (GIL)
+
+If the code you'try trying to run only does blocking network I/O, or file I/O, then
+you're better off using :doc:`worker thread <threads>` instead.
+
+This is done by using :func:`.interpreter.run_sync`::
+
+    import time
+
+    from anyio import run, to_interpreter
+
+    from yourothermodule import cpu_intensive_function
+
+    async def main():
+        result = await to_interpreter.run_sync(
+            cpu_intensive_function, 'Hello, ', 'world!'
+        )
+        print(result)
+
+    run(main)
+
+Limitations
+-----------
+
+* Subinterpreters are only supported on Python 3.13 or later
+* Code in the ``__main__`` module cannot be run with this (as a consequence, this
+  applies to any functions defined in the REPL)
+* The target functions cannot react to cancellation
+* Unlike with threads, the code running in the subinterpreter cannot share mutable data
+  with other interpreters/threads (however, sharing _immutable_ data is fine)
--- a/src/anyio/_core/_exceptions.py
+++ b/src/anyio/_core/_exceptions.py
@@ -49,12 +49,12 @@ class BrokenWorkerIntepreter(Exception):
         else:
             return dedent(
                 f"""
-    {super().__str__()}
+                {super().__str__()}
 
-    Uncaught in the interpreter:
+                Uncaught in the interpreter:
 
-    {formatted}
-                    """.strip()
+                {formatted}
+                """.strip()
             )
 
 
--- a/src/anyio/_core/_exceptions.py
+++ b/src/anyio/_core/_exceptions.py
@@ -2,6 +2,8 @@ from __future__ import annotations
 
 import sys
 from collections.abc import Generator
+from textwrap import dedent
+from typing import Any
 
 if sys.version_info < (3, 11):
     from exceptiongroup import BaseExceptionGroup
@@ -19,6 +21,41 @@ class BrokenWorkerProcess(Exception):
     Raised by :meth:`~anyio.to_process.run_sync` if the worker process terminates abruptly or
     otherwise misbehaves.
     """
+
+
+class BrokenWorkerIntepreter(Exception):
+    """
+    Raised by :meth:`~anyio.to_interpreter.run_sync` if an unexpected exception is
+    raised in the subinterpreter.
+    """
+
+    def __init__(self, excinfo: Any):
+        # This was adapted from concurrent.futures.interpreter.ExecutionFailed
+        msg = excinfo.formatted
+        if not msg:
+            if excinfo.type and excinfo.msg:
+                msg = f"{excinfo.type.__name__}: {excinfo.msg}"
+            else:
+                msg = excinfo.type.__name__ or excinfo.msg
+
+        super().__init__(msg)
+        self.excinfo = excinfo
+
+    def __str__(self) -> str:
+        try:
+            formatted = self.excinfo.errdisplay
+        except Exception:
+            return super().__str__()
+        else:
+            return dedent(
+                f"""
+    {super().__str__()}
+
+    Uncaught in the interpreter:
+
+    {formatted}
+                    """.strip()
+            )
 
 
 class BusyResourceError(Exception):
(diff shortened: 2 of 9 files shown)


src/anyio/_core/_exceptions.py
```diff
@@ -21,6 +23,41 @@ class BrokenWorkerProcess(Exception):
...
+
+    def __str__(self) -> str:
+        try:
+            formatted = self.excinfo.errdisplay
+        except Exception:
+            return super().__str__()
+        else:
+            return dedent(
+                f"""
+    {super().__str__()}
```
graingert, line 52: formatting

src/anyio/to_interpreter.py
```diff
@@ -0,0 +1,208 @@
...
+        dedent("""
+        import _interpqueues as queues
+        import _interpreters as interpreters
+        from pickle import loads, dumps, HIGHEST_PROTOCOL
+
+        item = queues.get(queue_id)[0]
+        try:
+            func, args, kwargs = loads(item)
+            retval = func(*args, **kwargs)
+        except Exception as exc:
```
graingert, line 45: BaseException?

src/anyio/to_interpreter.py
```diff
@@ -0,0 +1,208 @@
...
+            queues.put(queue_id, (retval, is_exception), FMT_PICKLED, UNBOUND)
+        """),
+        "<string>",
+        "exec",
+    )
+
+    _interpreter_id: int
+    _queue_id: int
+
+    async def initialize(self) -> None:
```
graingert, line 64:
```suggestion
    def initialize(self) -> None:
```

does this need to be async?

src/anyio/to_interpreter.py
```diff
@@ -0,0 +1,208 @@
...
+            self._interpreter_id,
+            {
+                "queue_id": self._queue_id,
+                "FMT_PICKLED": FMT_PICKLED,
+                "FMT_UNPICKLED": FMT_UNPICKLED,
+                "UNBOUND": UNBOUND,
+            },
+        )
+
+    def destroy(self) -> None:
```
graingert, line 80: do you need a `__del__` to destroy the interpreter, just in case atexit isn't called?

src/anyio/to_interpreter.py
agronholm: No, and in fact it should not set up the interpreter in the event loop thread either. I'll fix both issues at once.

src/anyio/to_interpreter.py
agronholm: Under what circumstances would the `atexit` hooks not be called?

graingert: Hmm

src/anyio/to_interpreter.py
agronholm: Changed.

src/anyio/_core/_exceptions.py
agronholm: Better now?

agronholm:
Do we need pruning of interpreters that have been unused for too long?

richardsheridan:
Actually I was going to comment on that. following the trio thread cache model has been standard. Unlike threads or processes though, it seems like your implementation would not be able to free up any resources of a timed-out worker until someone calls `run_sync`.  There might be a better way to free up after something caches a bunch of interpreters? otherwise  you need something like a background pruning task.

 Also, are there any implications for unexpected behavior as the subinterpreters jump around different threads?

agronholm:
> Actually I was going to comment on that. following the trio thread cache model has been standard. Unlike threads or processes though, it seems like your implementation would not be able to free up any resources of a timed-out worker until someone calls `run_sync`. There might be a better way to free up after something caches a bunch of interpreters? otherwise you need something like a background pruning task.

I was thinking of pruning unused workers before or after a call to `run_sync()`. IIRC we do the same with unused worker threads.
 
> Also, are there any implications for unexpected behavior as the subinterpreters jump around different threads?

None that I'm aware of.

richardsheridan:
Separately, the cancellation story of a busy worker seems poor here. if you abandon the thread, the `destroy` in the atexit handler will [fail](https://github.com/python/cpython/blob/b875917e214208f022c6a8a84e4c6945593ea7e2/Modules/_interpretersmodule.c#L702-L708). Not sure what the consequences will be!

agronholm:
> Separately, the cancellation story of a busy worker seems poor here. if you abandon the thread, the `destroy` in the atexit handler will [fail](https://github.com/python/cpython/blob/b875917e214208f022c6a8a84e4c6945593ea7e2/Modules/_interpretersmodule.c#L702-L708). Not sure what the consequences will be!

But the worker threads should all be gone by the time the atexit hooks are run?

agronholm:
> > Separately, the cancellation story of a busy worker seems poor here. if you abandon the thread, the `destroy` in the atexit handler will [fail](https://github.com/python/cpython/blob/b875917e214208f022c6a8a84e4c6945593ea7e2/Modules/_interpretersmodule.c#L702-L708). Not sure what the consequences will be!
> 
> But the worker threads should all be gone by the time the atexit hooks are run?

I tested with this:
```py
import time

import anyio
from anyio import to_interpreter


async def main():
    await to_interpreter.run_sync(time.sleep, 6, abandon_on_cancel=True)


anyio.run(main)
```
It won't exit the process until the worker thread has run its course.

richardsheridan:
> > Actually I was going to comment on that. following the trio thread cache model has been standard. Unlike threads or processes though, it seems like your implementation would not be able to free up any resources of a timed-out worker until someone calls `run_sync`. There might be a better way to free up after something caches a bunch of interpreters? otherwise you need something like a background pruning task.
> 
> I was thinking of pruning unused workers before or after a call to `run_sync()`. IIRC we do the same with unused worker threads.

I think you mean processes? anyway, the issue is that the processes can automatically free up most resources except the OS process table bits that need waiting. An interpreter might sit on gigabytes of memory or thousands of sockets until the next call rather than the prescribed timeout.

> > Separately, the cancellation story of a busy worker seems poor here. if you abandon the thread, the destroy in the atexit handler will [fail](https://github.com/python/cpython/blob/b875917e214208f022c6a8a84e4c6945593ea7e2/Modules/_interpretersmodule.c#L702-L708). Not sure what the consequences will be!

> But the worker threads should all be gone by the time the atexit hooks are run?

Not if the subinterpereter is deadlocked or something like that!  

Also, just realized that a cancelled subinterpreter worker should definitely not be returned to the idle worker queue.

agronholm:
> > > Actually I was going to comment on that. following the trio thread cache model has been standard. Unlike threads or processes though, it seems like your implementation would not be able to free up any resources of a timed-out worker until someone calls `run_sync`. There might be a better way to free up after something caches a bunch of interpreters? otherwise you need something like a background pruning task.
> > 
> > 
> > I was thinking of pruning unused workers before or after a call to `run_sync()`. IIRC we do the same with unused worker threads.
> 
> I think you mean processes? anyway, the issue is that the processes can automatically free up most resources except the OS process table bits that need waiting. An interpreter might sit on gigabytes of memory or thousands of sockets until the next call rather than the prescribed timeout.
> 
> > > Separately, the cancellation story of a busy worker seems poor here. if you abandon the thread, the destroy in the atexit handler will [fail](https://github.com/python/cpython/blob/b875917e214208f022c6a8a84e4c6945593ea7e2/Modules/_interpretersmodule.c#L702-L708). Not sure what the consequences will be!
> 
> > But the worker threads should all be gone by the time the atexit hooks are run?
> 
> Not if the subinterpereter is deadlocked or something like that!

Explain please. The _task_ might abandon the worker thread, but the worker thread won't abandon the subinterpreter; it will continue to run the given code until it completes.

> Also, just realized that a cancelled subinterpreter worker should definitely not be returned to the idle worker queue.

Yes, if `abandon_on_cancel=True`, we should not add the subinterpreter to the idle queue when cancelled. I will deal with this somehow.

richardsheridan:
> It won't exit the process until the worker thread has run its course.

Is this all happening in a non-daemon thread? or is the interpreter shutodwn logic of python doing this? Either way it seems problematic. Maybe an initial release could ignore "abandon_on_cancel"?

agronholm:
> > It won't exit the process until the worker thread has run its course.
> 
> Is this all happening in a non-daemon thread? or is the interpreter shutodwn logic of python doing this? Either way it seems problematic. Maybe an initial release could ignore "abandon_on_cancel"?

The worker threads are daemonic, but there is a hook added to the root task that will ensure the threads have finished before the event loop exits. It's not 100% foolproof (what is?) but good enough for the vast majority of cases.

But I'm okay with leaving out `abandon_on_cancel` in the initial release.

richardsheridan:
>>>But the worker threads should all be gone by the time the atexit hooks are run?

>>Not if the subinterpereter is deadlocked or something like that!

>Explain please. The task might abandon the worker thread, but the worker thread won't abandon the subinterpreter; it will continue to run the given code until it completes.

Maybe i don't understand the order of operations on exit. I thought the main script ends, then atexits are run, then daemon threads are held up acquiring the gil, then interpreters are destroyed, then python runtime ends. 

Reading your previous message, it sounds like anyio interjects another step up there. 

I think cancellation should wait until subinterpreters have a better interruption story. I tried and failed to make workers that run on  channels and thread/interpreter pairs.

docs/subinterpreters.rst
```diff
@@ -0,0 +1,51 @@
...
+Running a function in a worker interpreter
+------------------------------------------
+
+Running functions in a worker interpreter makes sense when:
+
+* The code you want to run in parallel is CPU intensive
+* The code is either pure Python code, or extension code that does not release the
+  Global Interpreter Lock (GIL)
+
+If the code you'try trying to run only does blocking network I/O, or file I/O, then
```
uSpike, line 24:
```suggestion
If the code you're trying to run only does blocking network I/O, or file I/O, then
```

richardsheridan:
> ... processes can automatically free up most resources except the OS process table bits that need waiting.

I just reviewed the `anyio.to_process` implementation and noticed it lacks this ability as well, so I guess there's no strong reason to try to implement it for interpreters in this pr.

graingert (approved):

PR #851, opened 2025-01-01
make start_soon always start soon - support eager tasks

graingert:
## Changes

Fixes #. 

## Checklist

features)

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -1045,7 +1045,7 @@ class TaskGroup(abc.TaskGroup):
         name: object,
         task_status_future: asyncio.Future | None = None,
     ) -> Callable[[], Coroutine[Any, Any, Any]]:
-        def get_exception(_task: asyncio.Task) -> BaseException:
+        def get_exception(_task: asyncio.Task) -> BaseException | None:
             try:
                 return _task.exception()
             except CancelledError as e:
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -1070,7 +1070,6 @@ class TaskGroup(abc.TaskGroup):
             else:
                 task_status_future.set_exception(exc)
 
-
         def simple_task_done(_task: asyncio.Task) -> None:
             self._tasks.discard(_task)
             exc = get_exception(_task)
@@ -1169,7 +1168,9 @@ class TaskGroup(abc.TaskGroup):
 
         task_factory = loop.get_task_factory()
         if is_eager(task_factory):
-            spawn_task: asyncio.Task = task_factory(loop, create(), name=f"spawn {name}")  # type: ignore[assignment, misc, call-arg]
+            spawn_task: asyncio.Task = task_factory(
+                loop, create(), name=f"spawn {name}"
+            )  # type: ignore[assignment, misc, call-arg]
             self._tasks.add(spawn_task)
             spawn_task.add_done_callback(simple_task_done)
             return await_task_cancel_and_wait
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -28,8 +28,6 @@ from collections.abc import (
     Collection,
     Coroutine,
     Iterable,
-    Iterator,
-    MutableMapping,
     Sequence,
 )
 from concurrent.futures import Future
@@ -54,6 +52,7 @@ from typing import (
     IO,
     TYPE_CHECKING,
     Any,
+    Literal,
     Optional,
     TypeVar,
     cast,
@@ -357,6 +356,9 @@ def _task_started(task: asyncio.Task) -> bool:
     """Return ``True`` if the task has been started and has not finished."""
     # The task coro should never be None here, as we never add finished tasks to the
     # task list
+    if task.done():
+        return False
+
     coro = task.get_coro()
     assert coro is not None
     try:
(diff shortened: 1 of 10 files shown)


graingert:
does this need documenting as a new feature?

agronholm:
I weep at the complexity. Is all this absolutely necessary?

graingert:
Hopefully we get a `loop.create_task(..., eager_start=False)` in 3.14 and we can ditch the complexity

This took a few rounds to squash down as much complexity as possible

graingert:
This probably needs a test with a fresh instance of create_eager_task_factory so it goes through the eagerness measurement process

tests/conftest.py
```diff
@@ -29,42 +29,27 @@
...
-    pytest.param(
-        ("asyncio", {"debug": True, "loop_factory": eager_task_loop_factory}),
-        marks=eager_marks,
-        id="asyncio+eager",
-    ),
 ]
+if sys.version_info >= (3, 12):
+
+    def eager_task_loop_factory() -> asyncio.AbstractEventLoop:
+        loop = asyncio.new_event_loop()
```
graingert, line 43: new_event_loop uses the deprecated policy it's best to use EventLoop on 3.13+

graingert: the new_event_loop will be a legacy alias for EventLoop that's kept forever so I suppose it doesn't matter

agronholm: Where is this information coming from? I'm not seeing anything in the 3.13 or 3.14 docs about `new_event_loop()` using a deprecated policy. Link?

graingert: https://github.com/python/cpython/blob/main/Lib/asyncio/events.py#L821 it calls get_event_loop_policy, which is deprecated. Eventually new_event_loop will be a deprecated alias for EventLoop

tests/test_taskgroups.py
```diff
@@ -1759,28 +1762,98 @@ async def typetest_optional_status(
...
             pass
 
-    loop = asyncio.get_running_loop()
-    old_task_factory = loop.get_task_factory()
-    loop.set_task_factory(asyncio.eager_task_factory)
-    request.addfinalizer(lambda: loop.set_task_factory(old_task_factory))
+    async with create_task_group() as tg:
+        tg.start_soon(sync_coro)
+        assert not ran
+        tg.cancel_scope.cancel()
```
agronholm, line 1778: This test feels redundant with the new one, yes?

tests/test_taskgroups.py
```diff
@@ -1759,28 +1762,98 @@ async def typetest_optional_status(
...
+            coro: Any,
+            *,
+            name: str | None = None,
+            context: contextvars.Context | None = None,
+        ) -> Any:
+            return custom_task_constructor(
+                coro, loop=loop, name=name, context=context, eager_start=True
+            )
+
+        return factory
```
agronholm, line 1807: Just what is the point of this?

tests/test_taskgroups.py
```diff
@@ -1759,28 +1762,98 @@ async def typetest_optional_status(
...
+                "asyncio",
+                {
+                    "debug": True,
+                    "loop_factory": task_factory_loop_factory_factory(
+                        asyncio.create_eager_task_factory(asyncio.Task)
+                    ),
+                },
+            ),
+            id="asyncio+stdlib-custom-eager",
+        ),
```
agronholm, line 1821: Isn't this covered by the other asyncio_params already?

tests/test_taskgroups.py
graingert: This is to make a task factory that's eager, but not the same as the built in eager task factory, so it can test the slow path that measures task factories for eagerness

tests/test_taskgroups.py
graingert: No this is different because it uses a totally custom eager task factory that takes the slow path in is_eager

src/anyio/_backends/_asyncio.py
```diff
@@ -745,6 +762,10 @@ def started(self, value: T_contra | None = None) -> None:
 
 async def _wait(tasks: Iterable[asyncio.Task[object]]) -> None:
     tasks = set(tasks)
+    if not tasks:
+        await sleep(0)
+        return
+
```
graingert, line 768:
```suggestion
```

graingert:
I pushed a simplification that gets rid of the two counts of pending_eager_tasks, and collects any errors from the eager task spawner

src/anyio/_backends/_asyncio.py
```diff
@@ -763,6 +780,194 @@ def on_completion(task: asyncio.Task[object]) -> None:
...
+        if tf is None:
+            return False
+        try:
+            return _TF_IS_EAGER_MAP[tf]
+        except KeyError:
+            pass
+
+        if getattr(tf, "__code__", object()) is _TF_CODE:
+            _TF_IS_EAGER_MAP[tf] = True
+            return True
```
agronholm, line 954: Why is it necessary to special case `asyncio.eager_task_factory`?

src/anyio/_backends/_asyncio.py
```diff
@@ -763,6 +780,194 @@ def on_completion(task: asyncio.Task[object]) -> None:
...
+        pass
+
+    def subprocess_shell(self, *args: object, **kwargs: object) -> Any:
+        pass
+
+    def time(self, *args: object, **kwargs: object) -> Any:
+        pass
+
+
+if sys.version_info >= (3, 12):
```
agronholm, line 938: Wouldn't it be possible to create an eager task factory on earlier Python versions?

src/anyio/_backends/_asyncio.py
```diff
@@ -763,6 +780,194 @@ def on_completion(task: asyncio.Task[object]) -> None:
...
+            _TF_IS_EAGER_MAP[tf] = True
+            return True
+
+        ran = False
+
+        async def corofn() -> None:
+            nonlocal ran
+            ran = True
+
+        tf(_ShimEventLoop(), corofn()).cancel()
```
agronholm, line 962: What do we need the shim event loop for really?

src/anyio/_backends/_asyncio.py
graingert: To avoid having to create a task and measure it in the default case

src/anyio/_backends/_asyncio.py
graingert: It's very difficult to get current_task to work without the 3.12 swap_task code

src/anyio/_backends/_asyncio.py
graingert: To avoid creating a task that gets destroyed when pending

graingert: I think I could refactor this to use a task in the TaskGroup

agronholm:
Closing in favor of #853.

PR #854, opened 2025-01-04
refactor waiting for tasks to complete from task group on the asyncio backend

graingert:
## Changes

This uses the approach that asyncio uses for asyncio.TaskGroup, this means we only have one done callback per task which on the C implementation uses the efficient first callback slot, and further callbacks need to be added to a list - which is slower.

The old implementation added and removed a lot of callbacks on initiation and cancellation and is slower than it needs to be

## Checklist
This is a simple refactor so no tests here, also no update to the docs - we probably save an event loop cycle here or there.

features)

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -18,8 +18,6 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
   (`#840 <https://github.com/agronholm/anyio/issues/840>`_)
 - Fixed return type annotation of various context managers' ``__exit__`` method
   (`#847 <https://github.com/agronholm/anyio/issues/847>`_; PR by @Enegg)
-- Refactored TaskGroup task waiting on the asyncio backend
-  (`#854 <https://github.com/agronholm/anyio/pull/854>`_; PR by @graingert)
 
 **4.7.0**
 
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -18,6 +18,8 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
   (`#840 <https://github.com/agronholm/anyio/issues/840>`_)
 - Fixed return type annotation of various context managers' ``__exit__`` method
   (`#847 <https://github.com/agronholm/anyio/issues/847>`_; PR by @Enegg)
+- Refactored TaskGroup task waiting on the asyncio backend
+  (`#854 <https://github.com/agronholm/anyio/pull/854>`_; PR by @graingert)
 
 **4.7.0**
 
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -737,8 +737,7 @@ class TaskGroup(abc.TaskGroup):
                 if self._tasks:
                     with CancelScope() as wait_scope:
                         while self._tasks:
-                            if self._on_completed_fut is None:
-                                self._on_completed_fut = loop.create_future()
+                            self._on_completed_fut = loop.create_future()
 
                             try:
                                 await self._on_completed_fut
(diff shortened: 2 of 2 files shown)


docs/versionhistory.rst
```diff
@@ -18,6 +18,8 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
   (`#840 <https://github.com/agronholm/anyio/issues/840>`_)
 - Fixed return type annotation of various context managers' ``__exit__`` method
   (`#847 <https://github.com/agronholm/anyio/issues/847>`_; PR by @Enegg)
+- Refactored TaskGroup task waiting on the asyncio backend
+  (`#854 <https://github.com/agronholm/anyio/pull/854>`_; PR by @graingert)
```
agronholm, line 22: If this is not a user-facing change, then there should be no changelog entry. Is it?

graingert:
```suggestion
```

src/anyio/_backends/_asyncio.py
```diff
@@ -808,6 +793,10 @@ def task_done(_task: asyncio.Task) -> None:
             self._tasks.remove(task)
             del _task_states[_task]
 
+            if self._on_completed_fut is not None and not self._tasks:
+                if not self._on_completed_fut.done():
```
agronholm, line 797: I'm curious – under what circumstances would this check be negative? That is, if there are no tasks, how could `_on_completed_fut` be set already?

src/anyio/_backends/_asyncio.py
```diff
@@ -751,12 +732,15 @@ async def __aexit__(
...
                     self._exceptions.append(exc_val)
 
+            loop = get_running_loop()
             try:
                 if self._tasks:
                     with CancelScope() as wait_scope:
                         while self._tasks:
+                            if self._on_completed_fut is None:
+                                self._on_completed_fut = loop.create_future()
                             try:
```
agronholm, line 742: Add a blank line after a control block.
```suggestion
                                self._on_completed_fut = loop.create_future()

                            try:
```

src/anyio/_backends/_asyncio.py
```diff
@@ -771,6 +755,7 @@ async def __aexit__(
                                     and not is_anyio_cancellation(exc)
                                 ):
                                     exc_val = exc
+                            self._on_completed_fut = None
```
agronholm, line 758: Add a blank line after a control block.
```suggestion

                            self._on_completed_fut = None
```

src/anyio/_backends/_asyncio.py
agronholm: Hm, perhaps if the host task (and thus the future) was cancelled?

agronholm (changes requested):
Formatting.

src/anyio/_backends/_asyncio.py
```diff
@@ -751,12 +732,16 @@ async def __aexit__(
                 if not isinstance(exc_val, CancelledError):
                     self._exceptions.append(exc_val)
 
+            loop = get_running_loop()
             try:
                 if self._tasks:
                     with CancelScope() as wait_scope:
                         while self._tasks:
+                            if self._on_completed_fut is None:
+                                self._on_completed_fut = loop.create_future()
```
agronholm, line 741: Is there any case where we _don't_ want to replace the future here?

graingert: Oh yeah, we can always set it

graingert:
```suggestion
                            self._on_completed_fut = loop.create_future()
```

src/anyio/_backends/_asyncio.py
graingert: yep, removing this check makes 18 tests fail - I've replaced it with a try/catch - should be faster on 3.11+

agronholm (approved):

PR #859, opened 2025-01-11
Make `anyio.run_process()` accept `stdin` argument

jmehnle:
## Changes

Make `anyio.run_process()` accept a `stdin` argument.

This is consistent with what `asyncio.create_subprocess_exec()`, `trio.run_process()`, and `subprocess.run()`, which accept such an argument.

The handling of conflicts between `stdin` and `input` arguments is simplistic: the latter currently overrides the former without warning.

## Checklist

features)

### DIFF
--- a/src/anyio/_core/_subprocesses.py
+++ b/src/anyio/_core/_subprocesses.py
@@ -4,7 +4,7 @@ import sys
 from collections.abc import AsyncIterable, Iterable, Mapping, Sequence
 from io import BytesIO
 from os import PathLike
-from subprocess import DEVNULL, PIPE, CalledProcessError, CompletedProcess
+from subprocess import PIPE, CalledProcessError, CompletedProcess
 from typing import IO, Any, Union, cast
 
 from ..abc import Process
@@ -85,9 +85,12 @@ async def run_process(
 
         stream_contents[index] = buffer.getvalue()
 
+    if stdin is not None and input is not None:
+        raise ValueError("only one of stdin and input is allowed")
+
     async with await open_process(
         command,
-        stdin=(input and PIPE) or stdin or DEVNULL,
+        stdin=PIPE if input else stdin,
         stdout=stdout,
         stderr=stderr,
         cwd=cwd,
--- a/src/anyio/_core/_subprocesses.py
+++ b/src/anyio/_core/_subprocesses.py
@@ -23,6 +23,7 @@ async def run_process(
     command: StrOrBytesPath | Sequence[StrOrBytesPath],
     *,
     input: bytes | None = None,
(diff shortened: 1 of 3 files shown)


agronholm (approved):
LGTM. I'll just ask around if anyone has any objections.

src/anyio/_core/_subprocesses.py
```diff
@@ -84,7 +87,7 @@ async def drain_stream(stream: AsyncIterable[bytes], index: int) -> None:
 
     async with await open_process(
         command,
-        stdin=PIPE if input else DEVNULL,
+        stdin=(input and PIPE) or stdin or DEVNULL,
```
jakkdl, line 90: I think this is tricky to quickly parse correctly, I'd go with a super explicit if-statement above:

```python
if input:
   open_process_stdin = PIPE
elif stdin:
   open_process_stdin = stdin
else:
   open_process_stdin = NEVNULL
```

jakkdl (changes requested):
> The handling of conflicts between stdin and input arguments is simplistic: the latter currently overrides the former without warning.

this feels like an unnecessary footgun with ~zero upside, I think you should add an
```python
if stdin is not None and input is not None:
    raise ValueError("only one of stdin and input is allowed"
```

agronholm:
> > > The handling of conflicts between stdin and input arguments is simplistic: the latter currently overrides the former without warning.
> > 
> > 
> > this feels like an unnecessary footgun with ~zero upside, I think you should add an
> > ```python
> > if stdin is not None and input is not None:
> >     raise ValueError("only one of stdin and input is allowed"
> > ```
> 
> How do the upstream APIs handle it when both are passed?

Oh, right, this is an AnyIO only thing that we have. Never mind.

src/anyio/_core/_subprocesses.py
agronholm: I pushed some changes to this PR. What do you think?

jakkdl: even better :+1:

src/anyio/_core/_subprocesses.py
```diff
@@ -82,9 +85,12 @@ async def drain_stream(stream: AsyncIterable[bytes], index: int) -> None:
 
         stream_contents[index] = buffer.getvalue()
 
+    if stdin is not None and input is not None:
+        raise ValueError("only one of stdin and input is allowed")
```
jakkdl, line 89: you probably want a test case for this.
(maybe look over coverage config to enforce 100% patch coverage?)

agronholm: Path coverage minimum set to 100% on Coveralls. Test case added.

PR #864, opened 2025-01-28
Avoid chaining current exception when exiting task-groups on asyncio

### LINKED ISSUE #863: Quadratic traceback-growth in TaskGroup nesting-level on asyncio backend
### Things to check first

### AnyIO version

4.8.0

### Python version

3.11

### What happened?

Since python implicitly adds a context linking the current Exception to the new one when raising while exiting a context manager due to an unhandled Exception and task groups always wrap errors in their own `BaseExceptionGroup` the resulting traceback grows quadratically with the level of nested task groups due to the "During handling of the above..."

This seems to be straightforward to fix by simply explicitly raising from None in https://github.com/agronholm/anyio/blob/897d69aca35ff38ad230ee65f63186e420677f25/src/anyio/_backends/_asyncio.py#L767-L769

It looks like we already have added the exception to the group so breaking the context should be ok without loss of information.

Simple example with 5 levels
```python
import anyio

async def main():
    async with(
        anyio.create_task_group(),
        anyio.create_task_group(),
        anyio.create_task_group(),
        anyio.create_task_group(),
        anyio.create_task_group(),
    ):
        raise Exception("Something went wrong")

if __name__ == "__main__":
    anyio.run(main)
```

Results in 
```
  File "/home/tobias/Projects/anyio/test_tb.py", line 12, in main
    raise Exception("Something went wrong")
Exception: Something went wrong

During handling of the above exception, another exception occurred:

  |   File "/home/tobias/Projects/anyio/test_tb.py", line 5, in main
  |     async with(
  |   File "/home/tobias/Projects/anyio/src/anyio/_backends/_asyncio.py", line 771, in __aexit__
    |   File "/home/tobias/Projects/anyio/test_tb.py", line 12, in main
    |     raise Exception("Something went wrong")
    | Exception: Something went wrong

During handling of the above exception, another exception occurred:

  |   File "/home/tobias/Projects/anyio/test_tb.py", line 5, in main
  |     async with(
  |   File "/home/tobias/Projects/anyio/src/anyio/_backends/_asyncio.py", line 771, in __aexit__
    | Exception Group Traceback (most recent call last):
    |   File "/home/tobias/Projects/anyio/test_tb.py", line 5, in main
    |     async with(
    |   File "/home/tobias/Projects/anyio/src/anyio/_backends/_asyncio.py", line 771, in __aexit__
      |   File "/home/tobias/Projects/anyio/test_tb.py", line 12, in main
      |     raise Exception("Something went wrong")
      | Exception: Something went wrong

During handling of the above exception, another exception occurred:

  |   File "/home/tobias/Projects/anyio/test_tb.py", line 5, in main
  |     async with(
  |   File "/home/tobias/Projects/anyio/src/anyio/_backends/_asyncio.py", line 771, in __aexit__
    | Exception Group Traceback (most recent call last):
    |   File "/home/tobias/Projects/anyio/test_tb.py", line 5, in main
    |     async with(
    |   File "/home/tobias/Projects/anyio/src/anyio/_backends/_asyncio.py", line 771, in __aexit__
      | Exception Group Traceback (most recent call last):
      |   File "/home/tobias/Projects/anyio/test_tb.py", line 5, in main
      |     async with(
      |   File "/home/tobias/Projects/anyio/src/anyio/_backends/_asyncio.py", line 771, in __aexit__
        |   File "/home/tobias/Projects/anyio/test_tb.py", line 12, in main
        |     raise Exception("Something went wrong")
        | Exception: Something went wrong

During handling of the above exception, another exception occurred:

  |   File "/home/tobias/Projects/anyio/test_tb.py", line 5, in main
  |     async with(
  |   File "/home/tobias/Projects/anyio/src/anyio/_backends/_asyncio.py", line 771, in __aexit__
    | Exception Group Traceback (most recent call last):
    |   File "/home/tobias/Projects/anyio/test_tb.py", line 5, in main
    |     async with(
    |   File "/home/tobias/Projects/anyio/src/anyio/_backends/_asyncio.py", line 771, in __aexit__
      | Exception Group Traceback (most recent call last):
      |   File "/home/tobias/Projects/anyio/test_tb.py", line 5, in main
      |     async with(
      |   File "/home/tobias/Projects/anyio/src/anyio/_backends/_asyncio.py", line 771, in __aexit__
        | Exception Group Traceback (most recent call last):
        |   File "/home/tobias/Projects/anyio/test_tb.py", line 5, in main
        |     async with(
        |   File "/home/tobias/Projects/anyio/src/anyio/_backends/_asyncio.py", line 771, in __aexit__
          |   File "/home/tobias/Projects/anyio/test_tb.py", line 12, in main
          |     raise Exception("Something went wrong")
          | Exception: Something went wrong

During handling of the above exception, another exception occurred:

  |   File "/home/tobias/Projects/anyio/test_tb.py", line 15, in <module>
  |     anyio.run(main)
  |   File "/home/tobias/Projects/anyio/src/anyio/_core/_eventloop.py", line 74, in run
  |     return async_backend.run(func, args, {}, backend_options)
  |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |   File "/home/tobias/Projects/anyio/src/anyio/_backends/_asyncio.py", line 2306, in run
  |     return runner.run(wrapper())
  |            ^^^^^^^^^^^^^^^^^^^^^
  |   File "/usr/lib/python3.11/asyncio/runners.py", line 118, in run
  |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |   File "/usr/lib/python3.11/asyncio/base_events.py", line 654, in run_until_complete
  |     return future.result()
  |            ^^^^^^^^^^^^^^^
  |   File "/home/tobias/Projects/anyio/src/anyio/_backends/_asyncio.py", line 2294, in wrapper
  |     return await func(*args)
  |            ^^^^^^^^^^^^^^^^^
  |   File "/home/tobias/Projects/anyio/test_tb.py", line 5, in main
  |     async with(
  |   File "/home/tobias/Projects/anyio/src/anyio/_backends/_asyncio.py", line 771, in __aexit__
    | Exception Group Traceback (most recent call last):
    |   File "/home/tobias/Projects/anyio/test_tb.py", line 5, in main
    |     async with(
    |   File "/home/tobias/Projects/anyio/src/anyio/_backends/_asyncio.py", line 771, in __aexit__
      | Exception Group Traceback (most recent call last):
      |   File "/home/tobias/Projects/anyio/test_tb.py", line 5, in main
      |     async with(
      |   File "/home/tobias/Projects/anyio/src/anyio/_backends/_asyncio.py", line 771, in __aexit__
        | Exception Group Traceback (most recent call last):
        |   File "/home/tobias/Projects/anyio/test_tb.py", line 5, in main
        |     async with(
        |   File "/home/tobias/Projects/anyio/src/anyio/_backends/_asyncio.py", line 771, in __aexit__
          | Exception Group Traceback (most recent call last):
          |   File "/home/tobias/Projects/anyio/test_tb.py", line 5, in main
          |     async with(
          |   File "/home/tobias/Projects/anyio/src/anyio/_backends/_asyncio.py", line 771, in __aexit__
            |   File "/home/tobias/Projects/anyio/test_tb.py", line 12, in main
            |     raise Exception("Something went wrong")
            | Exception: Something went wrong
```

tapetersen:
Fixes #863

## Changes

Fixes #863

Break exception context chaining when exiting `TaskGroup` on asyncio

## Checklist

features)

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -772,7 +772,7 @@ class TaskGroup(abc.TaskGroup):
                         "unhandled errors in a TaskGroup", self._exceptions
                     ) from None
                 elif exc_val:
-                    raise exc_val from None
+                    raise exc_val
             except BaseException as exc:
                 if self.cancel_scope.__exit__(type(exc), exc, exc.__traceback__):
                     return True
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -764,11 +764,15 @@ class TaskGroup(abc.TaskGroup):
 
                 self._active = False
                 if self._exceptions:
+                    # The exception that got us here should already have been
+                    # added to self._exceptions so it's ok to break exception
+                    # chaining and avoid adding a "During handling of above..."
+                    # for each nesting level.
                     raise BaseExceptionGroup(
                         "unhandled errors in a TaskGroup", self._exceptions
-                    )
+                    ) from None
                 elif exc_val:
-                    raise exc_val
+                    raise exc_val from None
             except BaseException as exc:
                 if self.cancel_scope.__exit__(type(exc), exc, exc.__traceback__):
                     return True
--- a/tests/test_taskgroups.py
+++ b/tests/test_taskgroups.py
@@ -1839,7 +1839,6 @@ async def test_patched_asyncio_task(monkeypatch: MonkeyPatch) -> None:
         tg.start_soon(sleep, 0)
 
 
-@pytest.mark.parametrize("anyio_backend", asyncio_params)
 async def test_exception_groups_suppresses_exc_context() -> None:
     with pytest.raises(
         cast(type[ExceptionGroup[Exception]], ExceptionGroup)
--- a/tests/test_taskgroups.py
+++ b/tests/test_taskgroups.py
@@ -1837,3 +1837,14 @@ async def test_patched_asyncio_task(monkeypatch: MonkeyPatch) -> None:
     )
     async with create_task_group() as tg:
         tg.start_soon(sleep, 0)
+
+
+@pytest.mark.parametrize("anyio_backend", asyncio_params)
+async def test_exception_groups_suppresses_exc_context() -> None:
+    with pytest.raises(
+        cast(type[ExceptionGroup[Exception]], ExceptionGroup)
+    ) as exc_info:
+        async with create_task_group():
+            raise Exception("Error")
+
+    assert exc_info.value.__suppress_context__
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -3,6 +3,12 @@ Version history
 
 This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
+ **UNRELEASED**
+- Fixed traceback formatting growing quadratically with level of ``TaskGroup``
+  nesting on asyncio due to exception chaining when raising ``ExceptionGroups``
+  in ``TaskGroup.__aexit__``
+  (`#863 <https://github.com/agronholm/anyio/issues/863>`_; PR by @tapetersen)
+
 **4.8.0**
 
 - Added **experimental** support for running functions in subinterpreters on Python
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -4,6 +4,7 @@ Version history
 This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
  **UNRELEASED**
+
 - Fixed traceback formatting growing quadratically with level of ``TaskGroup``
   nesting on asyncio due to exception chaining when raising ``ExceptionGroups``
   in ``TaskGroup.__aexit__``


agronholm:
`asyncio.TaskGroup` seems to do the same. @graingert I take it there's no reason for us not to do this too?

src/anyio/_backends/_asyncio.py
```diff
@@ -764,11 +764,15 @@ async def __aexit__(
...
+                    # added to self._exceptions so it's ok to break exception
+                    # chaining and avoid adding a "During handling of above..."
+                    # for each nesting level.
                     raise BaseExceptionGroup(
                         "unhandled errors in a TaskGroup", self._exceptions
-                    )
+                    ) from None
                 elif exc_val:
-                    raise exc_val
+                    raise exc_val from None
```
graingert, line 775: I think this needs to restore the cause/traceback/etc in a

```python
try:
    raise exc
finally:
    exc.__cause__ = cause
    exc.__traceback__ = ...
    ...
```

?

tests/test_taskgroups.py
```diff
@@ -1837,3 +1837,14 @@ async def test_patched_asyncio_task(monkeypatch: MonkeyPatch) -> None:
     )
     async with create_task_group() as tg:
         tg.start_soon(sleep, 0)
+
+
+@pytest.mark.parametrize("anyio_backend", asyncio_params)
```
graingert, line 1842: This should pass on trio and asyncio, we want to make sure the behavior is the same

graingert:
> `asyncio.TaskGroup` seems to do the same. @graingert I take it there's no reason for us not to do this too?

Yep makes sense here too.

What does trio do?

tests/test_taskgroups.py
agronholm: Fair. This test seems to have been added since I took a look at this changeset.

src/anyio/_backends/_asyncio.py
agronholm: This seems to defeat the purpose of `raise ... from None`?

tapetersen: I'm not sure about if that one is correct.

tapetersen: Hmm only way that is hit is if it's a CancelledError as otherwise it would've been added on 733.

It's probably correct to not suppress context in that case as that would lose the frames below the group.
Will revert that.

Using `raise ... from None` doesn't actually remove the `__context__` link though but only sets the `__suppress_context__` flag to avoid printing it in tracebacks by default so any code traversing that for cancelled-errors should work anyway.

tapetersen: After some testing it seems like re-raising a caught instance like this doesn't add any extra context and can just be left as is in any case.

tests/test_taskgroups.py
tapetersen: It already passes on trio so I'll just remove the marks here then.

tapetersen:
> It's not clear to me: https://github.com/python-trio/trio/blob/d7cb2fc3f96c57074678f3024c926fb4d2a99b6f/src/trio/_core/_run.py#L1046-L1066 Reading Trio code always makes my head hurt.

Hmm it looks like it's avoiding this problem by patching up the exception with the context it had before it was re-raised, effectively removing the context the `__aexit__` re-raise would've added.

I'm not sure about the `sys.exc_info` part in the finally if it just modified  the `combined_error_from_nursery` object anyway (as noted by the assert) unless the `sys.exc_info()` ball by itself (as opposed to the `raise`) is what triggers the modification of `__context__`

agronholm:
Whoops, I accidentally clicked on the "Ready for review" button.

agronholm:
Could you add a changelog entry? That should be the only thing missing.

agronholm (approved):

PR #867, opened 2025-02-10
Add async support for temporary file handling. Closes #344

### LINKED ISSUE #344: async tempfile support
eg https://github.com/Tinche/aiofiles/issues/20

specifically with AsyncSpooledTemporaryFile it doesn't run the executor unless the in-memory buffer is full:

https://github.com/Tinche/aiofiles/blob/3a66417d89e204aa7fa424b39b858126412e4d3c/src/aiofiles/tempfile/temptypes.py#L27-L36

11kkw:
## Changes

feat: Add async support for temporary file handling (#344)

- Implemented async support for TemporaryFile, NamedTemporaryFile, SpooledTemporaryFile, and TemporaryDirectory.
- Integrated `to_thread` for handling async operations with tempfiles.
- Added tests to ensure async compatibility.

## Checklist

- [x] You've added tests (in `tests/`) which would fail without your patch.
- [ ] You've updated the documentation (in `docs/`, in case of behavior changes or new features). _(Pending)_
- [ ] You've added a new changelog entry (in `docs/versionhistory.rst`). _(Pending)_

### Updating the changelog

_(Changelog not updated)_

---

## Additional Notes

I have implemented async support for temporary file handling.  
Currently, I haven't updated the documentation or changelog yet.  

I'm unsure if this implementation follows the best approach, so I would appreciate feedback on the current implementation first.  
Once I receive input from the review, I will update the documentation and changelog accordingly if needed.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -82,6 +82,23 @@ Async file I/O
 .. autoclass:: anyio.AsyncFile
 .. autoclass:: anyio.Path
 
+Temporary files and directories
+-------------------------------
+
+.. autofunction:: anyio.mkstemp
+.. autofunction:: anyio.mkdtemp
+.. autofunction:: anyio.gettempdir
+.. autofunction:: anyio.gettempdirb
+
+.. autoclass:: anyio.TemporaryFile
+
+.. autoclass:: anyio.NamedTemporaryFile
+
+.. autoclass:: anyio.SpooledTemporaryFile
+
+.. autoclass:: anyio.TemporaryDirectory
+
+
 Streams and stream wrappers
 ---------------------------
 
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -91,13 +91,9 @@ Temporary files and directories
 .. autofunction:: anyio.gettempdirb
 
 .. autoclass:: anyio.TemporaryFile
-
 .. autoclass:: anyio.NamedTemporaryFile
-
 .. autoclass:: anyio.SpooledTemporaryFile
-
 .. autoclass:: anyio.TemporaryDirectory
-
 
 Streams and stream wrappers
 ---------------------------
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -90,6 +90,9 @@ extend-select = [
 [tool.ruff.lint.isort]
 "required-imports" = ["from __future__ import annotations"]
 
+[tool.ruff.lint.per-file-ignores]
+"tests/test_tempfile.py" = ["ASYNC230"]
+
 [tool.mypy]
 python_version = "3.13"
 strict = true
--- a/src/anyio/_core/_tempfile.py
+++ b/src/anyio/_core/_tempfile.py
@@ -291,10 +291,10 @@ class SpooledTemporaryFile(AsyncFile[AnyStr]):
         }
         self._max_size = max_size
         if "b" in mode:
-            super().__init__(BytesIO())
+            super().__init__(BytesIO())  # type: ignore[arg-type]
         else:
             super().__init__(
-                TextIOWrapper(
+                TextIOWrapper(  # type: ignore[arg-type]
                     BytesIO(),
                     encoding=encoding,
                     errors=errors,
@@ -337,7 +337,7 @@ class SpooledTemporaryFile(AsyncFile[AnyStr]):
         if not self._rolled:
             return self._fp.read(size)
 
-        return await super().read(size)
+        return await super().read(size)  # type: ignore[return-value]
 
     async def read1(self: SpooledTemporaryFile[bytes], size: int = -1) -> bytes:
         if not self._rolled:
@@ -349,13 +349,13 @@ class SpooledTemporaryFile(AsyncFile[AnyStr]):
         if not self._rolled:
             return self._fp.readline()
 
-        return await super().readline()
+        return await super().readline()  # type: ignore[return-value]
 
     async def readlines(self) -> list[AnyStr]:
         if not self._rolled:
             return self._fp.readlines()
 
-        return await super().readlines()
+        return await super().readlines()  # type: ignore[return-value]
 
     async def readinto(self: SpooledTemporaryFile[bytes], b: WriteableBuffer) -> int:
         if not self._rolled:
@@ -388,11 +388,11 @@ class SpooledTemporaryFile(AsyncFile[AnyStr]):
         return await super().truncate(size)
 
     @overload
-    async def write(self: SpooledTemporaryFile[bytes], s: ReadableBuffer) -> int: ...
+    async def write(self: SpooledTemporaryFile[bytes], b: ReadableBuffer) -> int: ...
     @overload
-    async def write(self: SpooledTemporaryFile[str], s: str) -> int: ...
+    async def write(self: SpooledTemporaryFile[str], b: str) -> int: ...
 
-    async def write(self, s: str | ReadableBuffer) -> int:
+    async def write(self, b: ReadableBuffer | str) -> int:
         """
         Asynchronously write data to the spooled temporary file.
 
@@ -405,11 +405,11 @@ class SpooledTemporaryFile(AsyncFile[AnyStr]):
 
         """
         if not self._rolled:
-            result = self._fp.write(s)
+            result = self._fp.write(b)
             await self._check()
             return result
 
-        return await super().write(s)
+        return await super().write(b)  # type: ignore[misc]
 
     @overload
     async def writelines(
@@ -436,7 +436,7 @@ class SpooledTemporaryFile(AsyncFile[AnyStr]):
             await self._check()
             return result
 
-        return await super().writelines(lines)
+        return await super().writelines(lines)  # type: ignore[misc]
 
 
 class TemporaryDirectory(Generic[AnyStr]):
--- a/src/anyio/_core/_tempfile.py
+++ b/src/anyio/_core/_tempfile.py
@@ -4,7 +4,7 @@ import sys
 import tempfile
 from functools import partial
 from types import TracebackType
-from typing import Any, AnyStr, Callable, Generic, cast
+from typing import Any, AnyStr, Callable, Generic, cast, overload
 
 from .. import to_thread
 from .._core._fileio import AsyncFile
@@ -273,6 +273,24 @@ class TemporaryDirectory(Generic[AnyStr]):
             await to_thread.run_sync(self._tempdir.cleanup)
 
 
+@overload
+async def mkstemp(
+    suffix: str | None = None,
+    prefix: str | None = None,
+    dir: str | None = None,
+    text: bool = False,
+) -> tuple[int, str]: ...
+
+
+@overload
+async def mkstemp(
+    suffix: bytes | None = None,
+    prefix: bytes | None = None,
+    dir: bytes | None = None,
+    text: bool = False,
+) -> tuple[int, bytes]: ...
+
+
 async def mkstemp(
     suffix: AnyStr | None = None,
     prefix: AnyStr | None = None,
@@ -280,6 +298,22 @@ async def mkstemp(
     text: bool = False,
 ) -> tuple[int, str | bytes]:
     return await to_thread.run_sync(lambda: tempfile.mkstemp(suffix, prefix, dir, text))
+
+
+@overload
+async def mkdtemp(
+    suffix: str | None = None,
+    prefix: str | None = None,
+    dir: str | None = None,
+) -> str: ...
+
+
+@overload
+async def mkdtemp(
+    suffix: bytes | None = None,
+    prefix: bytes | None = None,
+    dir: bytes | None = None,
+) -> bytes: ...
 
 
 async def mkdtemp(
--- a/src/anyio/_core/_tempfile.py
+++ b/src/anyio/_core/_tempfile.py
@@ -3,11 +3,11 @@ from __future__ import annotations
 import sys
 import tempfile
 from functools import partial
-from os import PathLike
 from types import TracebackType
 from typing import Any, AnyStr, Callable, Generic, cast
 
-from .. import AsyncFile, to_thread
+from .. import to_thread
+from .._core._fileio import AsyncFile
 
 
 class TemporaryFile(Generic[AnyStr]):
@@ -68,9 +68,9 @@ class NamedTemporaryFile(Generic[AnyStr]):
         buffering: int = -1,
         encoding: str | None = None,
         newline: str | None = None,
-        suffix: (str | bytes) | None = None,
-        prefix: (str | bytes) | None = None,
-        dir: (str | bytes | PathLike[str]) | None = None,
+        suffix: AnyStr | None = None,
+        prefix: AnyStr | None = None,
+        dir: AnyStr | None = None,
         delete: bool = True,
         *,
         errors: str | None = None,
@@ -80,9 +80,9 @@ class NamedTemporaryFile(Generic[AnyStr]):
         self.buffering = buffering
         self.encoding = encoding
         self.newline = newline
-        self.suffix = suffix
-        self.prefix = prefix
-        self.dir = dir
+        self.suffix: AnyStr | None = suffix
+        self.prefix: AnyStr | None = prefix
+        self.dir: AnyStr | None = dir
         self.delete = delete
         self.errors = errors
         self.delete_on_close = delete_on_close
@@ -262,14 +262,6 @@ async def mkdtemp(
     dir: AnyStr | None = None,
 ) -> str | bytes:
     return await to_thread.run_sync(lambda: tempfile.mkdtemp(suffix, prefix, dir))
-
-
-async def gettempprefix() -> str:
-    return await to_thread.run_sync(tempfile.gettempprefix)
-
-
-async def gettempprefixb() -> bytes:
-    return await to_thread.run_sync(tempfile.gettempprefixb)
 
 
 async def gettempdir() -> str:
--- a/src/anyio/_core/_tempfile.py
+++ b/src/anyio/_core/_tempfile.py
@@ -208,8 +208,8 @@ class SpooledTemporaryFile(Generic[AnyStr]):
                 self._fp.rollover()
 
             return result
-        else:
-            return await to_thread.run_sync(self._fp.writelines, lines)
+
+        return await to_thread.run_sync(self._fp.writelines, lines)
 
     async def __aexit__(
         self,
--- a/src/anyio/_core/_tempfile.py
+++ b/src/anyio/_core/_tempfile.py
@@ -16,6 +16,7 @@ from typing import (
 
 from .. import to_thread
 from .._core._fileio import AsyncFile
+from ..lowlevel import checkpoint_if_cancelled
 
 if TYPE_CHECKING:
     from _typeshed import OpenBinaryMode, OpenTextMode, ReadableBuffer, WriteableBuffer
@@ -335,54 +336,63 @@ class SpooledTemporaryFile(AsyncFile[AnyStr]):
 
     async def read(self, size: int = -1) -> AnyStr:
         if not self._rolled:
+            await checkpoint_if_cancelled()
             return self._fp.read(size)
 
         return await super().read(size)  # type: ignore[return-value]
 
     async def read1(self: SpooledTemporaryFile[bytes], size: int = -1) -> bytes:
         if not self._rolled:
+            await checkpoint_if_cancelled()
             return self._fp.read1(size)
 
         return await super().read1(size)
 
     async def readline(self) -> AnyStr:
         if not self._rolled:
+            await checkpoint_if_cancelled()
             return self._fp.readline()
 
         return await super().readline()  # type: ignore[return-value]
 
     async def readlines(self) -> list[AnyStr]:
         if not self._rolled:
+            await checkpoint_if_cancelled()
             return self._fp.readlines()
 
         return await super().readlines()  # type: ignore[return-value]
 
     async def readinto(self: SpooledTemporaryFile[bytes], b: WriteableBuffer) -> int:
         if not self._rolled:
+            await checkpoint_if_cancelled()
             self._fp.readinto(b)
 
         return await super().readinto(b)
 
     async def readinto1(self: SpooledTemporaryFile[bytes], b: WriteableBuffer) -> int:
         if not self._rolled:
+            await checkpoint_if_cancelled()
             self._fp.readinto(b)
 
         return await super().readinto1(b)
 
     async def seek(self, offset: int, whence: int | None = os.SEEK_SET) -> int:
         if not self._rolled:
+            await checkpoint_if_cancelled()
             return self._fp.seek(offset, whence)
 
         return await super().seek(offset, whence)
 
     async def tell(self) -> int:
         if not self._rolled:
+            await checkpoint_if_cancelled()
             return self._fp.tell()
 
         return await super().tell()
 
     async def truncate(self, size: int | None = None) -> int:
         if not self._rolled:
+            await checkpoint_if_cancelled()
             return self._fp.truncate(size)
 
         return await super().truncate(size)
@@ -405,6 +415,7 @@ class SpooledTemporaryFile(AsyncFile[AnyStr]):
 
         """
         if not self._rolled:
+            await checkpoint_if_cancelled()
             result = self._fp.write(b)
             await self._check()
             return result
@@ -432,6 +443,7 @@ class SpooledTemporaryFile(AsyncFile[AnyStr]):
 
         """
         if not self._rolled:
+            await checkpoint_if_cancelled()
             result = self._fp.writelines(lines)
             await self._check()
             return result
@@ -540,6 +552,7 @@ async def mkstemp(
     :param dir: Directory in which the temporary file is created.
     :param text: Whether the file is opened in text mode.
     :return: A tuple containing the file descriptor and the file name.
+
     """
     return await to_thread.run_sync(tempfile.mkstemp, suffix, prefix, dir, text)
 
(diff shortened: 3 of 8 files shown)


src/anyio/_core/_tempfile.py
```diff
@@ -0,0 +1,280 @@
...
+) -> str | bytes:
+    return await to_thread.run_sync(lambda: tempfile.mkdtemp(suffix, prefix, dir))
+
+
+async def gettempprefix() -> str:
+    return await to_thread.run_sync(tempfile.gettempprefix)
+
+
+async def gettempprefixb() -> bytes:
+    return await to_thread.run_sync(tempfile.gettempprefixb)
```
agronholm, line 272: We don't need async versions of these, do we? They don't do any file I/O.

tests/test_tempfile.py
```diff
@@ -0,0 +1,185 @@
...
+
+
+@pytest.mark.anyio
+async def test_temporary_file() -> None:
+    data = b"temporary file data"
+    async with TemporaryFile[bytes]() as af:
+        await af.write(data)
+        await af.seek(0)
+        result = await af.read()
+    assert result == data
```
agronholm, line 34: Always add a blank line between the end of a control block and the code that follows.
```suggestion

    assert result == data
```

tests/test_tempfile.py
```diff
@@ -0,0 +1,185 @@
...
+    tdir = await gettempdir()
+    tdirb = await gettempdirb()
+    assert isinstance(pref, str)
+    assert isinstance(prefb, bytes)
+    assert isinstance(tdir, str)
+    assert isinstance(tdirb, bytes)
+    assert pref == tempfile.gettempprefix()
+    assert prefb == tempfile.gettempprefixb()
+    assert tdir == tempfile.gettempdir()
+    assert tdirb == tempfile.gettempdirb()
```
agronholm, line 143: These make the earlier `isinstance` checks redundant.

tests/test_tempfile.py
```diff
@@ -0,0 +1,185 @@
...
+async def test_mkstemp() -> None:
+    fd, path_ = await mkstemp(suffix=".txt", prefix="mkstemp_", text=True)
+    assert isinstance(fd, int)
+    assert isinstance(path_, str)
+
+    def write_file() -> None:
+        with os.fdopen(fd, "w", encoding="utf-8") as f:
+            f.write("mkstemp")
+
+    await to_thread.run_sync(write_file)
```
agronholm, line 107: Is there a reason to run this in a worker thread?

tests/test_tempfile.py
```diff
@@ -0,0 +1,185 @@
...
+        result = await af.read()
+    assert result == data
+    assert af.closed
+
+
+@pytest.mark.anyio
+async def test_named_temporary_file() -> None:
+    data = b"named temporary file data"
+    async with NamedTemporaryFile[bytes]() as af:
+        filename: str = str(af.name)
```
agronholm, line 42: Looks like a redundant cast?
```suggestion
        filename = str(af.name)
```

tests/test_tempfile.py
```diff
@@ -0,0 +1,185 @@
...
+        assert os.path.exists(filename)
+        await af.write(data)
+        await af.seek(0)
+        result = await af.read()
+    assert result == data
+    assert not os.path.exists(filename)
+
+
+@pytest.mark.anyio
+async def test_spooled_temporary_file_io_and_rollover() -> None:
```
agronholm, line 52: It might be better to organize multiple tests related to the same class under a test class (e.g. `TestSpooledTemporaryFile` and then remove the prefixes from the test methods appropriately.

tests/test_tempfile.py
```diff
@@ -0,0 +1,185 @@
...
+    file_path.write_text("cleanup test", encoding="utf-8")
+    await td.cleanup()
+    assert not td_path.exists()
+
+
+@pytest.mark.anyio
+async def test_mkstemp() -> None:
+    fd, path_ = await mkstemp(suffix=".txt", prefix="mkstemp_", text=True)
+    assert isinstance(fd, int)
+    assert isinstance(path_, str)
```
agronholm, line 101: Unless I'm missing something, there is no reason to add the underscore suffix to `path`.
```suggestion
    fd, path = await mkstemp(suffix=".txt", prefix="mkstemp_", text=True)
    assert isinstance(fd, int)
    assert isinstance(path, str)
```

tests/test_tempfile.py
```diff
@@ -0,0 +1,185 @@
...
+        with os.fdopen(fd, "w", encoding="utf-8") as f:
+            f.write("mkstemp")
+
+    await to_thread.run_sync(write_file)
+
+    def read_file() -> str:
+        with open(path_, encoding="utf-8") as f:
+            return f.read()
+
+    content = await to_thread.run_sync(read_file)
```
agronholm, line 113: Ditto

tests/test_tempfile.py
```diff
@@ -0,0 +1,185 @@
...
+    assert content == "mkstemp"
+    await to_thread.run_sync(lambda: os.remove(path_))
```
agronholm, line 115: Same

agronholm (changes requested):
Looks good in general, but I left several change requests.

tests/test_tempfile.py
11kkw: I updated this part because ruff raised ASYNC230, which restricts the use of blocking I/O (open(), read(), write()) inside async functions.
To resolve this, I wrapped the file operations inside to_thread.run_sync(), ensuring that synchronous operations run safely in a separate thread.
Let me know if you'd like a different approach!

agronholm: While this warning is relevant for run-time code, it's not so in the test suite. You can just add it to the ruff settings in `pyproject.toml` using [per-file ignores](https://docs.astral.sh/ruff/settings/#lint_extend-per-file-ignores). Let me know if you need help with that.

agronholm:
You also need to fix the pyright errors.

src/anyio/_core/_tempfile.py
```diff
@@ -0,0 +1,272 @@
...
+                errors=self.errors,
+            )
+        )
+        self._async_file = AsyncFile(fp)
+        return self
+
+    async def rollover(self) -> None:
+        if self._async_file is None:
+            raise RuntimeError("Internal file is not initialized.")
+        await to_thread.run_sync(cast(Callable[[], None], self._async_file.rollover))
```
agronholm, line 170: Add a blank line between the end of a control block and the next code line that follows it. Do this everywhere please.

11kkw:
> Given that this is a new feature, I think it warrants a new documentation chapter. The code examples don't need to be comprehensive, but should make readers aware of this feature.

I will add the document work and commit it as soon as possible. Thank you for your thorough review!

src/anyio/_core/_tempfile.py
```diff
@@ -0,0 +1,276 @@
...
+        exc_type: type[BaseException] | None,
+        exc_value: BaseException | None,
+        traceback: TracebackType | None,
+    ) -> None:
+        if self._async_file is not None:
+            await self._async_file.aclose()
+            self._async_file = None
+
+
+class SpooledTemporaryFile(Generic[AnyStr]):
```
graingert, line 122: Ideally for SpooledTemporaryFile is that writes/reads to/from the BytesIO buffer will not go to a thread

agronholm: @Kludex would you like to participate in the review of this feature too?

src/anyio/_core/_tempfile.py
```diff
@@ -0,0 +1,276 @@
...
+    def closed(self) -> bool:
+        if self._async_file is not None:
+            try:
+                return bool(self._async_file.closed)
+            except AttributeError:
+                f = getattr(self._async_file, "_f", None)
+                if f is None:
+                    return True
+                return bool(f.closed)
+        return True
```
agronholm, line 184:
```suggestion
        if self._async_file is None:
            return True
            
        try:
            return bool(self._async_file.closed)
        except AttributeError:
            f = getattr(self._async_file, "_f", None)
            return f is None or bool(f.closed)
```

agronholm: I'd also like to see a comment explaining the alternate code path with the `_f` attribute.

src/anyio/_core/_tempfile.py
11kkw: I've made the necessary changes based on your feedback, but I'm not entirely sure if I've implemented them correctly due to my limited experience. I'd really appreciate it if you could review my changes and provide any further suggestions. Thank you for your guidance!"

src/anyio/_core/_tempfile.py
```diff
@@ -0,0 +1,298 @@
...
+) -> tuple[int, str | bytes]:
+    return await to_thread.run_sync(lambda: tempfile.mkstemp(suffix, prefix, dir, text))
+
+
+async def mkdtemp(
+    suffix: AnyStr | None = None,
+    prefix: AnyStr | None = None,
+    dir: AnyStr | None = None,
+) -> str | bytes:
+    return await to_thread.run_sync(lambda: tempfile.mkdtemp(suffix, prefix, dir))
```
agronholm, line 290: The type annotations here could be slightly improved. Here's what Typeshed has: https://github.com/python/typeshed/blob/a410f251ee535950eb4440ad41ae4c00c87e6a67/stdlib/tempfile.pyi#L465-L478

In particular, the overloads would allow callers to get the right return type based on the `suffix`, `prefix` or `dir` arguments.

```py
@overload
async def mkstemp(
    suffix: str | None = None,
    prefix: str | None = None,
    dir: str | None = None,
    text: bool = False,
) -> tuple[int, str]: ...

@overload
async def mkstemp(
    suffix: bytes | None = None,
    prefix: bytes | None = None,
    dir: bytes | None = None,
    text: bool = False,
) -> tuple[int, bytes]: ...
```

agronholm (changes requested):

tests/test_tempfile.py
```diff
@@ -0,0 +1,208 @@
...
+            await af.write(data)
+            await af.seek(0)
+            result = await af.read()
+
+        assert result == data
+        assert not os.path.exists(filename)
+
+    async def test_exception_handling(self) -> None:
+        async with NamedTemporaryFile[bytes]() as af:
+            filename = str(af.name)
```
agronholm, line 52: Redundant conversion: `af.name` is already a string.
```suggestion
            filename = af.name
```

tests/test_tempfile.py
```diff
@@ -0,0 +1,208 @@
...
+
+        assert result == data
+        assert af.closed
+
+
+class TestNamedTemporaryFile:
+    async def test_named_temporary_file(self) -> None:
+        data = b"named temporary file data"
+        async with NamedTemporaryFile[bytes]() as af:
+            filename = str(af.name)
```
agronholm, line 40: Redundant conversion: af.name is already a string.
```suggestion
            filename = af.name
```

tests/test_tempfile.py
```diff
@@ -0,0 +1,208 @@
...
+            original_rollover = stf._fp.rollover
+
+            def fake_rollover() -> None:
+                nonlocal rollover_called
+                rollover_called = True
+                return original_rollover()
+
+            stf._fp.rollover = fake_rollover
+            n1 = await stf.write(b"12345")
+            assert n1 == 5
```
agronholm, line 105:
```suggestion
            assert await stf.write(b"12345") == 5
```

tests/test_tempfile.py
```diff
@@ -0,0 +1,208 @@
...
+                nonlocal rollover_called
+                rollover_called = True
+                return original_rollover()
+
+            stf._fp.rollover = fake_rollover
+            await stf.writelines([b"hello", b"world"])
+            await stf.seek(0)
+
+            content = await stf.read()
+            assert content == b"helloworld"
```
agronholm, line 127:
```suggestion
            assert await stf.read() == b"helloworld"
```

tests/test_tempfile.py
```diff
@@ -0,0 +1,208 @@
...
+
+class TestSpooledTemporaryFile:
+    async def test_io_and_rollover(self) -> None:
+        data = b"spooled temporary file data" * 3
+        async with SpooledTemporaryFile[bytes](max_size=10) as stf:
+            await stf.write(data)
+            await stf.seek(0)
+            result = await stf.read()
+
+            assert result == data
```
agronholm, line 69:
```suggestion
            assert await stf.read() == data
```

tests/test_tempfile.py
```diff
@@ -0,0 +1,208 @@
...
+        data = b"named temporary file data"
+        async with NamedTemporaryFile[bytes]() as af:
+            filename = str(af.name)
+            assert os.path.exists(filename)
+
+            await af.write(data)
+            await af.seek(0)
+            result = await af.read()
+
+        assert result == data
```
agronholm, line 47:
```suggestion
            assert await af.read() == data

```

agronholm (changes requested):
I added some minor clean-ups as suggestions. The only real piece missing is the additions to the API reference documentation, and then @graingert needs to re-check the work you did on `SpooledTemporaryFile`. We're almost done now!

docs/api.rst
```diff
@@ -82,6 +82,23 @@ Async file I/O
...
+.. autofunction:: anyio.gettempdirb
+
+.. autoclass:: anyio.TemporaryFile
+
+.. autoclass:: anyio.NamedTemporaryFile
+
+.. autoclass:: anyio.SpooledTemporaryFile
+
+.. autoclass:: anyio.TemporaryDirectory
+
```
agronholm, line 100:
```suggestion
.. autoclass:: anyio.NamedTemporaryFile
.. autoclass:: anyio.SpooledTemporaryFile
.. autoclass:: anyio.TemporaryDirectory
```

src/anyio/_core/_tempfile.py
```diff
@@ -0,0 +1,454 @@
...
+        :return: The number of bytes written.
+        :raises RuntimeError: If the underlying file is not initialized.
+        """
+        if self._fp is None:
+            raise RuntimeError("Underlying file is not initialized.")
+
+        if not getattr(self._fp, "_rolled", True):
+            result = self._fp.write(s)
+            if self._fp._max_size and self._fp.tell() > self._fp._max_size:
+                self._fp.rollover()
```
graingert, line 259: Doesn't this do blocking IO?

src/anyio/_core/_tempfile.py
```diff
@@ -0,0 +1,454 @@
...
+        self.mode = mode
+        self.buffering = buffering
+        self.encoding = encoding
+        self.newline = newline
+        self.suffix: AnyStr | None = suffix
+        self.prefix: AnyStr | None = prefix
+        self.dir: AnyStr | None = dir
+        self.errors = errors
+        self._async_file: AsyncFile | None = None
+        self._fp: Any = None
```
graingert, line 200: I think should be initialized here, how does the SpooledTemporyFile work?

src/anyio/_core/_tempfile.py
graingert: This class needs an aclose()

src/anyio/_core/_tempfile.py
```diff
@@ -0,0 +1,454 @@
...
+
+        :param s: The data to write.
+        :return: The number of bytes written.
+        :raises RuntimeError: If the underlying file is not initialized.
+        """
+        if self._fp is None:
+            raise RuntimeError("Underlying file is not initialized.")
+
+        if not getattr(self._fp, "_rolled", True):
+            result = self._fp.write(s)
```
graingert, line 257:
```suggestion
            return self._fp.write(s)
```

src/anyio/_core/_tempfile.py
graingert: write will do the rollover for us in this situation, which will result in blocking disk IO on the event loop there, so we need to make sure it won't.

You don't have to use SpooledTemporyFile here you can just copy the source code here with attribution, and modify it to use an async file

pyproject.toml
```diff
@@ -90,6 +90,9 @@ extend-select = [
 [tool.ruff.lint.isort]
 "required-imports" = ["from __future__ import annotations"]
 
+[tool.ruff.lint.per-file-ignores]
+"tests/test_tempfile.py" = ["ASYNC230"]
```
graingert, line 94: Which open call is blocking?

tests/test_tempfile.py
```diff
@@ -0,0 +1,225 @@
...
+            await af.write(data)
+            await af.seek(0)
+            assert await af.read() == data
+
+        assert not os.path.exists(filename)  # type: ignore[arg-type]
+
+    async def test_exception_handling(self) -> None:
+        async with NamedTemporaryFile[bytes]() as af:
+            filename = af.name
+            assert os.path.exists(filename)  # type: ignore[arg-type]
```
graingert, line 53: Why is this a mypy error?

src/anyio/_core/_tempfile.py
graingert, line 257: We need to have a bit of a think here. According to the rules of trio every async function should always checkpoint, however for SpooledTemporyFile to be useful it should be as close to a BytesIO in terms of performance at small sizes. The documentation should have a big red note here or write should include a checkpoint if we're writing to the memory buffer

src/anyio/_core/_tempfile.py
agronholm: It doesn't do any file I/O if that's what you mean, so yes, self._fp should be initialized to `tempfile.SpooledTemporaryFile` here.

src/anyio/_core/_tempfile.py
```diff
@@ -0,0 +1,454 @@
...
+        self.errors = errors
+        self._async_file: AsyncFile | None = None
+        self._fp: Any = None
+
+    async def __aenter__(
+        self: SpooledTemporaryFile[AnyStr],
+    ) -> SpooledTemporaryFile[AnyStr]:
+        fp = await to_thread.run_sync(
+            partial(
+                tempfile.SpooledTemporaryFile,
```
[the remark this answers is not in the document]
agronholm: Right, you already mentioned that ^ here.

tests/test_tempfile.py
```diff
@@ -0,0 +1,225 @@
...
+        with open(os.fsdecode(path), "rb") as f:
+            read_content = f.read()
+
+    assert read_content == content
+
+    os.remove(path)
+
+
+@pytest.mark.parametrize("prefix", [b"mkdtemp_", "mkdtemp_"])
+async def test_mkdtemp(prefix: AnyStr) -> None:
```
agronholm, line 207: The typevar is unwarranted here.
```suggestion
async def test_mkdtemp(prefix: str | bytes) -> None:
```

pyproject.toml
agronholm: The one in `test_mkstemp()`.

tests/test_tempfile.py
agronholm: Because the `name` attribute is retrieved using `__getattr__()` which is declared to return `object`. I think this has been used elsewhere in anyio too – perhaps we should declare it to return `Any` instead?

src/anyio/_core/_tempfile.py
agronholm: @11kkw Yeah, so instead of delegating to `tempfile.SpooledTemporaryFile`, @graingert is suggesting that you roll an independent implementation in order to get more control over async checkpoints and such.

tests/test_tempfile.py
agronholm: Of course, that won't be necessary if this is just reimplemented by copying stdlib code.

agronholm:
@11kkw Let us know if you need help. I understand all this back and forth may feel pretty exhausting!

src/anyio/_core/_tempfile.py
Kludex: Yep, it needs the `aclose()`.

src/anyio/_core/_tempfile.py
```diff
@@ -0,0 +1,454 @@
...
+        exc_type: type[BaseException] | None,
+        exc_value: BaseException | None,
+        traceback: TracebackType | None,
+    ) -> None:
+        if self._async_file is not None:
+            await self._async_file.aclose()
+            self._async_file = None
+
+
+class SpooledTemporaryFile(Generic[AnyStr]):
```
Kludex, line 158: Why is this a generic? The `tempfile.SpooledTemporaryFile` is not a generic.

agronholm: Looks pretty generic to me? https://github.com/python/typeshed/blob/ac8f2632ec37bb4a82ade0906e6ce9bdb33883d3/stdlib/tempfile.pyi#L277

Kludex: There's something different happening here...

```py
from tempfile import SpooledTemporaryFile

from anyio import SpooledTemporaryFile as anyio_SpooledTemporaryFile

a = SpooledTemporaryFile()
b = anyio_SpooledTemporaryFile()
```

I think it defaults to `bytes`. My IDE is running pyright (strict mode):

Screenshot 2025-02-22 at 13 53 19

agronholm: Curious, mypy reports the type as `SpooledTemporaryFile[builtins.str]` while pyright reports it as `SpooledTemporaryFile[Unknown]`.

agronholm:
```py
from tempfile import SpooledTemporaryFile

from anyio import SpooledTemporaryFile as AnyIOSpooledTemporaryFile

async def main() -> None:
    with SpooledTemporaryFile() as f1:
        reveal_type(f1)

    async with AnyIOSpooledTemporaryFile() as f2:
        reveal_type(f2)
```
Gives me:
```
  /home/alex/workspace/anyio/client.py:7:21 - information: Type of "f1" is "SpooledTemporaryFile[bytes]"
  /home/alex/workspace/anyio/client.py:10:21 - information: Type of "f2" is "SpooledTemporaryFile[Unknown]"
```

agronholm: Hm, somehow the `mode` argument should determine the value for the type parameter.

Kludex: Ah, I see those overloads in typeshed...

agronholm: I fixed the implicit type parameter assignment locally, but I see some oddities around the `suffix` and `prefix` arguments – why are they `str` for `SpooledTemporaryFile` and `AnyStr` for other classes? That makes no sense to me.

agronholm: As I understand it, the type parameter determines the return type of `read()` and the input type for `write()`, so annotating the `prefix`, `suffix` and `dir` arguments with `AnyStr` doesn't make a lick of sense to me. @Kludex ?

agronholm: Ahh...I didn't notice before that those `TemporaryFile` etc. in typeshed were actually **functions**! That explains a few things. I'll work on fixing the `SpooledTemporaryFile` implementation tomorrow.

agronholm: I refactored `SpooledTemporaryFile` in a way that should satisfy @graingert. However, this caused a lot fo mypy errors that I can't understand.

11kkw: I'm sorry for bringing this up so late. However, since the official documentation states, "The dir, prefix and suffix parameters have the same meaning and defaults as with mkstemp()," I assumed that the parameters for mkstemp are of type AnyStr, and I proceeded with that approach.

[https://docs.python.org/3/library/tempfile.html#tempfile.mkstemp](url)

agronholm: Whatever the reason, using `AnyStr` in both the class type argument AND the initializer arguments makes no sense because the type parameter in the class is supposed to determine whether the input and output type is `str` or `bytes`. But the reason for the initializer arguments to be parametrized is because they in part determine file name and you want them to be either `str` or `bytes`. Two different, unrelated uses.

11kkw: > Whatever the reason, using `AnyStr` in both the class type argument AND the initializer arguments makes no sense because the type parameter in the class is supposed to determine whether the input and output type is `str` or `bytes`. But the reason for the initializer arguments to be parametrized is because they in part determine file name and you want them to be either `str` or `bytes`. Two different, unrelated uses.

I agree with this change

src/anyio/_core/_tempfile.py
```diff
@@ -0,0 +1,600 @@
...
+import os
+import sys
+import tempfile
+from collections.abc import Iterable
+from io import BytesIO, TextIOWrapper
+from types import TracebackType
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    AnyStr,
```
decorator-factory, line 12: `AnyStr` seems to be deprecated (https://github.com/python/cpython/issues/105578), maybe it's a good idea to declare a local type variable like this, to avoid extra code churn in Python 3.16
```py
AnyStr = TypeVar("AnyStr", str, bytes)
```
(`TypeVar` itself isn't being deprecated any time soon even with the new generic syntax)

agronholm: I frankly don't see how that helps with anything. We'll of course switch to the new typevar syntax as soon as it becomes possible, but until then I don't think there's any reason whatsoever to shun `AnyStr`.

src/anyio/_core/_tempfile.py
agronholm: Based on the discussions I had on Discord, it would appear that neither mypy or pyright can handle overriding methods like `write()` or `writelines()` correctly. I'm therefore leaning towards adding `# type: ignore` comments to mute the errors about methods like `write()` and `writelines()`.

agronholm (approved):
Everything passes now. I'll just give the other reviewers a chance to take a final look at the end result.

agronholm:
Thanks for your initiative @11kkw, this is a great addition!

PR #875, opened 2025-02-21
Detect blocking calls in coroutines using BlockBuster

cbornet:
## Changes

This PR uses the [blockbuster](https://github.com/cbornet/blockbuster) library to detect blocking calls made in the asyncio event loop during unit tests.
Avoiding blocking calls is hard as these can be deeply buried in the code or made in 3rd party libraries.
Blockbuster makes it easier to detect them by raising an exception when a call is made to a known blocking function (eg: `time.sleep`).

## Checklist

features)

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -59,6 +59,9 @@ def blockbuster() -> Iterator[BlockBuster]:
         "anyio", excluded_modules=["anyio.pytest_plugin", "anyio._backends._asyncio"]
     ) as bb:
         bb.functions["os.stat"].can_block_in("anyio/streams/tls.py", "wrap")
+        bb.functions["socket.socket.accept"].can_block_in(
+            "anyio/_core/_asyncio_selector_thread.py", {"get_selector"}
+        )
         for func in ["os.stat", "os.unlink"]:
             bb.functions[func].can_block_in(
                 "anyio/_core/_sockets.py", "setup_unix_local_socket"
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -58,7 +58,6 @@ def blockbuster() -> Iterator[BlockBuster]:
     with blockbuster_ctx(
         "anyio", excluded_modules=["anyio.pytest_plugin", "anyio._backends._asyncio"]
     ) as bb:
-        bb.functions["os.stat"].can_block_in("anyio/streams/tls.py", "wrap")
         bb.functions["socket.socket.accept"].can_block_in(
             "anyio/_core/_asyncio_selector_thread.py", {"get_selector"}
         )
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import asyncio
 import ssl
 import sys
-from collections.abc import Generator
+from collections.abc import Generator, Iterator
 from ssl import SSLContext
 from typing import Any
 from unittest.mock import Mock
@@ -11,6 +11,7 @@ from unittest.mock import Mock
 import pytest
 import trustme
 from _pytest.fixtures import SubRequest
+from blockbuster import BlockBuster, blockbuster_ctx
 from trustme import CA
 
 uvloop_marks = []
@@ -50,6 +51,23 @@ if sys.version_info >= (3, 12):
             id="asyncio+eager",
         ),
     )
+
+
+@pytest.fixture(autouse=True)
+def blockbuster() -> Iterator[BlockBuster]:
+    with blockbuster_ctx(
+        "anyio", excluded_modules=["anyio.pytest_plugin", "anyio._backends._asyncio"]
+    ) as bb:
+        for func in ["os.stat", "os.unlink"]:
+            bb.functions[func].can_block_in(
+                "anyio/_core/_sockets.py", "setup_unix_local_socket"
+            )
+        yield bb
+
+
+@pytest.fixture
+def deactivate_blockbuster(blockbuster: BlockBuster) -> None:
+    blockbuster.deactivate()
 
 
 @pytest.fixture(params=[*asyncio_params, pytest.param("trio")])
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -65,6 +65,7 @@ def blockbuster() -> Iterator[BlockBuster]:
             bb.functions[func].can_block_in(
                 "anyio/_core/_sockets.py", "setup_unix_local_socket"
             )
+
         yield bb
 
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -58,6 +58,7 @@ def blockbuster() -> Iterator[BlockBuster]:
     with blockbuster_ctx(
         "anyio", excluded_modules=["anyio.pytest_plugin", "anyio._backends._asyncio"]
     ) as bb:
+        bb.functions["os.stat"].can_block_in("anyio/streams/tls.py", "wrap")
         for func in ["os.stat", "os.unlink"]:
             bb.functions[func].can_block_in(
                 "anyio/_core/_sockets.py", "setup_unix_local_socket"
--- a/tests/test_from_thread.py
+++ b/tests/test_from_thread.py
@@ -628,6 +628,7 @@ class TestBlockingPortal:
             assert exc.value.__context__ is None
 
     @pytest.mark.parametrize("portal_backend_name", get_all_backends())
+    @pytest.mark.usefixtures("deactivate_blockbuster")
     async def test_from_async(
         self, anyio_backend_name: str, portal_backend_name: str
     ) -> None:
--- a/tests/test_to_process.py
+++ b/tests/test_to_process.py
@@ -73,6 +73,7 @@ async def test_cancel_before() -> None:
     pytest.raises(LookupError, to_process._process_pool_workers.get)
 
 
+@pytest.mark.usefixtures("deactivate_blockbuster")
 async def test_cancel_during() -> None:
     """
     Test that cancelling an operation on the worker process causes the process to be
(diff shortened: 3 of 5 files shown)


tests/conftest.py
```diff
@@ -52,6 +53,26 @@ def eager_task_loop_factory() -> asyncio.AbstractEventLoop:
     )
 
 
+@pytest.fixture(autouse=True)
+def blockbuster() -> Iterator[BlockBuster]:
+    with blockbuster_ctx(
+        "anyio", excluded_modules=["anyio.pytest_plugin", "anyio._backends._asyncio"]
+    ) as bb:
+        bb.functions["os.read"].can_block_in(
+            "asyncio/base_events.py", "subprocess_exec"
```
cbornet, line 62: I'll upstream this one to blockbuster. `subprocess_shell` is already excluded for the same reason.

tests/conftest.py
```diff
@@ -52,6 +53,26 @@ def eager_task_loop_factory() -> asyncio.AbstractEventLoop:
...
+        )
+        for func in ["os.stat", "os.unlink"]:
+            bb.functions[func].can_block_in(
+                "anyio/_core/_sockets.py", "setup_unix_local_socket"
```
cbornet, line 66: `setup_unix_local_socket` is the only coroutine making blocking calls to `os.stat` and `os.unlink` .
Interestingly, asyncio's `create_unix_server` does exactly the same !
I did raise an issue for asyncio : https://github.com/python/cpython/issues/129807 but that was handled as `won't fix` by Guido himself 😄 . I let you decide if you want to do something about it. Blockbuster is about detecting potential issues, not forcing anything.

tests/test_from_thread.py
```diff
@@ -628,6 +628,7 @@ async def raise_baseexception() -> None:
             assert exc.value.__context__ is None
 
     @pytest.mark.parametrize("portal_backend_name", get_all_backends())
+    @pytest.mark.usefixtures("deactivate_blockbuster")
     async def test_from_async(
```
cbornet, line 632: Blockbuster makes this test fail.
At some point `run_sync_from_thread` is called which calls the blocking  `result()` on a future
I'm not sure exactly what to do.
So deactivating BlockBuster at the moment.
@agronholm Do you have insight ?

cbornet:
@agronholm about your question if it would have detected the `iterdir` issue: it would have.
I did on my local with Python 3.13 venv
```sh
git revert 38885f9
pytest "tests/test_fileio.py::TestPath::test_iterdir[asyncio]"
=================================================================== test session starts ====================================================================
platform darwin -- Python 3.13.2, pytest-8.3.4, pluggy-1.5.0
rootdir: /Users/christophe.bornet/workspace/anyio
configfile: pyproject.toml
plugins: anyio-0.0.post1087+dirty, hypothesis-6.126.0
collected 1 item

tests/test_fileio.py F                                                                                                                               [100%]

========================================================================= FAILURES =========================================================================
______________________________________________________________ TestPath.test_iterdir[asyncio] ______________________________________________________________
tests/test_fileio.py:409: in test_iterdir
    async for path in Path(populated_tmpdir).iterdir():
src/anyio/_core/_fileio.py:548: in iterdir
    gen = self._path.iterdir()
../../.pyenv/versions/3.13.2/lib/python3.13/pathlib/_local.py:576: in iterdir
    paths = [entry.path for entry in scandir_it]
.venv/lib/python3.13/site-packages/blockbuster/blockbuster.py:109: in wrapper
    raise BlockingError(func_name)
E   blockbuster.blockbuster.BlockingError: Blocking call to ScandirIterator.__next__
```

tests/test_to_process.py
```diff
@@ -73,6 +73,7 @@ async def test_cancel_before() -> None:
     pytest.raises(LookupError, to_process._process_pool_workers.get)
 
 
+@pytest.mark.usefixtures("deactivate_blockbuster")
 async def test_cancel_during() -> None:
```
cbornet, line 77: There's an error when calling 
```sh
  | exceptiongroup.ExceptionGroup: unhandled errors in a TaskGroup (1 sub-exception)
    |   File "/home/runner/work/anyio/anyio/src/anyio/to_process.py", line 95, in run_sync
    |     request = pickle.dumps(("run", func, args), protocol=pickle.HIGHEST_PROTOCOL)
    | AttributeError: Can't pickle local object '_wrap_blocking.<locals>.wrapper'
```

`_wrap_blocking.<locals>.wrapper` is a blockbuster function (the one that wraps the framework core functions).
Not sure what to do so just deactivating blockbuster for this test.

tests/test_from_thread.py
```diff
@@ -628,6 +628,7 @@ async def raise_baseexception() -> None:
             assert exc.value.__context__ is None
 
     @pytest.mark.parametrize("portal_backend_name", get_all_backends())
+    @pytest.mark.usefixtures("deactivate_blockbuster")
```
cbornet, line 631: That could also be a marker. LMK what you prefer.

tests/test_from_thread.py
smurfix: > At some point run_sync_from_thread is called which calls the blocking result() on a future

This is a context manager that's expected to run from a sync context. The test's conditions shouldn't even occur in normal usage.

Also, the fact that it "blocks" is a feature, because it tests for a regression that might cause a deadlock. See the comment at the test's beginning.

Thus, not a problem. Just deactivate the blockbuster.

tests/test_from_thread.py
agronholm: Nah, I don't see any real benefit from that.

agronholm:
There are test failures still. I'll take another look after you've fixed them.

agronholm:
> There are test failures still. I'll take another look after you've fixed them.

Oh, sorry, these are caused by the calls you asked about. Never mind.

tests/conftest.py
agronholm: While I usually like being pedantic about these things, I feel that in 99.999% of cases blocking will be a non-issue, and using threads will degrade performance. So add the exemptions please.

agronholm:
By the way, I suggest you refrain from force-pushing. It makes it harder to track the incremental changes you're making.

agronholm:
Are you sure you're not adding exemptions for legit errors reported by this tool?

cbornet:
> By the way, I suggest you refrain from force-pushing. It makes it harder to track the incremental changes you're making.

Sorry about that. Some commit/pushs I have done were to figure out how to fix. I was using the CI to test because I don't have a Windows machine to test locally.

cbornet:
> Are you sure you're not adding exemptions for legit errors reported by this tool?

Yes. These errors are maybe legit reports. I've put them as exemptions for now so we can discuss how to fix them.

tests/conftest.py
```diff
@@ -52,6 +53,27 @@ def eager_task_loop_factory() -> asyncio.AbstractEventLoop:
...
 
 
+@pytest.fixture(autouse=True)
+def blockbuster() -> Iterator[BlockBuster]:
+    with blockbuster_ctx(
+        "anyio", excluded_modules=["anyio.pytest_plugin", "anyio._backends._asyncio"]
+    ) as bb:
+        bb.functions["os.stat"].can_block_in("anyio/streams/tls.py", "wrap")
+        bb.functions["socket.socket.accept"].can_block_in(
+            "anyio/_core/_asyncio_selector_thread.py", {"get_selector"}
```
cbornet, line 63: `get_selector` initiates a `Selector()` which creates a socket pair.
Windows doesn't have a `socketpair` syscall so it it emulated by `socket._fallback_socketpair`.
[_fallback_socketpair](https://github.com/python/cpython/blob/86dab0e67e7e75c4c77e994b99a842f033ddbe46/Lib/socket.py#L598) does a `listen` on a server socket, creates a non-blocking client socket and connects to the server, then calls a blocking `accept` on the server socket. The accept wait time should be very short since the client socket is connecting but I'm not sure how long it is or if it's a problem.
Also the `Selector` created by `get_selector` is cached.
So maybe it can be kept as-is ?

agronholm: I don't see this as a problem, so let's just keep this exemption.

tests/conftest.py
```diff
@@ -52,6 +53,27 @@ def eager_task_loop_factory() -> asyncio.AbstractEventLoop:
     )
 
 
+@pytest.fixture(autouse=True)
+def blockbuster() -> Iterator[BlockBuster]:
+    with blockbuster_ctx(
+        "anyio", excluded_modules=["anyio.pytest_plugin", "anyio._backends._asyncio"]
+    ) as bb:
+        bb.functions["os.stat"].can_block_in("anyio/streams/tls.py", "wrap")
```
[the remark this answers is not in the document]
agronholm: I think Trio also defers this to a thread so it might be warranted here as well.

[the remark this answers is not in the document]
cbornet: Done in https://github.com/agronholm/anyio/pull/875/commits/27f8d566c6b1c8dc0276a477b0db71267bb6ce54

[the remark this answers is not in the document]
agronholm: Also why is `ssl.SSLContext` special cased? Won't that sometimes do file I/O too?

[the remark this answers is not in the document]
cbornet: > Your use of asyncio.to_thread() led to the failing of the test suite on Trio.

Of course! Bad habit... 😅 

> Also why is ssl.SSLContext special cased? Won't that sometimes do file I/O too?

I think it's OK at least for CPython: https://github.com/python/cpython/blob/3.13/Modules/_ssl.c#L4451
But if you prefer we can set to always defer to thread.

agronholm:
Thanks!

cbornet:
Awesome, thanks !

tests/conftest.py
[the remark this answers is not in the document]
graingert: ssl.create_default_context should probably be deferred to a thread (when it's a client side purpose)

PR #878, opened 2025-02-24
Feature/wrap socket

### LINKED ISSUE #845: add the ability to wrap stdlib sockets
### Things to check first

### Feature description

I'd like the ability to wrap a stdlib socket in a SocketStream/UNIXSocketStream/SocketListener/UDPSocket/ConnectedUDPSocket

Ideally it would be a synchronous function that takes a connected or listening socket of any kind and does the right thing, but asyncio only provides async methods to adopt a socket. It's also hard to detect if a socket is connected or connecting or listening securely for the purposes of wrapping with TLS

Perhaps the best interface for now would be two async functions, `async def wrap_client_sock -> SocketStream | UNIXSocketStream | ConnectedUDPSocket` and `async def wrap_server_sock -> SocketListener | UDPSocket` ?

### Use case

I plan to use it with socket.socketpair in the anyio.to_process.run_sync implementation to fix #516

leon1995:
## Changes

Fixes #845 

## Checklist

features)

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -89,8 +89,10 @@ jobs:
         allow-prereleases: true
         cache: pip
         cache-dependency-path: pyproject.toml
-    - name: Install dependencies
-      run: pip install -e .[test]
+    - name: Ensure pip >= v25.1
+      run: python -m pip install "pip >= 25.1"
+    - name: Install the project and its dependencies
+      run: pip install --group test -e .
     - name: Patch /etc/hosts
       if: runner.os != 'Windows'
       run: |
@@ -132,8 +134,10 @@ jobs:
         python-version: "3.11"
         cache: pip
         cache-dependency-path: pyproject.toml
-    - name: Install dependencies
-      run: pip install -e .[doc]
+    - name: Ensure pip >= v25.1
+      run: python -m pip install "pip >= 25.1"
+    - name: Install the project and its dependencies
+      run: pip install --group doc -e .
     - name: Build documentation
       run: sphinx-build -W docs build/sphinx
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,7 +5,7 @@
 # * Run "pre-commit install".
 repos:
   - repo: https://github.com/adrienverge/yamllint
-    rev: v1.37.0
+    rev: v1.37.1
     hooks:
       - id: yamllint
         args: ['-d {extends: relaxed, rules: {line-length: disable}}', '-s']
@@ -29,7 +29,7 @@ repos:
           - tomli
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.11.6
+    rev: v0.11.10
     hooks:
       - id: ruff
         args: [--fix, --show-fixes]
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -4,13 +4,11 @@ build:
   os: ubuntu-22.04
   tools:
     python: "3.11"
+  jobs:
+    install:
+      - python -m pip install --no-cache-dir "pip >= 25.1"
+      - python -m pip install --upgrade --upgrade-strategy only-if-needed --no-cache-dir --group doc .
 
 sphinx:
   configuration: docs/conf.py
   fail_on_warning: true
-
-python:
-  install:
-    - method: pip
-      path: .
-      extra_requirements: [doc]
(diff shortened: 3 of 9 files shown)


leon1995:
I dont know why mypy is complaining here because I just copied the call from the `connted_tcp` method (and swapped the parameters for the socket)

src/anyio/_backends/_asyncio.py
```diff
@@ -2646,6 +2646,43 @@ async def create_udp_socket(
...
+        cls, raw_socket: socket.socket
+    ) -> SocketStream | ConnectedUDPSocket:
+        if raw_socket.type == socket.SocketKind.SOCK_DGRAM:
+            transport, protocol = await get_running_loop().create_datagram_endpoint(
+                DatagramProtocol,
+                sock=raw_socket,
+            )
+            return ConnectedUDPSocket(transport, protocol)
+        elif raw_socket.type == socket.SocketKind.SOCK_STREAM:
+            transport, protocol = cast(  # type: ignore[assignment]  # mypy is not able to properly infer the type on unix
```
graingert, line 2660: will it work if you use:

```python
    SocketStream(
        *(
            await get_running_loop().create_connection(
                StreamProtocol,
                sock=raw_socket,
            )
        )
    )
```

tests/test_sockets.py
```diff
@@ -1526,6 +1530,65 @@ async def test_send_after_close(self, family: AnyIPAddressFamily) -> None:
             await udp.send(b"foo")
 
 
+@pytest.mark.network
+@pytest.mark.usefixtures("check_asyncio_bug")
```
graingert, line 1534: apparently this is fixed in https://github.com/python/cpython/issues/83329

leon1995: done

src/anyio/_backends/_asyncio.py
leon1995: done

tests/test_sockets.py
graingert: ah it's not actually fixed, I tried it out

leon1995: but the tests pass, so should I add it again or not?

agronholm: That doesn't explain why _all_ the tests pass.

leon1995:
> Before you go any further, I must point out that the notion of "client" UDP socket being always connected is fundamentally wrong, and the same function should not be used to wrap either a TCP socket or a UDP socket either. The notion of "connection" for a UDP socket just means that it has a default target address. An unconnected UDP socket could just as well be a client or a server socket.

What do you suggest then? I picked up the suggested interface from #845

agronholm:
> > Before you go any further, I must point out that the notion of "client" UDP socket being always connected is fundamentally wrong, and the same function should not be used to wrap either a TCP socket or a UDP socket either. The notion of "connection" for a UDP socket just means that it has a default target address. An unconnected UDP socket could just as well be a client or a server socket.
> 
> What do you suggest then? I picked up the suggested interface from #845

Well, that suggestion was also wrong then. I have to think about this for a little bit. Perhaps class methods in the respective socket stream/listener classes would do it? At any rate, let's reach a consensus before proceeding any further.

leon1995:
> > > Before you go any further, I must point out that the notion of "client" UDP socket being always connected is fundamentally wrong, and the same function should not be used to wrap either a TCP socket or a UDP socket either. The notion of "connection" for a UDP socket just means that it has a default target address. An unconnected UDP socket could just as well be a client or a server socket.
> > 
> > 
> > What do you suggest then? I picked up the suggested interface from #845
> 
> Well, that suggestion was also wrong then. I have to think about this for a little bit. Perhaps class methods in the respective socket stream/listener classes would do it? At any rate, let's reach a consensus before proceeding any further.

that might be a good idea. something like
```python
class SocketStream(ByteStream, _SocketProvider):
    """
    Transports bytes over a socket.

    Supports all relevant extra attributes from :class:`~SocketAttribute`.
    """
    @classmethod
    @abstractmethod
    def from_socket(cls, raw_socket: socket.socket) -> SocketStream: ...
```

agronholm:
We can't add new abstract methods to public classes as that would break compatibility for anyone implementing their own `SocketStream` subclass.

agronholm:
I'll have to think about this some more when I get the time and I'm not working on my day job.

leon1995:
> I'll have to think about this some more when I get the time and I'm not working on my day job.

Have you come to any conclusion yet?

agronholm:
I'll try to work on this issue this weekend.

agronholm:
This is still on track for v4.10, but I am not happy with the API. I think the right solution is to add static methods to the appropriate listener/stream interfaces which will then call the appropriate backend-specific method.

agronholm:
I'm working on my own PR to address this need.

PR #886, opened 2025-03-12
cheaper no_other_refs

graingert:
## Changes

Using the new 3.14a5 f_generator attribute for frames.
this is a trivial change, so no changelog or checklist


agronholm (approved):
Nothing looks out of place, and I defer to your expertise on the new 3.14 stuff.

PR #887, opened 2025-03-15
Fix(asyncio): to_thread.async_run holding ref of Context

Wankupi:
If the context had some objects,
after to_thread.async run,
the worker thread will keep one ref of the context,
which will cause the objects can not be released.

## Changes

This commit only fix the asyncio backend.
Create a new empty contextvars.Context when creating a new thread worker.

It should be checked for other backends.

## Checklist

features)

### DIFF
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -975,10 +975,10 @@ class WorkerThread(Thread):
                             self._report_result, future, result, exception
                         )
 
+                    del result, exception
+
                 self.queue.task_done()
-                # make sure no references after this epoch
-                # so that context can be garbage collected
-                del item, context, func, args, future, cancel_scope, result, exception
+                del item, context, func, args, future, cancel_scope
 
     def stop(self, f: asyncio.Task | None = None) -> None:
         self.stopping = True
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import array
 import asyncio
 import concurrent.futures
+import contextvars
 import math
 import os
 import socket
@@ -975,6 +976,9 @@ class WorkerThread(Thread):
                         )
 
                 self.queue.task_done()
+                # make sure no references after this epoch
+                # so that context can be garbage collected
+                del item, context, func, args, future, cancel_scope, result, exception
 
     def stop(self, f: asyncio.Task | None = None) -> None:
         self.stopping = True
@@ -2433,7 +2437,9 @@ class AsyncIOBackend(AsyncBackend):
                     worker = WorkerThread(root_task, workers, idle_workers)
                     worker.start()
                     workers.add(worker)
-                    root_task.add_done_callback(worker.stop)
+                    root_task.add_done_callback(
+                        worker.stop, context=contextvars.Context()
+                    )
                 else:
                     worker = idle_workers.pop()
 
--- a/tests/test_to_thread.py
+++ b/tests/test_to_thread.py
@@ -4,11 +4,10 @@ import asyncio
 import gc
 import threading
 import time
-from collections.abc import Callable, Coroutine
 from concurrent.futures import Future, ThreadPoolExecutor
 from contextvars import ContextVar
 from functools import partial
-from typing import Any, NoReturn, Optional
+from typing import Any, NoReturn
 
 import pytest
 import sniffio
@@ -25,7 +24,7 @@ from anyio import (
 )
 from anyio.from_thread import BlockingPortalProvider
 
-from .conftest import asyncio_params
+from .conftest import asyncio_params, no_other_refs
 
 pytestmark = pytest.mark.anyio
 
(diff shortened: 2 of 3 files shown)


src/anyio/_backends/_asyncio.py
```diff
@@ -975,6 +976,9 @@ def run(self) -> None:
                         )
 
                 self.queue.task_done()
+                # make sure no references after this epoch
+                # so that context can be garbage collected
+                del item, context, func, args, future, cancel_scope, result, exception
```
graingert, line 981: @agronholm I wonder if it would be better to use a NamedTuple for item instead of a regular tuple, then we only need to delete one reference ?

tests/test_to_thread.py
```diff
@@ -360,3 +360,30 @@ def dummy() -> None:
...
+    cvar = ContextVar[Optional[A]]("cvar", default=None)
+
+    def foo() -> None:
+        return None
+
+    async def one_request(t: int, res: list[int]) -> None:
+        a = A(t, res)
+        cvar.set(a)
+        await to_thread.run_sync(foo)
+        cvar.set(None)
```
graingert, line 384: you'll need to run a `gc.collect()` here to get this passing on pypy

tests/test_to_thread.py
```diff
@@ -360,3 +360,30 @@ def dummy() -> None:
...
+
+    cvar = ContextVar[Optional[A]]("cvar", default=None)
+
+    def foo() -> None:
+        return None
+
+    async def one_request(t: int, res: list[int]) -> None:
+        a = A(t, res)
+        cvar.set(a)
+        await to_thread.run_sync(foo)
```
graingert, line 383: your PR also fixes a problem where func and args are kept alive, can you add a test for this also?

graingert:
looks like it hung on `test_raise_baseexception_from_task[asyncio]` I'm retrying the CI

tests/test_to_thread.py
Wankupi: Sure. I already write two more tests about `function` and `args`, it does work.
Howerer for the `result`, current solution fails.
It seems to be the problem of asyncio itself.

tests/test_to_thread.py
Wankupi: Thanks to your suggestion. `gc.collect()` works. Now only the timeout problem in the CI/CD.

src/anyio/_backends/_asyncio.py
graingert: Looks like if the future is cancelled result is not bound maybe?

graingert: Probably we need this: 

```python
                result = None
                exception: BaseException | None = None
                if not future.cancelled():
                    threadlocals.current_cancel_scope = cancel_scope
```

graingert:
Hmm to make result get garbage collected we can pass a boxed [result] and just .pop() to get it. The other option is to run a checkpoint in the tests

graingert:
It's probably worth checking there's no cyclic references in any exceptions raised from run_sync also

src/anyio/_backends/_asyncio.py
```diff
@@ -2433,7 +2437,9 @@ async def run_sync_in_worker_thread(  # type: ignore[return]
                     worker = WorkerThread(root_task, workers, idle_workers)
                     worker.start()
                     workers.add(worker)
-                    root_task.add_done_callback(worker.stop)
+                    root_task.add_done_callback(
+                        worker.stop, context=contextvars.Context()
+                    )
```
agronholm, line 2442: What's this doing here?

src/anyio/_backends/_asyncio.py
agronholm: Or you could just `del result, exception` at the end of the `if` block.

src/anyio/_backends/_asyncio.py
Wankupi: When the first request arrival, and the worker created, the current context is copied by default. This adds the reference of objects in current context.
so it would looks like
1 2 -2 3 -3 4 -4 ... 10 -10 (program stop) -1

agronholm: I think you're answering a different question to what I asked. I was asking why you're using a blank context to call `worker.stop()` which doesn't access any context variables.

Wankupi: The reason using a blank context is, it is clean and have no impact to current context.
I was glad to see `worker.stop()` does not access any context variables. Otherwise it might be wrong to use a plain context. In that case I may have to copy these used variables one by one. So I just use a new Context and let it go.

src/anyio/_backends/_asyncio.py
```diff
@@ -974,7 +975,10 @@ def run(self) -> None:
                             self._report_result, future, result, exception
                         )
 
+                    del result, exception
+
                 self.queue.task_done()
+                del item, context, func, args, future, cancel_scope
```
graingert, line 981: item can be deleted up on line 961 after it's unpacked

agronholm: That just adds another code line with no benefit?

src/anyio/_backends/_asyncio.py
agronholm: We talked it over on Gitter and I now understand why you're doing it.

agronholm:
This needs a changelog entry.

agronholm:
The substance of this patch looks fine now, but I don't like the tests. I'll refactor them soon.

agronholm:
@graingert The test I replaced the OP's tests fails without the patch. Let me know if you think it's inadequate.

tests/test_to_thread.py
```diff
@@ -364,59 +363,18 @@ def dummy() -> None:
...
-    async def one_request(t: int, res: list[int]) -> None:
-        a = LifecycleIndicator(t, res)
-        await to_thread.run_sync(foo, a)
+async def test_run_sync_worker_cyclic_references() -> None:
+    def foo(_: str) -> None:
+        pass
 
-    await lifecycle_tester(one_request, [1, -1, 2, -2, 3, -3, 4, -4, 5, -5])
+    cvar = ContextVar[str]("cvar")
+    contextval = "ffaweaw"
```
graingert, line 371: I'd use a user defined class for this and arg:

```python
class Example:
    pass

contextval = Example()
arg = Example()
```

graingert (approved):

PR #888, opened 2025-03-15
Test selector thread with closed socket

davidbrochart:
## Changes

Fixes #. 

## Checklist

features)

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -6,8 +6,41 @@ on:
   pull_request:
 
 jobs:
+  changed-files:
+    runs-on: ubuntu-latest
+    outputs:
+      workflow-changed: ${{ steps.changed-files.outputs.workflow_any_changed }}
+      pyproject-changed: ${{ steps.changed-files.outputs.src_any_changed }}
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -60,13 +60,19 @@ jobs:
...
         - os: macos-latest
           python-version: "3.9"
         - os: macos-latest
           python-version: "3.13"
+        - os: windows-latest
+          python-version: "3.9"
+        - os: windows-latest
+          python-version: "3.13"
+        - os: windows-latest
+          python-version: pypy-3.11
```
agronholm, line 75: I may veto this as it has caused too many issues in the past.

davidbrochart: Yeah I just wanted to see if I can trigger the bug more easily on PyPy.

davidbrochart: Anyway it seems the test suite won't run if there is no file change?

agronholm: You can omit that from the test workflow for the time being. Just put it back when this is ready for review.

PR #890, opened 2025-03-16
Added concurrency helpers to retrieve task results

agronholm:
## Changes

Adds an enhanced version of the task group that allows task-by-task cancellation as well as awaiting on the results of individual tasks. Two other convenience functions are also provided:

- `amap()`: calls the given one-parameter coroutine function with each item from the given iterable of arguments and runs them concurrently in a task group
- `race()` launches all given coroutines as tasks in a task group and returns the return value of whichever task completes first

Concurrency and rate limiting is provided by both functions.

## Checklist

features)

these instructions.

### Updating the changelog

```
```


src/anyio/_core/_tasks.py
```diff
@@ -156,3 +180,248 @@ def create_task_group() -> TaskGroup:
 
     """
     return get_async_backend().create_task_group()
+
+
+class TaskHandle(Generic[T]):
```
davidbrochart, line 185: It looks like a `TaskHandle` could be used to implement a `create_future()` on the `EnhancedTaskGroup`, to have something like asyncio's `Future`.
What do you think?

davidbrochart:
A bit far-fetched, but what do you think of a free function `create_task()` that would use the current `EnhancedTaskGroup` if there is one, and errors out otherwise?
On one hand this is going against structured concurrency, but on the other hand it removes the need to pass a task group down the stack, when you know there must be one.

src/anyio/_core/_tasks.py
```diff
@@ -156,3 +180,248 @@ def create_task_group() -> TaskGroup:
...
+
+    def set_exception(self, exception: BaseException) -> None:
+        self._exception = exception
+        self._event.set()
+
+    def cancel(self) -> None:
+        self.cancel_scope.cancel()
+
+    @property
+    def start_value(self) -> Any:
```
davidbrochart, line 216: It seems that it's not possible to await the start value, should this be an async function instead?

src/anyio/_core/_tasks.py
```diff
@@ -156,3 +180,248 @@ def create_task_group() -> TaskGroup:
...
+
+    async def start(
+        self,
+        func: Callable[[Unpack[PosArgsT]], Coroutine[Any, Any, T]],
+        /,
+        *args: Unpack[PosArgsT],
+        kwargs: Mapping[str, Any] | None = None,
+    ) -> TaskHandle[T]:
+        handle = TaskHandle[T]()
+        handle._start_value = await self._task_group.start(
```
davidbrochart, line 284: If we want to be able to await the start value in the `TaskHandle`, I guess we should wrap `start()` with a `create_task()`?

smurfix:
> it removes the need to pass a task group down the stack,

On the other hand, that's the point of passing down the taskgroup -- when it's not there we know that the method we're calling won't start things it doesn't wait for.

If you want to circumvent that, for whatever reason, there's already a fairly-high-performance way to do it -- set a contextvar. So IMHO "explicit is better than implicit" and thus we shouldn't support that natively.

src/anyio/_core/_tasks.py
agronholm: I don't understand what you mean. The start value will already be available in the `TaskHandle` once `start()` returns.

agronholm:
> A bit far-fetched, but what do you think of a free function `create_task()` that would use the current `EnhancedTaskGroup` if there is one, and errors out otherwise? On one hand this is going against structured concurrency, but on the other hand it removes the need to pass a task group down the stack, when you know there must be one.

I'm -1 on an implicit task group.

src/anyio/_core/_tasks.py
agronholm: What exact addition are you suggesting?

davidbrochart: I was going to say something like that:
```py
class EnhancedTaskGroup:
    def create_future(self) -> TaskHandle[T]:
        handle = TaskHandle[T]()
        return handle
```
But now I realize a `Future` is just a `TaskHandle`, so there's nothing to do?

src/anyio/_core/_tasks.py
davidbrochart: Ah indeed :+1:

src/anyio/_core/_tasks.py
davidbrochart: My mistake, see https://github.com/agronholm/anyio/pull/890/files#r2006131909.

src/anyio/_core/_tasks.py
davidbrochart: I thought that the `TaskHandle` was returned immediately, but it's returned only when the task has started. Which means we cannot e.g. cancel the task before it has started. Not sure if that should be allowed?

dhirschfeld:
The one thing I'm after is a nursery/TaskGroup where I can start tasks and then iterate over the results asynchronously *as soon as they become available* so that I can interleave work and don't have to wait for the slowest task to start the next part of the pipeline.

With this API, `race` is trivial to implement - you just cancel the TaskGroup as soon as the first task is made available. `gather` is just collect all results in a list of length `n_inputs` in the order they were provided then return the results list when the TaskGroup exits.

dhirschfeld:
> > With this API, `race` is trivial to implement
> 
> And how do we deal with exceptions occurring in the child tasks?

In my implementation I actually return `Outcome` instances, so the user processes the `Outcome`'s as they're made available. It's then up to the user to decide what to do with any errors. If they blindly `unwrap` an error it will raise and cancel the TaskGroup.

agronholm:
Ok, so help me understand. What do you suggest `race()` returns then? What if the first result from a child task is an exception? Do you want to return the Outcome (or equivalent) of that?

dhirschfeld:
I think that depends on the use-case. If `amap` returned an asynchronous iterator that returned Outcome wrapped results as soon as they were ready I'd leave it at that and let users implement `race` themselves with the semantics that made sense for their problem.

Given the `amap` primitive, `race` becomes trivial to implement however you want so there's no need to provide an implementation which may not be ideal for all use-cases.

agronholm:
> I think that depends on the use-case. If `amap` returned an asynchronous iterator that returned Outcome wrapped results as soon as they were ready I'd leave it at that and let users implement `race` themselves with the semantics that made sense for their problem.
> 
> Given the `amap` primitive, `race` becomes trivial to implement however you want so there's no need to provide an implementation which may not be ideal for all use-cases.

But `as_completed()` does exactly that, doesn't it? I don't think `amap()` is suitable for implementing `race()`.

dhirschfeld:
> > I think that depends on the use-case. If `amap` returned an asynchronous iterator that returned Outcome wrapped results as soon as they were ready I'd leave it at that and let users implement `race` themselves with the semantics that made sense for their problem.
> > Given the `amap` primitive, `race` becomes trivial to implement however you want so there's no need to provide an implementation which may not be ideal for all use-cases.
> 
> But `as_completed()` does exactly that (with TaskHandles), doesn't it? I don't think `amap()` is suitable for implementing `race()`.

Ah yep, I missed the `as_completed` implementation - that does look like it does what I'm after!

dhirschfeld:
I'm actually pretty excited about this functionality as it will let me replace a bunch of custom code, so I'm curious if there are plans to land this in a release sometime soonish?

agronholm:
> I'm actually pretty excited about this functionality as it will let me replace a bunch of custom code, so I'm curious if there are plans to land this in a release sometime soonish?

I would like to get a fairly broad consensus on the API before moving forward with it. That means getting more people involved.

agronholm:
@Zac-HD I would like your 2 cents on this. Do you think the API would work as-is, or need changes?

Zac-HD:
Unfortunately there are a couple of parts of the design that make me pretty nervous.

- `TaskHandle` is very much like a future.  I worry about dropping captured exceptions if nobody awaits the handle; and in cases where that's certain not to happen I'm unclear on what it adds.  Capturing and later raising exceptions also makes tracebacks much less useful, which would be very very painful in large codebases (eg at work)
- `EnhancedTaskGroup` currently leans heavily on `TaskHandle`; if it didn't I'd still want to avoid interfaces which accept a coroutine object.
- `amap` and `as_completed` seem like useful helpers, though my instinct would be to implement them directly on top of a TaskGroup + cloneable streams.

agronholm:
> TaskHandle is very much like a future. I worry about dropping captured exceptions if nobody awaits the handle; and in cases where that's certain not to happen I'm unclear on what it adds. Capturing and later raising exceptions also

I wonder where the misunderstanding is coming from. Are you thinking that if the task handle isn't saved, the exception is silently dropped? No, `EnhancedTaskGroup` still enforces SC, just like `TaskGroup`. But it allows for the return value of a task to be retrieved without awkward user-side wrapping, unlike with normal AnyIO task groups or Trio nurseries. Thus I feel `EnhancedTaskGroup` combines the best features of AnyIO and AsyncIO task groups.

dhirschfeld:
I know Trio makes a point about not exposing coroutines to users. I'm not across the arguments *why*, but most things in Trio are very well thought out and very well designed.

In my own implementation I take zero-argument callables, so I force users to use `partial`. One benefit of this is that I can also take sync functions and run them using `anyio.to_thread.run_sync`

Zac-HD:
> > TaskHandle is very much like a future. I worry about dropping captured exceptions if nobody awaits the handle; and in cases where that's certain not to happen I'm unclear on what it adds. Capturing and later raising exceptions also
> 
> I wonder where the misunderstanding is coming from. Are you thinking that if the task handle isn't saved, the exception is silently dropped? No, `EnhancedTaskGroup` still enforces SC, just like `TaskGroup`. But it allows for the return value of a task to be retrieved without awkward user-side wrapping, unlike with normal AnyIO task groups or Trio nurseries. Thus I feel `EnhancedTaskGroup` combines the best features of AnyIO and AsyncIO task groups.

You're right of course!  I'd be more comfortable if `TaskHandle` didn't capture the task exception though - that'll already be raised from the TaskGroup, and so I'd have a `.wait()` method return the task return value or raise a `ValueError("task did not produce a return value")`-ish error.  Raising an exception object in two places just feels like a recipe for trouble with the context or traceback attributes (c.f. PEP-785).

agronholm:
> > > TaskHandle is very much like a future. I worry about dropping captured exceptions if nobody awaits the handle; and in cases where that's certain not to happen I'm unclear on what it adds. Capturing and later raising exceptions also
> > 
> > 
> > I wonder where the misunderstanding is coming from. Are you thinking that if the task handle isn't saved, the exception is silently dropped? No, `EnhancedTaskGroup` still enforces SC, just like `TaskGroup`. But it allows for the return value of a task to be retrieved without awkward user-side wrapping, unlike with normal AnyIO task groups or Trio nurseries. Thus I feel `EnhancedTaskGroup` combines the best features of AnyIO and AsyncIO task groups.
> 
> You're right of course! I'd be more comfortable if `TaskHandle` didn't capture the task exception though - that'll already be raised from the TaskGroup, and so I'd have a `.wait()` method return the task return value or raise a `ValueError("task did not produce a return value")`-ish error. Raising an exception object in two places just feels like a recipe for trouble with the context or traceback attributes (c.f. PEP-785).

If `TaskHandle` didn't capture the exception, what would happen if a task that isn't the task group's host task awaits for the task and the awaited task raises an exception?

agronholm:
> > > TaskHandle is very much like a future. I worry about dropping captured exceptions if nobody awaits the handle; and in cases where that's certain not to happen I'm unclear on what it adds. Capturing and later raising exceptions also
> > 
> > 
> > I wonder where the misunderstanding is coming from. Are you thinking that if the task handle isn't saved, the exception is silently dropped? No, `EnhancedTaskGroup` still enforces SC, just like `TaskGroup`. But it allows for the return value of a task to be retrieved without awkward user-side wrapping, unlike with normal AnyIO task groups or Trio nurseries. Thus I feel `EnhancedTaskGroup` combines the best features of AnyIO and AsyncIO task groups.
> 
> You're right of course! I'd be more comfortable if `TaskHandle` didn't capture the task exception though - that'll already be raised from the TaskGroup, and so I'd have a `.wait()` method return the task return value or raise a `ValueError("task did not produce a return value")`-ish error. Raising an exception object in two places just feels like a recipe for trouble with the context or traceback attributes (c.f. PEP-785).

I know that asyncio allows this. How do they deal with it?

agronholm:
```py
import asyncio
from asyncio import TaskGroup

async def fail():
    await asyncio.sleep(0.1)
    raise RuntimeError("foo")

async def main():
    async with TaskGroup() as tg:
        task = tg.create_task(fail())
        await asyncio.sleep(0.05)
        await task

asyncio.run(main())
```
Results in:
```pytb
  |   File "/tmp/a.py", line 11, in <module>
  |     asyncio.run(main())
  |     ~~~~~~~~~~~^^^^^^^^
  |   File "/usr/lib64/python3.13/asyncio/runners.py", line 195, in run
  |     return runner.run(main)
  |            ~~~~~~~~~~^^^^^^
  |   File "/usr/lib64/python3.13/asyncio/runners.py", line 118, in run
  |            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^
  |   File "/usr/lib64/python3.13/asyncio/base_events.py", line 719, in run_until_complete
  |     return future.result()
  |            ~~~~~~~~~~~~~^^
  |   File "/tmp/a.py", line 8, in main
  |     async with TaskGroup() as tg:
  |                ~~~~~~~~~^^
  |   File "/usr/lib64/python3.13/asyncio/taskgroups.py", line 71, in __aexit__
  |     return await self._aexit(et, exc)
  |            ^^^^^^^^^^^^^^^^^^^^^^^^^^
  |   File "/usr/lib64/python3.13/asyncio/taskgroups.py", line 173, in _aexit
  |     ...<2 lines>...
  |     ) from None
    |   File "/tmp/a.py", line 5, in fail
    |     raise RuntimeError("foo")
    | RuntimeError: foo
```

agronholm:
On asyncio, the `await` hits a `CancelledError` rather than `RuntimeError`, hence the one-exception group.

Graeme22:
This seems similar to the functionality provided by [aiometer](https://github.com/florimondmanca/aiometer/tree/master), which would be nice behavior to have by default. Would some of the other functions built on `amap` like `run_on_each` and `run_all` also add some utility here?

Also, if `EnhancedTaskGroup` provides the same SC guarantees, what is the reason for implementing this as a separate `EnhancedTaskGroup` class instead of modifying the existing `TaskGroup` class? Just backwards compatibility? Which of the features are not opt-in?

src/anyio/_core/_tasks.py
```diff
@@ -161,3 +185,359 @@ def create_task_group() -> TaskGroup:
...
+
+class AwaitedTaskCancelled(Exception):
+    """
+    Raised when awaiting on a :class:`TaskHandle` which was cancelled.
+
+    This exception class exists in order to differentiate between the cancellation of
+    the host task (the one awaiting on a task) and the cancellation of the task it's
+    awaiting on. Additionally, raising :class:`asyncio.CancelledError` when waiting on
+    a task would potentially cause cancellation`counters (Python 3.11 and later) to
+    """
```
m-u-xyz, line 223: Something seems to be missing here

src/anyio/_core/_tasks.py
```diff
@@ -161,3 +185,359 @@ def create_task_group() -> TaskGroup:
...
+        self._event = Event()
+        self._exception: BaseException | None = None
+        self._awaited = False
+
+    def set_return_value(self, value: T) -> None:
+        self._return_value = value
+        self._event.set()
+
+    def set_exception(self, exception: BaseException) -> None:
+        if isinstance(exception, get_cancelled_exc_class()):
```
m-u-xyz: My problem with this code is that, conceptually, a `BaseException` (which is not an `Exception`) is for leaving a scope, if not the whole program, in some controlled way, with the corollary that it shouldn't be swallowed or forwarded because that doesn't make sense in most cases.

As such I would strongly recommend to replace all `BaseException`s, not just cancellations,with some `Exception` subclass. (Doesn't have to be the same one.)

I created a PR for this: https://github.com/agronholm/anyio/pull/982

src/anyio/_core/_tasks.py
agronholm: Futures are a lower level primitive, not tied to any task, unlike `TaskHandle`. And a `Future` reports the exception it receives verbatim while `TaskHandle` handles base exceptions in a special manner.

src/anyio/_core/_tasks.py
```diff
@@ -161,3 +185,377 @@ def create_task_group() -> TaskGroup:
...
+        return self
+
+    async def __anext__(self) -> TaskHandle[Any]:
+        if not self._coros and not self._tasks:
+            raise StopAsyncIteration
+
+        return await self._receive.receive()
+
+
+async def amap(
```
Graeme22, line 537: In addition to this, a gather implementation would be super helpful. I have code like this in several libraries using anyio:

```python
T1 = TypeVar("T1")
T2 = TypeVar("T2")
T3 = TypeVar("T3")
T4 = TypeVar("T4")
T5 = TypeVar("T5")


@overload
async def gather(
    awaitable1: Awaitable[T1],
    awaitable2: Awaitable[T2],
    /,
) -> tuple[T1, T2]: ...


@overload
async def gather(
    awaitable1: Awaitable[T1],
    awaitable2: Awaitable[T2],
    awaitable3: Awaitable[T3],
    /,
) -> tuple[T1, T2, T3]: ...


@overload
async def gather(
    awaitable1: Awaitable[T1],
    awaitable2: Awaitable[T2],
    awaitable3: Awaitable[T3],
    awaitable4: Awaitable[T4],
    /,
) -> tuple[T1, T2, T3, T4]: ...


@overload
async def gather(
    awaitable1: Awaitable[T1],
    awaitable2: Awaitable[T2],
    awaitable3: Awaitable[T3],
    awaitable4: Awaitable[T4],
    awaitable5: Awaitable[T5],
    /,
) -> tuple[T1, T2, T3, T4, T5]: ...


@overload
async def gather(*awaitables: Awaitable[T1]) -> tuple[T1, ...]: ...


async def gather(*awaitables: Awaitable[Any]) -> tuple[Any, ...]:
    results: list[Any] = [None] * len(awaitables)

    async def runner(awaitable: Awaitable[Any], i: int) -> None:
        results[i] = await awaitable

        for i, awaitable in enumerate(awaitables):
            tg.start_soon(runner, awaitable, i)
    return tuple(results)
```

The important difference from something like aiometer's `run_all` being that you can mix things with different return types and maintain type safety.

smurfix:
Maybe `task_start` isn't too similar then.

I'm equally wary of a method that returns either X or Y depending on the class it's invoked on. While ETG and TG are currently not descended from a common subclass, ducktyping them is going to be … interesting.

src/anyio/_core/_tasks.py
```diff
@@ -161,3 +185,377 @@ def create_task_group() -> TaskGroup:
...
+                "'await tg.start(...)`)"
+            ) from None
+
+    def __await__(self) -> Generator[Any, None, T]:
+        if self._awaited:
+            raise RuntimeError(
+                f"{self.__class__.__qualname__} cannot be awaited multiple times"
+            )
+
+        if not self._event.is_set():
```
m-u-xyz, line 308: This sequence does not prevent two parallel callers to succeed when they both need to wait.
Suggested change:

        if _awaited: raise Owchie()
        _awaited = True
        try: wait
        except BaseException:
          _awaited = False
          raise

agronholm: Fixed.

src/anyio/_core/_tasks.py
Kludex: It would be nice to have this here.

agronholm:
Some kind of `gather()` implementation would be nice, as it's something that comes up quite often. I'm somewhat hesitant to name it that way, however, as the semantics will necessarily differ from `asyncio.gather()`.

agronholm:
One idea would be to provide these helper functions on their own without making `EnhancedTaskGroup` public, at least until we can come to an agreement about its API and other unsettled questions.

agronholm:
> I think users coming from asyncio are quite surprised to not be able to await the result of a task, or cancel it individually. I am personally eager to have this feature in.

Me too, but we need to settle the question of the API design. Whatever gets into the API will have to be supported for the foreseeable future which is why it's important to get this right.

agronholm:
> > Some kind of gather() implementation would be nice
> 
> At first glance, yes. However, that'd require us to expose `Awaitable`s to our users, which so far we've been avoiding rather studiously (as has Trio).
> 
> I'm _very_ reluctant to add any code to anyio that requires
> 
> ```
> awaitable = foo()
> [async] with bar:
>     await awaitable
> ```
> 
> (as an asyncio-style `gather` necessarily does). This pattern opens a big can of worms when `foo` is a sync function that happens to end with `return asyncfn()`.

I'm not sure I follow. Why would a `gather()` require an ACM to be used by the user?

agronholm:
> > Why would a gather() require an ACM to be used by the user?
> 
> It doesn't. The `gather()` implementation itself needs to use the ACM. The user simply writes code which depends on the call to a declared-as-async method to do nothing unless `await`ed. Or uses a library with a sync wrapper.
> 
> By the way: that includes not raising an exception, which is its own can of not-quite-deterministic worms when `gather`ing async results. Collecting the result of a bunch of callables is easy, you call `start_soon` on all of them and _always_ get a nice ExceptionGroup back if anything bad happens, including `TypeError` when one of the calls is wrong. Not so with asyncio's `gather`, which may result in a single exception plus a heap of random `RuntimeWarning: coroutine 'foo' was never awaited` messages instead. Your typical error handling code might not like that ambiguity.

I don't understand. You are implying that we would ignore exceptions? Under what circumstances would we ever do that?

smurfix:
> You are implying that we would ignore exceptions? 

Nope, I'm not. I'm implying that errors when calling a `gather`ed bunch of asyncio Awaitables may (and frequently do) result in either an ExceptionGroup, or (if the error occurs in the sync part of the function call, including `TypeError` when the call parameters don't match) a naked exception plus any number of ResourceWarnings.

Consider
```
async def foo(): pass
async def bar(wants_a:str): pass
async def baz(): pass

async def gather(): … # copy from above
async def main():
    await gather(foo(), bar(), baz())
anyio.run(main)
```
This will raise a `TypeError` (and emit a single spurious ResourceWarning for `foo`). **But** if `bar` changes to
```
async def bar(wants_a:str=None):
    if wants_a is None:
        raise TypeError("Hey I want that string")
```
which arguably is the exact same thing, semantically, you now get an ExceptionGroup instead.

We (and/or Trio) decided not to collapse single-exception ExceptionGroups for a reason … and IMHO that reason still holds here.

Thus I'm strongly in favor of not supplying any such thing as `gather`. IMHO it's not too onerous to use clean code, i.e.
```
async def gather(*funcs: Callable[[], Awaitable[Any]]) -> tuple[Any, ...]:
    results: list[Any] = [None] * len(funcs)

    async def runner(callable: Callable[[], Awaitable[Any]], i: int) -> None:
        results[i] = await callable()

        for i, callable in enumerate(callables):
            tg.start_soon(runner, callable, i)
    return tuple(results)

from functools import partial as fp
…
    await gather(foo, fp(bar, "Hello the string"), fp(baz))
```
instead. (Intentionally not renaming `gather` here, though I'm kindof +0 on using `collect` instead).

We can expand on that by _optionally_ allowing for a mix of sync and async functions (yes I know it's bad design, but so is requiring people to write an async wrapper for methods like `dict.set` when they design a state machine; been there done that) and/or use `Outcome` to store the results so the caller can handle partial errors as it sees fit instead of aborting the whole thing.

NB: just looking at asyncio's implementation of "gather" gives me the creeps. 100 lines instead of eight … I'm *so* glad we have decent taskgroups, enhanced or not.

agronholm:
Perhaps it'd be best if I added my own version of `gather()` to this PR. That would give us a better basis for a continued discussion.

agronholm:
OK, I just pushed my first shot at that.

src/anyio/_core/_tasks.py
```diff
@@ -565,7 +565,7 @@ async def as_completed(
         yield iterator
 
 
-async def gather(*coros: Coroutine[Any, Any, Any]) -> Sequence[Any]:
+async def gather(*coros: Coroutine[Any, Any, T]) -> Sequence[T]:
```
Graeme22, line 578: Eventually overrides will be necessary here to allow mixed return types. Also looks like it returns `list[T]` not `Sequence`

src/anyio/_core/_tasks.py
```diff
@@ -161,3 +185,405 @@ def create_task_group() -> TaskGroup:
...
+    """
+    Run a number of coroutines in a task group and return their results.
+
+    :param coros: coroutine objects to run as tasks
+    :return: a sequence of return values from the coroutines in the same order as they
+        were passed
+
+    """
+    task_handles: list[TaskHandle[T]] = []
```
Graeme22, line 587: It's probably worth checking for length 1 here as an optimization:
```python
    if len(coros) == 1:  # optimize for this case
        return [await coros[0]()]
```

src/anyio/_core/_tasks.py
m-u-xyz: Isn't every `List` a `Sequence`? Hence this won't hurt.

Graeme22: Yes, the point is to be more specific since this returns a list. However, now that I think about it, this should return a tuple.

agronholm: Why?

src/anyio/_core/_tasks.py
agronholm: No. That will raise the naked exception while users would expect an exception group.

agronholm:
> > One idea would be to provide these helper functions on their own without making `EnhancedTaskGroup` public, at least until we can come to an agreement about its API and other unsettled questions.
> 
> If we go this route we don't need EnhancedTaskGroup at all no? Since all the helpers could be built using normal task groups.

You still need some kind of a helper to run these and get the result.

Graeme22:
> You still need some kind of a helper to run these and get the result.

Right, I'm saying it could be a wrapper instead of the enhanced group.

src/anyio/_core/_tasks.py
Graeme22: https://github.com/python/typeshed/blob/main/stdlib/asyncio/tasks.pyi#L229

agronholm:
> > You still need some kind of a helper to run these and get the result.
> 
> Right, I'm saying it could be a wrapper instead of the enhanced group.

The enhanced group *IS* the wrapper really.

agronholm:
> > > > You still need some kind of a helper to run these and get the result.
> > > 
> > > 
> > > Right, I'm saying it could be a wrapper instead of the enhanced group.
> > 
> > 
> > The enhanced group _IS_ the wrapper really.
> 
> Fair enough, I just don't like the idea of exposing two different kinds of task groups to users.

The new task group class would not be exposed to users if we go with my suggestion of only exposing the helper functions.

src/anyio/_core/_tasks.py
agronholm: `asyncio.gather()` also always returns a list. The annotations are therefore deceptive.

m-u-xyz: Also frankly what's the point of converting the result to a tuple? There's no requirement for it to be read-only. In fact if the caller then needs to mangle the result somehow they'll need to convert it right back.

Graeme22: I would rather the types be consistent, but that's just my preference I guess

agronholm: Copying the types from `typeshed` would make them inconsistent.

agronholm: What I'm trying to say is that what I have here is the best we can reasonably do with Python's limited type system. What `typeshed` has is awkward at best and plain wrong at worst.

agronholm:
I think what was missed here is that `aiometer` provides not only capacity limiting, but time-based rate limiting as well. Does it make sense to bake into task groups like it does, or would another primitive work as well?

smurfix:
Time-based sequencing/rate limiting is easy, just take a semaphore and increment it with a timer task. You still need a well-implemented CapacityLimiter, to prevent accumulation of blocked tasks – but when you have that, adding a rate limit is non-intrusive and easy enough to not get wrong.

agronholm:
> Time-based sequencing/rate limiting is easy, just take a semaphore and increment it with a timer task

So basically:
```py
import time

from anyio import AsyncContextManagerMixin, Semaphore, create_task_group, sleep, run

class RateLimiter(AsyncContextManagerMixin):
    def __init__(self, max_per_second: int):
        self._delay = 1 / max_per_second
        self._semaphore = Semaphore(0, max_value=max_per_second)

    @property
    def max_per_second(self) -> int:
        return self._semaphore.max_value

    async def increment(self) -> None:
        while True:
            await sleep(self._delay)
            if self._semaphore.value < self._semaphore.max_value:
                self._semaphore.release()

    @asynccontextmanager
    async def __asynccontextmanager__(self) -> AsyncGenerator[Semaphore]:
            tg.start_soon(self.increment)
            yield self._semaphore
            tg.cancel_scope.cancel()


async def perform_task(num: int, start_time: float, ratelimit: Semaphore) -> None:
    await ratelimit.acquire()
    print(f"[{time.monotonic() - start_time:.6f}] performing task {num}")


async def main(count: int) -> None:
    async with RateLimiter(5) as ratelimit, create_task_group() as tg:
        start_time = time.monotonic()
        for num in range(1, count + 1):
            tg.start_soon(perform_task, num, start_time, ratelimit)

    end_time = time.monotonic()
    print(f"time elapsed: {end_time - start_time} s")
    print(f"total rate: {count / (end_time - start_time)} / s")

run(main, 20)
```

smurfix:
> So basically

Yes, except that I'd pass in delay and max value explicitly, instead of calculating both from a Req/s value. APIs with an allowed sustained rate of one request per minute, but an initial burst allowance of 20 or so, do exist out there. (Think "free weather forecast service").

Also, you want to limit the rate of task creation. As in

```python
# …

async def perform_task(num: int, start_time: float, ratelimit: Semaphore, *, task_status: TaskStatus) -> None:
    await ratelimit.acquire()
    task_status.started()
    print(f"[{time.monotonic() - start_time:.6f}] performing task {num}")

async def main(count: int) -> None:
    async with RateLimiter(5) as ratelimit, create_task_group() as tg:
        start_time = time.monotonic()
        for num in range(1, count + 1):
            await tg.start(perform_task, num, start_time, ratelimit)

```

agronholm:
The nested ACM may not be necessary, as I could just add an `acquire()` method to the `RateLimiter` class itself.

agronholm:
> Yes, except that I'd pass in delay and max value explicitly, instead of calculating both from a Req/s value. APIs with an allowed sustained rate of one request per minute, but an initial burst allowance of 20 or so, do exist out there. (Think "free weather forecast service").

Would you like to propose a change to my v2 API to accommodate this? I'm not confident about implementing this yet.

agronholm:
I think I'll stop spamming this PR and make a separate one for rate limiting, as it's not tied to the improved concurrency API.

agronholm:
For interested parties, here it is: https://github.com/agronholm/anyio/pull/989

agronholm:
Here's what I understand our options are:

1. Expose a new task group class (`EnhancedTaskGroup` or whatever)
Pros:

- It's a new API and we're free to shape it however we like

Cons:

- It's a new API which we have to support for the foreseeable future
- Having two distinct task groups complicates things for users too

2. Expose only the concurrency helpers but not a new task group API
Pros:

- Users get increased convenience without any duplication

Cons:

- The convenience functions don't cater to all use cases
- Only being able to retrieve task results via these concurrency helpers feels odd from an API design perspective

3. Shoehorn task handles into the existing API
Pros:

- Low-level APIs that actually let us use the existing task groups to return task results

Cons:

- Backwards compatibility will be tricky, particularly with the `start()` method which will likely require an extra keyword argument to return a task handle instead of the `task_status.started()` value
- If we mess this up, it will hurt both us and the users

src/anyio/_core/_tasks.py
Graeme22: My proposal is copy the types from typeshed, but then actually return a tuple to be consistent.

agronholm:
My own order of preference is 3, 2, 1. @Graeme22 what objections do you have against option 3? I feel that 2) is only kicking the can down the road. We need a final solution at some point.

Graeme22:
> My own order of preference is 3, 2, 1. @Graeme22 what objections do you have against option 3? I feel that 2) is only kicking the can down the road. We need a final solution at some point.

I don't have any strong objections to 3. If it's likely to be a breaking change anyway, maybe start could just return a `tuple[Any, TaskHandle[T]]`?

Something like this could also be considered, sort of a combination of 1 and 3 that maintains backwards compatibility:
```python
@classmethod
@override
def create_task_group(cls, collect_results: Literal[False] = False) -> TaskGroup: ...

@classmethod
@override
def create_task_group(cls, collect_results: Literal[True]) -> ResultCollectingTaskGroup: ...

class ResultCollectingTaskGroup:
    def start_soon(
        self,
        func: Callable[[Unpack[PosArgsT]], Awaitable[T]],
        *args: Unpack[PosArgsT],
        name: object = None,
    ) -> TaskHandle[T]:

    async def start(
        self,
        func: Callable[..., Awaitable[Any]],
        *args: object,
        name: object = None,
    ) -> tuple[Any, TaskHandle[T]]:
```

Regardless, I think the utility functions should definitely be present (maybe this was implied already idk)

agronholm:
One more thing: `TaskHandle` is already capable of storing the start value of a task, so a tuple is not necessary.

smurfix:
> If it's likely to be a breaking change anyway

There's a difference between "we need to change some detail and there's a deprecation notice and grace period where both work and everything" (like we had with `Event.set` and friends) and "we want to change something major, it's suddenly incompatible and you need to update your complete codebase all at once".

The latter is not happening. Unless it's for a very important overriding reason, which this clearly is not.

agronholm:
What bothers me a lot still is the `start()` mechanism where we cannot establish type safety for the start value. I have been unable to figure out a clearly better alternative for the `task_status.started()` call. The only idea I came up with is a contextvar-based mechanism where you call a free function that sets an event in the contextvar, but there is a downside for that, namely that it would allow starting functions that have no intention of ever calling that free function. Plus it would _still_ not give us type safety.

agronholm:
I have a new branch, `concurrency-helpers` which contains a reimagined implementation of the task group. It deals solely with coroutine objects and offers `start()` like functionality using an async generator rather than a keyword argument, thus offering a type safe way to provide a start value.

agronholm:
> Another advantage of # 3 is (most probably) that it's the least amount of additional work. # 1 will lead to either some code duplication or some additional method calls (including annoying wrappers to catch the results; while the user doesn't see them any more they'll be still there and slow things down). # 2 _definitely_ requires task wrappers.

No matter how we implement it, task wrappers will be needed as there is otherwise no way to extract the return value on Trio.

agronholm:
> > No matter how we implement it, task wrappers will be needed as there is otherwise no way to extract the return value on Trio.
> 
> … assuming the Trio people don't also go that route.
> 
> [python-trio/trio#2633](https://github.com/python-trio/trio/issues/2633)
> 
> Maybe you want to weigh in there.

Frankly I don't think this is going to be resolved upstream any time soon. But if Trio gets this feature, we have the option of changing our implementation behind the scenes. Plus my latest change would _still_ need the wrapper for releasing the concurrency limiter.

Graeme22:
Any updates on this?

Graeme22:
> So the thing is, after an extensive discussion, there was a loose consensus that we wanted to include rate limiting in these enhanced task groups as that is not properly doable from the outside. But that effort got stuck as there was no real consensus on _how_ the rate limiters should work. I have a couple related PRs open. I'm not sure how to get unstuck with this.

Couldn't the rate limiter code be separated out so we can get the task results implementation worked out?

agronholm:
> > So the thing is, after an extensive discussion, there was a loose consensus that we wanted to include rate limiting in these enhanced task groups as that is not properly doable from the outside. But that effort got stuck as there was no real consensus on _how_ the rate limiters should work. I have a couple related PRs open. I'm not sure how to get unstuck with this.
> 
> Couldn't the rate limiter code be separated out so we can get the task results implementation worked out?

If the rate limiters can be added later w/o breaking the API. There is also the possibility of using the rate limiting interface w/o a concrete implementation, though I think that would be an odd thing to provide.

agronholm:
Closing in favor of #1173.

PR #892, opened 2025-03-17
Use `list.extend()` with an async comprehension to create a transformed list

cclauss:
## Changes

These trivial changes only modify test code.

% `ruff rule PERF401 ` # [manual-list-comprehension](https://docs.astral.sh/ruff/rules/manual-list-comprehension/#manual-list-comprehension-perf401)
> List comprehensions are more readable and more performant.
If you're appending to an existing list, use the `.extend()` method instead:

## Checklist

features)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -78,15 +78,18 @@ src = ["src"]
 [tool.ruff.lint]
 extend-select = [
     "ASYNC",        # flake8-async
-    "W",            # pycodestyle warnings
+    "C4",           # flake8-comprehensions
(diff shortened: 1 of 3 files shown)


agronholm:
Why not just add the Ruff rule to `pyproject.toml` then?

cclauss:
% `ruff check --select=PERF --statistics`
```
14	PERF203	try-except-in-loop
```
but will add.

agronholm (approved):

agronholm:
Thanks!

PR #893, opened 2025-03-17
Make blockbuster test dependency optional

### LINKED ISSUE #891: Please make `blockbuster` test dependency optional
### Things to check first

### Feature description

anyio 4.9.0 introduced an unconditional test dependency on a third-party blockbuster package. However, this package is [broken on PyPy](https://github.com/cbornet/blockbuster/issues/47), and [it depends on a dead and unmaintained package on CPython](https://github.com/cbornet/blockbuster/issues/46), without which it is equally broken as on PyPy. Please consider making the new dependency optional, so that the test suite could be run successfully without it.

### Use case

I'm packaging anyio for Gentoo Linux, and running the test suite is an important part of this. This implies that the new release now forces me to package both `blockbuster` and `forbiddenfruit`, for all Python implementations that `anyio` supports. However, both packages as they are pose serious problems: `forbiddenfruit` is unmaintained for years and uses nosetests, which means even packaging it would require us to put a significant maintenance effort going forward. On the other hand, `blockbuster` has limited (if any) PyPy support, and at this point it is unclear if it does anything at all for this test suite.

mgorny:
## Changes

Fixes #891

Make it possible to run the test suite without `blockbuster` installed, by avoiding unguarded imports, and handling `ImportError` within the fixture.

## Checklist

features)

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -5,14 +5,16 @@ import ssl
 import sys
 from collections.abc import Generator, Iterator
 from ssl import SSLContext
-from typing import Any
+from typing import TYPE_CHECKING, Any
 from unittest.mock import Mock
 
 import pytest
 import trustme
 from _pytest.fixtures import SubRequest
-from blockbuster import BlockBuster, blockbuster_ctx
 from trustme import CA
+
+if TYPE_CHECKING:
+    from blockbuster import BlockBuster
 
 uvloop_marks = []
 try:
@@ -55,6 +57,12 @@ if sys.version_info >= (3, 12):
 
 @pytest.fixture(autouse=True)
 def blockbuster() -> Iterator[BlockBuster]:
+    try:
+        from blockbuster import blockbuster_ctx
+    except ImportError:
+        yield
+        return
+
     with blockbuster_ctx(
         "anyio", excluded_modules=["anyio.pytest_plugin", "anyio._backends._asyncio"]
     ) as bb:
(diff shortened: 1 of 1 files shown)


tests/conftest.py
```diff
@@ -55,6 +57,12 @@ def eager_task_loop_factory() -> asyncio.AbstractEventLoop:
 
 @pytest.fixture(autouse=True)
 def blockbuster() -> Iterator[BlockBuster]:
```
agronholm, line 59: Not sure why mypy didn't catch this.
```suggestion
def blockbuster() -> Iterator[BlockBuster | None]:
```

tests/conftest.py
```diff
@@ -71,7 +79,8 @@ def blockbuster() -> Iterator[BlockBuster]:
 
 @pytest.fixture
 def deactivate_blockbuster(blockbuster: BlockBuster) -> None:
```
agronholm, line 81:
```suggestion
def deactivate_blockbuster(blockbuster: BlockBuster | None) -> None:
```

tests/conftest.py
```diff
@@ -54,7 +56,13 @@ def eager_task_loop_factory() -> asyncio.AbstractEventLoop:
 
 
 @pytest.fixture(autouse=True)
-def blockbuster() -> Iterator[BlockBuster]:
+def blockbuster() -> Iterator[BlockBuster | None]:
+    try:
+        from blockbuster import blockbuster_ctx
+    except ImportError:
+        yield
```
agronholm, line 63:
```suggestion
        yield None
```

agronholm (approved):
Looks reasonable.

mgorny:
Thank you. I'm sorry it didn't occur to me to update the type hints.

PR #896, opened 2025-03-17
add notify_closing

graingert:
## Changes

* add anyio.notify_closing
* adjust wait_readable/wait_writable to checkpoint after registration and before waiting on the event.

anyio-zmq needs this feature to handle ZmqSocket.close calls

we expose the trio implemenation which is simple enough, and emulate the feature on asyncio. If someone is using add_reader themselves on an FD then you can't use it with anyio wait_readable/wait_writiable/notify_closing

## Checklist

features)

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -2707,10 +2707,9 @@ class AsyncIOBackend(AsyncBackend):
             read_events = {}
             _read_events.set(read_events)
 
-        if not isinstance(obj, int):
-            obj = obj.fileno()
+        fd = obj if isinstance(obj, int) else obj.fileno()
 
-        if read_events.get(obj):
+        if read_events.get(fd):
             raise BusyResourceError("reading from")
 
         loop = get_running_loop()
@@ -2718,31 +2717,37 @@ class AsyncIOBackend(AsyncBackend):
 
         def cb() -> None:
             try:
+                del read_events[fd]
+            except KeyError:
+                pass
+            else:
+                remove_reader(fd)
+            try:
                 fut.set_result(True)
             except asyncio.InvalidStateError:
                 pass
 
         try:
-            loop.add_reader(obj, cb)
+            loop.add_reader(fd, cb)
         except NotImplementedError:
             from anyio._core._asyncio_selector_thread import get_selector
 
             selector = get_selector()
-            selector.add_reader(obj, cb)
+            selector.add_reader(fd, cb)
             remove_reader = selector.remove_reader
         else:
             remove_reader = loop.remove_reader
 
-        read_events[obj] = fut
+        read_events[fd] = fut
         try:
             success = await fut
         finally:
             try:
-                del read_events[obj]
+                del read_events[fd]
             except KeyError:
                 pass
             else:
-                remove_reader(obj)
+                remove_reader(fd)
         if not success:
             raise ClosedResourceError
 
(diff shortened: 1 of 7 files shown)


src/anyio/_backends/_asyncio.py
```diff
@@ -2729,14 +2746,14 @@ async def wait_readable(cls, obj: FileDescriptorLike) -> None:
 
         read_events[obj] = event
         try:
+            await cls.checkpoint()
```
graingert, line 2749: I think this checkpoint is redundant because wait_readable won't callback without an event loop cycle so it runs selector.select()

src/anyio/_backends/_asyncio.py
```diff
@@ -2764,11 +2781,57 @@ async def wait_writable(cls, obj: FileDescriptorLike) -> None:
 
         write_events[obj] = event
         try:
+            await cls.checkpoint()
```
graingert, line 2784:
```suggestion
```

tests/test_sockets.py
```diff
@@ -1932,3 +1933,42 @@ async def test_deprecated_wait_socket(anyio_backend_name: str) -> None:
...
+
+        async def writer() -> None:
+            with pytest.raises(ClosedResourceError):
+                await wait_writable(a)
+
+        try:
+            while True:
+                a_sock.send(b"x" * 65536)
+        except BlockingIOError:
+            pass
```
agronholm, line 1967: Did you intend to actually use that function?
```suggestion
        fill_socket(a_sock)
```

src/anyio/_core/_sockets.py
```diff
@@ -702,6 +702,35 @@ def wait_writable(obj: FileDescriptorLike) -> Awaitable[None]:
...
+    3. Actually close the object.
+
+    It's also possible to do them in a different order if that's more
+    convenient, *but only if* you make sure not to have any checkpoints in
+    between the steps. This way they all happen in a single atomic
+    step, so other tasks won't be able to tell what order they happened
+    in anyway.
+
+    :param obj: an object with a ``.fileno()`` method or an integer handle
+    """
```
agronholm, line 730:
```suggestion

    """
```

src/anyio/abc/_eventloop.py
```diff
@@ -343,6 +343,11 @@ async def wait_readable(cls, obj: HasFileno | int) -> None:
     async def wait_writable(cls, obj: HasFileno | int) -> None:
         pass
 
+    @classmethod
+    @abstractmethod
+    def notify_closing(cls, obj: HasFileno | int) -> None:
```
agronholm, line 348: Should we switch these to `FileDescriptorLike` too?

agronholm (changes requested):

src/anyio/_backends/_asyncio.py
```diff
@@ -1745,8 +1745,26 @@ async def send(self, item: bytes) -> None:
...
+    _closed: bool = field(default=False, init=False)
+    _event: asyncio.Event = field(default_factory=asyncio.Event, init=False)
+
+    def set(self) -> None:
+        self._event.set()
+
+    async def wait(self) -> None:
+        await self._event.wait()
+        if self._closed:
+            raise ClosedResourceError
```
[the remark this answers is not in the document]
graingert: You need to wait until it's closed

src/anyio/abc/_eventloop.py
graingert: Yep

agronholm:
So, two questions came to mind:

1. Is there a reason we're not just popping the fd from the dict when we have a result?
2. Is there a reason why you're not simply setting the `ClosedResourceError` as the future's exception?

graingert:
> So, two questions came to mind:
> 
>     1. Is there a reason we're not just popping the fd from the dict when we have a result?

that seems nicer, and allows new readers/writers slightly sooner. I've pushed that - it's a bit more complex though
 
>     2. Is there a reason why you're not simply setting the `ClosedResourceError` as the future's exception?

using a future for handling exceptions results in a reference cycle attached to a traceback and I'd like to avoid this. Also the traceback is harder to debug

agronholm (approved):

PR #900, opened 2025-03-20
Added an optional thread `name` parameter to `start_blocking_portal()`

### LINKED ISSUE #818: add the ability to name the thread used by start_blocking_portal, and give it a default name
### Things to check first

### Feature description

I'd like to be able to call `anyio.to_thread.start_blocking_portal(..., thread_name="example name")` I'd also like a default name so I know what's going on in `threading.enumerate()`

### Use case

I was porting a bespoke run-trio-in-thread code to BlockingPortal, the original code named their thread - but with a blocking portal I don't have that ability

davidbrochart:
## Changes

Fixes #818. 

## Checklist

features)

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -5,7 +5,7 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
 **UNRELEASED**
 
-- Added optional thread ``name`` parameter to ``start_blocking_portal()``
+- Added the ability to specify the thread name in ``start_blocking_portal()``
   (`#818 <https://github.com/agronholm/anyio/issues/818>`_; PR by @davidbrochart)
 
 **4.9.0**
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -3,6 +3,11 @@ Version history
 
 This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
+**UNRELEASED**
+
+- Added optional thread ``name`` parameter to ``start_blocking_portal()``
+  (`#900 <https://github.com/agronholm/anyio/issues/818>`_; PR by @davidbrochart)
+
 **4.9.0**
 
 - Added async support for temporary file handling
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -6,7 +6,7 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 **UNRELEASED**
 
 - Added the ability to specify the thread name in ``start_blocking_portal()``
-  (`#900 <https://github.com/agronholm/anyio/pull/900>`_; PR by @davidbrochart)
+  (`#818 <https://github.com/agronholm/anyio/issues/818>`_; PR by @davidbrochart)
 
 **4.9.0**
 
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -6,7 +6,7 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 **UNRELEASED**
 
 - Added optional thread ``name`` parameter to ``start_blocking_portal()``
-  (`#900 <https://github.com/agronholm/anyio/issues/818>`_; PR by @davidbrochart)
+  (`#818 <https://github.com/agronholm/anyio/issues/818>`_; PR by @davidbrochart)
 
 **4.9.0**
 
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -6,7 +6,7 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 **UNRELEASED**
 
 - Added the ability to specify the thread name in ``start_blocking_portal()``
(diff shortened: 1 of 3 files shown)


agronholm:
I think so, yes - but it shouldn't be hard to write one.

davidbrochart:
Will do it later.

src/anyio/from_thread.py
```diff
@@ -454,7 +454,9 @@ def __exit__(
 
 @contextmanager
 def start_blocking_portal(
-    backend: str = "asyncio", backend_options: dict[str, Any] | None = None
+    backend: str = "asyncio",
+    backend_options: dict[str, Any] | None = None,
+    name: str | None = None,
```
agronholm, line 459: I would prefer this to be a keyword-only argument.
```suggestion
    *,
    name: str | None = None,
```

graingert:
The thread gets an OK ish name based on the target function 

If a default name is added I'd like it to be an f-string with the backend_name templated in

agronholm:
> The thread gets an OK ish name based on the target function
> 
> If a default name is added I'd like it to be an f-string with the backend_name templated in

Yeah, that's precisely what I was thinking of.

11kkw:
It looks like this PR is implementing the feature requested in issue #818.
Should we link them so that the issue gets closed automatically upon merging?"

@davidbrochart @agronholm

agronholm:
> It looks like this PR is implementing the feature requested in issue #818. Should we link them so that the issue gets closed automatically upon merging?"
> 
> @davidbrochart @agronholm

I've edited the description.

src/anyio/from_thread.py
```diff
@@ -486,7 +498,7 @@ def run_blocking_portal() -> None:
                     future.set_exception(exc)
 
     future: Future[BlockingPortal] = Future()
-    thread = Thread(target=run_blocking_portal, daemon=True)
+    thread = NamedThread(target=run_blocking_portal, daemon=True, name=name)
```
agronholm, line 501: The idea was to use the ID of the blocking portal, not the thread. The thread ID is already evident from the list of threads.

davidbrochart: You mean the ID as `id(blocking_portal)`?

agronholm: Yes (presented in the hexadecimal format) as `repr()` does for objects not having their own `__repr__()` method.

graingert: But we don't have the instance of blocking_portal before starting the thread?

agronholm: That's why we're setting the thread name after the portal instance is created.

graingert: Oh you set it later when the thread boots

tests/test_from_thread.py
```diff
@@ -371,6 +371,30 @@ async def async_get_thread_id() -> int:
...
+    def test_start_without_thread_name(
+        self, anyio_backend_name: str, anyio_backend_options: dict[str, Any]
+    ) -> None:
+        def get_thread_name() -> str:
+            return threading.current_thread().name
+
+        with start_blocking_portal(anyio_backend_name, anyio_backend_options) as portal:
+            thread_name = portal.call(get_thread_name)
+
+        assert thread_name == f"{anyio_backend_name}-portal-{hex(id(portal))}"
```
agronholm, line 396:
```suggestion
        assert thread_name == f"{anyio_backend_name}-portal-{id(portal):x}"
```

agronholm:
This just needs a changelog entry now.

docs/versionhistory.rst
```diff
@@ -3,6 +3,11 @@ Version history
 
 This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
+**UNRELEASED**
+
+- Added optional thread ``name`` parameter to ``start_blocking_portal()``
```
```suggestion
- Added ability to specify thread name in ``start_blocking_portal()``
```

docs/versionhistory.rst
```diff
@@ -3,6 +3,11 @@ Version history
...
+  (`#900 <https://github.com/agronholm/anyio/issues/818>`_; PR by @davidbrochart)
```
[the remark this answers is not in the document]
davidbrochart: Yes, I noticed the links below are wrong too?

docs/versionhistory.rst
```diff
@@ -3,6 +3,11 @@ Version history
 
 This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
+**UNRELEASED**
+
+- Added the ability to specify the thread name in ``start_blocking_portal()``
+  (`#900 <https://github.com/agronholm/anyio/pull/900>`_; PR by @davidbrochart)
```
agronholm, line 9: Don't override my changes please
```suggestion
  (`#818 <https://github.com/agronholm/anyio/issues/818>`_; PR by @davidbrochart)
```

agronholm (approved):
LGTM. @graingert anything to add?

graingert (approved):

graingert:
Oh it's not passing

agronholm:
> Oh it's not passing

This seems to be a coveralls issue rather than a test failure.

PR #905, opened 2025-04-02
Added context manager mix-in classes

agronholm:
## Changes

This adds two new mix-in classes: `ContextManagerMixin` and `AsyncContextManagerMixin`. They're meant to enable users to write context manager classes in the same way as `@contextmanager` and `@asynccontextmanager`, respectively. The aim is to enable users to more easily embed cancel scopes and task groups into other context manager classes.

## Checklist

features)

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/docs/cancellation.rst
+++ b/docs/cancellation.rst
@@ -216,10 +216,16 @@ context manager needs to use other context managers, you may find it convenient 
 :class:`AsyncContextManagerMixin` in order to avoid cumbersome code that calls
 ``__aenter__()`` and ``__aexit__()`` directly::
 
+    from __future__ import annotations
+
+    from collections.abc import AsyncGenerator
+
     from anyio import AsyncContextManagerMixin, create_task_group
 
 
-    class MyAsyncContextManager(AsyncContextManagerMixin):
-        async def __asynccontextmanager__(self):
+    # AsyncContextManagerMixin is parametrized this way because it returns 'self'
+    class MyAsyncContextManager(AsyncContextManagerMixin["MyAsyncContextManager"]):
+        async def __asynccontextmanager__(self) -> AsyncGenerator[MyAsyncContextManager]:
             async with create_task_group() as tg:
                 ...  # launch tasks
+                yield self
--- a/docs/cancellation.rst
+++ b/docs/cancellation.rst
@@ -157,6 +157,8 @@ cancelled scope::
 
             raise
 
+.. _cancel_scope_stack_corruption:
+
 Avoiding cancel scope stack corruption
 --------------------------------------
 
@@ -219,13 +221,16 @@ context manager needs to use other context managers, you may find it convenient 
     from __future__ import annotations
 
     from collections.abc import AsyncGenerator
+    from typing import Self
 
     from anyio import AsyncContextManagerMixin, create_task_group
 
 
-    # AsyncContextManagerMixin is parametrized this way because it returns 'self'
-    class MyAsyncContextManager(AsyncContextManagerMixin["MyAsyncContextManager"]):
-        async def __asynccontextmanager__(self) -> AsyncGenerator[MyAsyncContextManager]:
+    class MyAsyncContextManager(AsyncContextManagerMixin):
+        @asynccontextmanager
+        async def __asynccontextmanager__(self) -> AsyncGenerator[Self]:
             async with create_task_group() as tg:
                 ...  # launch tasks
                 yield self
+
+.. seealso:: :doc:`contextmanagers`
--- a/docs/cancellation.rst
+++ b/docs/cancellation.rst
@@ -196,8 +196,11 @@ Depending on how they are used, this pattern is, however, *usually* safe to use 
 asynchronous context managers, so long as you make sure that the same host task keeps
 running throughout the entire enclosed code block::
 
+    from contextlib import asynccontextmanager
+
+
     # Okay in most cases!
-    @async_context_manager
+    @asynccontextmanager
     async def some_context_manager():
         async with create_task_group() as tg:
             tg.start_soon(foo)
@@ -209,24 +212,14 @@ start to end in the same task, making it possible to have task groups or cancel 
 safely straddle the ``yield``.
 
 When you're implementing the async context manager protocol manually and your async
-context manager needs to use other context managers, you may find it necessary to call
-their ``__aenter__()`` and ``__aexit__()`` directly. In such cases, it is absolutely
-vital to ensure that their ``__aexit__()`` methods are called in the exact reverse order
-of the ``__aenter__()`` calls. To this end, you may find the
-:class:`~contextlib.AsyncExitStack` class very useful::
+context manager needs to use other context managers, you may find it convenient to use
+:class:`AsyncContextManagerMixin` in order to avoid cumbersome code that calls
+``__aenter__()`` and ``__aexit__()`` directly::
 
-    from contextlib import AsyncExitStack
-
-    from anyio import create_task_group
+    from anyio import AsyncContextManagerMixin, create_task_group
 
 
-    class MyAsyncContextManager:
-        async def __aenter__(self):
-            self._exitstack = AsyncExitStack()
-            await self._exitstack.__aenter__()
-            self._task_group = await self._exitstack.enter_async_context(
-                create_task_group()
-            )
-
-        async def __aexit__(self, exc_type, exc_val, exc_tb):
-            return await self._exitstack.__aexit__(exc_type, exc_val, exc_tb)
+    class MyAsyncContextManager(AsyncContextManagerMixin):
+        async def __asynccontextmanager__(self):
+            async with create_task_group() as tg:
+                ...  # launch tasks
--- a/src/anyio/_core/_contextmanagers.py
+++ b/src/anyio/_core/_contextmanagers.py
@@ -4,12 +4,23 @@ from abc import abstractmethod
 from contextlib import AbstractAsyncContextManager, AbstractContextManager
 from inspect import isasyncgen, iscoroutine, isgenerator
 from types import TracebackType
-from typing import Generic, TypeVar, final
+from typing import Protocol, TypeVar, cast, final
 
 _T_co = TypeVar("_T_co", covariant=True)
+_ExitT_co = TypeVar("_ExitT_co", covariant=True, bound="bool | None")
 
 
-class ContextManagerMixin(Generic[_T_co]):
+class _SupportsCtxMgr(Protocol[_T_co, _ExitT_co]):
+    def __contextmanager__(self) -> AbstractContextManager[_T_co, _ExitT_co]: ...
+
+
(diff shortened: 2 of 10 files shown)


agronholm:
Open issues:

- Can't type annotate to return `Self` (probably unsolvable)
- This may require a new documentation section
- We should probably check for corner cases in the `@asynccontextmanager` implementation

src/anyio/_core/_contextmanagers.py
```diff
@@ -0,0 +1,136 @@
...
+    Mixin class providing context manager functionality via a generator-based
+    implementation.
+
+    This class allows you to implement a context manager using a generator, similar to
+    :func:`contextlib.contextmanager`, but for classes that need to embed other context
+    managers.
+
+    This class is designed to streamline the use of context management by
+    requiring the implementation of the `__contextmanager__` method, which
+    should yield instances of the class itself. It then wraps this generator
```
smurfix: Well, there's no requirement for a context manager to return Self and the wrapper shouldn't enforce that. It should accept / forward whatever type `__contextmanager__` is declared to return.

On the other hand, what this wrapper does not do is to ensure that the context is not entered twice, and *that* **really** is a bug.

agronholm: > Well, there's no requirement for a context manager to return Self and the wrapper shouldn't enforce that. It should accept / forward whatever type __contextmanager__ is declared to return.

I've already updated the docstring before you posted this. This version was never intended to be pushed to GH.

> On the other hand, what this wrapper does not do is to ensure that the context is not entered twice, and that really is a bug.

Hm, I see what you mean. But rather than raise an error here, perhaps reentrancy should be properly supported instead?

agronholm: I've thus disallowed reentrancy.

smurfix:
I have been doing something like this in my own codebase for ages by now.

One problem, though: in a subclass, how do I "cleanly" call the superclass's `__asynccontextmanager__`? The best way I can come up with is
```
async with asynccontextmanager(super().__asynccontextmanager__)():
    ...
```
which is … meh. Anyway, whichever way we come up with needs documentation.

agronholm:
@Zac-HD mentioned elsewhere that he wanted to enforce the use of `@asynccontextmanager` on this special method, for the benefit of [PEP 789](https://peps.python.org/pep-0789/) linter checks. If we do that, then this gets considerably cleaner: `async with super().__asynccontextmanager__() as obj:`

agronholm:
I pushed those changes, but I encountered a different problem: How can we override the type parameter when subclassing? The pre-commit checks fail for this reason, and I don't know how to fix this.

agronholm:
> Huh. pyright doesn't understand this either, which frankly surprises me somewhat. Maybe ask there?

Doesn't understand what? The parent class pins the type variable to itself, and that class does not have the `child_started` and `child_finished` attributes, so it seems pretty clear to me why the mypy checks fail here. What I don't understand is how to improve the type annotations to get around this.

agronholm:
There appears to be no way to pin the type variable to `Self`, which is exactly what we'd need here.

tapetersen:
The problem here is that `DummyContextManager` binds the typevar concretely in its definition which sets the type for `__enter__` unless we override that as well.

If it was allowed/defined and possible we would want to parametrize it directly with `Self` as:
```python
class DummyContextManager(ContextManagerMixin[Self]):
    ...
    @contextmanager
    def __contextmanager__(self) -> Generator[Self]:
        ...
```

What does work here is the old workaround with a typevar with a bound as:
```
_TSelf = TypeVar("_TSelf", bound="DummyContextManager")
class DummyContextManager(ContextManagerMixin[_TSelf]):
    def __init__(self, handle_exc: bool = False) -> None:
        self.started = False
        self.finished = False
        self.handle_exc = handle_exc

    @contextmanager
    def __contextmanager__(self: _TSelf) -> Generator[_TSelf]:
        self.started = True
        try:
            yield self
        except RuntimeError:
            if not self.handle_exc:
                raise

        self.finished = True
 ```
 
 That's of course not ideal, needing a declared typevar with correct bound for every class more or less, but the case of further extending the contextmanagers overriding the returned type maybe isn't that common either?

**Comment by agronholm** [comment]
> The problem here is that `DummyContextManager` binds the typevar concretely in its definition which sets the type for `__enter__` unless we override that as well.
> 
> If it was allowed/defined and possible we would want to parametrize it directly with `Self` as:
> 
> ```python
> class DummyContextManager(ContextManagerMixin[Self]):
>     ...
>     @contextmanager
>     def __contextmanager__(self) -> Generator[Self]:
>         ...
> ```
> 
> What does work here is the old workaround with a typevar with a bound as:
> 
> ```python
> _TSelf = TypeVar("_TSelf", bound="DummyContextManager")
> class DummyContextManager(ContextManagerMixin[_TSelf]):
>     def __init__(self, handle_exc: bool = False) -> None:
>         self.started = False
>         self.finished = False
>         self.handle_exc = handle_exc
> 
>     @contextmanager
>     def __contextmanager__(self: _TSelf) -> Generator[_TSelf]:
>         self.started = True
>         try:
>             yield self
>         except RuntimeError:
>             if not self.handle_exc:
>                 raise
> 
>         self.finished = True
> ```
> 
> That's of course not ideal, needing a declared typevar with correct bound for every class more or less, but the case of further extending the contextmanagers overriding the returned type maybe isn't that common either?

By "does work" do you mean you get the correct type out of `reveal_type()` in an inherited CM? Can you show me?

**Comment by agronholm** [comment]
I don't think forcing context manager authors to use a typevar just to enable proper inheritance is a proper solution to the problem. Let's see what the pyright folks have to say about this.

**Comment by tapetersen** [comment]
As in I checked out your branch but there are some issues with `pyproject.toml` parsing (the license checks but also some things about the tox config) so it was a bit hard to get the pre-commit to run.

Looking closer I can't get it to work without explicitly passing through the leaf-class as a typevar.

I suspect that getting this to work without explicit annotation when inheriting is not really possible with current typing system.

(The return type of `__aenter__` in the mixin is basically dependent on a possibly non-self type-parameter in a future child-class)

Have added it as a pull request to your branch anyway but it's not really any better except maybe the default (and maybe demonstrating that it works *if* you specify the return type explicitly for every child)

**Comment by agronholm** [comment]
> As in I checked out your branch but there are some issues with pyproject.toml parsing (the license checks but also some things about the tox config) so it was a bit hard to get the pre-commit to run.

What license checks? Pre-commit has no issues running over here, nor on CI.

> I suspect that getting this to work without explicit annotation when inheriting is not really possible with current typing system.

That was my conclusion too.

**Comment by tapetersen** [comment]
(this is very possibly older versions of tox and/or problems with this specific environment and I can sort it out and report as a real issue when/if i figure out if is some requirement that needs to be bumped)

Trying to run tox for me right now results in 
```
$ tox 
ROOT: No tox.ini or setup.cfg or pyproject.toml found, assuming empty tox.ini at /home/tobias/Projects/anyio
.pkg: _optional_hooks> python /home/tobias/.local/lib/python3.10/site-packages/pyproject_api/_backend.py True setuptools.build_meta
.pkg: get_requires_for_build_sdist> python /home/tobias/.local/lib/python3.10/site-packages/pyproject_api/_backend.py True setuptools.build_meta
py: packaging backend failed (code=1), with ValueError: invalid pyproject.toml config: `project.license`.
configuration error: `project.license` must be valid exactly by one definition (2 matches found):

    - keys:
        'file': {type: string}
      required: ['file']
    - keys:
        'text': {type: string}
      required: ['text']
```

**Comment by agronholm** [comment]
Yeah, `tox` doesn't seems to understand the SPDX license specifiers.

**Comment by tapetersen** [comment]
tox still can't find it's settings from the `pyproject.toml` file for me though

```
ROOT: No tox.ini or setup.cfg or pyproject.toml found, assuming empty tox.ini at /home/tobias/Projects/anyio
```

**Comment by tapetersen** [comment]
It was just an old version of tox (4.5) that was picked up due to vscode's restoration of env-vars that messed up some things.

May still not hurt to add a requires in the tox section as specified here: https://tox.wiki/en/latest/config.html#pyproject-toml-native
Wouldn't have solved this as the old version couldn't parse it anyway but for other incompatibilites it may be a good idea.
```
[tool.tox]
requires = ["tox>=4.19"]
env_list = ["pre-commit", "py39", "py310", "py311", "py312", "py313", "py314", "pypy3"]
skip_missing_interpreters = true
```

**Comment by agronholm** [comment]
> It was just an old version of tox (4.5) that was picked up from outside the venv due to vscode's restoration of env-vars that messed up some things.
> 
> May still not hurt to add a requires in the tox section as specified here: https://tox.wiki/en/latest/config.html#pyproject-toml-native Wouldn't have solved this as the old version couldn't parse it anyway but for other incompatibilites it may be a good idea.

I originally skipped specifying that for precisely that reason.

File: docs/cancellation.rst [hunk 2]
```diff
@@ -209,24 +214,23 @@ start to end in the same task, making it possible to have task groups or cancel
...
+    from typing import Self
 
-    from anyio import create_task_group
+    from anyio import AsyncContextManagerMixin, create_task_group
 
 
-    class MyAsyncContextManager:
-        async def __aenter__(self):
-            self._exitstack = AsyncExitStack()
-            await self._exitstack.__aenter__()
```
graingert, line 226: oh this was slightly wrong it should have used pop_all()

I think it's worth including an ExitStack instruction on how to do this

src/anyio/_core/_contextmanagers.py
```diff
@@ -0,0 +1,200 @@
...
+        that once you enter it, you can't re-enter before first exiting it.
+
+    .. seealso:: :doc:`contextmanagers`
+    """
+
+    __cm: AbstractContextManager[object, bool | None] | None = None
+
+    @final
+    def __enter__(self: _SupportsCtxMgr[_T_co, bool | None]) -> _T_co:
+        # Needed for mypy to assume self still has the __cm member
```
graingert, line 42: I think you can alternatively put _ContextManagerMixin__cm on the _SupportsCtxMgr protocol

tests/test_contextmanagers.py
```diff
@@ -0,0 +1,256 @@
...
+        try:
+            yield self
+        except RuntimeError:
+            if not self.handle_exc:
+                raise
+
+        self.finished = True
+
+
+class DummyAsyncContextManager(AsyncContextManagerMixin):
```
graingert, line 43: might be worth a test for a class that implements both AsyncContextManagerMixin and ContextManagerMixin

graingert (approved):

tests/test_contextmanagers.py
agronholm: I'm not sure why? The mix-in classes have no overlap whatsoever.

agronholm: Oh, I just realized: that's the problem. One could enter the CM synchronously AND asynchronously without first exiting.

PR #906, opened 2025-04-05
add create_priority_memory_object_stream

graingert:
## Changes

Previously this was required for a Textual port to AnyIO. This is just a sketch to see how extensive the changes would be.

## Checklist

features)

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/src/anyio/_core/_streams.py
+++ b/src/anyio/_core/_streams.py
@@ -8,6 +8,7 @@ from ..streams.memory import (
     MemoryObjectReceiveStream,
     MemoryObjectSendStream,
     MemoryObjectStreamState,
+    PriorityMemoryObjectStreamState,
 )
 
 T_Item = TypeVar("T_Item")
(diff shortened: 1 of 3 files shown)


src/anyio/_core/_streams.py
```diff
@@ -50,3 +51,27 @@ def __new__(  # type: ignore[misc]
...
+
+    The stream's item type can be annotated like
+    :func:`create_priority_memory_object_stream[T_Item]`.
+
+    :param max_buffer_size: number of items held in the buffer until ``send()`` starts
+        blocking
+    """
+
+    def __new__(  # type: ignore[misc]
+        cls, max_buffer_size: float = 0
```
graingert, line 70: not sure if max_buffer_size should be allowed to be set to 0 or 1 for create_priority_memory_object_stream, it would behave the same as create_memory_object_stream, but be slower for no purpose.

we could set the default to be math.inf and require it to be more than 2

src/anyio/streams/memory.py
```diff
@@ -57,6 +77,7 @@ class MemoryObjectStreamState(Generic[T_Item]):
     waiting_senders: OrderedDict[Event, T_Item] = field(
         init=False, default_factory=OrderedDict
     )
+    buffer: Buffer[T_Item] = field(init=False, default_factory=deque)
```
graingert, line 80:
```suggestion
```

src/anyio/_core/_streams.py
tapetersen: Is any limited buffer size really a good and intended idea here (at least when using the blocking send)?

From what I can understand of the semantics as soon as the buffer becomes full you lose the priority guaranteee as a blocked sender with a low priority task will not get processed at all until there is room in the queue even if the item in front is of lower priority.

It feels like the kind of construct that will behave as someone expected until a critical rare corner case where someone will spend a long time debugging why the priority isn't respected.

I would probably have preferred primitives that either always block but respect priority order (would more or less behave as a priority-based lock instead) or raise an exception if it couldn't guarantee priority (which you would get here if you always use non-blocking send.

I may overreact though and just a note about the caveats that priority may be violated if using limited capacity (and that 0 or 1 therefore are meaningless)  may be enough.

PR #912, opened 2025-04-16
Fix contextmanager mixin typing

tapetersen:
## Changes

Fixes #. 

## Checklist

features)

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/src/anyio/_core/_contextmanagers.py
+++ b/src/anyio/_core/_contextmanagers.py
@@ -4,17 +4,20 @@ from abc import abstractmethod
 from contextlib import AbstractAsyncContextManager, AbstractContextManager
 from inspect import isasyncgen, iscoroutine, isgenerator
 from types import TracebackType
-from typing import Protocol, TypeVar, final
+from typing import Protocol, TypeVar, cast, final
 
 _T_co = TypeVar("_T_co", covariant=True)
+_ExitT_co = TypeVar("_ExitT_co", covariant=True, bound="bool | None")
 
 
-class _SupportsCtxMgr(Protocol[_T_co]):
-    def __contextmanager__(self) -> AbstractContextManager[_T_co]: ...
+class _SupportsCtxMgr(Protocol[_T_co, _ExitT_co]):
+    def __contextmanager__(self) -> AbstractContextManager[_T_co, _ExitT_co]: ...
 
 
-class _SupportsAsyncCtxMgr(Protocol[_T_co]):
-    def __asynccontextmanager__(self) -> AbstractAsyncContextManager[_T_co]: ...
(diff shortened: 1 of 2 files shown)


agronholm:
Just so you know, I really don't think I will merge this as I have an alternate solution on the way.

agronholm:
Oh, you improved it? In that case I'll take a closer look.

tapetersen:
Yeah, as I wrote in the other thread I didn't really think the old was an improvement but I have used the trick of typing self before for other cases and thought to give it a try.

I'm somewhat surprised it worked as well as it did in mypy since I've mainly used these tricks with pyright and expected it to be possibly problematic.

The lack of type-intersection at least for protocols is really hurting here though... (it would allow this to be implemented with a decorator instead of a mixin but there is no way to specify the return type of that correctly unfortunately)

src/anyio/_core/_contextmanagers.py
```diff
@@ -4,12 +4,20 @@
...
-class ContextManagerMixin(Generic[_T_co]):
+class _SupportsCtxMgr(Protocol[_T_co]):
+    def __contextmanager__(self) -> AbstractContextManager[_T_co]: ...
+
+
+class _SupportsAsyncCtxMgr(Protocol[_T_co]):
+    def __asynccontextmanager__(self) -> AbstractAsyncContextManager[_T_co]: ...
+
+
+class ContextManagerMixin:
```
tapetersen, line 20: Removing the typevar from the class level here is not necessary but avoids having to declare another one to use in the method and since the mixin class doesn't really depend on the value type it's not needed.

src/anyio/_core/_contextmanagers.py
```diff
@@ -102,10 +112,12 @@ class AsyncContextManagerMixin(Generic[_T_co]):
...
     """
 
-    __cm: AbstractAsyncContextManager[_T_co] | None = None
+    __cm: AbstractAsyncContextManager[object] | None = None
 
     @final
-    async def __aenter__(self) -> _T_co:
+    async def __aenter__(self: _SupportsAsyncCtxMgr[_T_co]) -> _T_co:
+        # Needed for mypy to assume self still has the __cm member
+        assert isinstance(self, AsyncContextManagerMixin)
```
tapetersen, line 120: pyright treats the self typing as an implicit intersection with the containing class but mypy doesn't (even if it does other magic inference fine).

agronholm:
I'm leaning towards accepting this, as it seems to work exactly the way I wanted! Thanks! I'll give this a more thorough review within the next couple days.

agronholm (approved):
I can't find fault in this, so I'm merging it.

PR #915, opened 2025-04-23
fix(tempfile): correct SpooledTemporaryFile rollover boundary

11kkw:
## Changes

- Change `_check` condition from `< self._max_size` to `<= self._max_size` so that rollover only occurs when size > max_size  
- Add `test_exact_boundary_no_rollover` to verify no rollover at the exact boundary and rollover on overflow

## Checklist

- [ ] You've updated the documentation (in `docs/`, in case of behavior changes or new features)  
- [ ] You've added a new changelog entry (in `docs/versionhistory.rst`)

### DIFF
--- a/tests/test_tempfile.py
+++ b/tests/test_tempfile.py
@@ -102,7 +102,6 @@ class TestSpooledTemporaryFile:
 
         assert stf.closed
 
-
     async def test_exact_boundary_no_rollover(self) -> None:
         async with SpooledTemporaryFile(max_size=10) as stf:
             await stf.write(b"0123456789")
--- a/tests/test_tempfile.py
+++ b/tests/test_tempfile.py
@@ -102,6 +102,24 @@ class TestSpooledTemporaryFile:
 
         assert stf.closed
 
+    async def test_exact_boundary_no_rollover(self) -> None:
+        rollover_called = False
+
+        async def fake_rollover() -> None:
+            nonlocal rollover_called
+            rollover_called = True
+            await original_rollover()
+
+        async with SpooledTemporaryFile(max_size=10) as stf:
+            original_rollover = stf.rollover
+            with patch.object(stf, "rollover", fake_rollover):
+                await stf.write(b"0123456789")
+                assert not rollover_called
+
+                more = b"x"
+                await stf.write(more)
+                assert rollover_called
+
 
 class TestTemporaryDirectory:
     async def test_context_manager(self) -> None:
--- a/tests/test_tempfile.py
+++ b/tests/test_tempfile.py
@@ -40,20 +40,20 @@ class TestNamedTemporaryFile:
         data = b"named temporary file data"
         async with NamedTemporaryFile[bytes]() as af:
             filename = af.name
-            assert os.path.exists(filename)  # type: ignore[arg-type]
+            assert os.path.exists(filename)
 
             await af.write(data)
             await af.seek(0)
             assert await af.read() == data
 
-        assert not os.path.exists(filename)  # type: ignore[arg-type]
+        assert not os.path.exists(filename)
 
     async def test_exception_handling(self) -> None:
         async with NamedTemporaryFile[bytes]() as af:
             filename = af.name
-            assert os.path.exists(filename)  # type: ignore[arg-type]
+            assert os.path.exists(filename)
 
-        assert not os.path.exists(filename)  # type: ignore[arg-type]
+        assert not os.path.exists(filename)
 
         with pytest.raises(ValueError):
             await af.write(b"should fail")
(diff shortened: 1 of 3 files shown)


tests/test_tempfile.py
```diff
@@ -102,6 +102,24 @@ async def test_closed_state(self) -> None:
 
         assert stf.closed
 
+    async def test_exact_boundary_no_rollover(self) -> None:
+        rollover_called = False
+
+        async def fake_rollover() -> None:
+            nonlocal rollover_called
+            rollover_called = True
+            await original_rollover()
```
agronholm, line 111: Why do you need to call the original rollover here?

tests/test_tempfile.py
```diff
@@ -102,6 +102,24 @@ async def test_closed_state(self) -> None:
...
+
+        async with SpooledTemporaryFile(max_size=10) as stf:
+            original_rollover = stf.rollover
+            with patch.object(stf, "rollover", fake_rollover):
+                await stf.write(b"0123456789")
+                assert not rollover_called
+
+                more = b"x"
+                await stf.write(more)
```
agronholm, line 120: Why do you have to store the "x" separately into a variable that isn't even used anywhere else?
```suggestion
                await stf.write(b"x")
```

tests/test_tempfile.py
```diff
@@ -102,6 +102,24 @@ async def test_closed_state(self) -> None:
...
+        rollover_called = False
+
+        async def fake_rollover() -> None:
+            nonlocal rollover_called
+            rollover_called = True
+            await original_rollover()
+
+        async with SpooledTemporaryFile(max_size=10) as stf:
+            original_rollover = stf.rollover
+            with patch.object(stf, "rollover", fake_rollover):
```
agronholm, line 115: Maybe you could just check `stf._rolled` to see if the rollover happened?

agronholm (changes requested):

agronholm:
Why did you mess with existing tests? Mypy was passing just fine for them before. Now it's not.

11kkw:
> Why did you mess with existing tests? Mypy was passing just fine for them before. Now it's not.

I’ve only modified the following test:

`
async def test_exact_boundary_no_rollover(self) -> None:
    async with SpooledTemporaryFile(max_size=10) as stf:
        await stf.write(b"0123456789")
        assert not stf._rolled

        await stf.write(b"x")
        assert stf._rolled
`
But now both mypy and the test suite are failing. What am I doing wrong?

agronholm:
The changelog says that you also modified `test_mkstemp()`. Are you saying that you didn't do that?

smurfix:
What do you mean, what are you doing wrong? You removed the assignment to `read_content` but not the subsequent test. See [tests/test_tempfile.py](https://github.com/agronholm/anyio/pull/915/commits/957730880569bd9bb11e650ce871f9f74af9cf49#diff-526cd08084beeacee14277bcb801cc797f173cece3b57aee9700ad7e9fe17290) line 180.

agronholm:
It looks like you also reverted the test simplifications you made previously?

11kkw:
> It looks like you also reverted the test simplifications you made previously?

I updated the test to this:

```
async def test_exact_boundary_no_rollover(self) -> None:
    async with SpooledTemporaryFile(max_size=10) as stf:
        await stf.write(b"0123456789")
        assert not stf._rolled

        await stf.write(b"x")
        assert stf._rolled

```

```
tests/test_tempfile.py:43: error: Unused "type: ignore" comment  [unused-ignore]
tests/test_tempfile.py:49: error: Unused "type: ignore" comment  [unused-ignore]
tests/test_tempfile.py:54: error: Unused "type: ignore" comment  [unused-ignore]
tests/test_tempfile.py:56: error: Unused "type: ignore" comment  [unused-ignore]
tests/test_tempfile.py:171: error: Argument 1 to "write" of "_TextIOBase" has incompatible type "bytes"; expected "str"  [arg-type]
tests/test_tempfile.py:176: error: Argument 1 to "write" of "BufferedWriter" has incompatible type "str"; expected "Buffer"  [arg-type]
tests/test_tempfile.py:178: error: Incompatible types in assignment (expression has type "bytes", variable has type "str")  [assignment]
Found 7 errors in 1 file (checked 1 source file)
```
mypy to fail.

tests/test_tempfile.py
```diff
@@ -102,6 +102,24 @@ async def test_closed_state(self) -> None:
...
+                await stf.write(b"0123456789")
+                assert not rollover_called
+
+                more = b"x"
+                await stf.write(more)
+                assert rollover_called
+
```
agronholm, line 122:
```suggestion
async def test_exact_boundary_no_rollover(self) -> None:
    async with SpooledTemporaryFile(max_size=10) as stf:
        await stf.write(b"0123456789")
        assert not stf._rolled

        await stf.write(b"x")
        assert stf._rolled

```

tests/test_tempfile.py
```diff
@@ -103,6 +103,15 @@ async def test_closed_state(self) -> None:
         assert stf.closed
 
 
+async def test_exact_boundary_no_rollover(self) -> None:
+    async with SpooledTemporaryFile(max_size=10) as stf:
+        await stf.write(b"0123456789")
+        assert not stf._rolled
+
+        await stf.write(b"x")
+        assert stf._rolled
```
agronholm, line 112:
```suggestion

    async def test_exact_boundary_no_rollover(self) -> None:
        async with SpooledTemporaryFile(max_size=10) as stf:
            await stf.write(b"0123456789")
            assert not stf._rolled

            await stf.write(b"x")
            assert stf._rolled
```

agronholm:
I just applied the simplifications myself now.

agronholm:
The cause of these mypy confusion stems from a mypy bug which causes it to report nonexistent issues when run against a subset of the source files. But we can't yet switch to pyright either, as it's reporting tons of issues (some of which are _technically_ legitimate).

agronholm:
Now this just needs a changelog entry.

agronholm (approved):

PR #921, opened 2025-05-19
Added connectable stream classes and connectors

agronholm:
## Changes

This adds a number of classes for abstracting how outgoing connections are made. This makes it much easier to add support for various transports, including mock transports without having to hard code such behavior into network clients.

It also allows for customizing connection strategies.

## Checklist

features)

these instructions.

### Updating the changelog

```
```


smurfix:
While you're at it … it'd be a good idea to actually include the endpoint which the failed connection was trying to reach, ideally in some machine readable form (as in `raise ConnectionFailed("some_error_message", dest.host, dest.port)`).

agronholm:
> While you're at it … it'd be a good idea to actually include the endpoint which the failed connection was trying to reach, ideally in some machine readable form (as in `raise ConnectionFailed("some_error_message", dest.host, dest.port)`).

How would this work with `MultiConnector` or other advanced connectors? They could also be wrapped in several layers.

smurfix:
If it's too complicated (for now) doing this with `MultiConnector` then it's too complicated. 🤷🏼‍♂️
That doesn't negate the simple cases.

agronholm:
I've made some changes, hopefully to your liking. I figured that machine-readable parameters would be problematic assuming a single `ConnectionFailed` exception was used.

agronholm:
Most likely the resource/connection pool stuff should be deferred to a follow-up PR. I just stuffed everything in here that I had been planning to add.

src/anyio/_core/_streams.py
```diff
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import math
+import sys
```
graingert, line 4:
```suggestion
```

src/anyio/abc/_streams.py
```diff
@@ -209,3 +209,42 @@ async def serve(
         :param task_group: the task group that will be used to start tasks for handling
             each accepted connection (if omitted, an ad-hoc task group will be created)
         """
+
+
+class ObjectStreamConnectable(Generic[T_co]):
```
graingert, line 214: should this set the metaclass to abc.ABC?

agronholm: No, as `Generic` already uses `ABCMeta` as its metaclass.

src/anyio/abc/_streams.py
```diff
@@ -209,3 +209,42 @@ async def serve(
...
+    async def connect(self) -> ObjectStream[T_co]:
+        """
+        Connect to the remote endpoint.
+
+        :return: an object stream connected to the remote end
+        :raises ConnectionFailed: if the connection fails
+        """
+
+
+class ByteStreamConnectable:
```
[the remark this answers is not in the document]
agronholm: Yes, I'll do that.

src/anyio/abc/_streams.py
agronholm: No, I was wrong about that. I'll add it here too.

src/anyio/abc/_streams.py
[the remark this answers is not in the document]
agronholm: Done.

docs/networking.rst
```diff
@@ -240,3 +240,60 @@ having to pass the path every time you send data to the peer::
...
+    from os import PathLike
+    from ssl import SSLContext
+
+    from anyio.abc import ByteStreamConnectable, as_connectable
+
+
+    class MyNetworkClient:
+        def __init__(
+            self,
+            connectable: ByteStreamConnectable | tuple[str, int] | str | PathLike[str],
```
graingert, line 261: do we want to provide a TypeAlias for this?

graingert (approved):

docs/networking.rst
agronholm: Perhaps as a follow-up. What would I call it?

Issue #926, opened 2025-05-27
Test failures with Python 3.14.0b2: ValueError: unsupported unboundop 0


hroncok:
### Things to check first

- [x] I have searched the existing issues and didn't find my bug already reported there

- [x] I have checked that my bug is still present in the latest release

### AnyIO version

4.9.0.post32

### Python version

3.14.0b2

### What happened?

```
============================= test session starts ==============================
platform linux -- Python 3.14.0b2, pytest-8.3.5, pluggy-1.6.0
rootdir: .../anyio
configfile: pyproject.toml
plugins: anyio-4.9.0.post32, hypothesis-6.131.28
collected 12 items

tests/test_to_interpreter.py FsFFFsFFFsFF                                [100%]

=================================== FAILURES ===================================
____________________________ test_run_sync[asyncio] ____________________________
tests/test_to_interpreter.py:37: in test_run_sync
    interpreter_id, _ = await to_interpreter.run_sync(_interpreters.get_current)
venv/lib64/python3.14/site-packages/anyio/to_interpreter.py:189: in run_sync
    return await worker.call(func, args, limiter)
venv/lib64/python3.14/site-packages/anyio/to_interpreter.py:126: in call
    result, is_exception = await to_thread.run_sync(
venv/lib64/python3.14/site-packages/anyio/to_thread.py:56: in run_sync
    return await get_async_backend().run_sync_in_worker_thread(
venv/lib64/python3.14/site-packages/anyio/_backends/_asyncio.py:2470: in run_sync_in_worker_thread
    return await future
venv/lib64/python3.14/site-packages/anyio/_backends/_asyncio.py:967: in run
    result = context.run(func, *args)
venv/lib64/python3.14/site-packages/anyio/to_interpreter.py:104: in _call
    self.initialize()
venv/lib64/python3.14/site-packages/anyio/to_interpreter.py:75: in initialize
    self._queue_id = queues.create(2, FMT_UNPICKLED, UNBOUND)
E   ValueError: unsupported unboundop 0
_________________________ test_run_sync[asyncio+eager] _________________________
tests/test_to_interpreter.py:37: in test_run_sync
    interpreter_id, _ = await to_interpreter.run_sync(_interpreters.get_current)
venv/lib64/python3.14/site-packages/anyio/to_interpreter.py:189: in run_sync
    return await worker.call(func, args, limiter)
venv/lib64/python3.14/site-packages/anyio/to_interpreter.py:126: in call
[...]

PR #927, opened 2025-05-27
Adjust to _interpqueues API changes in Python 3.14.0b2+

hroncok:
## Changes

Fixes https://github.com/agronholm/anyio/issues/926

## Checklist

features)

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/src/anyio/to_interpreter.py
+++ b/src/anyio/to_interpreter.py
@@ -86,9 +86,6 @@ class Worker:
             self._interpreter_id,
             {
                 "queue_id": self._queue_id,
-                "FMT_PICKLED": FMT_PICKLED,
-                "FMT_UNPICKLED": FMT_UNPICKLED,
-                "UNBOUND": UNBOUND,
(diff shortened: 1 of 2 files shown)


agronholm:
Doesn't look like the runner has 3.14.0b2 yet.

src/anyio/to_interpreter.py
```diff
@@ -81,6 +89,8 @@ def initialize(self) -> None:
                 "FMT_PICKLED": FMT_PICKLED,
                 "FMT_UNPICKLED": FMT_UNPICKLED,
                 "UNBOUND": UNBOUND,
```
agronholm, line 91: We don't need these here anymore.
```suggestion
```

agronholm:
I verified this locally with a hand built 3.14.0b2. Now you just need to add the changelog entry and mark this ready for review.

hroncok:
Should I squash?

agronholm (approved):
Thanks!

PR #928, opened 2025-05-31
Enabled flake8-bugbear rules for Ruff

### LINKED ISSUE #816: return in finally swallows exceptions
### Things to check first

### AnyIO version

mater branch

### Python version

NA

### What happened?

In https://github.com/agronholm/anyio/blob/3a62738be0ba8d7934bd447e3833d644d414d49a/src/anyio/from_thread.py#L119 there is a `return` statement in a `finally` block, which would swallow any in-flight exception.

This means that if an unhandled exception (including a `BaseException` such as `KeyboardInterrupt`) is raised from the `try` body, it will not propagate on as expected.

If the intention is to suppress all exceptions, I would propose to make this clear by using `except BaseException`.

See also https://docs.python.org/3/tutorial/errors.html#defining-clean-up-actions.

NA

agronholm:
## Changes
This enables the flake8-bugbear rules in Ruff and fixes the errors it reported.

Fixes #816. 

## Checklist

features)

these instructions.

### Updating the changelog

```
```


src/anyio/_backends/_asyncio.py
```diff
@@ -219,7 +219,7 @@ def run(self, coro: Coroutine[T_Retval], *, context=None) -> T_Retval:
                 if self._interrupt_count > 0:
                     uncancel = getattr(task, "uncancel", None)
                     if uncancel is not None and uncancel() == 0:
-                        raise KeyboardInterrupt()
+                        raise KeyboardInterrupt from None
```
graingert: Isn't this supposed to behave the same as asyncio.run? I want to see the context of the cancellation here

src/anyio/from_thread.py
```diff
@@ -116,7 +116,8 @@ async def run_async_cm(self) -> bool | None:
             # `_BlockingAsyncContextManager.__exit__` is called, and an
             # `_exit_exc_info` has been set.
             result = await self._async_cm.__aexit__(*self._exit_exc_info)
-            return result
+
+        return result
```
graingert, line 120: Do you think it's worth it to have a test for this?

src/anyio/_backends/_asyncio.py
agronholm: Perhaps I should set the `__cause__` then rather than `__context__`?

graingert: I'd be tempted to ignore the lint here, maybe even backport Runner and use that?

src/anyio/from_thread.py
agronholm: In `tests/test_from_thread.py`, there are the `test_async_context_manager*` tests that should cover this. But I'll check the coverage directly too.

src/anyio/_backends/_asyncio.py
agronholm: Haven't I already backported the `Runner`, at the top of the module?

agronholm: I've reverted this particular change and added the #noqa.

src/anyio/from_thread.py
agronholm: As mentioned on Gitter, these cases are indeed covered by tests.

graingert (approved):

PR #929, opened 2025-06-01
Added class methods for wrapping sockets

### LINKED ISSUE #845: add the ability to wrap stdlib sockets

agronholm:
## Changes

Allows wrapping existing sockets as AnyIO streams or listeners.

Fixes #845. 

## Checklist

features)

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/docs/networking.rst
+++ b/docs/networking.rst
@@ -241,6 +241,20 @@ having to pass the path every time you send data to the peer::
 
     run(main)
 
+Wrapping existing sockets as streams or listeners
+-------------------------------------------------
+
+In some cases, you might want to create a socket in third party code and wrap that as an
+AnyIO stream or socket listener. For that, various class methods exists:
+
+* :meth:`.abc.SocketListener.from_socket`
+* :meth:`.abc.SocketStream.from_socket`
+* :meth:`.abc.UNIXSocketStream.from_socket`
+* :meth:`.abc.UDPSocket.from_socket`
+* :meth:`.abc.ConnectedUDPSocket.from_socket`
+* :meth:`.abc.UNIXDatagramSocket.from_socket`
+* :meth:`.abc.ConnectedUNIXDatagramSocket.from_socket`
+
 Abstracting remote connections using Connectables
 -------------------------------------------------
 
--- a/docs/networking.rst
+++ b/docs/networking.rst
@@ -245,7 +245,7 @@ Wrapping existing sockets as streams or listeners
 -------------------------------------------------
 
 In some cases, you might want to create a socket in third party code and wrap that as an
-AnyIO stream or socket listener. For that, various class methods exists:
+AnyIO stream or socket listener. For that, various class methods exist:
 
 * :meth:`.abc.SocketListener.from_socket`
 * :meth:`.abc.SocketStream.from_socket`
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -5,6 +5,15 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
 **UNRELEASED**
 
+- Added various class methods to wrap existing sockets as listeners or socket streams:
+
+  * ``SocketListener.wrap_socket()``
+  * ``SocketStream.wrap_socket()``
+  * ``UNIXSocketStream.wrap_socket()``
+  * ``UDPSocket.wrap_socket()``
+  * ``ConnectedUDPSocket.wrap_socket()``
+  * ``UNIXDatagramSocket.wrap_socket()``
+  * ``ConnectedUNIXDatagramSocket.wrap_socket()``
 - Added a hierarchy of connectable stream classes for transparently connecting to
   various remote or local endpoints for exchanging bytes or objects
 - Added context manager mix-in classes (``anyio.ContextManagerMixin`` and
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -7,13 +7,13 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
 - Added various class methods to wrap existing sockets as listeners or socket streams:
 
-  * ``SocketListener.wrap_socket()``
-  * ``SocketStream.wrap_socket()``
-  * ``UNIXSocketStream.wrap_socket()``
-  * ``UDPSocket.wrap_socket()``
-  * ``ConnectedUDPSocket.wrap_socket()``
-  * ``UNIXDatagramSocket.wrap_socket()``
-  * ``ConnectedUNIXDatagramSocket.wrap_socket()``
+  * ``SocketListener.from_socket()``
+  * ``SocketStream.from_socket()``
+  * ``UNIXSocketStream.from_socket()``
+  * ``UDPSocket.from_socket()``
+  * ``ConnectedUDPSocket.from_socket()``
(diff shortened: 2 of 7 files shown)


docs/networking.rst
```diff
@@ -241,6 +241,20 @@ having to pass the path every time you send data to the peer::
 
     run(main)
 
+Wrapping existing sockets as streams or listeners
+-------------------------------------------------
+
+In some cases, you might want to create a socket in third party code and wrap that as an
+AnyIO stream or socket listener. For that, various class methods exists:
```
graingert:
```suggestion
AnyIO stream or socket listener. For that, various class methods exist:
```

docs/versionhistory.rst
```diff
@@ -5,6 +5,15 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
 **UNRELEASED**
 
+- Added various class methods to wrap existing sockets as listeners or socket streams:
+
+  * ``SocketListener.wrap_socket()``
```
graingert: from_socket or wrap_socket?

src/anyio/abc/_sockets.py
```diff
@@ -31,6 +32,64 @@
...
+    if sock.type != sock_type:
+        raise ValueError(
+            f"socket type mismatch: expected {sock_type.name}, got {sock.type.name}"
+        )
+
+    if require_connected:
+        try:
+            sock.getpeername()
+        except OSError:
+            raise ValueError("the socket must be connected") from None
```
graingert: Do you need to close the sock in this case to avoid a ResourceWarning?

src/anyio/abc/_sockets.py
```diff
@@ -31,6 +32,64 @@
...
+    if require_bound:
+        try:
+            if sock.family in (socket.AF_INET, socket.AF_INET6):
+                bound_addr = sock.getsockname()[1]
+            else:
+                bound_addr = sock.getsockname()
+        except OSError:
+            bound_addr = None
+
+        if not bound_addr:
```
graingert: And this one

src/anyio/abc/_sockets.py
graingert: Eg if it's an fd being passed the sock will get closed by the GC, so maybe you need to detach the fd again?

docs/versionhistory.rst
agronholm: Fixed.

src/anyio/abc/_sockets.py
agronholm: I'm not sure I follow. If there's a problem wrapping the socket, then I don't think that we should be closing the fd.

src/anyio/abc/_sockets.py
agronholm: As above.

src/anyio/abc/_sockets.py
```diff
@@ -31,6 +32,71 @@
...
+        raise TypeError(
+            f"expected an int or socket, got {type(sock_or_fd).__qualname__} instead"
+        )
+
+    try:
+        if require_connected:
+            try:
+                sock.getpeername()
+            except OSError:
+                raise ValueError("the socket must be connected") from None
```
leon1995: why raise `from None` instead of the `OSError` as done above?

agronholm: I felt that the `OSError` was an unnecessary background detail.

leon1995: I can understand that, but you may hide the information why the socket is not connected. It can be a useful hint for the user.

agronholm: WDYM, why? I'm not following.

leon1995: e.g. if the user expects the socket to be connected because of some precondition, but then the creation fails and just the value error is raised. Then, the user would not know why the socket is not connected.

Does the [`sock.getpeername()`](https://docs.python.org/3/library/socket.html#socket.socket.getpeername) method returns always the remote peer name once it has been connected at least once and does not reflect the current state of the connection? Because if so, then the `ValueError` is just fine

agronholm: `sock.getpeername()` returns the remote address only if the socket is currently in a connected state. Otherwise `OSError` is raised. If the socket is not connected, the `OSError` does not give away any additional information. That's why I chose to omit it.

leon1995: ok. thanks for the explanation

agronholm: Hmm – the socket could also be in a closed state, rather than just not connected. That could justify adding the `OSError` as the `__cause__`.

leon1995: I think it does not "hurt" to add the information. Then you have it when you need it and can be ignored when not needed

agronholm: I updated the code now.

agronholm: You were right about one thing – after calling `sock.shutdown(SHUT_RDWR)`, you can still call `sock.getpeername()` to get the host it used to be connected to. After closing the socket this no longer works however.

agronholm:
> would be nice, if this could be merged soon :)

As soon as it gets an approving review.

smurfix:
> > would be nice, if this could be merged soon :)
> 
> As soon as it gets an approving review.

Looks good to me. 🙂

agronholm:
> > > would be nice, if this could be merged soon :)
> > 
> > 
> > As soon as it gets an approving review.
> 
> Looks good to me. 🙂

@graingert needs to confirm that I've addressed his concerns appropriately.

src/anyio/abc/_sockets.py
```diff
@@ -31,6 +32,71 @@
...
+    if isinstance(sock_or_fd, int):
+        try:
+            sock = socket.socket(fileno=sock_or_fd)
+        except OSError as exc:
+            if require_connected:
+                raise ValueError("the socket must be connected") from exc
+            elif require_bound:
+                raise ValueError("the socket must be bound to a local address") from exc
+            else:
+                raise
```
Zac-HD: This seems likely to cause confusion in cases where e.g. `exc.errno == errno.ENOTSOCK`, if the user pased the fd for a file rather than a socket.  Maybe use `exc.add_note()` rather than `raise ... from exc`?

Zac-HD (approved):
Small note below, otherwise looks good (though I'm not a sockets expert).

src/anyio/abc/_sockets.py
agronholm: Good catch. I've added an explicit message (and an accompanying test) to cover this case.

PR #930, opened 2025-06-04
implement tg.start_context

### LINKED ISSUE #678: Stricter typing for `TaskGroup.start()`

rmorshea:
## Changes

Fixes #678

This is accomplished by adding a `TaskGroup.start_context` method as a better-typed alternative to `TaskGroup.start`. The implementation for `start_context` is essentially:

```python
class TaskGroup:
    async def start_context(self, ctx: AbstractAsyncContextManager[T], *, name=None) -> T:
        async def wrapper(*, task_status):
            async with ctx as ret:
                task_status.started(ret)
        return await self.start(wrapper, ctx, name=name)
```

This also has the additional benefit of not being limited to positional only args since the user gets to construct the context manager however they'd like:

```python
    value = await tg.start_context(my_context(*a, **kw))
```

## Checklist

features)

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -54,7 +54,7 @@ jobs:
     - name: Install dependencies
       run: pip install -e . pyright pytest
     - name: Run pyright
-      run: pyright --verifytypes anyio
+      run: pyright --ignoreexternal --verifytypes anyio
 
   test:
     strategy:
--- a/docs/tasks.rst
+++ b/docs/tasks.rst
@@ -91,8 +91,47 @@ that is calling with :meth:`TaskGroup.start() <.abc.TaskGroup.start>` will be bl
 until then. If the spawned task never calls it, then the
 :meth:`TaskGroup.start() <.abc.TaskGroup.start>` call will raise a ``RuntimeError``.
 
-.. note:: Unlike :meth:`~.abc.TaskGroup.start_soon`, :meth:`~.abc.TaskGroup.start` needs
-   an ``await``.
+Alternatively, you can use :meth:`TaskGroup.start_context() <.abc.TaskGroup.start_context>`
+to achieve the same thing without needing to accept a ``task_status`` argument in the
+target coroutine function. Instead, initialization is signaled by the context
+manager having been entered. Further the value returned by the ``tg.start_context()``
+call is also the value returned by the context manager's ``__aenter__()`` method. When
+using the ``@asynccontextmanager`` decorator, this is the value yielded by the
+async generator function::
+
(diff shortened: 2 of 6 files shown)


rmorshea:
Not really sure what's up with the failure. Seems like Pyright is interrogating PyTest's type rather than AnyIO's. For the Python 3.11 I ran that on my machine successfully - a re-run would likely fix that but I don't seem able to via the GitHub actions UI for some reason.

rmorshea:
Added `--ignoreexternal` to the Pyright check which is used to "Ignore external imports for --verifytypes" according to the [Pyright docs](https://microsoft.github.io/pyright/#/command-line?id=pyright-command-line-options).

rmorshea:
On second thought, if anyio is being checked by Pyright now perhaps we can just switch to the annotated suggested at the top of #678 instead.

PR #931, opened 2025-06-04
Stricter typing for TaskGroup.start()

### LINKED ISSUE #678: Stricter typing for `TaskGroup.start()`

rmorshea:
## Changes

Fixes #678

## Checklist

features)

these instructions.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -54,7 +54,7 @@ jobs:
     - name: Install dependencies
       run: pip install -e . pyright pytest
     - name: Run pyright
-      run: pyright --verifytypes anyio
+      run: pyright --ignoreexternal --verifytypes anyio
 
   test:
     strategy:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -55,6 +55,8 @@ jobs:
       run: pip install -e . pyright pytest
     - name: Run pyright
       run: pyright --ignoreexternal --verifytypes anyio
+    - name: Run pyright tests
+      run: pyright tests
 
   test:
     strategy:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,49 +4,45 @@
 # * Copy this file as ".pre-commit-config.yaml"
 # * Run "pre-commit install".
 repos:
-  - repo: https://github.com/adrienverge/yamllint
-    rev: v1.37.1
-    hooks:
-      - id: yamllint
-        args: ['-d {extends: relaxed, rules: {line-length: disable}}', '-s']
+    - repo: https://github.com/adrienverge/yamllint
+      rev: v1.37.1
+      hooks:
+          - id: yamllint
+            args: ["-d {extends: relaxed, rules: {line-length: disable}}", "-s"]
 
-  - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v5.0.0
-    hooks:
-      - id: check-toml
-      - id: check-yaml
-      - id: debug-statements
-      - id: end-of-file-fixer
-      - id: mixed-line-ending
-        args: [ "--fix=lf" ]
-      - id: trailing-whitespace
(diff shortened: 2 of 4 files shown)


.github/workflows/test.yml
```diff
@@ -54,7 +54,9 @@ jobs:
     - name: Install dependencies
       run: pip install -e . pyright pytest
     - name: Run pyright
-      run: pyright --verifytypes anyio
+      run: pyright --ignoreexternal --verifytypes anyio
```
agronholm, line 57: What's the extra option for?

.pre-commit-config.yaml
```diff
@@ -4,49 +4,45 @@
...
-    hooks:
-      - id: rst-backticks
-      - id: rst-directive-colons
-      - id: rst-inline-touching-normal
+    - repo: https://github.com/pre-commit/pygrep-hooks
+      rev: v1.10.0
+      hooks:
+          - id: rst-backticks
+          - id: rst-directive-colons
+          - id: rst-inline-touching-normal
```
agronholm, line 48: Please undo this. It's uncalled for.

docs/versionhistory.rst
```diff
@@ -30,6 +30,8 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
   (`#926 <https://github.com/agronholm/anyio/issues/926>`_; PR by @hroncok)
 - Fixed ``SyntaxWarning`` on Python 3.14 about ``return`` in ``finally``
   (`#816 <https://github.com/agronholm/anyio/issues/816>`_)
+- Added stricter type annotations for ``TaskGroup.start()``.
+  Fixes `#678 <https://github.com/agronholm/anyio/issue/678>`_.
```
agronholm, line 34: Needs a rebase against `master`.

agronholm (changes requested):
Looks good save for the pointers I left. Sorry I took this long to review.

agronholm:
I took a better look, and there are showstopper problems with this:

1. The protocol requires `task_status` to be keyword-only argument which we don't currently require
2. The protocol is incompatible with the `TASK_STATUS_IGNORED` default value
3. The protocol requires a return type of `None` which we don't currently require

agronholm:
I don't think it's possible with the current typing system of Python :face_with_diagonal_mouth:

PR #938, opened 2025-07-03
fix: typo in BrokenWorkerIntepreter

ayussh-verma:
## Changes

Renames BrokenWorkerIntepreter to BrokenWorkerInterpreter;

## Checklist

features)

### DIFF
--- a/src/anyio/__init__.py
+++ b/src/anyio/__init__.py
@@ -89,6 +89,18 @@ for __value in list(locals().values()):
     if getattr(__value, "__module__", "").startswith("anyio."):
         __value.__module__ = __name__
 
-del __value
 
-BrokenWorkerIntepreter = BrokenWorkerInterpreter
+def __getattr__(attr: str):  # type: ignore[no-untyped-def]
+    """Support deprecated aliases."""
+    if attr == "BrokenWorkerIntepreter":
+        replaced_class_name = BrokenWorkerInterpreter.__name__
+        import warnings  # A lazy import here cuts the overall import time
+
+        warnings.warn(
+            f"The 'BrokenWorkerIntepreter' alias is deprecated, use {replaced_class_name!r} instead.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        return BrokenWorkerInterpreter
+
+    raise AttributeError(f"module {__name__!r} has no attribute {attr!r}")
--- a/src/anyio/__init__.py
+++ b/src/anyio/__init__.py
@@ -96,11 +96,10 @@ del __value
 def __getattr__(attr: str) -> type[BrokenWorkerInterpreter]:
     """Support deprecated aliases."""
     if attr == "BrokenWorkerIntepreter":
-        replaced_class_name = BrokenWorkerInterpreter.__name__
         import warnings
 
         warnings.warn(
-            f"The 'BrokenWorkerIntepreter' alias is deprecated, use {replaced_class_name!r} instead.",
+            "The 'BrokenWorkerIntepreter' alias is deprecated, use 'BrokenWorkerInterpreter' instead.",
             DeprecationWarning,
             stacklevel=2,
         )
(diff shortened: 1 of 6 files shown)


src/anyio/__init__.py
```diff
@@ -10,7 +10,7 @@
 from ._core._eventloop import sleep_forever as sleep_forever
 from ._core._eventloop import sleep_until as sleep_until
 from ._core._exceptions import BrokenResourceError as BrokenResourceError
-from ._core._exceptions import BrokenWorkerIntepreter as BrokenWorkerIntepreter
```
agronholm, line 13: This must be also made available with the old name for backwards compatibility.

ayussh-verma: Hi Alex, that's already done. It's re-assigned at the bottom of the file

ayussh-verma: I missed the `del` logic though, I think the reassignment needs to be done before it
edit: no looks like its not needed

agronholm: Hm, how about implementing this alias with `__getattribute__()` to emit a deprecation warning when the old name is accessed?

ayussh-verma: Sounds good, will make the change tomorrow 👍

src/anyio/__init__.py
```diff
@@ -89,6 +89,18 @@
...
         __value.__module__ = __name__
 
-del __value
 
-BrokenWorkerIntepreter = BrokenWorkerInterpreter
+def __getattr__(attr: str):  # type: ignore[no-untyped-def]
+    """Support deprecated aliases."""
+    if attr == "BrokenWorkerIntepreter":
+        replaced_class_name = BrokenWorkerInterpreter.__name__
+        import warnings  # A lazy import here cuts the overall import time
```
agronholm: False: several AnyIO modules that are imported in `__init__.py` already import `warnings`.

src/anyio/__init__.py
```diff
@@ -89,4 +89,18 @@
     if getattr(__value, "__module__", "").startswith("anyio."):
         __value.__module__ = __name__
 
-del __value
```
agronholm, line 92: Why did you remove this?

src/anyio/__init__.py
```diff
@@ -89,4 +89,18 @@
...
+
+def __getattr__(attr: str):  # type: ignore[no-untyped-def]
```
agronholm: This is the only thing it returns, might as well use it as a return type hint?
```suggestion
def __getattr__(attr: str) -> type[BrokenWorkerInterpreter]:
```

agronholm:
Please also add a test for the import, using `with pytest.warns(DeprecationWarning, ...):`

src/anyio/__init__.py
agronholm: That said, importing it lazily avoids having it exposed in the module namespace.

ayussh-verma: Removed the comment for accuracy 👍

src/anyio/__init__.py
```diff
@@ -89,4 +89,21 @@
...
+
+
+def __getattr__(attr: str) -> type[BrokenWorkerInterpreter]:
+    """Support deprecated aliases."""
+    if attr == "BrokenWorkerIntepreter":
+        replaced_class_name = BrokenWorkerInterpreter.__name__
+        import warnings
+
+        warnings.warn(
+            f"The 'BrokenWorkerIntepreter' alias is deprecated, use {replaced_class_name!r} instead.",
```
agronholm: There is no point making this dynamic.
```suggestion
        import warnings

        warnings.warn(
            "The 'BrokenWorkerIntepreter' alias is deprecated, use 'BrokenWorkerInterpreter' instead.",
```

ayussh-verma:
Just waiting for the CI/CD to pass, it failed randomly last time

ayussh-verma:
Albeit the timeout looked unrelated

agronholm (approved):
Thanks!

PR #956, opened 2025-08-03
Updated the to_interpreter module to use the public API on Python 3.14

agronholm:
## Changes

The `to_interpreter` module has been refactored to use the new public API from `concurrent.interpreters` on Python 3.14 and later.

## Checklist

features)

these instructions.

### Updating the changelog

```
```


richardsheridan (approved):
Looks like the new API makes things much cleaner, so not much to comment on here. Maybe processes and interpreters should share the same default limiter, since presumably users will be using both tools primarily to dodge the GIL?

agronholm:
> Looks like the new API makes things much cleaner, so not much to comment on here. Maybe processes and interpreters should share the same default limiter, since presumably users will be using both tools primarily to dodge the GIL?

Combining them now would constitute a behavioral change that might, in theory, negatively affect some apps that *do* use both mechanisms.

PR #960, opened 2025-08-05
add winloop for windows if use_uvloop is used on windows

Vizonex:
## Changes

Fixes #959 

This attempts to add winloop support to anyio. And is my second attempt. This time I not hook pre-commit so that stressful things don't happen again. I am going to be doing these steps one at a time so that I can be guided in the right direction instead of screwing up.

## Checklist

features)

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -49,7 +49,7 @@ asyncio_params = [
     pytest.param(
         ("asyncio", {"debug": True, "loop_factory": winloop.new_event_loop}),
         marks=winloop_marks,
-        id="asyncio+winloop",
+        id="asyncio+uvloop",
     ),
 ]
 if sys.version_info >= (3, 12):
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -47,7 +47,7 @@ asyncio_params = [
         id="asyncio+uvloop",
     ),
     pytest.param(
-        ("winloop", {"debug": True, "loop_factory": winloop.new_event_loop}),
+        ("asyncio", {"debug": True, "loop_factory": winloop.new_event_loop}),
         marks=winloop_marks,
         id="asyncio+winloop",
     ),
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -30,6 +30,20 @@ else:
             pytest.mark.skip(reason="uvloop is missing shutdown_default_executor()")
         )
 
+winloop_marks = []
+try:
+    import winloop
+except ImportError:
+    winloop_marks.append(pytest.mark.skip(reason="winloop not available"))
+    winloop = Mock()
+else:
+    if hasattr(asyncio.AbstractEventLoop, "shutdown_default_executor") and not hasattr(
+        winloop.loop.Loop, "shutdown_default_executor"
+    ):
+        winloop_marks.append(
+            pytest.mark.skip(reason="winloop is missing shutdown_default_executor()")
+        )
+
 pytest_plugins = ["pytester"]
 
 asyncio_params = [
@@ -39,6 +53,11 @@ asyncio_params = [
         marks=uvloop_marks,
         id="asyncio+uvloop",
     ),
+    pytest.param(
+        ("winloop", {"debug": True, "loop_factory": winloop.new_event_loop}),
+        marks=winloop_marks,
+        id="asyncio+winloop"
+    )
 ]
 if sys.version_info >= (3, 12):
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -91,13 +91,18 @@ def blockbuster() -> Iterator[BlockBuster | None]:
 
         yield bb
 
+
 @pytest.fixture
 async def skip_if_winloop():
-    import sniffio
     import platform
+
+    import sniffio
+
     if sniffio.current_async_library() == "asyncio" and platform.system() == "Windows":
         import asyncio
+
         import winloop
+
         if isinstance(asyncio.get_running_loop(), winloop.Loop):
             pytest.skip("winloop is currently broken")
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import asyncio
+import platform
 import ssl
 import sys
 from collections.abc import Generator, Iterator
@@ -9,6 +10,7 @@ from typing import TYPE_CHECKING, Any
 from unittest.mock import Mock
 
 import pytest
+import sniffio
 import trustme
 from _pytest.fixtures import SubRequest
 from trustme import CA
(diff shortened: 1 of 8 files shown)


Vizonex:
@agronholm I'll make sure to grab what I previously had for documentation about winloop real quick.

tests/conftest.py
```diff
@@ -39,6 +53,11 @@
         marks=uvloop_marks,
         id="asyncio+uvloop",
     ),
+    pytest.param(
+        ("winloop", {"debug": True, "loop_factory": winloop.new_event_loop}),
```
agronholm: There's no backend named "winloop".
```suggestion
        ("asyncio", {"debug": True, "loop_factory": winloop.new_event_loop}),
```

tests/conftest.py
```diff
@@ -30,6 +30,20 @@
...
+except ImportError:
+    winloop_marks.append(pytest.mark.skip(reason="winloop not available"))
+    winloop = Mock()
+else:
+    if hasattr(asyncio.AbstractEventLoop, "shutdown_default_executor") and not hasattr(
+        winloop.loop.Loop, "shutdown_default_executor"
+    ):
+        winloop_marks.append(
+            pytest.mark.skip(reason="winloop is missing shutdown_default_executor()")
+        )
```
agronholm: Does `winloop` have releases that lack `shutdown_default_executor()`?

Vizonex: Actually I don't think it does. Let me remove this

tests/test_sockets.py
```diff
@@ -2195,6 +2195,7 @@ async def test_getaddrinfo() -> None:
 
 
 @pytest.mark.parametrize("sock_type", [socket.SOCK_STREAM, socket.SOCK_STREAM])
+@pytest.mark.skipif(sys.platform == "win32", reason="getaddrinfo can be incorrect on winloop")
```
agronholm: Only skip it on winloop, not on vanilla asyncio

agronholm: Better yet, fix the test logic to check for the correct values

Vizonex: Oh ok my bad Maybe I should make a wrapper that skips if on winloop then?

agronholm: Why can't you just fix the logic on line 2205?

tests/test_subprocesses.py
```diff
@@ -24,6 +24,7 @@
 pytestmark = pytest.mark.anyio
 
 
+@pytest.mark.skipif(sys.platform == "win32", reason="winloop fails this shell test")
```
agronholm: You shouldn't skip this universally on Windows if it only fails on winloop

tests/test_subprocesses.py
```diff
@@ -275,6 +276,9 @@ async def test_process_aexit_cancellation_closes_standard_streams(
         await process.stderr.receive(1)
 
 
+@pytest.mark.skipif(
+    sys.platform == "win32", reason="winloop doesn't have have umask or special kwargs"
+)
```
agronholm: As above, no skipping this on all loop implementations

tests/test_sockets.py
```diff
@@ -2195,11 +2197,9 @@ async def test_getaddrinfo() -> None:
 
 
 @pytest.mark.parametrize("sock_type", [socket.SOCK_STREAM, socket.SOCK_STREAM])
-@pytest.mark.skipif(
-    sys.platform == "win32", reason="getaddrinfo can be incorrect on winloop"
-)
 async def test_getaddrinfo_ipv6addr(
     sock_type: Literal[socket.SocketKind.SOCK_STREAM],
+    skip_if_winloop
```
agronholm: You could've just followed the existing examples of skipping when uvloop is detected, yes?

tests/test_sockets.py
```diff
@@ -2196,7 +2196,7 @@ async def test_getaddrinfo() -> None:
 
 @pytest.mark.parametrize("sock_type", [socket.SOCK_STREAM, socket.SOCK_STREAM])
 async def test_getaddrinfo_ipv6addr(
-    sock_type: Literal[socket.SocketKind.SOCK_STREAM],
+    sock_type: Literal[socket.SocketKind.SOCK_STREAM], skip_if_winloop
 ) -> None:
     # IDNA trips up over raw IPv6 addresses
     proto = 0 if platform.system() == "Windows" else 6
```
agronholm: Winloop does this correctly so the expected value should be 6

tests/test_subprocesses.py
```diff
@@ -309,6 +312,7 @@ async def test_py39_arguments(
     argvalue_factory: Callable[[], Any],
     anyio_backend_name: str,
     anyio_backend_options: dict[str, Any],
+    skip_if_winloop: None,
```
agronholm: Instead of this, check for winloop like we're already checking for uvloop:
`and anyio_backend_options["loop_factory"].__module__ in ("uvloop", "winloop")`

agronholm:
I will approve this once you change the test skipping to use the same method as with uvloop. It's okay to add a utility function in the test suite to check for winloop/uvloop as you would now need to duplicate the same nontrivial logic.

tests/test_subprocesses.py
```diff
@@ -42,6 +42,11 @@
 async def test_run_process(
     shell: bool, command: str | list[str], anyio_backend_name: str
 ) -> None:
+    if anyio_backend_name == "asyncio" and sys.platform == "win32":
+        import asyncio
+
+        if asyncio.get_event_loop().__module__ == "winloop":
+            pytest.skip(reason="winloop produces a non-zero exit with status 1")
```
agronholm: Why would winloop fail with this basic functionality?

Vizonex:
is there anything left or is this mergeable?

Vizonex:
> Why is there an added winloop skip in `test_run_process()`? It makes no sense. Does running subprocesses really not work on winloop?
there seems to be a bug that doesn't want it to pass and seems to be a problem with the library itself maybe. I'll see about making a duplicate of the test but basically the one that is failing is tests/test_subprocesses.py::test_run_process[asyncio+uvloop-shell] for some strange reason. I'll see about patching it in a later update.

agronholm:
> > Why is there an added winloop skip in `test_run_process()`? It makes no sense. Does running subprocesses really not work on winloop?
> 
> @agronholm there seems to be a bug that doesn't want it to pass and seems to be a problem with the library itself maybe. I'll see about making a duplicate of the test but basically the one that is failing is tests/test_subprocesses.py::test_run_process[asyncio+uvloop-shell] for some strange reason. I'll see about patching it in a later update.

I'd really like to see this fixed in winloop before this PR is merged. The next release of AnyIO is some time away given how v4.10.0 was just released, so there's plenty of time.

agronholm (approved):

Vizonex:
@agronholm Looks like my long 2 month journey to get this in is finally finished. What a relief. I'll be able to focus on some of my other goals now thanks to this being completed.

agronholm:
Thanks!

PR #961, opened 2025-08-06
Use cancel_shielded_checkpoint in MemoryObjectStream.send() to prevent cancellation race conditions

ShijunDeng:
#Description
The send() method in MemoryObjectStream currently uses a standard await checkpoint(), which allows cancellation to propagate immediately after the checkpoint. This can lead to race conditions where the task is cancelled after the checkpoint but before critical internal state updates (e.g., modifying the stream's internal queue or signaling waiting receivers). This may leave the stream in an inconsistent state or trigger unhandled exceptions.

# Problem
If the task is cancelled between await checkpoint() and the completion of internal state updates (e.g., appending an item to the stream or notifying receivers), the following issues may occur:

- The stream's internal state (e.g., queues or event counters) may become inconsistent.
- Waiting receivers may miss updates or raise unexpected errors.
- Unhandled CancelError exceptions may propagate up the call stack, causing crashes.

# Solution
Replace await checkpoint() with await anyio.lowlevel.cancel_shielded_checkpoint() to ensure cancellation is deferred until the internal state updates are safely completed. This guarantees atomicity for critical operations and prevents the stream from entering an inconsistent state during cancellation.

# Related Issues
https://github.com/modelcontextprotocol/python-sdk/issues/144

# Program error message
```python
  File "D:\Python3.11\Lib\site-packages\mcp\client\session.py", line 181, in send_ping
    return await self.send_request(
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\Python3.11\Lib\site-packages\mcp\shared\session.py", line 261, in send_request
    await self._write_stream.send(SessionMessage(message=JSONRPCMessage(jsonrpc_request), metadata=metadata))
  File "D:\Python3.11\Lib\site-packages\anyio\streams\memory.py", line 240, in send
    await checkpoint()
  File "D:\Python3.11\Lib\site-packages\anyio\lowlevel.py", line 27, in checkpoint
    await get_async_backend().checkpoint()
  File "D:\Python3.11\Lib\site-packages\anyio\_backends\_asyncio.py", line 2326, in checkpoint
    await sleep(0)
  File "D:\Python3.11\Lib\asyncio\tasks.py", line 630, in sleep
    await __sleep0()
  File "D:\Python3.11\Lib\asyncio\tasks.py", line 624, in __sleep0
    yield
asyncio.exceptions.CancelledError: Cancelled by cancel scope 278ccfc1fd0
```

## Changes

Fixes #. 

## Checklist

features)

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/tests/streams/test_memory.py
+++ b/tests/streams/test_memory.py
@@ -187,7 +187,7 @@ async def test_cancel_send() -> None:
     async with create_task_group() as tg:
         tg.start_soon(send.send, "hello")
         await wait_all_tasks_blocked()
-        tg.cancel_scope.cancel()  
+        tg.cancel_scope.cancel()
 
     assert send._state.is_closed
     assert receive._state.is_closed


agronholm:
Did you just delete the PR template text?

agronholm:
Is this not reproduceable without looking at private attributes? Also, if you wanted to replace `checkpoint()` with something else, the way forward would be to call `checkpoint_if_cancelled()` before the operation, and then calling `cancel_shielded_checkpoint()` if no other yield point was called before the coroutine function returns.

agronholm:
Did the test suite pass on your local machine?

tests/streams/test_memory.py
```diff
@@ -169,9 +169,15 @@ async def test_cancel_receive() -> None:
...
         tg.cancel_scope.cancel()
 
+    assert receive._state.is_closed
+    assert send._state.is_closed
+
     with pytest.raises(WouldBlock):
         send.send_nowait("hello")
 
+    with pytest.raises(EndOfStream):
+        await receive.receive()
```
agronholm, line 179: This is just plain **wrong**. The send stream was never closed, so why would `receive.receive()` raise `EndOfStream`?

PR #964, opened 2025-08-13
Set `None` as the default type argument for `TaskStatus`

agronholm:
## Changes

This makes `None` the default type argument for the `TaskStatus` class, forcing users to specify an explicit type argument unless they plan to call `task_status.started()` without arguments.

This will likely break static type checks in existing projects, but nothing at run-time.

## Checklist

features)

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -3,6 +3,10 @@ Version history
 
 This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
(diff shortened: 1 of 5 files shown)


agronholm:
This makes our `TaskStatus` match Trio: https://github.com/python-trio/trio/blob/393d0ccaedd8adf4d3a7aca52fd6610d7f00b762/src/trio/_core/_run.py#L82

PR #979, opened 2025-09-16
fix(sockets): ensure same port is used for IPv4/IPv6 when local_port=0 (#857)

### LINKED ISSUE #857: `create_tcp_listener(local_port=0)`: return listeners with the same port
### Things to check first

### Feature description

> it would be nice if `multilistener = await create_tcp_listener(local_port=0)` used the same port for both of the listeners. This would be [implemented via something roughly] like
> 
> ```python
> def get_free_ipv4_ipv6_port(kind, bind_ipv4_addr, bind_ipv6_addr):
>     if platform_has_dualstack and bind_ipv4_addr == bind_ipv6_addr == "":
>         # We can use the dualstack_sock.bind(("", 0)) trick
>         ...
>     else:
>         ...
> ```
> 
> Currently, folks who need this have to implement it downstream.)

Ow, that's definitely something I would like to fix! Strange that nobody has reported that as an issue. I definitely meant for the listener to bind to the same ephemeral port if the local port was omitted.

_Originally posted by @agronholm in https://github.com/agronholm/anyio/pull/856#discussion_r1907675607_

### Use case

.

11kkw:
## Changes

Fixes #857

- Updated `create_tcp_listener` so that when `local_port=0`, IPv4 and IPv6 listeners are guaranteed to use the same ephemeral port.
- Added a guard to skip creating a separate IPv4 listener if an IPv6 dualstack listener was already bound.
- Ensured consistent behavior on platforms without dualstack support by reusing the same ephemeral port.
- Added new tests in `tests/test_sockets.py` to verify the same-port guarantee.

## Checklist

- [ ] You've updated the documentation (in `docs/`, in case of behavior changes or new features)  

### Updating the changelog

- Fix TCP listener behavior to guarantee same ephemeral port for IPv4 and IPv6 when `local_port=0`
  ([#857](https://github.com/agronholm/anyio/issues/857); PR by @yourgithubaccount)

### DIFF
--- a/tests/test_sockets.py
+++ b/tests/test_sockets.py
@@ -953,8 +953,10 @@ class TestTCPListener:
             assert len(ports) == 1
             if local_port != 0:
                 assert ports == {local_port}
+
             assert len(multi.listeners) == expected_listeners
 
+    @skip_ipv6_mark
     async def test_tcp_listener_dualstack_disabled(
         self, monkeypatch: MonkeyPatch
     ) -> None:


agronholm:
Oh, this is fantastic! This was precisely the next issue I was going to tackle for v4.11.

src/anyio/_core/_sockets.py
```diff
@@ -301,7 +301,6 @@ async def create_tcp_listener(
     :param reuse_port: ``True`` to allow multiple sockets to bind to the same
         address/port (not supported on Windows)
     :return: a multi-listener object containing one or more socket listeners
-
```
agronholm, line 304: The blank line in the docstring was intentional.

src/anyio/_core/_sockets.py
```diff
@@ -340,21 +344,39 @@ async def create_tcp_listener(
...
+                ephemeral_port = raw_socket.getsockname()[1]
+
             raw_socket.listen(backlog)
             listener = asynclib.create_tcp_listener(raw_socket)
             listeners.append(listener)
+
     except BaseException:
         for listener in listeners:
             await listener.aclose()
-
```
agronholm, line 357: I always add a blank line after a control block ends. Please don't "fix" my current code formatting.

src/anyio/_core/_sockets.py
```diff
@@ -340,21 +344,39 @@ async def create_tcp_listener(
...
+
             raw_socket.bind(sockaddr)
+
+            if local_port == 0 and ephemeral_port is None:
+                ephemeral_port = raw_socket.getsockname()[1]
+
             raw_socket.listen(backlog)
             listener = asynclib.create_tcp_listener(raw_socket)
             listeners.append(listener)
+
```
agronholm, line 410: This blank line is uncalled for.
```suggestion
```

src/anyio/_core/_sockets.py
```diff
@@ -340,21 +344,39 @@ async def create_tcp_listener(
 
             # If only IPv6 was requested, disable dual stack operation
             if fam == socket.AF_INET6:
-                raw_socket.setsockopt(IPPROTO_IPV6, socket.IPV6_V6ONLY, 1)
+                if (
+                    local_port == 0
+                    and local_host in (None, "::")
+                    and socket.has_dualstack_ipv6()
+                ):
+                    raw_socket.setsockopt(IPPROTO_IPV6, socket.IPV6_V6ONLY, 0)
```
agronholm: Isn't this option off by default?

agronholm:
Digging deeper, it appears that every operating system we can test on supports dual-stack IPv6, making this a non-issue really.

agronholm:
You also need to add the changelog entry (the third checkbox).

agronholm:
I think we can only use dual-stack binding when the local address is either `None`, `localhost` or `::`, yes? Assuming no specific IP family is requested of course.

src/anyio/_core/_sockets.py
11kkw: > Isn't this option off by default?

You’re right that on Linux the default is 0, but my understanding is that on Windows the default is actually 1 (IPv6-only), so I set it explicitly to 0 here to make dual-stack binding work consistently. According to the [Microsoft docs](https://learn.microsoft.com/en-us/windows/win32/winsock/ipproto-ipv6-socket-options?utm_source=chatgpt.com)
, the default on Windows is nonzero. That said, I may be mistaken, so please correct me if I’ve misunderstood.

agronholm: The docs speak for themselves, so it seems you were correct. I think we need a test for this nonetheless. For dual-stack bindings we would only have one listener, and I don't think we're testing for that anywhere.

agronholm: We always do name resolution for both protocols, and I think the local addresses `(None, "::", "localhost")` should be special cased to skip that and do dual-stack binding instead, wouldn't you agree?

tests/test_sockets.py
```diff
@@ -902,6 +915,48 @@ async def test_from_socket_not_bound(
...
+        ],
+    )
+    async def test_tcp_listener_dualstack_expected_listener_count(
+        self, local_host: str | None, expect_dualstack: bool
+    ) -> None:
+        async with await create_tcp_listener(local_host=local_host) as multi:
+            if expect_dualstack:
+                assert len(multi.listeners) == 1
+            else:
+                assert len(multi.listeners) >= 1
```
agronholm: I think more than 1 listener would be an error.
```suggestion
                assert len(multi.listeners) == 2
```

tests/test_sockets.py
```diff
@@ -1438,7 +1493,18 @@ async def handle(stream: SocketStream) -> None:
...
+
+                        # On Windows, connecting to ANY (:: or 0.0.0.0) is invalid (WinError 10049).
+                        # Replace it with loopback (::1 or 127.0.0.1) to make the test portable.
+                        if sys.platform == "win32" and host in ("::", "0.0.0.0"):
+                            family = listener.extra(SocketAttribute.family)
+                            if family == socket.AF_INET6:
+                                host = "::1"
+                            elif family == socket.AF_INET:
+                                host = "127.0.0.1"
+
```
agronholm: Let's simplify this a bit.
```suggestion
                        host, port = local_address
                        host = "::1" if listener.extra(SocketAttribute.family) == socket.AF_INET6 else "127.0.0.1"
```

src/anyio/_core/_sockets.py
```diff
@@ -306,6 +306,36 @@ async def create_tcp_listener(
...
+        else:
+            raw_socket.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+
+        if reuse_port:
+            raw_socket.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEPORT, 1)
+        return raw_socket
+
+    if (
+        local_host in (None, "::", "localhost")
+        and local_port == 0
```
agronholm: Why is dual-stack restricted to ephemeral ports?

src/anyio/_core/_sockets.py
```diff
@@ -306,6 +306,36 @@ async def create_tcp_listener(
...
+        and family == socket.AddressFamily.AF_UNSPEC
+        and socket.has_dualstack_ipv6()
+    ):
+        raw_socket = _setup_raw_socket(socket.AF_INET6)
+        raw_socket.setsockopt(IPPROTO_IPV6, socket.IPV6_V6ONLY, 0)
+        raw_socket.bind(("::", 0))
```
agronholm: Why does it bind to `::` even when the local address is `localhost`?

11kkw: > Why does it bind to `::` even when the local address is `localhost`?

I chose to bind to :: because localhost can resolve differently across platforms, sometimes to IPv4 only or IPv6 only, and I wanted to ensure both families were covered consistently. But maybe this isn’t the right approach — would you prefer it to follow the hostname resolution of localhost more strictly?

src/anyio/_core/_sockets.py
11kkw: > Why is dual-stack restricted to ephemeral ports?

I limited dual-stack to the local_port=0 case because the original request in issue #857 was specifically about ephemeral ports, and I was concerned that forcing dual-stack on explicit ports might cause platform-specific conflicts. But perhaps I’m overthinking this — do you think it would make more sense to allow dual-stack for fixed ports as well?

src/anyio/_core/_sockets.py
agronholm: Binding to `::` when `local_host == 'localhost'` is not only flat out wrong, it's dangerous. It opens the port to the outside world which is a security issue.

agronholm: In what case does `localhost` not resolve to `::1` or `127.0.0.1`?

src/anyio/_core/_sockets.py
agronholm: Is there any any reason *not to* use dual stack binding whenever possible?

src/anyio/_core/_sockets.py
agronholm: I did some digging and apparently this is a non-starter for `localhost` anyway, so dual-stack binding only works on the "any" address.

src/anyio/_core/_sockets.py
agronholm: Sorry for flip-flopping on this issue, but these nuances were apparently not clear to me either. I think it would make sense if only `local_host=None` used dual-stack binding, as it wouldn't make a lot of sense for `0.0.0.0` to only bind to IPv4 while `::` would then bind to both. Thoughts?

agronholm: Absolutely. But see my other comments elsewhere too.

agronholm:
I now remembered why I had that mandatory `getaddrinfo()` call: to handle situations where there were no IPv6 interfaces present. Even if `socket.has_dualstack_ipv6()`  is true, all it means is that the *kernel* supports this option, but it doesn't mean there is an IPv6 interface to bind to. The trouble is, I cannot find an environment where IPv6 is actually not available in order to test this properly.

tests/test_sockets.py
```diff
@@ -918,32 +918,108 @@ async def test_from_socket_not_bound(
...
-            local_host=local_host, family=family
+            local_host=local_host, family=family, local_port=local_port
         ) as multi:
             ports = {
                 listener.extra(SocketAttribute.local_port)
                 for listener in multi.listeners
             }
             assert len(ports) == 1
+            if local_port != 0:
+                assert list(ports)[0] == local_port
```
agronholm: Why are you only checking this if `local_port == 0`?

11kkw: > Why are you only checking this if `local_port == 0`?

As I understood it, when `local_port=0` the OS will assign an ephemeral port,  
so checking against the requested value wouldn’t make sense.  
That’s why I only added the assertion when `local_port != 0`.  
Or am I misunderstanding this?

agronholm: Sorry, I misspoke. My actual concern was that you're only checking the first port and not all of them.

11kkw: > Sorry, I misspoke. My actual concern was that you're only checking the first port and not all of them.

You're right — checking against the full set is clearer.  
I'll update it to `assert ports == {local_port}`. Thanks!

tests/test_sockets.py
```diff
@@ -902,6 +917,137 @@ async def test_from_socket_not_bound(
...
+            if bind_count == 1 and port == 0:
+                return real_bind(self, addr)
+
+            if fail_once and port != 0:
+                fail_once = False
+                raise OSError(errno.EADDRINUSE, "simulated collision on second bind")
+
+            return real_bind(self, addr)
+
+        monkeypatch.setattr(socket.socket, "bind", fake_bind)
```
agronholm: This is my last issue in this PR: the various backends **really** don't like it when you monkey patch the `socket` module. That's why I had the `with patch(...)` in the other test. Please refactor this one to do the same.

agronholm: I only ended up using `unittest.mock.patch()` directly because the asyncio backend was reporting failures in callbacks outside of the test, but if the actual `socket` module was not touched then this isn't strictly necessary, and the `mocker` fixture can be used for greater comfort. Just LMK if you want me to do this. I want this merged today.

11kkw: Thanks for pointing that out — I’ve updated the test to use `with patch(...)` instead of `monkeypatch`.  
Sorry for the extra round, and really appreciate your patience!

tests/test_sockets.py
```diff
@@ -902,6 +917,134 @@ async def test_from_socket_not_bound(
...
+            if bind_count == 1 and port == 0:
+                return real_bind(self, addr)
+
+            if fail_once and port != 0:
+                fail_once = False
+                raise OSError(errno.EADDRINUSE, "simulated collision on second bind")
+
+            return real_bind(self, addr)
+
+        with patch.object(socket.socket, "bind", new=fake_bind):
```
agronholm, line 1007: You're still patching the `socket` module here?

agronholm: Just copy my setup from `test_tcp_listener_total_bind_failure`. The tests are quite similar otherwise.

11kkw: > You're still patching the `socket` module here?

would you prefer me to patch the AnyIO-internal path, e.g.  
`with patch("anyio._core._sockets.socket.socket.bind", new=fake_bind):` ?

agronholm: No, that would still patch the real `sockets` module. Just please do what I suggested and copy the setup from the other test. It replaces the `socket` name in `anyio._core._sockets` with a fake sockets module.

agronholm: LMK if you want me to do this.

11kkw: > LMK if you want me to do this.

Thanks, I think I’ll just learn from this one instead.

agronholm: Does that mean you'll do it today, or...?

agronholm: The recipe in the other test didn't work here, and I'm not seeing obvious practical problems, so I'm leaving this as is.

agronholm (approved):
LGTM. Thank your for sticking with me!

PR #982, opened 2025-09-23
Added an `AwaitedTaskTerminated` exception class.

smurfix:
This class proxies *all* `BaseException`s.

`AwaitedTaskCancelled` is now a subclass of this.

### DIFF
--- a/src/anyio/_core/_tasks.py
+++ b/src/anyio/_core/_tasks.py
@@ -192,8 +192,9 @@ class AwaitedTaskTerminated(Exception):
     Raised when awaiting on a :class:`TaskHandle` which was terminated by a
     `BaseException`.
 
-    This exception class exists because a `BaseException` (i.e. one that is
-    not an `Exception`) should not be suppressed, nor forwarded to another scope.
+    This exception class exists because a :exc:`BaseException` (i.e. one
+    that is not an :exc:`Exception`) should not be suppressed, nor
+    forwarded to another scope.
     """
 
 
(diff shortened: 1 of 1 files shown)


src/anyio/_core/_tasks.py
```diff
@@ -187,13 +187,25 @@ def create_task_group() -> TaskGroup:
...
+
+
+class AwaitedTaskCancelled(AwaitedTaskTerminated):
     """
     Raised when awaiting on a :class:`TaskHandle` which was cancelled.
 
-    This exception class exists in order to differentiate between the cancellation of
-    the host task (the one awaiting on a task) and the cancellation of the task it's
-    awaiting on. Additionally, raising :class:`asyncio.CancelledError` when waiting on
+    This subclass of `AwaitedTaskTerminated` exists in order to
```
agronholm:
```suggestion
    This subclass of :exc:`AwaitedTaskTerminated` exists in order to
```

src/anyio/_core/_tasks.py
```diff
@@ -187,13 +187,25 @@ def create_task_group() -> TaskGroup:
...
+    differentiate between the cancellation of the host task (the one
+    awaiting on a task) and the cancellation of the task it's awaiting on.
+
+    Additionally, raising :class:`asyncio.CancelledError` when waiting on
     a task would potentially cause cancellation`counters (Python 3.11 and later) to
```
agronholm:
```suggestion
    a task would potentially cause cancellation counters (Python 3.11 and later) to
    be incorrectly decremented, as they should only be decremented when the task
    itself has been cancelled.
```

src/anyio/_core/_tasks.py
```diff
@@ -187,13 +187,25 @@ def create_task_group() -> TaskGroup:
     return get_async_backend().create_task_group()
 
 
-class AwaitedTaskCancelled(Exception):
+class AwaitedTaskTerminated(Exception):
```
agronholm, line 190: I'm not 100% confident this should be the final name of the exception class, but I can't think of an alternative right now.

m-u-xyz: Same here.
…Exited? …Died? …Killed? …Quit?

agronholm: We can figure it out later.

agronholm: ...Aborted?

m-u-xyz: > … Aborted

Ah. Best idea so far.

PR #987, opened 2025-09-26
Do not call getaddrinfo() with port=0 and fix related tests on illumos

### LINKED ISSUE #692: service name not available for the specified socket type
### Things to check first

### AnyIO version

4.3.0

### Python version

3.9.18

### What happened?

Several `anyio` tests fails with this error:
```
_________________ TestTLSListener.test_handshake_fail[asyncio] _________________
tests/streams/test_tls.py:417: in test_handshake_fail
    listener = await create_tcp_listener(local_host="127.0.0.1")
../prototype/i386/usr/lib/python3.9/vendor-packages/anyio/_core/_sockets.py:290: in create_tcp_listener
    gai_res = await getaddrinfo(
../prototype/i386/usr/lib/python3.9/vendor-packages/anyio/_core/_sockets.py:573: in getaddrinfo
    gai_res = await get_async_backend().getaddrinfo(
../prototype/i386/usr/lib/python3.9/vendor-packages/anyio/_backends/_asyncio.py:2370: in getaddrinfo
    return await get_running_loop().getaddrinfo(
/usr/lib/python3.9/asyncio/base_events.py:861: in getaddrinfo
    return await self.run_in_executor(
/usr/lib/python3.9/concurrent/futures/thread.py:58: in run
    result = self.fn(*self.args, **self.kwargs)
/usr/lib/python3.9/asyncio/base_events.py:844: in _getaddrinfo_debug
    addrinfo = socket.getaddrinfo(host, port, family, type, proto, flags)
/usr/lib/python3.9/socket.py:954: in getaddrinfo
    for res in _socket.getaddrinfo(host, port, family, type, proto, flags):
E   socket.gaierror: [Errno 9] service name not available for the specified socket type
```
Here is the complete list of tests with this failure:
```
TestTLSListener.test_handshake_fail[asyncio]
TestTLSListener.test_extra_attributes[asyncio]
TestTCPListener.test_extra_attributes[asyncio-ipv4]
TestTCPListener.test_extra_attributes[asyncio-ipv6]
TestTCPListener.test_accept[asyncio-ipv4]
TestTCPListener.test_accept[asyncio-ipv6]
TestTCPListener.test_accept[asyncio-both]
TestTCPListener.test_accept_after_close[asyncio-ipv4]
TestTCPListener.test_accept_after_close[asyncio-ipv6]
TestTCPListener.test_socket_options[asyncio-ipv4]
TestTCPListener.test_socket_options[asyncio-ipv6]
TestTCPListener.test_close_from_other_task[asyncio-ipv4]
TestTCPListener.test_close_from_other_task[asyncio-ipv6]
TestTCPListener.test_send_after_eof[asyncio-ipv4]
TestTCPListener.test_send_after_eof[asyncio-ipv6]
TestTCPListener.test_bind_link_local[asyncio]
test_multi_listener[asyncio]
test_getaddrinfo[asyncio]
```
The problem seems to be that `getaddrinfo` is called with `port=0`.  The similar issue was [already fixed in eventlet](https://github.com/eventlet/eventlet/issues/791) (please read the link for more details about the issue).

Run `anyio` tests on OpenIndiana.

mtelka:
Fixes #692

## Changes

Fixes #692

## Checklist

features)

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/src/anyio/_core/_sockets.py
+++ b/src/anyio/_core/_sockets.py
@@ -350,14 +350,9 @@ async def create_tcp_listener(
     # IPv6 link-local addresses when passing type=socket.SOCK_STREAM to
     # getaddrinfo():
     # https://github.com/MagicStack/uvloop/issues/539
-    #
-    # It is undocumented what should happen when getaddrinfo() is called with
-    # port=0 and actually it fails on illumos.  So convert port=0 to port=None
-    # which is documented and works well.
-    # See also: https://github.com/agronholm/anyio/issues/692
     gai_res = await getaddrinfo(
         local_host,
-        local_port or None,
+        local_port,
         family=family,
         type=socket.SOCK_STREAM if sys.platform in ("win32", "sunos5") else 0,
         flags=socket.AI_PASSIVE | socket.AI_ADDRCONFIG,
@@ -652,8 +647,12 @@ async def getaddrinfo(
     else:
         encoded_host = host
 
+    # It is undocumented what should happen when getaddrinfo() is called with
+    # port=0 and actually it fails on illumos.  So convert port=0 to port=None
+    # which is documented and works well.
+    # See also: https://github.com/agronholm/anyio/issues/692
     gai_res = await get_async_backend().getaddrinfo(
-        encoded_host, port, family=family, type=type, proto=proto, flags=flags
+        encoded_host, port or None, family=family, type=type, proto=proto, flags=flags
     )
     return [
         (family, type, proto, canonname, convert_ipv6_sockaddr(sockaddr))
--- a/src/anyio/_core/_sockets.py
+++ b/src/anyio/_core/_sockets.py
@@ -353,7 +353,9 @@ async def create_tcp_listener(
         local_host,
         local_port or None,
         family=family,
-        type=socket.SOCK_STREAM if sys.platform == "win32" or sys.platform == "sunos5" else 0,
+        type=socket.SOCK_STREAM
+        if sys.platform == "win32" or sys.platform == "sunos5"
+        else 0,
         flags=socket.AI_PASSIVE | socket.AI_ADDRCONFIG,
     )
 
--- a/src/anyio/_core/_sockets.py
+++ b/src/anyio/_core/_sockets.py
@@ -353,9 +353,7 @@ async def create_tcp_listener(
         local_host,
         local_port or None,
         family=family,
-        type=socket.SOCK_STREAM
-        if sys.platform in ("win32", "sunos5")
-        else 0,
+        type=socket.SOCK_STREAM if sys.platform in ("win32", "sunos5") else 0,
         flags=socket.AI_PASSIVE | socket.AI_ADDRCONFIG,
     )
 
--- a/src/anyio/_core/_sockets.py
+++ b/src/anyio/_core/_sockets.py
@@ -345,10 +345,16 @@ async def create_tcp_listener(
 
         return sock
 
-    # We passing type=0 on non-Windows platforms as a workaround for a uvloop bug
-    # where we don't get the correct scope ID for IPv6 link-local addresses when passing
-    # type=socket.SOCK_STREAM to getaddrinfo():
+    # We passing type=0 on non-Windows and non-illumos platforms as a
+    # workaround for a uvloop bug where we don't get the correct scope ID for
+    # IPv6 link-local addresses when passing type=socket.SOCK_STREAM to
+    # getaddrinfo():
     # https://github.com/MagicStack/uvloop/issues/539
+    #
+    # It is undocumented what should happen when getaddrinfo() is called with
+    # port=0 and actually it fails on illumos.  So convert port=0 to port=None
+    # which is documented and works well.
+    # See also: https://github.com/agronholm/anyio/issues/692
     gai_res = await getaddrinfo(
         local_host,
         local_port or None,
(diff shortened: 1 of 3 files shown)


agronholm:
Every fix requires an accompanying regression test. Some kind of comment should be left to explain the extra logic.

tests/test_sockets.py
```diff
@@ -971,27 +970,6 @@ async def test_tcp_listener_same_port(
...
-            }
-            assert families == {socket.AF_INET, socket.AF_INET6}
-            assert len(multi.listeners) == 2
-
-            ports = {
-                listener.extra(SocketAttribute.local_port)
-                for listener in multi.listeners
-            }
-            assert len(ports) == 1
-
```
agronholm, line 994: You said in the issue that you left this test intact, but clearly you didn't? What's going on?

tests/test_sockets.py
```diff
@@ -935,7 +935,6 @@ async def test_from_socket_not_bound(
     @pytest.mark.parametrize(
         "local_host,family,expected_listeners,local_port",
         [
-            (None, AddressFamily.AF_UNSPEC, 1 if socket.has_dualstack_ipv6() else 2, 0),
```
agronholm, line 938: Why did you delete this?

src/anyio/_core/_sockets.py
```diff
@@ -351,9 +351,11 @@ def setup_raw_socket(
     # https://github.com/MagicStack/uvloop/issues/539
     gai_res = await getaddrinfo(
         local_host,
-        local_port,
+        local_port or None,
         family=family,
-        type=socket.SOCK_STREAM if sys.platform == "win32" else 0,
+        type=socket.SOCK_STREAM
+        if sys.platform == "win32" or sys.platform == "sunos5"
```
agronholm:
```suggestion
        if sys.platform in ("win32", "sunos5")
```

tests/test_sockets.py
mtelka: That was a typo, I meant `test_tcp_listener_total_bind_failure()`.  I updated the comment there.

tests/test_sockets.py
mtelka: Because this caused the `socket.getaddrinfo()` call with both `host` and `port` set to `None`, which in turn caused the libc call of `getaddrinfo()` with both `nodename` and `servname` set to `NULL` which is unsupported and so it failed.

For the same reason I removed the whole `test_tcp_listener_dualstack_disabled` test.

src/anyio/_core/_sockets.py
```diff
@@ -345,15 +345,21 @@ def setup_raw_socket(
...
     # https://github.com/MagicStack/uvloop/issues/539
+    #
+    # It is undocumented what should happen when getaddrinfo() is called with
+    # port=0 and actually it fails on illumos.  So convert port=0 to port=None
+    # which is documented and works well.
+    # See also: https://github.com/agronholm/anyio/issues/692
     gai_res = await getaddrinfo(
         local_host,
-        local_port,
+        local_port or None,
```
mtelka: While looking closely at this file maybe it would be better to move this change down here:
```
    gai_res = await get_async_backend().getaddrinfo(
        encoded_host, port, family=family, type=type, proto=proto, flags=flags
    )
```
@agronholm what do you think?

mtelka: Yes, this seems to be far better and more generic fix.  It also allowed to drop a change in `test_getaddrinfo()`.

src/anyio/_core/_sockets.py
```diff
@@ -345,15 +345,16 @@ def setup_raw_socket(
...
+    # workaround for a uvloop bug where we don't get the correct scope ID for
+    # IPv6 link-local addresses when passing type=socket.SOCK_STREAM to
+    # getaddrinfo():
     # https://github.com/MagicStack/uvloop/issues/539
     gai_res = await getaddrinfo(
         local_host,
         local_port,
         family=family,
-        type=socket.SOCK_STREAM if sys.platform == "win32" else 0,
+        type=socket.SOCK_STREAM if sys.platform in ("win32", "sunos5") else 0,
```
mtelka, line 357: Hmmm, this change is unrelated to the `port=0` issue.  It is just a fix to make this working properly (or better :-)) on illumos.  Maybe I should split out this to a new PR?

agronholm:
> > Every fix requires an accompanying regression test. Some kind of comment should be left to explain the extra logic.
> 
> Any suggestion what exactly should I test? These changes are primarily to make sure current tests pass.
> 
> I updated the comment in `setup_raw_socket()`.

They already pass in CI, as we don't have any other OS available to test against. The point of regression tests is to ensure that the problem stays away, so when a fix is made, a test that fails without the associated fix must be added. Since we can't reproduce the problem normally, I suggested monkey patching `getaddrinfo()` in a new test to reproduce the behavior from Illumos. If that proves unfeasible, then we can look for alternative solutions.

src/anyio/_core/_sockets.py
agronholm: I'm okay with including it in this one, so long as the changelog reflects all the changes made.

mtelka:
This is illumos bug and it should be fixed there (see https://github.com/agronholm/anyio/issues/692).  Sorry for confusion and thank you for your help to narrow it down @agronholm.

PR #989, opened 2025-09-29
Added the RateLimiter class

agronholm:
## Changes

Adds a built-in rate limiting feature.

## Checklist

features)

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/docs/synchronization.rst
+++ b/docs/synchronization.rst
@@ -226,6 +226,79 @@ Example::
     # Task number 8 is now working with the shared resource
     # Task number 9 is now working with the shared resource
 
+Rate limiters
+-------------
+
+Rate limiters allow you to limit the rate of your asynchronous actions, in terms of
+actions per second. They can be configured with two parameters:
+
+* ``allowance``: how many operation are allowed within the time window
+* ``delay``: the delay after which the used operation slots are restored
+
+The :meth:`RateLimiter.max_per_second` class method is provided as a convenience to
+create a rate limiter with the specified rate of allowed operations per second.
+
+Here's a simple example::
+
+    from anyio import RateLimiter, current_time, run
+
+
+    async def main(count: int, max_per_second: int) -> None:
+        async with RateLimiter.from_max_per_second(max_per_second) as limiter:
+            start_time = current_time()
+            for num in range(count):
+                await limiter.acquire()
+                duration = current_time() - start_time
+                print(f"[{duration:.3f}] Performed action")
+
+    run(main, 10, 3)
+
+    # Output:
+    # [0.000] Performed action
+    # [0.335] Performed action
+    # [0.669] Performed action
+    # [1.004] Performed action
+    # [1.339] Performed action
+    # [1.674] Performed action
+    # [2.008] Performed action
+    # [2.343] Performed action
+    # [2.678] Performed action
+    # [3.012] Performed action
+
+Here's a more advanced example that uses explicit ``allowance`` and ``delay`` to specify
+that 5 operations are allowed within a 2 second period::
+
+    from anyio import RateLimiter, create_task_group, current_time, run, sleep
+
+
+    async def your_task_function(start_time: float) -> None:
+        duration = current_time() - start_time
+        await sleep(2)  # Simulate some activity
+        print(f"[{duration:.3f}] Performed action")
+
+    async def main(count: int, slots: int, delay: int) -> None:
+        async with (
+            RateLimiter(slots, delay) as limiter,
+            create_task_group() as tg
+        ):
+            start_time = current_time()
+            for num in range(count):
+                await limiter.acquire()
+                tg.start_soon(your_task_function, start_time)
+
+    run(main, 10, 5, 2)
+
+    # Output:
+    # [0.000] Performed action
+    # [0.000] Performed action
+    # [0.000] Performed action
+    # [0.000] Performed action
+    # [0.000] Performed action
+    # [2.003] Performed action
+    # [2.003] Performed action
+    # [2.003] Performed action
+    # [2.004] Performed action
+
 Resource guards
 ---------------
 
--- a/docs/synchronization.rst
+++ b/docs/synchronization.rst
@@ -232,8 +232,8 @@ Rate limiters
 Rate limiters allow you to limit the rate of your asynchronous actions, in terms of
 actions per second. They can be configured with two parameters:
 
-* ``allowance``: how many operation are allowed within the time window
-* ``delay``: the delay after which the used operation slots are restored
+* ``tokens``: how many operation are allowed within the time window
+* ``interval``: the delay after which the used operation slots are restored
 
 The :meth:`RateLimiter.max_per_second` class method is provided as a convenience to
 create a rate limiter with the specified rate of allowed operations per second.
@@ -265,7 +265,7 @@ Here's a simple example::
     # [2.678] Performed action
     # [3.012] Performed action
 
-Here's a more advanced example that uses explicit ``allowance`` and ``delay`` to specify
+Here's a more advanced example that uses explicit ``tokens`` and ``interval`` to specify
 that 5 operations are allowed within a 2 second period::
 
     from anyio import RateLimiter, create_task_group, current_time, run, sleep
@@ -276,9 +276,9 @@ that 5 operations are allowed within a 2 second period::
         await sleep(2)  # Simulate some activity
         print(f"[{duration:.3f}] Performed action")
 
-    async def main(count: int, slots: int, delay: int) -> None:
+    async def main(count: int, tokens: int, interval: int) -> None:
         async with (
-            RateLimiter(slots, delay) as limiter,
+            RateLimiter(tokens, interval) as limiter,
             create_task_group() as tg
         ):
             start_time = current_time()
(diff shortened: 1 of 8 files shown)


docs/synchronization.rst
```diff
@@ -226,6 +226,79 @@ Example::
...
     # Task number 9 is now working with the shared resource
 
+Rate limiters
+-------------
+
+Rate limiters allow you to limit the rate of your asynchronous actions, in terms of
+actions per second. They can be configured with two parameters:
+
+* ``allowance``: how many operation are allowed within the time window
+* ``delay``: the delay after which the used operation slots are restored
```
dhirschfeld: This seems like great functionality that I'm excited to have in `anyio` core!

Just bikeshedding on the names, would `limit`/`interval` perhaps be clearer?
```suggestion
* ``limit``: The number of tokens allowed in the specified interval
* ``interval``: The length of time (in seconds) before tokens are added back to the pool
```

agronholm: "interval" seems clearly better, not sure about "limit" vs "tokens".

dhirschfeld: > *"interval" seems clearly better, not sure about "limit" vs "tokens".*

Yeah, something with tokens rather than `limit` could be better - it's common terminology from the `CapacityLimiter` and also shows up in the statistics.

dhirschfeld:
I have a use-case where I have multiple limits - e.g. you can make 10 requests per second but only 200 per minute. Implementation-wise I think you have to have the shorter interval bucket draw tokens from the longer interval bucket. 

I'm not sure if something like that should necessarily be supported in core, it's potentially too niche of a use-case, but I'll put it out there to consider - it's an interesting problem!

dhirschfeld:
> *That sounds like you need two rate limiters then, and acquire both before starting an operation?*

I think I might have been over-complicating it - I think that should work! :thinking:

docs/synchronization.rst
agronholm: OK, it's `tokens` and `interval` now.

agronholm:
I made a bunch of updates to the logic and so far it looks good to me.

agronholm:
As a feasibility test, I tried implementing a `race()` function to use with `connect_tcp()`, but ran into a design issue where we want to ignore all `OSErrors` raised from the tasks. Do you have ideas on how to go about this? I feel uncertain about just adding some `ignore_exceptions=...` kwarg.

agronholm:
> Umm, yes. Consider an API that limits me to 1000 API calls per day. It'd be foolish to run the limiter with these settings, however, because after a too-large burst the system would be blocked for too long.

So what are the actual constraints you want to place? If you want to do additional rate limiting, you can use a second rate limiter, yes?

smurfix:
If I stack two limiters the way this might work (e.g. a limiter 1000/24h and then another one 1/86sec with 100 initial behind it), I still have the problem that these 100 initial requests will not refresh themselves when the rate drops (say, at night).

Even for basic rate limiting it's better to do "send five requests at once, but then no more than one every 10th of a second" than "send five concurrent requests at once every half second". The former cannot be implemented with your current code because if I happen to not send a request at all for more than a second I want to be back in the ten-at-once state.

smurfix:
My point is that after I happen to not need any requests for half an hour I want that initial allowance back.

agronholm:
> My point is that after I happen to not need any requests for half an hour I want that initial allowance back.

It looks like the problem class is a bit more complex than I initially thought. I found a list of different rate limiting algorithms: https://blog.algomaster.io/p/rate-limiting-algorithms-explained-with-code

So, one size might not fit all when it comes to rate limiters. It's not clear to me if your PR implements one of these algorithms or not. It seemed pretty complex and I was frankly a bit overwhelmed by the number of parameters.

agronholm:
I found a bunch of interesting real-world code examples here: https://github.com/aio-libs/aiobotocore/tree/master/aiobotocore/retries
It just goes to show that there is great variety in rate limiting algorithms, and we need to make sure that users can substitute this with their own implementations when calling our API.

agronholm:
Another interesting feature of their `AsyncTokenBucket` is that they allow acquiring more than one token. It's not clear to me how it works when there are not enough tokens and multiple tasks are waiting to acquire some.

smurfix:
OK, so if we're going to make this pluggable anyway, I vote that out basic implementation should not do anything fancy (*maybe* support `initial_tokens` but not my burst add-on) and instead allow for subclassing and/or plug-ins. Their design looks reasonable at first glance.

dhirschfeld:
When I deploy my code it's in containers on k8s. There is usually more than 1 replica but the rate limit for the external API applies to all replicas so I need a distributed rate-limiter.

My plan of attack is to use Redis (or similar) as the token source, so it would be great if there we some interface for the token bucket and you could plug in different implementations, including e.g. a Redis one for distributed use-cases.

smurfix:
> Why do we now need a release() method when we didn't before?

An adaptive limiter might want to change the number of concurrent tokens it hands out, based on how much time a request currently takes.

Also, a Redis/Valkey backend needs to send a friendly message to the server to release the token.

agronholm:
It's also possible to create a combined limiter ABC which supports all of this and is maximally pluggable too.

agronholm:
OK, I've added a new ABC and a concrete implementation for it.

x42005e1f:
I think I can add my two cents here. In the past, I implemented a fairly simple rate limiter that also works as a concurrency limiter on AnyIO - [x42005e1f/976f84ac395ae2b51f14c8507be0ea34](https://gist.github.com/x42005e1f/976f84ac395ae2b51f14c8507be0ea34). It works with two limits (weak and strong) at the same time and provides the maximum number of requests per second. It has been tested on a real API, so you may find it useful.

agronholm:
> In my previous comment (which I deleted; available via the Wayback Machine), I described my old implementation of rate limiters, but okay, that is beside the point. Suffice it to say that calculating the deadline on `acquire()` is not reliable when working with external services due to the fact that concurrent requests can actually reach the server in the next window. I think this could become an issue in the future, because I personally have experienced exceeding the limit with similar algorithms, and it was solved by moving the deadline calculation to `release()`.

What occurred to me is that you might be talking about concurrency limiting and not rate limiting here? As `release()` would be relevant to those.

x42005e1f:
> What occurred to me is that you might be talking about concurrency limiting and not rate limiting here? As `release()` would be relevant to those.

Your `RateLimiter` is already a concurrency limiter, is not it? Or what do you mean by a concurrency limiter? It allows concurrent entries by decrementing the counter for the current window and blocking when it reaches zero. So, for the current window, it theoretically allows no more than *tokens* tasks.

> The `RateLimiter` class here does not have a `release()` method anyway?

Yes, and that is a problem. But unlike the discussion above, it is not because of adaptability. It is because one of the popular use cases for a rate limiter is to ensure that the maximum number of requests for one client is not exceeded. Note, there does not necessarily have to be more than one client here; it is enough to use the same authorization token for concurrent requests, which is common when an external service's API allows 100 requests per second, the ping is 0.05 seconds, and thus without concurrency, you cannot reach the maximum number of requests per second (only 20 due to the overhead on network operations).

If you send, say, 100 requests within one window, but there are more than 100 tasks, your rate limiter will still allow the next requests in the next window regardless of whether the previous requests reached the server or not. As a result, if those requests were temporarily held back during sending, they may reach the server at the same time as the next ones. The server will then consider that you sent more than 100 requests within the next window, which will result in a limit exceeded error.

x42005e1f:
> Just what code are you looking at?

I think we have different understandings of the terms. I apologize for not being able to convey my point to you.

agronholm:
If I'm understanding this correctly, it would mean that the client code would have to somehow signal that it's done? Or how would that work? I'm confused.

x42005e1f:
Consider a task in which regular network requests are sent, for example via HTTPX. The code with a rate limiter might look something like this:

```python
async with rate_limiter:
    response = await httpx.get("...")  # via REST API
```

I would not say that this is a rarely used pattern.

x42005e1f:
Yes, mine. I described it in the deleted comment. Here is a link to the Wayback Machine, if you are interested: [09:25:04 UTC](https://web.archive.org/web/20251110092504/https://github.com/agronholm/anyio/pull/989).

agronholm:
But surely, if the problem is as widespread as you said, someone else must have written a solution too?

x42005e1f:
No, that was my own research aimed at achieving maximum performance. If you are asking about credible sources (and I am certainly not one in the eyes of many), unfortunately, I cannot provide any.

I just tried to quickly search for alternative solutions, but found nothing similar. Perhaps someone else will be able to do so.

**Note**
I am not requesting such a feature. I am merely warning that the rate limiter will behave flaky with concurrent requests, which may be an unpleasant effect for the user. So if you decide not to solve the problem, that is fine.

agronholm:
I've been reading your code, and I think I have achieved at least a partial understanding of it. I have some follow-up questions:

1. I didn't see any place where you're removing timestamps from the queue. Did you not experience any increasing memory footprint, or was the code simply not running long enough or high volume enough for the problem to become noticeable?
2. Where is the rate limiter context manager supposed to be used? In the parent task that launches the tasks that perform the outbound requests or in the child task code itself? I'm trying to achieve a solution for the former.

PR #990, opened 2025-09-29
Rate limiter: Drop the semaphore

smurfix:
The semaphore and the timer task is not necessary, easier to just use a lock and a timestamp. Additional advantage, we don't twiddle our thumbs trying to re-fill an already-full bucket.

I don't have the mental bandwidth right now to clean up the test further, maybe the statistics getter should check the timestamp and report the full number of tokens if the current time is past the next token update.

As to an initial / refilling burst capacity, that's easy enough to add but I don't have the mental bandwidth right now to figure out the exact semantics. Will send another PR for that, tomorrow or so.

### DIFF
--- a/src/anyio/_core/_synchronization.py
+++ b/src/anyio/_core/_synchronization.py
@@ -8,7 +8,7 @@ from contextlib import asynccontextmanager
 from dataclasses import dataclass
 from datetime import timedelta
 from types import TracebackType
-from typing import TypeVar, cast
+from typing import TypeVar
 
 from sniffio import AsyncLibraryNotFoundError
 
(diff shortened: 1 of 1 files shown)


tests/test_synchronization.py
```diff
@@ -938,20 +938,23 @@ async def test_acquire(self, fake_sleep: FakeSleeper) -> None:
...
-            assert stats.available_tokens == 0
+            # assert stats.available_tokens == 0
             assert stats.tasks_waiting == 0
 
             # With all the tokens in use, the next acquire() should block
             with pytest.raises(TimeoutError), fail_after(0.1):
                 await limiter.acquire()
+                await limiter.acquire()
+                await limiter.acquire()
+                await limiter.acquire()
```
agronholm, line 957: Why these extra acquires?

m-u-xyz: There are still two or three open slots in the limiter at this point because the refill didn't happen in the `sleep` but during the subsequent countdown.

On second thought this is a problem because when you set the slot to five you want five and not anywhere from five to nine. Will fix tomorrow.

PR #992, opened 2025-09-30
Ratelimiter: refactor, add burst parameters

smurfix:
This PR adds comprehensive parameters to control the limiter's burst behavior.

* initial: the operations allowed in the first time window. May be zero, which is useful when you start a program and can't risk that the previous run has just filled its slot and exited.
* burst_rate: the percentage of "unused" operations that carry over from the previous timeslot(s).
* burst_max: the maximum for carry-over, to avoid large surges after doing nothing for some time.

This also fixes a race condition: if you exactly exhaust a time slot and then resume after the slot has ended, the previous version waited for the next slot unnecessarily.

TODO: tests for all of the above.

### DIFF
--- a/src/anyio/_core/_synchronization.py
+++ b/src/anyio/_core/_synchronization.py
@@ -741,21 +741,40 @@ class RateLimiter:
     """
     Provides rate limiting via an internal semaphore which is periodically incremented.
 
-    :param tokens: number of operations allowed within the time window
-    :param interval: the time window, in seconds (or a :class:`~datetime.timedelta),
+    :param tokens: number of operations allowed within each time window
+    :param interval: the time window, in seconds (or a :class:`~datetime.timedelta`),
         that ``tokens`` applies to
+    :param initial: number of operations allowed in the first time window.
+                    The default is `tokens`. May be zero.
+    :param burst_rate: percentage of a slot's unused operations that are
+                       carried over to the next interval (must be 0 ... 1)
+    :param burst_max: upper limit for carried-over operations
+
+    Each time window starts with its first operation.
     """
 
     def __init__(
         self,
         tokens: int,
         interval: float | timedelta = 1,
+        burst_rate: float = 0,
+        burst_max: int | None = None,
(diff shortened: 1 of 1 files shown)


agronholm:
Did you see the changes I already made today?

agronholm:
I'm feeling inclined to merge this, but I'll take a closer look and may adjust some minor details.

src/anyio/_core/_synchronization.py
```diff
@@ -792,41 +813,72 @@ def from_max_per_second(cls, max_per_second: int, /) -> Self:
...
-                tasks_waiting=0,
-                available_tokens=self._available,
-            )
+        """Return the statistics for the underlying lock."""
 
-        if current_time() >= self._next:
-            self._available = self._tokens
-            self._next = None
+        # Refresh token availability
+        self._update_available()
```
agronholm, line 877: I don't like `statistics()` affecting the internal state of the limiter. I did away with that in my own recent changes to the rate limiter PR.

src/anyio/_core/_synchronization.py
```diff
@@ -792,41 +813,72 @@ def from_max_per_second(cls, max_per_second: int, /) -> Self:
         """
         return cls(1, 1 / max_per_second)
 
+    def _update_available(self, slot_time: float | None = None) -> bool:
```
agronholm, line 816: I think this method should only perform the availability computation logic and return the new available slots and next time frame, and not change the internal state. Only `acquire()` should change the state.

src/anyio/_core/_synchronization.py
```diff
@@ -741,30 +741,51 @@ class RateLimiter:
...
+    Each time window starts with its first operation.
     """
 
     def __init__(
         self,
         tokens: int,
         interval: float | timedelta = 1,
+        burst_rate: float = 0,
+        burst_max: int | None = None,
+        initial: int | None = None,
```
agronholm, line 762: Please rename this to `initial_tokens`.

Issue #783, opened 2024-09-05
Look into creating a shared subprocess implementation between backends


agronholm:
### Things to check first

- [X] I have searched the existing issues and didn't find my feature already requested there

### Feature description

We should take a stab at creating a more low-level subprocess implementation that only relies on a minimum set of functionality from each backend, just like the threading API.

### Use case

Currently the asyncio subprocess API is a bit iffy from AnyIO's perspective. There are a lot of moving parts which we don't control, and cases where we use undocumented methods (`StreamReader.set_exception()` as an example). A new implementation, if possible, might make it easier to further develop the AnyIO subprocess APIs.

PR #993, opened 2025-10-02
Allowed passing keyword arguments to subprocesses

davidbrochart:
## Changes

This PR makes it possible to pass keyword arguments to [Popen](https://docs.python.org/3/library/subprocess.html#popen-constructor), through `run_process`, `open_process` or `to_process.run_sync`. In the later case, a new subprocess is always created, since it wouldn't be possible to reuse a worker process created with different keyword arguments.

## Checklist

features)

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/docs/subprocesses.rst
+++ b/docs/subprocesses.rst
@@ -98,7 +98,6 @@ in this case a new subprocess will always be created, as it would not be possibl
 to reuse a worker process created with different keyword arguments::
 
   import os
-  import time
   from functools import partial
 
   from anyio import create_task_group, run, to_process, to_thread
--- a/docs/subprocesses.rst
+++ b/docs/subprocesses.rst
@@ -118,7 +118,6 @@ to reuse a worker process created with different keyword arguments::
           data = await to_thread.run_sync(os.read, receiver1, 1024)
       print(data)  # b'Hello, World!'
 
-
   # This check is important when the application uses to_process.run_sync()
   if __name__ == '__main__':
       run(main)
--- a/docs/subprocesses.rst
+++ b/docs/subprocesses.rst
@@ -95,7 +95,8 @@ This is done by using :func:`.to_process.run_sync`::
 
 You can pass keyword arguments directly to :class:`subprocess.Popen`, but note that
 in this case a new subprocess will always be created, as it would not be possible
-to reuse a worker process created with different keyword arguments::
+to reuse a worker process created with different keyword arguments. For instance,
+the following example works (on Linux) because ``close_fds=False`` is passed::
 
   import os
   from functools import partial
--- a/docs/subprocesses.rst
+++ b/docs/subprocesses.rst
@@ -93,10 +93,10 @@ This is done by using :func:`.to_process.run_sync`::
     if __name__ == '__main__':
         run(main)
 
-Any keyword argument to :func:`.to_process.run_sync` (other than ``cancellable`` and
-``limiter``) will be passed to :class:`subprocess.Popen`, but note that in this case
-a new subprocess will always be created, as it would not be possible to reuse a worker
-process created with different keyword arguments. For instance, the following example
+If you want to pass arguments down to :class:`subprocess.Popen` (which is called under the hood),
+you can do so by calling :func:`.to_process.run_sync` with a ``popen_args`` dictionary.
+Note that in this case a new subprocess will always be created, as it would not be possible
+to reuse a worker process created with different arguments. For instance, the following example
 works (on Linux) because ``close_fds=False`` is passed::
 
   import os
@@ -115,7 +115,15 @@ works (on Linux) because ``close_fds=False`` is passed::
       os.set_inheritable(receiver0, True)
       os.set_inheritable(sender1, True)
       async with create_task_group() as tg:
-          tg.start_soon(partial(to_process.run_sync, cpu_intensive_function, receiver0, sender1, close_fds=False))
+          tg.start_soon(
+              partial(
+                  to_process.run_sync,
+                  cpu_intensive_function,
+                  receiver0,
+                  sender1,
+                  popen_args={"close_fds": False},
+              )
+          )
           os.write(sender0, b"Hello")
           data = await to_thread.run_sync(os.read, receiver1, 1024)
       print(data)  # b'Hello, World!'
--- a/docs/subprocesses.rst
+++ b/docs/subprocesses.rst
@@ -93,6 +93,37 @@ This is done by using :func:`.to_process.run_sync`::
     if __name__ == '__main__':
         run(main)
 
+You can pass keyword arguments directly to :class:`subprocess.Popen`, but note that
+in this case a new subprocess will always be created, as it would not be possible
+to reuse a worker process created with different keyword arguments::
+
+  import os
+  import time
+  from functools import partial
+
+  from anyio import create_task_group, run, to_process, to_thread
+
+
+  def cpu_intensive_function(receiver, sender):
+      data = os.read(receiver, 1024)
+      os.write(sender, data + b", World!")
+
+  async def main():
+      receiver0, sender0 = os.pipe()
+      receiver1, sender1 = os.pipe()
+      os.set_inheritable(receiver0, True)
+      os.set_inheritable(sender1, True)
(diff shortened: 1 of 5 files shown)


tests/test_to_process.py
```diff
@@ -48,6 +48,7 @@ def process_func(receiver: int, sender: int) -> None:
     os.write(sender, data + b", World!")
 
 
+@pytest.mark.parametrize("anyio_backend", ["asyncio"])
```
agronholm: Wait, why it this only tested with asyncio? I'm not willing to support any feature that isn't available across the different back-ends.

davidbrochart: I'm not sure if the test was hanging for other event loops or because of the platform, I'll try to remove this line.

davidbrochart: So after https://github.com/agronholm/anyio/pull/993/commits/7c3f196eef39229ac1e0b54c1d81304be4def918 it seems that the test hangs with `asyncio+uvloop`.

davidbrochart: And I'm not sure why the test hangs on Windows. [Pipes](https://docs.python.org/3/library/os.html#os.pipe) are supposed to work, and [inheritance of file descriptors](https://docs.python.org/3/library/os.html#inheritance-of-file-descriptors) should be supported.

agronholm: This will nonetheless have to be resolved before I will consider merging this.

davidbrochart: For `uvloop` too? Maybe a simpler test would be better.

agronholm: At the minimum I need to know what's going on before making any decision on how to proceed.

davidbrochart: At least https://github.com/agronholm/anyio/pull/993/commits/cb1bd23481c5a8db5feb38c6a8754d52b1cb683a shows an error instead of hanging (bad file descriptor).

agronholm: This could be done step by step, first making sure it works with the stdlib and Trio and then adding `uvloop` / `winloop` to the mix. The latter at least just got some fixes in regarding its subprocess support.

davidbrochart: I think I misread the [documentation](https://docs.python.org/3/library/os.html#inheritance-of-file-descriptors), on Windows: "using the [subprocess](https://docs.python.org/3/library/subprocess.html#module-subprocess) module, all file descriptors except standard streams are closed".
I think that explains why this test cannot run on Windows?

agronholm: Can you alter the test to make it work on Windows?

davidbrochart: It's working on Windows now. Let's see what happens with `uvloop`.

davidbrochart: So the issue is only with `uvloop` on macOS (bad file descriptor). Not sure what to do.

agronholm: Well...we're already skipping some uvloop related test combinations due to uvloop bugs. The whole project is abandonware by now. I'm fine with adding one more skip.

docs/subprocesses.rst
```diff
@@ -93,6 +93,35 @@ This is done by using :func:`.to_process.run_sync`::
     if __name__ == '__main__':
         run(main)
 
+You can pass keyword arguments directly to :class:`subprocess.Popen`, but note that
+in this case a new subprocess will always be created, as it would not be possible
+to reuse a worker process created with different keyword arguments::
```
davidbrochart: Should we mention that this example doesn't work on Windows and macOS/uvloop?

agronholm: Yeah. And I wanted to ask, why is this mentioning `subprocess.Popen` in the first place?

davidbrochart: Because that's what is called under the hood, both in asyncio and Trio? Do you prefer to not mention that at all, or on the contrary say that this is what is called eventually?

agronholm: This feels confusing to me because the wording implies that the user needs to call `subprocess.Popen` directly. This needs to be reworded to dispel any such confusion.

src/anyio/to_process.py
```diff
@@ -40,6 +40,7 @@ async def run_sync(  # type: ignore[return]
     *args: Unpack[PosArgsT],
     cancellable: bool = False,
     limiter: CapacityLimiter | None = None,
+    **kwargs: Any,
```
agronholm: This is absolutely a no-no. Having this here means we lose the freedom to define new kwargs of our own, plus any future Popen kwarg might conflict with ours.

agronholm: To be specific, this needs to be something like `popen_args: Mapping[str, Any]`, and not a variable keyword argument.

src/anyio/_core/_subprocesses.py
```diff
@@ -171,10 +175,14 @@ async def open_process(
     :param extra_groups: supplementary groups to set in the subprocess (POSIX only)
     :param umask: if not negative, this umask is applied in the child process before
         running the given command (POSIX only)
+    :param popen_args: arguments passed to :class:`subprocess.Popen`
```
gschaffner, line 172: This implies to me that all `Popen` kwargs are supported here, but on asyncio these kwargs go to `asyncio.create_subprocess_exec`/`asyncio.create_subprocess_shell`, which do not support all of the same kwargs that `Popen` does.

#783 seems related.

src/anyio/_core/_subprocesses.py
```diff
@@ -141,6 +144,7 @@ async def open_process(
     group: str | int | None = None,
     extra_groups: Iterable[str | int] | None = None,
     umask: int = -1,
+    popen_args: dict[str, Any] | None = None,
```
gschaffner, line 141: @davidbrochart I'm curious which kwargs are useful to pass via `popen_args` here. In https://github.com/agronholm/anyio/pull/749#issuecomment-2321935742 it seemed like there were only a few that are possible that aren't already supported by `open_process`, but this info. may be outdated. If we could add the few parameters directly instead of doing a `dict[str, Any]`, it would be a bit nicer for users, assuming that annotating them isn't that hard.

`TypedDict` could also be nice here.

davidbrochart: > I'm curious which kwargs are useful to pass via `popen_args` here

For instance `close_fds` (see the documentation example in this PR, or the tests), if you want to send/receive data to a subprocess. It seems that `pass_fds` doesn't work on Windows, so we cannot use that instead.

gschaffner: Ah—didn't see the docs/tests. I wrote `pass_fds`(-like) Windows support for work; perhaps it's useful to you too? Here's the code:

```python
async def _open_process(
    ...
    *,
    pass_fds: Sequence[int],
) -> Process:
    if sys.platform != "win32":
        return await anyio.open_process(
            ...
            pass_fds=pass_fds,
        )
    else:
        with ExitStack() as exit_stack:
            # N.B.: See the warning at
            # https://docs.python.org/3/library/subprocess.html#subprocess.STARTUPINFO.lpAttributeList.
            for handle in pass_fds:
                if not os.get_handle_inheritable(handle):
                    os.set_handle_inheritable(handle, True)
                    exit_stack.callback(os.set_handle_inheritable, handle, False)

            return await anyio.open_process(
                ...
                startupinfo=subprocesslib.STARTUPINFO(
                    lpAttributeList={"handle_list": pass_fds}
                ),
            )
```

IIRC using `STARTUPINFO.lpAttributeList["handle_list"]` seemed preferable over `close_fds=False` on Windows because fd inheritance is unsafe in concurrent programs. `handle_list` still has some unsafety, but IIUC it's safer than `close_fds=False` because it restricts inheritance to only the specific set of fds we are concerned with rather than it involving every fd.

davidbrochart: Thanks, yes that could work. There is still the issue that `to_process.run_sync` doesn't expose these parameters though.

src/anyio/_core/_subprocesses.py
agronholm: @davidbrochart I recall this is the exact reason why I originally opted not to pass through arbitrary keyword arguments.

PR #1001, opened 2025-10-15
Added an async version of the functools module

### LINKED ISSUE #999: Add an asynchronous functools implementation
### Things to check first

### Feature description

The need to cache the results of expensive calls is not unique to synchronous functions. This is why libraries like [async_lru](https://pypi.org/project/async-lru/) exist.

I would like to implement two functions from the `functools` module: `lru_cache` and `reduce`.

### Use case

An async `@lru_cache` is needed whenever one needs to cache the result of an expensive function. The `reduce` function can be done cheaply for completeness if no other reason.

agronholm:
## Changes

Fixes #999. 

## Checklist

- [X] You've updated the documentation (in `docs/`), in case of behavior changes or new
features

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -12,6 +12,12 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 - Added support for use as a context manager to ``anyio.lowlevel.RunVar``
 - Added ``__all__`` declarations to public submodules (``anyio.lowlevel`` etc.)
   (`#1009 <https://github.com/agronholm/anyio/pull/1009>`_)
+- Fixed ``Process.stdin.send()`` not raising ``ClosedResourceError`` and
+  ``BrokenResourceError`` on asyncio. Previously, a non-AnyIO exception was raised in
+  such cases (`#671 <https://github.com/agronholm/anyio/issues/671>`_; PR by
+  @gschaffner)
+- Fixed ``Process.stdin.send()`` not checkpointing before writing data on asyncio
+  (`#1002 <https://github.com/agronholm/anyio/issues/1002>`_; PR by @gschaffner)
 
 .. _winloop: https://github.com/Vizonex/Winloop
 
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -34,7 +34,7 @@ from collections.abc import (
 from concurrent.futures import Future
 from contextlib import AbstractContextManager, suppress
 from contextvars import Context, copy_context
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from functools import partial, wraps
 from inspect import (
     CORO_RUNNING,
@@ -151,18 +151,18 @@ else:
 
         def __exit__(
             self,
-            exc_type: type[BaseException],
-            exc_val: BaseException,
-            exc_tb: TracebackType,
+            exc_type: type[BaseException] | None,
+            exc_val: BaseException | None,
+            exc_tb: TracebackType | None,
         ) -> None:
             self.close()
 
         def close(self) -> None:
             """Shutdown and close event loop."""
-            if self._state is not _State.INITIALIZED:
+            loop = self._loop
+            if self._state is not _State.INITIALIZED or loop is None:
                 return
             try:
-                loop = self._loop
                 _cancel_all_tasks(loop)
                 loop.run_until_complete(loop.shutdown_asyncgens())
                 if hasattr(loop, "shutdown_default_executor"):
@@ -1052,12 +1052,30 @@ class StreamReaderWrapper(abc.ByteReceiveStream):
 @dataclass(eq=False)
 class StreamWriterWrapper(abc.ByteSendStream):
     _stream: asyncio.StreamWriter
+    _closed: bool = field(init=False, default=False)
 
     async def send(self, item: bytes) -> None:
-        self._stream.write(item)
-        await self._stream.drain()
+        await AsyncIOBackend.checkpoint_if_cancelled()
+        stream_paused = self._stream._protocol._paused  # type: ignore[attr-defined]
+        try:
+            self._stream.write(item)
+            await self._stream.drain()
+        except (ConnectionResetError, BrokenPipeError, RuntimeError) as exc:
+            # If closed by us and/or the peer:
+            # * on stdlib, drain() raises ConnectionResetError or BrokenPipeError
+            # * on uvloop and Winloop, write() eventually starts raising RuntimeError
+            if self._closed:
+                raise ClosedResourceError from exc
+            elif self._stream.is_closing():
+                raise BrokenResourceError from exc
+
+            raise
+
+        if not stream_paused:
+            await AsyncIOBackend.cancel_shielded_checkpoint()
 
     async def aclose(self) -> None:
+        self._closed = True
         self._stream.close()
         await AsyncIOBackend.checkpoint()
 
@@ -1599,8 +1617,8 @@ class UDPSocket(abc.UDPSocket):
         return self._transport.get_extra_info("socket")
 
     async def aclose(self) -> None:
+        self._closed = True
         if not self._transport.is_closing():
-            self._closed = True
             self._transport.close()
 
     async def receive(self) -> tuple[bytes, IPSockAddrType]:
@@ -1647,8 +1665,8 @@ class ConnectedUDPSocket(abc.ConnectedUDPSocket):
         return self._transport.get_extra_info("socket")
 
     async def aclose(self) -> None:
+        self._closed = True
         if not self._transport.is_closing():
-            self._closed = True
             self._transport.close()
 
     async def receive(self) -> bytes:
--- a/tests/test_subprocesses.py
+++ b/tests/test_subprocesses.py
@@ -10,11 +10,12 @@ from textwrap import dedent
 from typing import Any
 
 import pytest
-from pytest import FixtureRequest
 
 from anyio import (
+    BrokenResourceError,
     CancelScope,
     ClosedResourceError,
+    EndOfStream,
     create_task_group,
     open_process,
     run_process,
@@ -236,10 +237,7 @@ async def test_process_aexit_cancellation_doesnt_orphan_process() -> None:
     assert process.returncode != 0
 
 
-async def test_process_aexit_cancellation_closes_standard_streams(
-    request: FixtureRequest,
-    anyio_backend_name: str,
-) -> None:
+async def test_process_aexit_cancellation_closes_standard_streams() -> None:
     """
     Regression test for #669.
 
@@ -248,12 +246,6 @@ async def test_process_aexit_cancellation_closes_standard_streams(
     closed stream.
 
     """
-    if anyio_backend_name == "asyncio":
-        # Avoid pytest.xfail here due to https://github.com/pytest-dev/pytest/issues/9027
-        request.node.add_marker(
-            pytest.mark.xfail(reason="#671 needs to be resolved first")
-        )
-
     with CancelScope() as scope:
         async with await open_process(
             [sys.executable, "-c", "import time; time.sleep(1)"]
@@ -271,6 +263,48 @@ async def test_process_aexit_cancellation_closes_standard_streams(
         await process.stdout.receive(1)
 
     assert process.stderr is not None
+
+    with pytest.raises(ClosedResourceError):
+        await process.stderr.receive(1)
+
+
+async def test_exceptions_after_subprocess_closes_standard_streams() -> None:
+    async with await open_process([sys.executable, "-c", ""]) as process:
+        await process.wait()
+        assert process.stdin is not None
+        assert process.stderr is not None
+        assert process.stdout is not None
+        with pytest.raises(BrokenResourceError):
+            # * On Trio, stdin.send() will always raise if the peer's end of the pipe is
+            #   closed.
+            # * On asyncio, even though process.wait() finished, some event loop
+            #   implementations do not yet inform us that the peer's end of the pipe is
+            #   closed; the event loop needs to run some rounds of callbacks before it
+            #   will get that information to us. In particular:
+            #   * On stdlib asyncio, stdin.send() will now always raise
+            #     BrokenResourceError.
+            #   * On uvloop, the second stdin.send() will raise BrokenResourceError.
+            #   * On Winloop 0.3.0, the third stdin.send() will raise
+            #     BrokenResourceError. The second also might, but it depends on
+            #     scheduling order.
+            #   * On Winloop 0.3.1, the second stdin() will raise BrokenResourceError.
+            for _ in range(3):
+                await process.stdin.send(b"foo")
+
+        with pytest.raises(BrokenResourceError):
+            await process.stdin.send(b"foo")
+
+        with pytest.raises(EndOfStream):
+            await process.stdout.receive(1)
+
+        with pytest.raises(EndOfStream):
+            await process.stderr.receive(1)
+
+    with pytest.raises(ClosedResourceError):
+        await process.stdin.send(b"foo")
+
+    with pytest.raises(ClosedResourceError):
+        await process.stdout.receive(1)
 
     with pytest.raises(ClosedResourceError):
         await process.stderr.receive(1)


src/anyio/functools.py
```diff
@@ -0,0 +1,247 @@
...
+    async def __call__(self, *args: P.args, **kwargs: P.kwargs) -> T:
+        # Easy case first: if maxsize == 0, no caching is done
+        if self._maxsize == 0:
+            value = await self.__wrapped__(*args, **kwargs)
+            self._misses += 1
+            return value
+
+        if self._typed:
+            key: Hashable = tuple((type(arg), arg) for arg in args) + tuple(
+                (type(val), key, val) for key, val in kwargs.items()
```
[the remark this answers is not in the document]
agronholm: And why is this bad, given how nothing is being overwritten?

src/anyio/functools.py
```diff
@@ -0,0 +1,253 @@
+from __future__ import annotations
+
+__all__ = ("cache", "lru_cache", "reduce")
```
gschaffner: We don't have any other `__all__`s in AnyIO. Do you anticipate folks will want to wildcard import this?

(Not an important comment—mostly just curiosity.)

src/anyio/functools.py
```diff
@@ -0,0 +1,253 @@
...
+) -> T: ...
+
+
+async def reduce(  # type: ignore[misc]
+    function: Callable[[T, T], Awaitable[T]] | Callable[[T, S], Awaitable[T]],
+    iterable: Iterable[T] | Iterable[S] | AsyncIterable[T] | AsyncIterable[S],
+    /,
+    initial: T | _InitialMissingType = initial_missing,
+) -> T:
+    """Asynchronous version of :func:`functools.reduce`."""
```
gschaffner: Perhaps note in the docstring that it can take both sync and async iterables? The generated API docs don't show the annotations.

tests/test_functools.py
```diff
@@ -0,0 +1,249 @@
...
+    async def test_not_iterable(self) -> None:
+        with pytest.raises(
+            TypeError, match="argument 2 must be an iterable or async iterable"
+        ):
+            await reduce(lambda x, y: x + y, 1)  # type: ignore[call-overload]
+
+    async def test_no_initial(self) -> None:
+        async def func(x: int, y: int) -> int:
+            await checkpoint()
+            return x + int(y)
```
gschaffner:
```suggestion
        async def func(x: int, y: int) -> int:
            await checkpoint()
            return x + y
```

tests/test_functools.py
```diff
@@ -0,0 +1,249 @@
...
+        async def func(x: int, y: str) -> int:
+            await checkpoint()
+            return x + int(y)
+
+        assert await reduce(func, ["1", "2", "3"], 2) == 8
+
+    async def test_asynciter_no_initial(self) -> None:
+        async def func(x: int, y: int) -> int:
+            await checkpoint()
+            return x + int(y)
```
gschaffner:
```suggestion
        async def func(x: int, y: int) -> int:
            await checkpoint()
            return x + y
```

tests/test_functools.py
```diff
@@ -0,0 +1,249 @@
...
+
+        assert await reduce(func, [1, 2, 3]) == 6
+
+    async def test_has_initial(self) -> None:
+        async def func(x: int, y: str) -> int:
+            await checkpoint()
+            return x + int(y)
+
+        assert await reduce(func, ["1", "2", "3"], 2) == 8
+
```
gschaffner, line 290: stdlib `functools.reduce` converts `StopIteration` to `TypeError`:

```suggestion

    async def test_empty_iter_no_initial(self) -> None:
        with pytest.raises(
            TypeError, match=r"reduce\(\) of empty iterable with no initial value"
        ):
            await reduce(lambda x, y: x + y, [])

```

tests/test_functools.py
```diff
@@ -0,0 +1,249 @@
...
+        async def func(x: int, y: str) -> int:
+            await checkpoint()
+            return x + int(y)
+
+        async def asyncgen() -> AsyncIterator[str]:
+            yield "1"
+            yield "2"
+            yield "3"
+
+        assert await reduce(func, asyncgen(), 2) == 8
```
gschaffner, line 319: continued:

```suggestion
        assert await reduce(func, asyncgen(), 2) == 8

    async def test_empty_asynciter_no_initial(self) -> None:
        class AIter:
            def __aiter__(self) -> Self:
                return self

            async def __anext__(self) -> NoReturn:
                raise StopAsyncIteration

        with pytest.raises(
            TypeError, match=r"reduce\(\) of empty async iterable with no initial value"
        ):
            await reduce(lambda x, y: x + y, AIter())
```

src/anyio/functools.py
```diff
@@ -0,0 +1,253 @@
...
+    element: Any
+    if isinstance(iterable, AsyncIterable):
+        async_it = iterable.__aiter__()
+        if initial is initial_missing:
+            value = cast(T, await async_it.__anext__())
```
gschaffner: continued:

```suggestion
            try:
                value = cast(T, await async_it.__anext__())
            except StopAsyncIteration:
                raise TypeError(
                    "reduce() of empty async iterable with no initial value"
                ) from None
```

src/anyio/functools.py
```diff
@@ -0,0 +1,253 @@
...
+        else:
+            value = cast(T, initial)
+
+        async for element in async_it:
+            value = await function(value, element)
+    elif isinstance(iterable, Iterable):
+        it = iter(iterable)
+        if initial is initial_missing:
+            value = cast(T, next(it))
```
gschaffner: continued:

```suggestion
            try:
                value = cast(T, next(it))
            except StopIteration:
                raise TypeError(
                    "reduce() of empty iterable with no initial value"
                ) from None
```

src/anyio/functools.py
```diff
@@ -0,0 +1,253 @@
...
+async def reduce(
+    function: Callable[[T, T], Awaitable[T]],
+    iterable: Iterable[T] | AsyncIterable[T],
+    /,
+) -> T: ...
+
+
+async def reduce(  # type: ignore[misc]
+    function: Callable[[T, T], Awaitable[T]] | Callable[[T, S], Awaitable[T]],
+    iterable: Iterable[T] | Iterable[S] | AsyncIterable[T] | AsyncIterable[S],
```
gschaffner, line 293: Somewhat off topic: If we are getting into `functools.reduce`/`itertools`, I think that we should add `anyio.as_safe_stream` (a port of `trio.as_safe_channel`). We could mention it in the "Avoiding cancel scope stack corruption" section of the docs.

src/anyio/functools.py
```diff
@@ -0,0 +1,253 @@
...
+
+    @overload
+    def __call__(
+        self, func: Callable[..., T], /
+    ) -> functools._lru_cache_wrapper[T]: ...
+
+    def __call__(
+        self, f: Callable[P, Coroutine[Any, Any, T]] | Callable[..., T], /
+    ) -> AsyncLRUCacheWrapper[P, T] | functools._lru_cache_wrapper[T]:
+        if iscoroutinefunction(f):
```
gschaffner, line 197: IIRC we still need to do

```python
    from inspect import iscoroutinefunction
else:
    from asyncio import iscoroutinefunction
```

to cover `asyncio.coroutine`-based coros. https://docs.python.org/3.10/library/asyncio-task.html?highlight=iscoroutinefunction#asyncio.iscoroutinefunction

src/anyio/functools.py
```diff
@@ -0,0 +1,253 @@
...
+    def __init__(
+        self, func: Callable[..., Awaitable[T]], maxsize: int | None, typed: bool
+    ):
+        self.__wrapped__ = func
+        self._hits: int = 0
+        self._misses: int = 0
+        self._maxsize = max(maxsize, 0) if maxsize is not None else None
+        self._currsize: int = 0
+        self._typed = typed
+        update_wrapper(self, func)
```
gschaffner, line 92: If you like:

```suggestion
        self.__wrapped__: Final = func
        self._hits: int = 0
        self._misses: int = 0
        self._maxsize: Final = max(maxsize, 0) if maxsize is not None else None
        self._currsize: int = 0
        self._typed: Final = typed
        update_wrapper(self, func)
```

src/anyio/functools.py
```diff
@@ -0,0 +1,253 @@
...
+
+    This is the asynchronous equivalent to :func:`functools.cache`.
+
+    """
+    return lru_cache(maxsize=None)(func)
+
+
+@overload
+def lru_cache(
+    *, maxsize: int | None = 128, typed: bool = False
```
gschaffner: It looks like these are not keyword-only in the stdlib version. I like making them kw-only here though.

I checked and we already give a clear `TypeError` when someone who isn't type-checking tries `lru_cache(10)`. We might want to add a test for this though, since currently this behavior relies on `10` getting dispatched to the stdlib as `stdlib.lru_cache(max_size=self._maxsize, ...)(func=10)`, which gives the `TypeError`. If a future refactor/whatever caused `10` to get dispatched to `AsyncLRUCacheWrapper(10, maxsize=self._maxsize, ...)` instead, it would result in a more cryptic error.

src/anyio/functools.py
```diff
@@ -0,0 +1,253 @@
...
+            if kwargs:
+                key += (initial_missing,) + tuple(type(val) for val in kwargs.values())
+
+        try:
+            cache = lru_cache_items.get()
+        except LookupError:
+            cache = defaultdict(lambda: (OrderedDict(), Lock()))
+            lru_cache_items.set(cache)
+
+        cache_entry, lock = cache[self.__wrapped__]
```
gschaffner: Perhaps key this on `self` instead of `self.__wrapped__`, in case the user callable isn't hashable? I don't think that stdlib requires the user function to be hashable.

src/anyio/functools.py
```diff
@@ -0,0 +1,253 @@
...
+        if self._maxsize == 0:
+            value = await self.__wrapped__(*args, **kwargs)
+            self._misses += 1
+            return value
+
+        # The key is constructed as a flat tuple to avoid memory overhead
+        key: tuple[Any, ...] = args
+        if kwargs:
+            # initial_missing is used as a separator
+            key += (initial_missing,) + sum(kwargs.items(), ())
```
gschaffner: I think that we need to sort `kwargs` by parameter name. Otherwise `f(a=1, b=2)` and `f(b=2, a=1)` will be cached separately.

src/anyio/functools.py
```diff
@@ -0,0 +1,253 @@
...
+            try:
+                value = cache_entry[key]
+            except KeyError:
+                self._misses += 1
+                if self._maxsize is not None and len(cache_entry) >= self._maxsize:
+                    cache_entry.popitem(last=False)
+                else:
+                    self._currsize += 1
+
+                cache_entry[key] = value = await self.__wrapped__(*args, **kwargs)
```
gschaffner: I don't think that this lock should be held while `__wrapped__` is called. This causes all callers of the function to go single-file. stdlib `lru_cache` releases its lock while calling the user function.

src/anyio/functools.py
gschaffner: Caching on `self.__wrapped__` instead of `self` is also buggy in another way: if I do

```python
a = lru_cache(maxsize=5)(f)
b = lru_cache(maxsize=10)(f)
```

then `a` and `b` must not share the same cache. If they share a cache, and I call `b(i) for i in range(10)` to make the shared cache size 10, and then call `a(20)`, we don't want to evict from `b`'s cache or update `b`'s cache's order. Here's a test that currently fails (but passes on stdlib `lru_cache`):


```python
    async def test_independent_caches(self) -> None:
        async def f(x: int) -> int:
            await checkpoint()
            return x

        a = lru_cache(maxsize=5)(f)
        b = lru_cache(maxsize=10)(f)
        for i in range(10):
            await b(i)
        await a(20)  # the test passes if you comment out this line
        await b(0)
        assert b.cache_info().hits == 1
```

src/anyio/functools.py
```diff
@@ -0,0 +1,253 @@
...
+        self._typed = typed
+        update_wrapper(self, func)
+
+    def cache_info(self) -> AsyncCacheInfo:
+        return AsyncCacheInfo(self._hits, self._misses, self._maxsize, self._currsize)
+
+    def cache_parameters(self) -> AsyncCacheParameters:
+        return AsyncCacheParameters(maxsize=self._maxsize, typed=self._typed)
+
+    def cache_clear(self) -> None:
```
gschaffner, line 104: Perhaps add a `__del__` that calls `cache_clear`?

gschaffner: Nevermind, I don't think we want `__del__`. `RunVar`s are backed by a `WeakKeyDict[NativeToken, ...]` already, and `lru_cache_items.get()` would not be well-defined in `__del__` anyway.

I think that instead of having `RunVar[defaultdict[...]]`, it should be `RunVar[WeakKeyDict[...]]`. Currently, it looks like a cache for a deleted `AsyncLRUCacheWrapper` gets kept around until the corresponding `NativeToken` reaches 0 refcounts.

src/anyio/functools.py
agronholm: The `__all__` serves another function – specifying what names are "exported" from the module. Given Python's unfortunate lack for ECMAScript's `export` statement, this is the best we can do at the moment.

src/anyio/functools.py
agronholm: We don't use these elsewhere as I'm assuming people have migrated to proper coroutines a long ago. If there is a case for this, I'm waiting for someone to open an issue and giving a sensible rationale for still supporting this.

src/anyio/functools.py
agronholm: Not sure I'd bother, given that these are undocumented private fields anyway.

src/anyio/functools.py
agronholm: What is the lock supposed to guard against then? The whole point of the lock is to prevent two tasks from concurrently calling `__wrapped__` when there's no cache entry. If I release the lock before `__wrapped__` is called, how is this goal achieved?

src/anyio/functools.py
agronholm: While somewhat of an edge case, this is true, and I want to have parity with stdlib wherever possible.

agronholm: Fixed.

src/anyio/functools.py
agronholm: Done.

src/anyio/functools.py
agronholm: Let's discuss on Gitter/Matrix.

src/anyio/functools.py
agronholm: Added a test.

src/anyio/functools.py
agronholm: Handled.

src/anyio/functools.py
agronholm: Oh, I think I missed this one before requesting a review.

agronholm: OK, also done now.

src/anyio/functools.py
```diff
@@ -0,0 +1,281 @@
...
+        if cache := lru_cache_items.get(None):
+            cache.pop(self.__wrapped__, None)
```
gschaffner:
```suggestion
            cache.pop(self, None)
```

src/anyio/functools.py
```diff
@@ -0,0 +1,281 @@
...
+if sys.version_info >= (3, 11):
+    from typing import ParamSpec
+else:
+    from typing_extensions import ParamSpec
+
+T = TypeVar("T")
+S = TypeVar("S")
+P = ParamSpec("P")
+lru_cache_items: RunVar[
+    WeakKeyDictionary[Callable[..., Any], tuple[OrderedDict[Hashable, Any], Lock]]
```
gschaffner:
```suggestion
    WeakKeyDictionary[AsyncLRUCacheWrapper, tuple[OrderedDict[Hashable, Any], Lock]]
```

src/anyio/functools.py
gschaffner: The `lru_cache` in the stdlib doesn't have that goal. From the docs:

> It is possible for the wrapped function to be called more than once if another thread makes an additional call before the initial call has been completed and cached.

I can think of a few options for `anyio.functools.lru_cache`:

1.  When there are two concurrent calls of a not-yet-cached call, both will run the user function. This behavior is like the stdlib. `test_concurrent_access` will fail. If we did this, we wouldn't have any locks—it's safe to query/update a cache without a lock because each cache is per-event-loop and querying and updating a cache are synchronous operations.

    I.e. `lru_cache_items: RunVar[WeakKeyDictionary[AsyncLRUCacheWrapper, OrderedDict[ArgsKwargsKey, RetVal]]]`.

2.  Have one lock per cache. Hold the lock during `await self.__wrapped__(anything)`. `test_concurrent_access` will pass. However, `await f(1)` and `await f(2)` cannot execute concurrently; all calls to `f()` are put in a single-file line and made to wait for each other before starting, even when the arguments are different.

    I.e. `lru_cache_items: RunVar[WeakKeyDictionary[AsyncLRUCacheWrapper, tuple[Lock, OrderedDict[ArgsKwargsKey, RetVal]]]]`.

3.  Have one lock (or other synchronization primitive) per `(*args, **kwargs)`. Hold the `(*args, **kwargs)` lock during `await self.__wrapped__(*args, **kwargs)`. `test_concurrent_access` will pass. `await f(1)` and `await f(2)` can execute concurrently.
    
    I.e. `lru_cache_items: RunVar[WeakKeyDictionary[AsyncLRUCacheWrapper, OrderedDict[ArgsKwargsKey, tuple[Lock, RetVal]]]]`.

tests/test_functools.py
```diff
@@ -0,0 +1,294 @@
...
+            def __aiter__(self) -> AIter:
+                return self
+
+            async def __anext__(self) -> NoReturn:
+                raise StopAsyncIteration
+
+        with pytest.raises(
+            TypeError, match=r"reduce\(\) of empty sequence with no initial value"
+        ):
+            await reduce(lambda x, y: x + y, AIter())
```
gschaffner, line 338: A bit of a pathological case, but `reduce()` doesn't currently visit a checkpoint when the iterable is empty. I think that we should add a checkpoint if the iterable is empty.

```suggestion
            await reduce(lambda x, y: x + y, AIter())

    async def test_checkpoints_empty_iterable(self) -> None:
        async def func(x: int, y: int) -> int:
            await checkpoint()
            return x + y

        with CancelScope() as cs:
            cs.cancel()
            with pytest.raises(get_cancelled_exc_class()):
                await reduce(func, [], 1)

    async def test_checkpoints_empty_async_iterable(self) -> None:
        async def func(x: int, y: int) -> int:
            await checkpoint()
            return x + y

        class AIter:
            def __aiter__(self) -> AIter:
                return self

            async def __anext__(self) -> NoReturn:
                raise StopAsyncIteration

        with CancelScope() as cs:
            cs.cancel()
            with pytest.raises(get_cancelled_exc_class()):
                await reduce(func, AIter(), 1)
```

agronholm: It also might not visit a checkpoint if there are none in the coroutine function and async iterable.

gschaffner: Yeah. If there are no checkpoints in the user's function then there is a bug in the user's function though.

In contrast: if `await reduce(function_that_checkpoints, ...)` returns without visiting a checkpoint then I think there is a bug in `reduce`. I don't think that users should have to write

```python
async def foo():
    if not iter_is_empty:
        return await reduce(function_that_checkpoints, iter_, init)
    else:
        # We can't just `return await reduce(function_that_checkpoints, iter_, init)`
        # because `reduce` doesn't checkpoint in this case. We need to add a checkpoint.
        return init
```

agronholm: > Yeah. If there are no checkpoints in the user's function then there is a bug in the user's function though.

From everything I've heard so far from the asyncio devs and users, just about **nobody** thinks this way, and it's very rarely a problem in practice. In contrast, every checkpoint is super expensive due to at least one system call being made. [This](https://github.com/agronholm/anyio/discussions/1007) is a very recent example that shows both the performance cost and also the almost complete lack of awareness of the issue among the users.

The problem for me is the perceived poor performance of AnyIO APIs, given my continuous push for SC and better APIs in the asyncio space. This is a very difficult problem to deal with. Trio has the advantage of being in control of the event loop so they can conditionally allow the task to continue from an `await` w/o necessarily going through the system call, if they so choose. With asyncio that is not an option, leading to a performance hit which is why I'm being so wary of adding any unnecessary checkpoints, and why the `fast_acquire` option was added to locks and semaphores a while back.

src/anyio/functools.py
agronholm: OK, I understand now – it's about the granularity of the locks. I'll fix that.

agronholm: I've fixed the problem and added a test that fails without said fix.

tests/test_functools.py
agronholm: All that said, I've added code to ensure that there is a checkpoint even in the case of an initial value and an empty iterable.

gschaffner: Thanks, I greatly appreciate you taking the time to write that longer message. I see how this problem is difficult. It reminds me of the saying "perfect is the enemy of good." Thank you for pushing for SC/level-cancellation/other ease-of-use improvements in asyncio-land.

src/anyio/functools.py
```diff
@@ -0,0 +1,315 @@
...
+        cached_value: T | _InitialMissingType
+        try:
+            cached_value, lock = cache_entry[key]
+        except KeyError:
+            cached_value, lock = initial_missing, Lock(fast_acquire=True)
+            cache_entry[key] = cached_value, lock
+
+        async with lock:
+            # Check if another task filled the cache while we awaited the lock
+            if (value := cache_entry[key][0]) is initial_missing:
```
gschaffner: This could potentially be optimized further to reduce lock contention and reduce memory usage, but I don't know if optimizing it further would be worth the added code complexity. Potential optimizations:

1.  Bypass the lock when the result is already cached. Roughly:

    ```python
    # Avoid lock contention if we can
    if (value := cache_entry[key][0]) is not initial_missing:
        self._hits += 1
        cache_entry.move_to_end(key)
        return value
    
    async with lock:
        ...
    ```

2.  When writing a new result to the cache, set `cache_entry[key][1] = None` to release memory. Existing waiters of the lock will continue waiting, but the existing waiters will finish and the lock will get deleted shortly.

src/anyio/functools.py
```diff
@@ -0,0 +1,315 @@
...
+        try:
+            cache_entry = cache[self]
+        except KeyError:
+            cache_entry = cache[self] = OrderedDict()
+
+        cached_value: T | _InitialMissingType
+        try:
+            cached_value, lock = cache_entry[key]
+        except KeyError:
+            cached_value, lock = initial_missing, Lock(fast_acquire=True)
```
gschaffner: The change to `fast_acquire=True` gives me pause, although I understand that it's a difficult situation/a difficult tradeoff to decide.

Could it make sense to make `fast_acquire` configurable here, similar to `Lock`? Like `Lock`, users would select correctness when they do not have a performance bottleneck, and select performance when they have a bottleneck that needs it. This would allow users of `lru_cache` to decide the tradeoff themselves based on their application's needs. Since this is a new API without backwards-compatibility concerns like `Lock` had, the default could even be `lru_cache(fast_acquire=True)`.

To be clear, this is entirely at your discretion; I only mean to propose a hypothetical.

agronholm: OK so I think this is yet another case of "should we checkpoint here just because the user does an `await`". It's not clear whether the expectation is that a cached function like this would always checkpoint or not. Looking at existing implementations, I see:

[async_lru_cache](https://github.com/aio-libs/async-lru/blob/master/async_lru/__init__.py): does **not** checkpoint when the result is cached and ready
[asyncstdlib](https://github.com/maxfischer2781/asyncstdlib/blob/master/asyncstdlib/_lrucache.py): also does **not** checkpoint when the result is cached

I have to note that the first implementation *kind of* does locking by creative use of futures and shielding while the second implementation does not even bother preventing multiple tasks from generating the value concurrently.

If I were to apply your other suggestion of avoiding the lock entirely, then `fast_acquire=True` would be moot from the correctness PoV unless we *also* forced a checkpoint on a cache hit (which, given the nature of this whole thing, feels dubious).

src/anyio/functools.py
agronholm: 1. If the result is cached, where would the lock contention come from, given that the task acquires and releases the lock without yielding?
2. Ah, the lock is indeed uselessly taking up memory once the result is cached. And if the entry is ever evicted, a new lock would be created anyway, so getting rid of the lock once we put the result in the cache seems easy enough.

agronholm: I just pushed a change that avoids acquiring the lock entirely when the value is cached.

src/anyio/functools.py
gschaffner: > OK so I think this is yet another case of "should we checkpoint here just because the user does an `await`".

Indeed. I mentioned in [another comment](https://github.com/agronholm/anyio/pull/1001#discussion_r2454094150) that the checkpointing situation on asyncio does seem like a difficult "perfect is the enemy of good" situation. I agree with the conclusion that you and Zac reached in https://github.com/agronholm/anyio/pull/761#issuecomment-2329787356.

> It's not clear whether the expectation is that a cached function like this would always checkpoint or not.

This is exactly why I suspect that allowing users to choose which semantics they expect/want/need would be useful.

> If I were to apply your other suggestion of avoiding the lock entirely, then `fast_acquire=True` would be moot from the correctness PoV unless we _also_ forced a checkpoint on a cache hit (which, given the nature of this whole thing, feels dubious).

Yes, this was somewhat confusing since this comment about checkpoints was somewhat related to the optimization comment. Sorry about that. I think that my comment above would be clearer if written in terms of `lru_cache(always_checkpoint: bool = False)` instead of `lru_cache(fast_acquire: bool = True)`:

We could, in principle, allow users to select `lru_cache(always_checkpoint=True)` without negatively affecting the performance for other users. Adding an `always_checkpoint: bool = False` parameter to `lru_cache` would not be a breaking change since the default would still be to not checkpoint. `Lock(fast_acquire: bool)` is configurable; allowing `lru_cache` to be similarly configurable could allow users to select the tradeoff that suits their project/function/backend best.

gschaffner (approved):

src/anyio/functools.py
agronholm: I've now added the `always_checkpoint` parameter which disables `fast_acquire` and also checkpoints on a cache hit where the lock acquisition is skipped.

agronholm:
Thanks for the review! This was a big help.

PR #752, opened 2024-06-29
Fixed feed_data after feed_eof assertion errors on asyncio

agronholm:
## Changes

This removes the call to `self._stream.feed_eof()`, as it should not be done when the protocol is still sending data (and we can't prevent it from doing so). Instead, we set an internal flag and raise `ClosedResourceError` if someone tries to read from the stream after it's closed.

Fixes #490. 

## Checklist

If this is a user-facing code change, like a bugfix or a new feature, please ensure that
you've fulfilled the following conditions (where applicable):

- [X] You've added tests (in `tests/`) added which would fail without your patch
- [ ] You've updated the documentation (in `docs/`, in case of behavior changes or new
features)
- [X] You've added a new changelog entry (in `docs/versionhistory.rst`).

If this is a trivial change, like a typo fix or a code reformatting, then you can ignore
these instructions.

### Updating the changelog

If there are no entries after the last release, use `**UNRELEASED**` as the version.
If, say, your patch fixes issue #123, the entry should look like this:

`* Fix big bad boo-boo in task groups (#123
<https://github.com/agronholm/anyio/issues/123>_; PR by @yourgithubaccount)`

If there's no issue linked, just link to your pull request instead by updating the
changelog after you've created the PR.

Issue #671, opened 2024-01-17
`ClosedResourceError` vs. `BrokenResourceError` is sometimes backend-dependent, and is sometimes not raised at all (in favor of a non-AnyIO exception type)


gschaffner:
### Things to check first

- [X] I have searched the existing issues and didn't find my bug already reported there

- [X] I have checked that my bug is still present in the latest release

### AnyIO version

main

### Python version

CPython 3.12.1

### What happened?

see the title. here are some reproducers for some issues that i noticed while working on a fix for #669: https://github.com/agronholm/anyio/compare/c6f0334e67818b90540dac20815cad9e0b2c7eee...b6576ae16a9b055e8109c8d2ca81e00bf439cb3f

the main question that these tests raise is: if a send stream is in a `Broken` state (i.e. our side of the connection learned that it was closed by the peer, or otherwise broken due to something external) and then is explicitly `aclose()`d by our side, should subsequent calls raise `BrokenResourceError` or should they raise `ClosedResourceError`? i.e. if you `aclose()` a `Broken` send stream, does that "clear" its `Broken`ness and convert it to just being `Closed`?

the documentation does not seem to discuss what should happen if the conditions for `ClosedResourceError` and `BrokenResourceError` are _both_ met. i am not sure what the behavior here is intended to be (or if it's intended to be undefined behavior?), as currently different streams do different things in this situation, with the behavior sometimes also varying between backends.

my initial intuition was that the intended behavior was to give precedence to `raise BrokenResourceError` over `raise ClosedResourceError` (this choice prevents the stream from changing from `Broken` to `Closed`). I thought this becuase this behavior looks like it was rather explicitly chosen when implementing the "important" asyncio-backend streams (TCP and UDP): they explicitly do _not_ set `self._closed` if they are already closing due to an external cause:

*   `SocketStream` https://github.com/agronholm/anyio/blob/c6efbe352705529123d55f87d6dbb366a3e0612f/src/anyio/_backends/_asyncio.py#L1173-L1184
*   `UDPStream` https://github.com/agronholm/anyio/blob/c6efbe352705529123d55f87d6dbb366a3e0612f/src/anyio/_backends/_asyncio.py#L1459-L1463

so I started to implement it: here is most of an implementation of that behavior: https://github.com/agronholm/anyio/compare/c6f0334e67818b90540dac20815cad9e0b2c7eee...1be403d20f94a8a6522f27f24a1830f2351aab3b [^1]

[...]

PR #1002, opened 2025-10-17
Fix `Process.stdin.send` exceptions and checkpointing on asyncio

gschaffner:
## Changes

Fixes the remaining issue reported in #671:

`_asyncio.Process.stdin.send` not raising `ClosedResourceError` or `BrokenResourceError`, but rather non-AnyIO exceptions. (The analogous `std{out,err}` case was already fixed in #752.)

I also fixed a missing checkpoint before writing to a subprocess send buffer on asyncio.

I also made `_close = False` unconditional in asyncio UDP `aclose`, similar to #980. I'm not sure that this is strictly necessary, but it seems safer and I don't see any downsides. See details in commit message.

## Checklist

features

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -29,7 +29,7 @@ repos:
           - tomli
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.14.0
+    rev: v0.14.1
     hooks:
       - id: ruff-check
         args: [--fix, --show-fixes]
--- a/tests/test_from_thread.py
+++ b/tests/test_from_thread.py
@@ -690,6 +690,7 @@ class TestBlockingPortal:
 
         assert not caplog.text
 
+    @pytest.mark.skip("This test is known to hang often on CI")
     def test_raise_baseexception_from_task(
         self, anyio_backend_name: str, anyio_backend_options: dict[str, Any]
     ) -> None:


src/anyio/_backends/_asyncio.py
```diff
@@ -1052,12 +1052,26 @@ async def aclose(self) -> None:
...
+        except (ConnectionResetError, BrokenPipeError, RuntimeError) as exc:
+            # If closed by us and/or the peer:
+            # * on stdlib, drain() raises ConnectionResetError or BrokenPipeError
+            # * on uvloop, write() raises RuntimeError. drain() probably raises other
+            #   things too, but it's hard to test.
+            if self._closed:
+                raise ClosedResourceError from exc.__cause__
+            elif self._stream.is_closing():
+                raise BrokenResourceError from exc.__cause__
+            raise
```
gschaffner, line 1072: We could also use a custom `Protocol.connection_lost` instead of catching/remapping these exceptions, i.e. use `BaseEventLoop.subprocess_exec` instead of the higher-level `asyncio.create_subprocess_exec`. That seems a bit annoying to do however, and it would just be deleted if/when we pursue #783 anyway. In the meantime, catching the exceptions seems to be an okay fix...it's possible that we have missed some exception types in the `except` tuple, but it seems to me that the exception semantics of this function should be no worse than they are without this exception remapping.

src/anyio/_backends/_asyncio.py
```diff
@@ -1052,12 +1052,26 @@ async def aclose(self) -> None:
 @dataclass(eq=False)
 class StreamWriterWrapper(abc.ByteSendStream):
```
gschaffner, line 1053: Off topic: I noticed that we don't have `ResourceGuard`s in `_asyncio.Stream{Reader,Writer}Wrapper` and `_trio.{Receive,Send}StreamWrapper`. Should we add guards?

src/anyio/_backends/_asyncio.py
```diff
@@ -1052,12 +1052,26 @@ async def aclose(self) -> None:
...
     _stream: asyncio.StreamWriter
+    _closed: bool = field(init=False, default=False)
 
     async def send(self, item: bytes) -> None:
-        self._stream.write(item)
-        await self._stream.drain()
+        await AsyncIOBackend.checkpoint()
```
agronholm: We already have `self._stream.drain()`, so I don't think we need a second full checkpoint here. What we should add instead is `AsyncIOBackend.checkpoint_if_cancelled()` to respond to cancellation before the synchronous `stream.write()`.

src/anyio/_backends/_asyncio.py
```diff
@@ -1052,12 +1052,26 @@ async def aclose(self) -> None:
...
+        except (ConnectionResetError, BrokenPipeError, RuntimeError) as exc:
+            # If closed by us and/or the peer:
+            # * on stdlib, drain() raises ConnectionResetError or BrokenPipeError
+            # * on uvloop, write() raises RuntimeError and drain() doesn't appear to
+            #   raise anything
+            if self._closed:
+                raise ClosedResourceError from exc.__cause__
+            elif self._stream.is_closing():
+                raise BrokenResourceError from exc.__cause__
+            raise
```
agronholm, line 1072: Add a blank line after a control block ends.
```suggestion

            raise
```

src/anyio/_backends/_asyncio.py
agronholm: Why are we chaining these from `exc.__cause__` and not `exc` directly?

tests/test_subprocesses.py
```diff
@@ -276,6 +268,27 @@ async def test_process_aexit_cancellation_closes_standard_streams(
...
+            await process.stdout.receive(1)
+        with pytest.raises(EndOfStream):
+            await process.stderr.receive(1)
+
+    with pytest.raises(ClosedResourceError):
+        await process.stdin.send(b"foo")
+    with pytest.raises(ClosedResourceError):
+        await process.stdout.receive(1)
+    with pytest.raises(ClosedResourceError):
+        await process.stderr.receive(1)
```
agronholm, line 310: Add a blank line after a control block ends.
```suggestion
        with pytest.raises(BrokenResourceError):
            await process.stdin.send(b"foo")

        with pytest.raises(EndOfStream):
            await process.stdout.receive(1)

        with pytest.raises(EndOfStream):
            await process.stderr.receive(1)

    with pytest.raises(ClosedResourceError):
        await process.stdin.send(b"foo")

    with pytest.raises(ClosedResourceError):
        await process.stdout.receive(1)

    with pytest.raises(ClosedResourceError):
        await process.stderr.receive(1)
```

src/anyio/_backends/_asyncio.py
agronholm: Perhaps that would be wise. I'm not sure if Trio doesn't already have a mechanism in place to guard against concurrent access. It would be good to check first.

src/anyio/_backends/_asyncio.py
gschaffner: It looks like `drain()` might not checkpoint in all branches: https://github.com/python/cpython/blob/v3.14.0/Lib/asyncio/streams.py#L167-L168

src/anyio/_backends/_asyncio.py
gschaffner: Fixed. I had incorrectly borrowed that from

```python
except trio.ClosedResourceError as exc:
    raise ClosedResourceError from exc.__cause__
```

src/anyio/_backends/_asyncio.py
gschaffner: Trio already has its own `ConflictDetector`s in standard streams (`FdStream`, `PipeSendStream`, and `PipeReceiveStream`), but they raise `trio.BusyResourceError`, not `anyio.BusyResourceError`, so I think we should to add our own guards on the Trio backend too.

Perhaps let's handle this in a separate PR after this one. It looks like there are also a few other places where one can currently get a `ConflictDetector`'s `trio.BusyResourceError` on AnyIO instead of an `anyio.BusyResourceError`.

Vizonex:
> > @Vizonex there's a new test here that is only failing on Winloop—would you be able to take a look? I will take a look tomorrow too.
> 
> > Note that a test on Windows + Python 3.9 is genuinely failing.
> 
> See [Vizonex/Winloop#84](https://github.com/Vizonex/Winloop/issues/84)

@gschaffner @agronholm I saw it and I have now made a fork. Thanks for writing me a reproducible test those really help me understand what it is that is going on. I'll try and come up with a template as well when the bug for winloop is fixed so people know I want reproducible tests =)

agronholm:
> @agronholm I fixed the bug on winloop's end there should be a release for 0.3.1 shortly here later today.

There seems to be a failure on 3.13 now with winloop. 3.9 passes though.

src/anyio/_backends/_asyncio.py
agronholm: OK, how about checkpointing based on `self._stream._paused`?
```py
   async def send(self, item: bytes) -> None:
        await checkpoint_if_cancelled()
        stream_paused = self._stream._paused
        try:
            self._stream.write(item)
            await self._stream.drain()
        except (ConnectionResetError, BrokenPipeError, RuntimeError) as exc:
            # If closed by us and/or the peer:
            # * on stdlib, drain() raises ConnectionResetError or BrokenPipeError
            # * on uvloop, write() raises RuntimeError and drain() doesn't appear to
            #   raise anything
            if self._closed:
                raise ClosedResourceError from exc
            elif self._stream.is_closing():
                raise BrokenResourceError from exc

            raise

        if not stream_paused:
            await cancel_shielded_checkpoint()
```
I know it's a private attribute but our CI should alert us if it's ever changed.

gschaffner:
It looks like `test_exceptions_after_subprocess_closes_standard_streams[asyncio+winloop]` is still failing sporadically on Winloop 0.3.1, on both 3.9 and 3.13. Looking into it.

src/anyio/_backends/_asyncio.py
gschaffner: Sounds good. I'm assuming that you meant `self._stream._protocol._paused` since `self._stream._paused` does not exist.

However: when I do make this change, `test_exceptions_after_subprocess_closes_standard_streams[asyncio+uvloop]` starts failing. (It's fine on stdlib asyncio.) Without the checkpoint, it seems like we might be hitting a uvloop bug that is similar to the Winloop bug. Looking into it.

gschaffner:
When I remove the schedule point at the start of `send()`, the test that is sporadically failing on `asyncio+winloop` (`test_exceptions_after_subprocess_closes_standard_streams`) starts failing on `asyncio+uvloop` too (but not stdlib asyncio). I'm wondering if there is a problem with the approach or the test here.

I suppose it's just a race? Perhaps it's fine for `subprocess.stdin.send()` to return success once or twice after the peer has closed their end of the pipe—even if the event loop[^1] already knows that the peer closed their end of the pipe—as long as `send()` *eventually* starts raising `BrokenResourceError`?

[^1]: or at least some parts of the event loop

gschaffner:
I.e. perhaps the test is too strict, and we should amend it?:

```diff
diff --git a/tests/test_subprocesses.py b/tests/test_subprocesses.py
index c18df94..736ef8c 100644
--- a/tests/test_subprocesses.py
+++ b/tests/test_subprocesses.py
@@ -275,7 +275,8 @@ async def test_exceptions_after_subprocess_closes_standard_streams() -> None:
         assert process.stderr is not None
         assert process.stdout is not None
         with pytest.raises(BrokenResourceError):
-            await process.stdin.send(b"foo")
+            for _ in range(5):
+                await process.stdin.send(b"foo")
 
         with pytest.raises(EndOfStream):
             await process.stdout.receive(1)
```

src/anyio/_backends/_asyncio.py
gschaffner: Done.

PR #1009, opened 2025-10-21
Added `__all__` to all public modules

agronholm:
## Changes

This adds the `__all__` attribute to all documented `anyio` submodules. This is not to encourage wildcard imports, but offer some structure to IDEs and other tools to determine which items are "exported" from the modules. This is helpful, for example, in narrowing down the suggestions for automatic imports.

## Checklist

features

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/tests/test_sockets.py
+++ b/tests/test_sockets.py
@@ -12,6 +12,7 @@ import sys
 import tempfile
 import threading
 import time
+import warnings
 from collections.abc import Generator, Iterable, Iterator
 from contextlib import suppress
 from ipaddress import IPv4Address, IPv6Address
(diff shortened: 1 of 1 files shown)


src/anyio/_core/_streams.py
```diff
@@ -7,7 +7,7 @@
 from ..streams.memory import (
     MemoryObjectReceiveStream,
     MemoryObjectSendStream,
-    MemoryObjectStreamState,
+    _MemoryObjectStreamState,
```
davidbrochart, line 10: Does this mean that it will be a breaking change?

agronholm: No, this class was not part of the API in the first place. It was never in the documentation, and was not intended to be used by users.

PR #1010, opened 2025-10-23
Added a decorator for safe use of async generators

agronholm:
## Changes

This is the Trio counterpart for `@as_safe_channel`, but exposes async generators as async generators, not as channels/streams.

## Checklist

features

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/src/anyio/_core/_utils.py
+++ b/src/anyio/_core/_utils.py
@@ -1,0 +1,168 @@
+from __future__ import annotations
+
+import sys
+from collections.abc import AsyncGenerator, Callable
+from contextlib import AbstractAsyncContextManager, asynccontextmanager
+from dataclasses import dataclass
+from functools import wraps
+from types import TracebackType
+from typing import overload
+
+from anyio import Lock, create_memory_object_stream, create_task_group
+from anyio.abc import TaskStatus
+from anyio.streams.memory import MemoryObjectReceiveStream, MemoryObjectSendStream
+
+if sys.version_info >= (3, 13):
+    from typing import TypeVar
+else:
+    from typing_extensions import TypeVar
+
+if sys.version_info < (3, 11):
+    from exceptiongroup import BaseExceptionGroup
+
+if sys.version_info >= (3, 10):
+    from typing import ParamSpec
+else:
+    from typing_extensions import ParamSpec
+
+YieldT_co = TypeVar("YieldT_co", covariant=True)
+SendT_contra = TypeVar("SendT_contra", contravariant=True, default=None)
+P = ParamSpec("P")
+
+
+@dataclass(frozen=True)
+class _ExceptionWrapper:
+    exc: BaseException
+
+
+class _GeneratorRunner(AsyncGenerator[YieldT_co, SendT_contra]):
+    __slots__ = (
+        "_asyncgen",
+        "_send_to_asyncgen",
+        "_receive_from_asyncgen",
+        "_lock",
+        "_exc",
+        "_item",
+    )
+
+    _send_to_asyncgen: MemoryObjectSendStream[SendT_contra | _ExceptionWrapper | None]
+    _receive_from_asyncgen: MemoryObjectReceiveStream[YieldT_co | _ExceptionWrapper]
+
+    def __init__(self, asyncgen: AsyncGenerator[YieldT_co, SendT_contra]) -> None:
+        self._asyncgen = asyncgen
+        self._lock = Lock()
+        self._exc: BaseException | None = None
+
+    def __aiter__(self) -> AsyncGenerator[YieldT_co, SendT_contra]:
+        return self
+
+    async def __anext__(self) -> YieldT_co:
+        return await self._send_item_to_asyncgen(None)
+
+    async def _send_item_to_asyncgen(
+        self, item: SendT_contra | _ExceptionWrapper | None
+    ) -> YieldT_co:
+        async with self._lock:
+            await self._send_to_asyncgen.send(item)
+            received_item = await self._receive_from_asyncgen.receive()
+            if isinstance(received_item, _ExceptionWrapper):
+                raise received_item.exc from None
+            else:
+                return received_item
+
+    async def asend(self, item: SendT_contra) -> YieldT_co:
+        return await self._send_item_to_asyncgen(item)
+
+    @overload
+    async def athrow(
+        self,
+        typ: type[BaseException],
+        val: BaseException | object = None,
+        tb: TracebackType | None = None,
+        /,
+    ) -> YieldT_co: ...
+
+    @overload
+    async def athrow(
+        self, typ: BaseException, val: None = None, tb: TracebackType | None = None, /
+    ) -> YieldT_co: ...
+
+    async def athrow(
+        self,
+        typ: type[BaseException] | BaseException,
+        val: BaseException | object | None = None,
+        tb: TracebackType | None = None,
+        /,
+    ) -> YieldT_co:
+        if isinstance(val, BaseException):
+            exc = val
+        elif isinstance(typ, BaseException):
+            exc = typ
+        else:
+            exc = typ(val) if val is not None else typ()
+
+        return await self._send_item_to_asyncgen(_ExceptionWrapper(exc))
+
+    async def aclose(self) -> None:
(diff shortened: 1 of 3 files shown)


src/anyio/_core/_utils.py
```diff
@@ -0,0 +1,168 @@
...
+                            item_from_agen = await self._asyncgen.athrow(
+                                item_to_send.exc
+                            )
+                        elif item_to_send is None:
+                            item_from_agen = await self._asyncgen.__anext__()
+                        else:
+                            item_from_agen = await self._asyncgen.asend(item_to_send)
+                    except BaseException as exc:
+                        send_to_outside.send_nowait(_ExceptionWrapper(exc))
+                        if not isinstance(exc, Exception):
```
m-u-xyz, line 132: Don't re-raise `BaseException`s. That tends not to end well.

agronholm: Huh? Swallowing base exceptions like `CancelledError` is a frequent source of bugs and weird behavior. You're suggesting to swallow cancellation exceptions? How then could the host task be cancelled?

m-u-xyz: Sorry, my fault, I tend to be entirely too good at sloppy writing.

What I *meant* was that posting a `BaseException` to the `send_to_outside` queue (and thus re-raising it in the outside context as well as inside) is a bad idea.

m-u-xyz: The exception send to the outside should be some non-Base exception (probably with the `BaseException` set as its cause) instead.

agronholm: Well, we need to communicate _somehow_ that something unexpected happened in the generator, right?

m-u-xyz: > Well, we need to communicate _somehow_ that something unexpected happened in the generator, right?

Of course. But duplicating and forwarding a `CancelledError` to a context which might not be contained in the frame that issued the cancellation looks like a bug waiting to happen.

agronholm: Sure, but this is still a draft for a reason.

graingert:
Reading through this, I think the bidirectional generator interface (`asend`/`athrow`) is what's driving most of the complexity here — the two-stream request/response, the `_ExceptionWrapper` round-tripping, and the `BaseException`-across-the-boundary issue @smurfix flagged all fall out of supporting send/throw into the wrapped generator.

Worth noting that `trio.as_safe_channel` deliberately *doesn't* do that. It exposes a receive-only `ReceiveChannel` and nothing else, which sidesteps the whole category:

- Only yielded values ever cross the task boundary, so there's no need to marshal exceptions in-band and re-raise them in the consumer's task. A backend `Cancelled` stays in the background task's scope where it belongs — which is exactly the resurrection hazard @smurfix is pointing at (`raise received_item.exc` re-raising a `Cancelled`/`CancelledError` outside its scope → `TrioInternalError` on trio).
- A generator exception just propagates out of the background task and through the task group, where it gets flattened back to a bare exception. No `_ExceptionWrapper`, no `send_nowait`-into-a-zero-buffer edge case.

If `asend`/`athrow` aren't a hard requirement for this API, mirroring trio more closely gets a lot simpler. The structure that maps over directly:

- one memory object stream + a `Semaphore` for lazy pull (only advance the generator when the consumer actually calls `receive()` — no read-ahead),
- a small receive-side wrapper that releases the semaphore and uses a **sync** `with`/`close()` (trio uses a sync CM here specifically because `async with` can swallow exceptions, cf. python-trio/trio#1559),
- `tg.start()` so the generator/stream teardown is installed before the receive end is handed out,
- flatten the task group's `BaseExceptionGroup` back to a single exception, cancel-aware, à la trio's `raise_single_exception_from_group`,
- the `GeneratorExit`-out-of-`aclose` unwrap (python-trio/trio#3324).

The only backend substitution that matters for correctness is `trio.Cancelled` → `anyio.get_cancelled_exc_class()` so the flatten step is cancel-aware on both backends.

<details>
<summary>Sketch (receive-only, mirrors <code>trio._channel.as_safe_channel</code>) — passes on asyncio + trio</summary>

```python

import sys
from collections.abc import AsyncGenerator, Callable
from contextlib import AbstractAsyncContextManager, asynccontextmanager
from functools import wraps
from typing import NoReturn

import anyio
from anyio import BrokenResourceError, create_memory_object_stream
from anyio.abc import ObjectReceiveStream, TaskStatus
from anyio.streams.memory import MemoryObjectReceiveStream, MemoryObjectSendStream

if sys.version_info >= (3, 10):
    from typing import ParamSpec, TypeVar
else:
    from typing_extensions import ParamSpec, TypeVar

T = TypeVar("T")
P = ParamSpec("P")


def _raise_single_exception_from_group(eg: BaseExceptionGroup) -> NoReturn:
    cancelled_type = anyio.get_cancelled_exc_class()
    for e in eg.exceptions:
        if isinstance(e, (KeyboardInterrupt, SystemExit)):
            raise type(e)(*e.args) from eg
    cancelled = noncancelled = None
    for e in eg.exceptions:
        if isinstance(e, cancelled_type):
            cancelled = cancelled or e
        elif noncancelled is None:
            noncancelled = e
        else:
            raise eg  # >1 non-cancelled: don't pretend we can unwrap
    exc = noncancelled if noncancelled is not None else cancelled
    assert exc is not None
    context = exc.__context__  # don't let the group clobber __context__
    try:
        raise exc
    finally:
        exc.__context__ = context
        del exc, context


class _RecvStreamWrapper(ObjectReceiveStream[T]):
    def __init__(self, recv: MemoryObjectReceiveStream[T], sem: anyio.Semaphore) -> None:
        self._recv = recv
        self._sem = sem

    async def receive(self) -> T:
        self._sem.release()
        return await self._recv.receive()

        await self._recv.aclose()

    def __enter__(self):  # sync cm: async with could swallow exceptions
        return self

    def __exit__(self, *exc: object) -> None:
        self._recv.close()


def safe_asyncgen(
    func: Callable[P, AsyncGenerator[T, None]],
) -> Callable[P, AbstractAsyncContextManager[ObjectReceiveStream[T]]]:
    @asynccontextmanager
    @wraps(func)
    async def cm(*args: P.args, **kwargs: P.kwargs):
        send, recv = create_memory_object_stream[T](0)
        try:
                agen = func(*args, **kwargs)
                sem = anyio.Semaphore(0)
                await tg.start(_pump, agen, send, sem)
                with _RecvStreamWrapper(recv, sem) as wrapped:
                    yield wrapped
                tg.cancel_scope.cancel()
        except BaseExceptionGroup as eg:
            _raise_single_exception_from_group(eg)

    async def _pump(
        agen: AsyncGenerator[T, None],
        send: MemoryObjectSendStream[T],
        sem: anyio.Semaphore,
        *,
        task_status: TaskStatus[None] = anyio.TASK_STATUS_IGNORED,
    ) -> None:
        with send:  # sync close so it can't eat exceptions
            try:
                task_status.started()
                while True:
                    await sem.acquire()
                    if not send._state.open_receive_channels:
                        break
                    try:
                        value = await agen.__anext__()
                    except StopAsyncIteration:
                        return
                    try:
                        await send.send(value)
                    except BrokenResourceError:
                        break
            finally:
                try:
                    await agen.aclose()
                except BaseExceptionGroup as exceptions:
                    _removed, narrowed = exceptions.split(GeneratorExit)
                    if narrowed is not None:
                        raise narrowed from None

    return cm
```

</details>

A generator that raises surfaces as a bare exception at the `async for`, early-break runs the generator's `finally`, and the semaphore gives the same no-read-ahead behaviour trio has. One caveat carried over verbatim from trio: it pokes `send._state.open_receive_channels` to skip a possibly-expensive generator step when the receiver is already gone — same private-attr reliance trio has, so if that bothers us it's worth solving for both.

So I guess the real question is scope: is bidirectional `asend`/`athrow` something we want to commit to supporting here, or is a receive-only "safe producer" enough? If it's the former, I'd still suggest lifting trio's exception helpers (`raise_single_exception_from_group` + the context-preserving re-raise + the `aclose`/`GeneratorExit` unwrap) rather than the one-level `excgrp.exceptions[0]` flatten, since the cancel/KI handling is the subtle part. Happy to push either direction into a branch with tests.

Issue #357, opened 2021-08-10
RuntimeError when cancelling asyncio.Task running within a BlockingPortal.


jonathanslenders:
Hi, I'm facing some issues with cancellation and blocking portals and wondering whether that's a bug in anyio, or in my code.

To reproduce:
- Create a blocking portal.
- Schedule a coroutine in the blocking portal.
- This coroutine creates an ``asyncio.Task`` and waits for it.
- This ``asyncio.Task`` gets cancelled from elsewhere, within the event loop. So, we don't cancel the ``concurrent.futures.Future`` that gets returned by the ``BlockingPortal.start_task_soon``.
- A ``RuntimeError`` gets logged.

I don't have a code snippet to reproduce, but if required, I can probably create it. To me, the ``RuntimeError`` doesn't make sense.

```python
Traceback (most recent call last):
  File "/home/jonathan/.pyenv/versions/3.9.5/lib/python3.9/concurrent/futures/_base.py", line 329, in _invoke_callbacks
    callback(self)
  File "/home/jonathan/git/anyio/src/anyio/from_thread.py", line 166, in callback
    self.call(scope.cancel)
  File "/home/jonathan/git/anyio/src/anyio/from_thread.py", line 230, in call
    return cast(T_Retval, self.start_task_soon(func, *args).result())
  File "/home/jonathan/git/anyio/src/anyio/from_thread.py", line 291, in start_task_soon
    self._check_running()
  File "/home/jonathan/git/anyio/src/anyio/from_thread.py", line 140, in _check_running
    raise RuntimeError('This method cannot be called from the event loop thread')
RuntimeError: This method cannot be called from the event loop thread
```

agronholm:
Yes, please create a minimal reproducing case. This traceback doesn't show where the call really originated from.

jonathanslenders:
This should do it:

```python
import asyncio
from anyio.from_thread import BlockingPortal


async def main():
    async with BlockingPortal() as portal:
        await asyncio.get_running_loop().run_in_executor(
            None, lambda: in_thread(portal)
        )


def in_thread(portal):
    async def coro():
        pass

    async def in_loop():
        t = asyncio.create_task(coro())
        t.cancel()
        await t

    portal.start_task_soon(in_loop)


asyncio.run(main())
```

Output:
```
 $ python /tmp/test.py
exception calling callback for <Future at 0x7f9ebd9b4280 state=cancelled>
Traceback (most recent call last):
[...]

PR #1012, opened 2025-10-25
Fix `BlockingPortal.start_task_soon` `Future.cancel` TOCTOU bug

### LINKED ISSUE #1011: Cancelling a `Future` from `BlockingPortal.start_task_soon()` occasionally doesn't cancel the function
### Things to check first

### AnyIO version

master (14356be497c892abf372599efe84dd27ab9c4f09)

### Python version

CPython 3.13.7 & PyPy 3.11.13 (7.3.20)

### What happened?

Flaky test failures of `TestBlockingPortal.test_start_task_soon_cancel_immediately`. Failures happen on both asyncio and Trio, and on both CPython and PyPy.

For example https://github.com/agronholm/anyio/actions/runs/18794158431/job/53630833111 on asyncio + PyPy 3.11.

Run `while pytest -k 'test_start_task_soon_cancel_immediately' -v -r=A; do true; done` until it fails.

On my machine it seems to usually fail quicker (5–60 seconds) on PyPy3.11 than on CPython 3.13.7.

Analysis: The test failures happen when `BlockingPortal._call_func.callback` is called by `future.add_done_callback(callback)` in the event loop thread. The failures are caused by a TOCTOU bug here:

https://github.com/agronholm/anyio/blob/14356be497c892abf372599efe84dd27ab9c4f09/src/anyio/from_thread.py#L256-L259

The test fails when the following scheduling order happens: `future.cancelled()` returns `False` in the event loop thread, then `future.cancel()` happens in the sync thread, then `future.add_done_callback(callback)` runs in the event loop thread, calling `callback` in the event loop thread.

gschaffner:
## Changes

Fixes #1011.

## Checklist

features

### DIFF
--- a/src/anyio/from_thread.py
+++ b/src/anyio/from_thread.py
@@ -244,9 +244,9 @@ class BlockingPortal:
     ) -> None:
         def callback(f: Future[T_Retval]) -> None:
             if f.cancelled():
-                if self._event_loop_thread_id in (None, get_ident()):
+                if self._event_loop_thread_id == get_ident():
                     scope.cancel("the future was cancelled")
-                else:
+                elif self._event_loop_thread_id is not None:
                     self.call(scope.cancel, "the future was cancelled")
 
         try:


src/anyio/from_thread.py
```diff
@@ -243,21 +243,17 @@ async def _call_func(
         future: Future[T_Retval],
     ) -> None:
         def callback(f: Future[T_Retval]) -> None:
-            if f.cancelled() and self._event_loop_thread_id not in (
-                None,
-                get_ident(),
-            ):
-                self.call(scope.cancel, "the future was cancelled")
+            if f.cancelled():
+                if self._event_loop_thread_id in (None, get_ident()):
```
gschaffner: I'm not exactly sure what we should do when `_event_loop_thread_id is None`. When this is `None`, we don't know what thread we're in. We might be in the event loop thread or a different thread, so we can't tell if we are supposed to do `call_in_loop_thread(scope.cancel)` or `scope.cancel()`.

See also #357.

gschaffner: Okay, I have adjusted this PR so that it preserves the existing behavior that we have when `_event_loop_thread_id is None`, and I've opened #1013 so that we can figure out the `_event_loop_thread_id is None` situation separately.

agronholm:
I can't reproduce it locally. Did you do something special?

gschaffner:
I just ran it in a loop. It consistently reproduces in ~seconds (on master and on this branch) if I apply the following patch and then run `tox -e py312 -- -k 'test_raise_baseexception_from_task' -v -r=A -o=faulthandler_timeout=5`:

```diff
diff --git a/tests/test_from_thread.py b/tests/test_from_thread.py
index 3ac480e..62dc70e 100644
--- a/tests/test_from_thread.py
+++ b/tests/test_from_thread.py
@@ -690,9 +690,12 @@ class TestBlockingPortal:
 
         assert not caplog.text
 
-    @pytest.mark.skip("This test is known to hang often on CI")
+    @pytest.mark.parametrize("execution_number", range(1_000))
     def test_raise_baseexception_from_task(
-        self, anyio_backend_name: str, anyio_backend_options: dict[str, Any]
+        self,
+        anyio_backend_name: str,
+        anyio_backend_options: dict[str, Any],
+        execution_number: int,
     ) -> None:
         """
         Test that when a task raises a BaseException, it does not trigger additional
```

agronholm:
Alright, I'll try to get this PR review on Sunday.

agronholm (approved):

PR #1019, opened 2025-11-02
Allowed 0 tokens in a CapacityLimiter

agronholm:
This matches the new behavior in Trio.

## Checklist

- [X] You've updated the documentation (in `docs/`), in case of behavior changes or new
features

these instructions.

### Updating the changelog

```
```


src/anyio/_backends/_asyncio.py
```diff
@@ -1990,7 +1990,7 @@ def total_tokens(self, value: float) -> None:
         if not isinstance(value, int) and not math.isinf(value):
             raise TypeError("total_tokens must be an int or math.inf")
         if value < 1:
```
davidbrochart: Should it be:
```suggestion
        if value < 0:
```

agronholm: Good catch, fixed!

davidbrochart (approved):

PR #1021, opened 2025-11-02
Dropped sniffio as a hard dependency

agronholm:
## Changes

This drops sniffio as a hard dependency. It's unnecessary baggage for downstream users who only care about the asyncio backend.

## Checklist

features

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -29,7 +29,7 @@ repos:
           - tomli
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.14.2
+    rev: v0.14.3
     hooks:
       - id: ruff-check
         args: [--fix, --show-fixes]


src/anyio/_core/_eventloop.py
```diff
@@ -164,3 +189,40 @@ def get_async_backend(asynclib_name: str | None = None) -> type[AsyncBackend]:
         module = import_module(f"anyio._backends._{asynclib_name}")
         loaded_backends[asynclib_name] = module.backend_class
         return module.backend_class
+
+
+def current_async_library() -> str | None:
+    try:
+        import sniffio
```
x42005e1f: I think that importing an uninstalled module on every call is a bad idea. It will lead to a significant drop in performance for all operations:

```python
>>> import sys
>>> from timeit import timeit
>>> def a():
...     return "uninstalled" in sys.modules
>>> def b():
...     try:
...             import uninstalled
...     except ImportError:
...             return False
...     else:
...             return True
>>> timeit(a, number=1000)
0.0002316311001777649
>>> timeit(b, number=1000)
0.10281949606724083
```

I have shown one alternative here, and it is a simple check of `sys.modules`. This is sufficient for Trio support, but for more complex cases, smarter approaches may be useful. These are described [in my documentation](https://aiologic.readthedocs.io/latest/advanced-topics/libraries.html#how-does-aiologic-import-libraries).

agronholm: I just pushed a change that does an eager conditional import of `sniffio`.

x42005e1f: Yes, the eager import will also be sufficient. AnyIO uses a fairly old sniffio API, so there should be no conflicts with the installed sniffio version. It does not support runtime installation, but this is a rather unjustified option.

In fact, performance can be improved further, but only slightly. And more global optimizations are beyond the scope of this PR.

PR #1023, opened 2025-11-04
Fix Python 3.14 compatibility

mgrossu:
## Changes
Fix python3.14 compatibiliy by adjusting pathlib.PurePath.relative_to() and guarding removed asyncio.get_child_watcher() API
Building version 3.7.1 with Python 3.14 fails due to API removals:
- asyncio get_child_watcher() function have been removed
- pathlib.PurePath.relative_to() no longer supports multiple arguments

### DIFF
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -1061,9 +1061,21 @@ def _forcibly_shutdown_process_pool_on_exit(
     Forcibly shuts down worker processes belonging to this event loop."""
     child_watcher: asyncio.AbstractChildWatcher | None
     try:
-        child_watcher = asyncio.get_event_loop_policy().get_child_watcher()
-    except NotImplementedError:
+        policy = asyncio.get_event_loop_policy()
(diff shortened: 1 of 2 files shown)


MeggyCal:
Hi, we are from openSUSE. We are maintaining 3.x release because some not-so-fast packages are still incompatible with 4.x.

As most of the python world we are working on 3.14 compatibility and we added this patch. We only wanted to contribute to the wider community by sharing it with you, because I (incorrectly) thought 3.x is still maintained. If you don't wan't it, feel free to close this PR.

agronholm:
I'll consider it but I'm leaning towards not making another 3.x release. Which packages are holding you back?

MeggyCal:
Hi, watchfiles is the package which still needs anyio v3: https://github.com/samuelcolvin/watchfiles/issues/254

But during the process I found out that it could be updated in the distribution, I'll try that.

agronholm:
Indeed, `watchfiles` does not have an upper bound for its `anyio` dependency.

PR #1031, opened 2025-11-13
Added a new test job to test against the oldest allowed dependencies

agronholm:
## Changes

This tests that our lowest bound dependencies still work.

This change also makes us use `uv` on GitHub actions for all commands involving Python.

## Checklist

features

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -106,12 +106,12 @@ jobs:
         cache-dependency-path: pyproject.toml
     - name: Install the project and its dependencies
       if: matrix.lowest-bound == false
-      run: pip install --group test -e .
+      run: pip install --group test -e .${{ matrix.extras }}
     - name: Install the project and its dependencies (lowest bound)
       if: matrix.lowest-bound == true
       run: |
         pip install uv
-        uv pip install --system --resolution lowest-direct --group test -e .
+        uv pip install --system --resolution lowest-direct --group test -e .${{ matrix.extras }}
     - name: Patch /etc/hosts
       if: runner.os != 'Windows'
       run: |


.github/workflows/test.yml
```diff
@@ -110,13 +112,15 @@ jobs:
       run: |
         Add-Content -Path C:\Windows\System32\drivers\etc\hosts -Value "1.2.3.4 xn--fa-hia.de"
         Add-Content -Path C:\Windows\System32\drivers\etc\hosts -Value "5.6.7.8 fass.de"
+    - name: Install uv
+      run: pip install uv
```
davidbrochart: Why not using https://github.com/astral-sh/setup-uv?

agronholm: I forgot it existed. I'll change the PR to use that.

davidbrochart (approved):

agronholm:
I'm unconvinced that `uv` is actually installing the lowest bound dependencies. I checked and it installed pytest 8.4.2 instead of 7.0.0.

agronholm:
Never mind, there was a bug in my workflow. It installs the stuff I expect now.

.github/workflows/test.yml
```diff
@@ -63,6 +63,7 @@ jobs:
         os: [ubuntu-latest]
         python-version: ["3.9", "3.10", "3.11", "3.12", "3.13", "3.14", pypy-3.11]
         extras: ["[trio]"]
+        lowest-bound: [false]
```
davidbrochart, line 66: Maybe?
```suggestion
        lowest-bound: false
```

.github/workflows/test.yml
```diff
@@ -99,7 +105,13 @@ jobs:
         cache: pip
         cache-dependency-path: pyproject.toml
     - name: Install the project and its dependencies
+      if: matrix.lowest-bound == false
       run: pip install --group test -e .${{ matrix.extras }}
+    - name: Install the project and its dependencies (lowest bound)
+      if: matrix.lowest-bound == true
+      run: |
+        pip install uv
```
davidbrochart, line 113: I thought you were using https://github.com/astral-sh/setup-uv?

.github/workflows/test.yml
```diff
@@ -99,7 +105,13 @@ jobs:
         cache: pip
         cache-dependency-path: pyproject.toml
     - name: Install the project and its dependencies
+      if: matrix.lowest-bound == false
       run: pip install --group test -e .${{ matrix.extras }}
```
davidbrochart, line 109: Aren't you using `uv`?

.github/workflows/test.yml
agronholm: This would be a syntax error. The values in the matrix must be in a list.

.github/workflows/test.yml
agronholm: I only needed uv for its pip-like interface which can do lowest-bound resolution. Turns out this is the simplest way to achieve that. No need to go all in on `uv`.

.github/workflows/test.yml
agronholm: No need to if I don't have to.

agronholm:
Thanks!

PR #1049, opened 2025-12-15
Replaced the NoCurrentAsyncBackend exception with NoEventLoopError

### LINKED ISSUE #1048: Regression in Async Libraries discovery
### Things to check first

### AnyIO version

4.12.0

### Python version

3.10.17

### What happened?

Hello,

Before version 4.12.0, AnyIO used sniffio to detect available async libraries in the Python environment. Sniffio raised an `AsyncLibraryNotFoundError` when no async library was available, which was very convenient in some cases (for example, when running asynchronously code in production but synchronously in tests, with asynchronous libs in the classpath).

Starting with version 4.12, sniffio was removed and async library detection is now handled by custom code ([PR #1021](https://github.com/agronholm/anyio/pull/1021)￼). However, this introduces two issues:
- The new `NoCurrentAsyncBackend` exception, which is raised when no event loop is present, lives in the protected _core module and therefore should not be accessed from outside the library.
- The newly added `get_available_backends` function only checks which backends are available using `import`; it does not verify whether an event loop is actually running.

I would recommend either exposing this exception so it is accessible to end users, or updating the backend availability check to reflect real availability (i.e., whether it is being called from within an active event loop).

```
anyio==4.12.0
asyncio==4.0.0
```

```
import anyio

if __name__ == '__main__':
    print(anyio.get_available_backends()) # ('asyncio',)
    anyio.to_thread.current_default_thread_limiter() # Trigger anyio._core._eventloop.NoCurrentAsyncBackend (protected member)
```

agronholm:
## Changes
This replaces the private `NoCurrentAsyncBackend` exception with the existing `NoEventLoopError`.

Fixes #1048.

## Checklist

features

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/src/anyio/_core/_eventloop.py
+++ b/src/anyio/_core/_eventloop.py
@@ -120,6 +120,8 @@ def current_time() -> float:
     Return the current value of the event loop's internal clock.
 
     :return: the clock value (seconds)
+    :raises NoEventLoopError: if no supported asynchronous event loop is running in the
+        current thread
 
     """
     return get_async_backend().current_time()
@@ -152,7 +154,13 @@ def get_available_backends() -> tuple[str, ...]:
 
 
 def get_cancelled_exc_class() -> type[BaseException]:
-    """Return the current async library's cancellation exception class."""
+    """
+    Return the current async library's cancellation exception class.
+
+    :raises NoEventLoopError: if no supported asynchronous event loop is running in the
+        current thread
+
+    """
     return get_async_backend().cancelled_exception_class()
 
 
--- a/src/anyio/_core/_signals.py
+++ b/src/anyio/_core/_signals.py
@@ -14,10 +14,10 @@ def open_signal_receiver(
     Start receiving operating system signals.
 
     :param signals: signals to receive (e.g. ``signal.SIGINT``)
-    :raises NoEventLoopError: if no supported asynchronous event loop is running in the
-        current thread
     :return: an asynchronous context manager for an asynchronous iterator which yields
         signal numbers
+    :raises NoEventLoopError: if no supported asynchronous event loop is running in the
+        current thread
 
     .. warning:: Windows does not support signals natively so it is best to avoid
         relying on this in cross-platform applications.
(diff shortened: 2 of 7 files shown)


docs/versionhistory.rst
```diff
@@ -6,6 +6,9 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 **UNRELEASED**
 
 - Dropped support for Python 3.9
+- Changed all functions currently raising the private ``NoAsyncBackend`` exception
+  (since v4.12.0) to instead raise the public ``NoEventLoopError`` exception
+  (`#1048 <https://github.com/agronholm/anyio/issues/1048>`_)
```
gschaffner:
```suggestion
- Changed all functions currently raising the private ``NoCurrentAsyncBackend``
  exception (since v4.12.0) to instead raise the public ``NoEventLoopError`` exception
  (`#1048 <https://github.com/agronholm/anyio/issues/1048>`_)
```

m-u-xyz: > Would it make sense to just leave off :raises `NoEventLoopError:`?

It should certainly be "undocumented" for any async function or context manager, because while you could provoke the exception by manually iterating the awaitable, that doesn't happen in practice.

However, I'm not so sure about sync functions. For instance, if `anyio.Event` always used the back-end implementation you couldn't pre-set it without raising that exception. (Fortunately anyio is smarter than that.)

Thus the existence of this `raises:` can be a low-visual-clutter reminder that you need to be in async context to call that function/method, without the docs having to mention this explicitly everywhere it's relevant.

agronholm: OK, so the consensus is that it's better to leave it out of the docs altogether?

agronholm: Alright, I removed the `:raises NoEventLoopError` from the docstrings.

gschaffner (approved):
Looks good to me. Only comment: Re:

> However, I'm not so sure about sync functions. For instance, if `anyio.Event` always used the back-end implementation you couldn't pre-set it without raising that exception. (Fortunately anyio is smarter than that.)
> 
> Thus the existence of this `raises:` can be a low-visual-clutter reminder that you need to be in async context to call that function/method, without the docs having to mention this explicitly everywhere it's relevant.

> OK, so the consensus is that it's better to leave it out of the docs altogether?

Matthias's point regarding `:raises NoEventLoopError:` on sync functions[^1] makes sense to me. However it has some maintenance cost for little user benefit, so leaving it out entirely also seems fine to me.

[^1]: I.e. a bunch of constructors, `get_cancelled_exc_class`, `current_time`, etc.

smurfix:
> I added the :raises to all relevant sync functions I could find.

Thanks. LGTM.

PR #1056, opened 2025-12-27
Fixed `anyio.functools.lru_cache` not working with methods

### LINKED ISSUE #1042: lru_cache is not compatible with mypy when decorated method
### Things to check first

### AnyIO version

4.12.0

### Python version

3.13.8

### What happened?

Hi. When decorating a method on a subclass, mypy raises an error. async_lru is fine with it.

```python
import abc

from anyio.functools import lru_cache
from async_lru import alru_cache


class Iterface(abc.ABC):
    @abc.abstractmethod
    async def method(self, value: str) -> str: ...


class AnyIO(Iterface):
    @lru_cache()
    async def method(self, value: str) -> str:  # error
        return value


class AsyncLru(Iterface):
    @alru_cache()
    async def method(self, value: str) -> str:
        return value


async def main() -> None:
    anyio_value = await AnyIO().method("test")  # error
    async_lru_value = await AsyncLru().method("test")

    assert anyio_value == async_lru_value
```

```
test.py:14: error: Signature of "method" incompatible with supertype "Iterface"  [override]
test.py:14: note:      Superclass:
test.py:14: note:          def method(self, value: str) -> Coroutine[Any, Any, str]
test.py:14: note:      Subclass:
test.py:14: note:          AsyncLRUCacheWrapper[[AnyIO, str], Any]
test.py:25: error: Missing positional argument "value" in call to "__call__" of "AsyncLRUCacheWrapper"  [call-arg]
test.py:25: error: Argument 1 to "__call__" of "AsyncLRUCacheWrapper" has incompatible type "str"; expected "AnyIO"  [arg-type]
```

mypy 1.18.2 (compiled: yes)

agronholm:
## Changes

Fixes #1042. 

## Checklist

features

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/src/anyio/functools.py
+++ b/src/anyio/functools.py
@@ -38,9 +38,9 @@ from ._core._synchronization import Lock
 from .lowlevel import RunVar, checkpoint
 
 if sys.version_info >= (3, 11):
-    from typing import ParamSpec, Self
+    from typing import ParamSpec
 else:
-    from typing_extensions import ParamSpec, Self
+    from typing_extensions import ParamSpec
 
 T = TypeVar("T")
 S = TypeVar("S")
@@ -88,6 +88,9 @@ class _LRUMethodWrapper(Generic[T]):
         self.__wrapper.cache_clear()
 
     async def __call__(self, *args: Any, **kwargs: Any) -> T:
+        if self.__instance is None:
+            return await self.__wrapper(*args, **kwargs)
+
         return await self.__wrapper(self.__instance, *args, **kwargs)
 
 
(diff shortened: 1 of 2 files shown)


tapetersen:
@agronholm Holidays so was kind of busy but would gladly do. I keep an eye on here regularly and will add some questions at least for other MR's when I get the chance.

I don't mind but am somewhat surprised you went for the route of ditching preserving type-support for arguments using ParamSpecs.

It does make some sense to do the same as the existing hints for the sync version and it would indeed not be possible to annotate the classmehtod/staticmethod cases correctly for current checkers but I would probably have taken the tradeoff of those being rare special cases, with reasonably simple workarounds of simply using a module level function instead, in return for getting correct typing for useage of the function/method in the common case.

agronholm:
> > How else was I going to prevent false type checking errors in all cases (free functions, static methods, class methods, instance methods)? I believe we tried everything we could, and I could not find any examples out there which accomplishes all that.
> 
> Correct type-checking in the staticmethod/classmethod cases was indeed what I would've given up with, my possibly very wrong, assumption that these are uncommon niche-cases that can be very reasonably worked around with module-level functions instead.
> 
> If requested enough added separate decorators could also be added.

I spent considerable time trying everything I could to make all the cases work with strict type checking. The trouble is that type checkers can't tell the difference between class methods and static methods, as they simply disregard the `@classmethod` decorator. But if you have some idea on how to at least achieve partial strict argument type checks, I'm all ears.

PR #1059, opened 2026-01-20
Stop requiring exceptiongroup on Python ≥ 3.11

McSinyx:
## Changes

Make exceptiongroup not a test requirement on Python ≥ 3.11.  [In Guix we want to remove exceptiongroup originally because it depends on the inactive flit-scm project](https://codeberg.org/guix/guix/issues/2360), but fewer dependencies is always better

## Checklist

Please let me know if any of the following is applicable, as the patch isn't user-facing.

- [ ] I've added tests (in `tests/`) which would fail without my patch
- [ ] I've updated the documentation (in `docs/`), in case of behavior changes or new
features
- [ ] I've added a new changelog entry (in `docs/versionhistory.rst`).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -48,7 +48,7 @@ pytest11 = {anyio = "anyio.pytest_plugin"}
 test = [
     "blockbuster >= 1.5.23",
     "coverage[toml] >= 7",
-    "exceptiongroup >= 1.2.0",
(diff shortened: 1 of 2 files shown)


agronholm:
Given that the tests fail on 3.10 in this PR, I think you can see why it's still there.

PR #1063, opened 2026-02-06
don't use exceptiongroup.catch() in test_reraise_cancelled_in_excgroup

dotlambda:
## Changes
There's no need to rely on exceptiongroup for Python >= 3.11.


dotlambda:
> You're still using `exceptiongroup.catch()` there, so what's the point of this PR?

This PR allows running the tests on Python >= 3.11 without exceptiongroup.

agronholm:
Never mind, I misread. All good.

PR #1064, opened 2026-02-08
fix: preserve exception cause in asyncio SocketStream

### LINKED ISSUE #1055: Asyncio SocketStream suppresses exception cause
### Things to check first

### AnyIO version

4.11.0

### Python version

3.11.2

### What happened?

https://github.com/agronholm/anyio/commit/9a464bf68a2e7cc9aed1db8ee50c6aad47fa4fe6 updated the Asyncio SocketStream receive exception handling to `raise self._protocol.exception from None`.  Raising from None suppresses the cause of the exception, so users of the asyncio SocketStream are not able to differentiate the underlying cause of the BrokenResourceError, e.g. library users may want to implement specific handling for different types of ConnectionErrors:

- `ConnectionAbortedError`
- `ConnectionRefusedError`
- `ConnectionResetError`

Using the trio backend, the SocketStream [maintains the exception cause](https://github.com/agronholm/anyio/blob/master/src/anyio/_backends/_trio.py#L374) (which is preferable IMO) so the two backend implementations are not at parity in this aspect.

The following example should raise a `ConnectionResetError` on Linux, though the exception cause is lost as the `exception.__cause__` is set to None.  Using the trio backend, the  `ConnectionResetError`  can still be found in the cause of the `BrokenResourcError`

```python
import socket
import struct

import anyio
from anyio.abc import SocketStream


async def handler(socket: SocketStream) -> None:
    try:
        async for _ in socket:
            ...
    except anyio.BrokenResourceError as exc:
        print(type(exc.__cause__))
        raise

async def main() -> None:
    server = await anyio.create_tcp_listener(local_port=12345)
        tg.start_soon(server.serve, handler)

        s = socket.socket()
        s.connect(("localhost", 12345))
        # Create a ConnectionResetError
        s.setsockopt(socket.SOL_SOCKET, socket.SO_LINGER, struct.pack('ii', 1, 0))
        s.close()

```

Asyncio traceback snippet:
```
<class 'NoneType'>
...
        |   File "/home/user/SW/anyio/socket_test.py", line 12, in handler
        |     async for _ in socket:
        |   File "/home/user/SW/anyio/.venv/lib/python3.11/site-packages/anyio/abc/_streams.py", line 131, in __anext__
        |     return await self.receive()
        |            ^^^^^^^^^^^^^^^^^^^^
        |   File "/home/user/SW/anyio/.venv/lib/python3.11/site-packages/anyio/_backends/_asyncio.py", line 1275, in receive
        |     raise self._protocol.exception from None
        | anyio.BrokenResourceError
```

Trio traceback snippet:

```
<class 'ConnectionResetError'>
...
      | ExceptionGroup: Exceptions from Trio nursery (1 sub-exception)
        |   File "/home/user/SW/anyio/.venv/lib/python3.11/site-packages/anyio/_backends/_trio.py", line 388, in receive
        |     data = await self._trio_socket.recv(max_bytes)
        |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        |   File "/home/user/SW/anyio/.venv/lib/python3.11/site-packages/trio/_socket.py", line 440, in wrapper
        |     return await self._nonblocking_helper(wait_fn, fn, *args, **kwargs)
        |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        |   File "/home/user/SW/anyio/.venv/lib/python3.11/site-packages/trio/_socket.py", line 1014, in _nonblocking_helper
        |     return fn(self._sock, *args, **kwargs)
        |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        | ConnectionResetError: [Errno 104] Connection reset by peer
        | 
        | The above exception was the direct cause of the following exception:
        | 
        |   File "/home/user/SW/anyio/socket_test.py", line 10, in handler
        |     async for _ in socket:
        |   File "/home/user/SW/anyio/.venv/lib/python3.11/site-packages/anyio/abc/_streams.py", line 131, in __anext__
        |     return await self.receive()
        |            ^^^^^^^^^^^^^^^^^^^^
        |   File "/home/user/SW/anyio/.venv/lib/python3.11/site-packages/anyio/_backends/_trio.py", line 390, in receive
        |     self._convert_socket_error(exc)
        |   File "/home/user/SW/anyio/.venv/lib/python3.11/site-packages/anyio/_backends/_trio.py", line 374, in _convert_socket_error
        |     raise BrokenResourceError from exc
        | anyio.BrokenResourceError
```

veeceey:
## Summary
- Remove `from None` in the asyncio `SocketStream.receive` method so the underlying exception cause (e.g., `ConnectionResetError`) set by `connection_lost` is preserved in `BrokenResourceError.__cause__`
- This makes the asyncio backend consistent with the Trio backend, which already preserves the exception cause via `raise BrokenResourceError from exc`
- The `send` method in the same class already correctly preserves the exception cause; only `receive` had this issue
- Refactored `connection_lost` to store the raw exception instead of wrapping it in `BrokenResourceError`, and moved the `BrokenResourceError` construction to the `receive` and `send` methods using `raise ... from`

Fixes #1055

## Checklist
- [x] Changelog entry added in `docs/versionhistory.rst`
- [x] Test coverage added for both `receive` and `send` exception cause preservation

## Test plan
- Added `test_receive_exception_cause_preserved` - verifies that `BrokenResourceError.__cause__` is not `None` when receiving from a stream whose connection was reset (RST)
- Added `test_send_exception_cause_preserved` - verifies that `BrokenResourceError.__cause__` is not `None` when sending to a stream whose connection was reset (RST)
- Both tests use `SO_LINGER` with a zero timeout to force a TCP RST, ensuring `connection_lost` is called with an exception
- Tests are parametrized across asyncio backends only (asyncio, asyncio+uvloop, asyncio+eager) since this is an asyncio-specific fix
- All existing `TestTCPStream` tests continue to pass

### DIFF
--- a/tests/test_sockets.py
+++ b/tests/test_sockets.py
@@ -555,7 +555,7 @@ class TestTCPStream:
             await stream.send(b"foo")
 
     async def test_receive_exception_cause_preserved(
-        self, family: AnyIPAddressFamily
+        self, family: AnyIPAddressFamily, request: FixtureRequest
     ) -> None:
         """
         Test that when connection_lost is called with an exception, the resulting
@@ -563,35 +563,24 @@ class TestTCPStream:
 
         Regression test for https://github.com/agronholm/anyio/issues/1055.
         """
+        server_sock = socket.create_server(("localhost", 0), family=family)
+        request.addfinalizer(server_sock.close)
+        server_sock.settimeout(1)
+        server_addr = server_sock.getsockname()[:2]
 
-        def serve() -> None:
-            sock, addr = server_sock.accept()
-            event.wait(3)
-            sock.setsockopt(
+        async with await connect_tcp(*server_addr) as stream:
+            client_sock, _ = server_sock.accept()
+            client_sock.setsockopt(
                 socket.SOL_SOCKET, socket.SO_LINGER, struct.pack("ii", 1, 0)
             )
-            sock.close()
+            client_sock.close()
+            with pytest.raises(BrokenResourceError) as exc_info:
+                await stream.receive()
 
-        with socket.socket(family, socket.SOCK_STREAM) as server_sock:
-            server_sock.settimeout(1)
-            server_sock.bind(("localhost", 0))
-            server_sock.listen()
-            server_addr = server_sock.getsockname()[:2]
-            event = threading.Event()
-            thread = Thread(target=serve)
-            thread.start()
-            async with await connect_tcp(*server_addr) as stream:
-                await stream.send(b"GET")
-                event.set()
-                with pytest.raises(BrokenResourceError) as exc_info:
-                    await stream.receive()
-
-                assert exc_info.value.__cause__ is not None
-
-            thread.join()
+            assert exc_info.value.__cause__ is not None
 
     async def test_send_exception_cause_preserved(
-        self, family: AnyIPAddressFamily
+        self, family: AnyIPAddressFamily, request: FixtureRequest
     ) -> None:
         """
         Test that when connection_lost is called with an exception, the resulting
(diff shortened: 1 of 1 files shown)


veeceey:
Good point — you're right that explicit chaining is the better approach. I've updated the PR accordingly.

**What changed:**

1. `connection_lost` now stores the raw exception (`self.exception = exc`) instead of manually constructing a `BrokenResourceError` and setting `__cause__`.
2. Both `receive` and `send` now use explicit chaining: `raise BrokenResourceError from self._protocol.exception`.

This is consistent with how the Trio backend handles it:
```python
except trio.BrokenResourceError as exc:
    raise BrokenResourceError from exc.__cause__
```

And it follows the same `raise X from Y` pattern used throughout the anyio codebase (e.g., `raise BrokenResourceError from exc` in the asyncio backend's other exception handling paths).

The previous approach of just removing `from None` (implicit chaining) would have preserved the cause, but explicit chaining is cleaner and makes the intent unambiguous. I also applied the same fix to `send` for consistency, since it had the same pattern of bare-raising the pre-constructed exception.

tests/test_sockets.py
```diff
@@ -553,6 +554,81 @@ async def test_send_after_peer_closed(
         with pytest.raises(ClosedResourceError):
             await stream.send(b"foo")
 
+    @pytest.mark.parametrize("anyio_backend", asyncio_params)
```
graingert:
```suggestion
```

tests/test_sockets.py
```diff
@@ -553,6 +554,81 @@ async def test_send_after_peer_closed(
...
+                await stream.send(b"GET")
+                event.set()
+                with pytest.raises(BrokenResourceError) as exc_info:
+                    await stream.receive()
+
+                assert exc_info.value.__cause__ is not None
+
+            thread.join()
+
+    @pytest.mark.parametrize("anyio_backend", asyncio_params)
```
graingert:
```suggestion
```

tests/test_sockets.py
```diff
@@ -553,6 +554,79 @@ async def test_send_after_peer_closed(
...
+            thread.start()
+            async with await connect_tcp(*server_addr) as stream:
+                await stream.send(b"GET")
+                event.set()
+                with pytest.raises(BrokenResourceError) as exc_info:
+                    await stream.receive()
+
+                assert exc_info.value.__cause__ is not None
+
+            thread.join()
```
agronholm: Does this test actually need threads? I'm not 100% sure but it seems to me like it doesn't.

tests/test_sockets.py
```diff
@@ -553,6 +554,79 @@ async def test_send_after_peer_closed(
...
+            async with await connect_tcp(*server_addr) as stream:
+                await stream.send(b"GET")
+                event.set()
+                with pytest.raises(BrokenResourceError) as exc_info:
+                    for _ in range(1000):
+                        await stream.send(b"foo")
+
+                assert exc_info.value.__cause__ is not None
+
+            thread.join()
```
agronholm: As above.

agronholm (commented):
One question about the tests.

PR #1068, opened 2026-02-12
Fix parameter name "func" in parametrize breaking pytest

JohnnyDeuss:
## Changes

When using the pytest plugin combined with `pytest.mark.parametrize`, using the parameter name "func" breaks due to the signature of `_call_in_runner_task` mixing the positional-or-keyword parameter `func` with `**kwargs` which will have `func` in it if used as a parameter name in `pytest.mark.parametrize`. Adding the `/` in the function signature makes the `func` parameter positional-only and fixes the resulting `TypeError`.

Example of problematic test:

```python
@pytest.mark.parametrize("func", [1])
async def test_func(func: int) -> None:
    pass
```

## Checklist

features

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/tests/test_pytest_plugin.py
+++ b/tests/test_pytest_plugin.py
@@ -678,10 +678,7 @@ class TestFreePortFactory:
                 sock.bind((addr, free_udp_port))
 
 
-@pytest.mark.parametrize("anyio_backend", get_all_backends(), indirect=True)
-def test_func_as_parametrize_param_name(
-    anyio_backend_name: str, testdir: Pytester
-) -> None:
+def test_func_as_parametrize_param_name(testdir: Pytester) -> None:
     """
     Test that "func" can be used as a parameter name in
     `pytest.mark.parametrize` when using the pytest plugin.


tests/test_pytest_plugin.py
```diff
@@ -676,3 +676,38 @@ async def test_free_udp_port(
...
+        from anyio import sleep
+        from anyio._core._eventloop import current_async_library
+
+        @pytest.fixture
+        async def async_fixture():
+            await sleep(0)
+            return current_async_library()
+        """
+    )
+
```
agronholm: Doesn't look like this is actually used, so it could be dropped, yes?
```suggestion
```

tests/test_pytest_plugin.py
```diff
@@ -676,3 +676,38 @@ async def test_free_udp_port(
...
+    testdir.makepyfile('''
+        import pytest
+
+        @pytest.mark.parametrize("func", [1])
+        @pytest.mark.anyio
+        async def test_func_as_parametrize_param_name(func: int) -> None:
+            """
+            Test that "func" can be used as a parameter name in
+            `pytest.mark.parametrize` when using the pytest plugin.
+            """
```
agronholm: If you want to document the rationale for the test, put it in the docstring of the enclosing test (`test_func_as_parametrize_param_name`).

tests/test_pytest_plugin.py
```diff
@@ -676,3 +676,38 @@ async def test_free_udp_port(
         for family, addr in families:
             with socket.socket(family, socket.SOCK_DGRAM) as sock:
                 sock.bind((addr, free_udp_port))
+
+
+@pytest.mark.parametrize("anyio_backend", get_all_backends(), indirect=True)
+def test_func_as_parametrize_param_name(
+    anyio_backend_name: str, testdir: Pytester
+) -> None:
```
agronholm: Why did you add the `anyio_backend` parameter here? No other test in this module has it.
```suggestion
def test_func_as_parametrize_param_name(testdir: Pytester) -> None:
```

JohnnyDeuss: Sorry, I wasn't too clear on how to test pytest inside pytest and ensuring the plugin is enabled and a certain backend is used. Since I altered both the trio and asyncio backends, I wanted the test to run against both backend and assumed this is how to achieve that.

agronholm: Tests are run on all back-ends by default. Some tests, like `test_asyncio`, choose a specific back-end in their `conftest` modules. The `result.assert_outcomes(passed=len(get_available_backends()))` looks like that because it assumes N tests passing where N is the number of available backends.

PR #1071, opened 2026-02-13
Fix some `TaskGroup` exit issues on asyncio

### LINKED ISSUE #1069: `TaskGroup.__aexit__` visits a cancel point on asyncio
### Things to check first

### AnyIO version

main (a2a07c7c7af5d0f154b770f280d8772299a20a38)

### Python version

CPython 3.10.19, 3.11.14, 3.12.12, 3.13.11, 3.14.2

### What happened?

In a cancelled scope, or with a pending native cancellation, `TaskGroup` on asyncio injects a cancellation exception even when no child exited with a cancelled exception. This violates Trio's cancellation safety semantics: see https://github.com/python-trio/trio/blob/b0732873e851b6c6f75d5b819ea4f8a063029f52/newsfragments/1457.feature.rst and https://github.com/python-trio/trio/commit/451393acec35c3a01cba0c19de6bf8f6f23dbc7b.

f637fb6c93cf625b9615c260767e5e21b231e381

### LINKED ISSUE #1070: `TaskGroup` doesn't raise cancelled when it should on asyncio
### Things to check first

### AnyIO version

main (a2a07c7c7af5d0f154b770f280d8772299a20a38)

### Python version

CPython 3.10.19, 3.11.14, 3.12.12, 3.13.11, 3.14.2

### What happened?

On asyncio, when

1.  a task group is in a level-cancelled context

1.  a spawned child of the task group exits raising a level-`CancelledError` due to the aforementioned level cancellation from above

1.  after that child has exited, the body of the `async with tg:` block finishes running without raising any exception

then the task group incorrectly exits without raising a cancelled.

(FWIW, this is not a bug that I encountered IRL; I noticed it while reading the implementation while looking into #1069.)

The `start_soon_and_wait-False` and `start_soon_and_wait_and_shielded_checkpoint-False` parameterizations of 1e3122a6c0ee9e0e55054c535c6158734f823108.

gschaffner:
## Changes

Fixes #1069.

Fixes #1070.

## Checklist

features

### DIFF
--- a/tests/test_taskgroups.py
+++ b/tests/test_taskgroups.py
@@ -1651,26 +1651,27 @@ async def test_outer_cancellation_propagated_by_task_group_aexit(
                     # of the `if not self._tasks` branch.
                     tg.start_soon(taskfunc2)
 
-                if child == "host_task":
-                    await taskfunc1()
-                elif child == "start_soon":
-                    tg.start_soon(taskfunc1)
-                elif child == "start_soon_and_wait":
-                    tg.start_soon(taskfunc1)
-                    with CancelScope(shield=True):
-                        await taskfunc_exited.wait()
-                elif child == "start_soon_and_wait_and_checkpoint":
-                    tg.start_soon(taskfunc1)
-                    with CancelScope(shield=True):
-                        await taskfunc_exited.wait()
-                    await checkpoint()
-                elif child == "start_soon_and_wait_and_shielded_checkpoint":
-                    tg.start_soon(taskfunc1)
-                    with CancelScope(shield=True):
-                        await taskfunc_exited.wait()
+                match child:
+                    case "host_task":
+                        await taskfunc1()
+                    case "start_soon":
+                        tg.start_soon(taskfunc1)
+                    case "start_soon_and_wait":
+                        tg.start_soon(taskfunc1)
+                        with CancelScope(shield=True):
+                            await taskfunc_exited.wait()
+                    case "start_soon_and_wait_and_checkpoint":
+                        tg.start_soon(taskfunc1)
+                        with CancelScope(shield=True):
+                            await taskfunc_exited.wait()
                         await checkpoint()
-                else:
-                    assert_never(child)
+                    case "start_soon_and_wait_and_shielded_checkpoint":
+                        tg.start_soon(taskfunc1)
+                        with CancelScope(shield=True):
+                            await taskfunc_exited.wait()
+                            await checkpoint()
+                    case _:
+                        assert_never(child)
         except get_cancelled_exc_class() as exc:
             if isinstance(exc, get_cancelled_exc_class()):
                 raise
--- a/tests/test_taskgroups.py
+++ b/tests/test_taskgroups.py
@@ -1351,6 +1351,8 @@ async def test_single_cancellation_exc() -> None:
             raise
         except Exception as exc:
             pytest.fail(f"Raised the wrong type of exception: {exc.__class__.__name__}")
+        else:
+            pytest.fail("Did not raise a cancellation exception")
 
 
 async def test_start_soon_parent_id() -> None:
@@ -1663,6 +1665,8 @@ async def test_outer_cancellation_propagated_by_task_group_aexit(
             raise
         except Exception as exc:
             pytest.fail(f"Raised the wrong type of exception: {exc.__class__.__name__}")
+        else:
+            pytest.fail("Did not raise a cancellation exception")
 
     assert not tg.cancel_scope.cancelled_caught
     assert cs.cancelled_caught
--- a/tests/test_taskgroups.py
+++ b/tests/test_taskgroups.py
@@ -1549,6 +1549,62 @@ async def test_cancel_before_entering_task_group() -> None:
             pytest.fail("This should not raise a cancellation exception")
 
 
+async def test_no_new_cancellation_from_task_group_aexit() -> None:
+    """
+    Test that `TaskGroup.__aexit__` does not visit a cancel point. I.e. test that
+    `TaskGroup.__aexit__` does not raise `get_cancelled_exc_class()` if none of its
+    children raised `get_cancelled_exc_class()`. See also
+    https://github.com/python-trio/trio/pull/1696 and
+    https://github.com/python-trio/trio/pull/3011.
+    """
+    started = False
+
+    async def taskfunc() -> None:
+        nonlocal started
+        started = True
+        with CancelScope(shield=True):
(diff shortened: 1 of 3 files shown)


tests/test_taskgroups.py
```diff
@@ -1549,6 +1553,136 @@ async def test_cancel_before_entering_task_group() -> None:
...
+        "start_soon_and_wait_and_shielded_checkpoint",
+    ],
+)
+async def test_outer_cancellation_propagated_by_task_group_aexit(
+    child: Literal[
+        "host_task",
+        "start_soon",
+        "start_soon_and_wait",
+        "start_soon_and_wait_and_checkpoint",
+        "start_soon_and_wait_and_shielded_checkpoint",
```
gschaffner: A note: the `start_soon_and_wait-False` and `start_soon_and_wait_and_shielded_checkpoint-False` cases are very similar, but I'd like to add regression tests for both, because it is possible to have one pass and the other fail. My initial attempt at fixing #1070 managed to only fix the `start_soon_and_wait_and_shielded_checkpoint` case but failed for the `start_soon_and_wait` case, which required me to modify my approach.

agronholm:
Thanks. I'll get this reviewed during the weekend. This is one of those times when I get a sudden flurry of PRs and issues in several projects all of a sudden.

tests/test_taskgroups.py
```diff
@@ -1549,6 +1553,136 @@ async def test_cancel_before_entering_task_group() -> None:
...
+                    with CancelScope(shield=True):
+                        await taskfunc_exited.wait()
+                        await checkpoint()
+                else:
+                    assert_never(child)
+        except get_cancelled_exc_class() as exc:
+            if isinstance(exc, get_cancelled_exc_class()):
+                raise
+
+            pytest.fail(f"Raised the wrong type of exception: {exc}")
```
agronholm: How would we ever end up here?

tests/test_taskgroups.py
```diff
@@ -1549,6 +1553,136 @@ async def test_cancel_before_entering_task_group() -> None:
...
+                    with CancelScope(shield=True):
+                        await taskfunc_exited.wait()
+                    await checkpoint()
+                elif child == "start_soon_and_wait_and_shielded_checkpoint":
+                    tg.start_soon(taskfunc1)
+                    with CancelScope(shield=True):
+                        await taskfunc_exited.wait()
+                        await checkpoint()
+                else:
+                    assert_never(child)
```
agronholm: Use a `match` block here?

tests/test_taskgroups.py
```diff
@@ -1549,6 +1553,136 @@ async def test_cancel_before_entering_task_group() -> None:
...
+    ],
+)
+async def test_outer_cancellation_propagated_by_task_group_aexit(
+    child: Literal[
+        "host_task",
+        "start_soon",
+        "start_soon_and_wait",
+        "start_soon_and_wait_and_checkpoint",
+        "start_soon_and_wait_and_shielded_checkpoint",
+    ],
```
agronholm: Maybe the annotation is unnecessarily heavy-handed here; `str` should be sufficient, yes?

agronholm (commented):
I added a few comments. I haven't come up with a full verdict yet, but it's looking reasonable.

tests/test_taskgroups.py
gschaffner: Due to some unexpected bug, I suppose. I just copied this `pytest.fail` logic from [another test](https://github.com/agronholm/anyio/blob/1c403b1a7bf278455dbf0abd4d475e8de3193683/tests/test_taskgroups.py#L1336) because I needed a version of `pytest.raises(get_cancelled_exc_class())` that didn't swallow the `get_cancelled_exc_class()`.

tests/test_taskgroups.py
gschaffner: Done. 12300a74adf07097a1f729a3ce56fb1326b71892

tests/test_taskgroups.py
gschaffner: Done. 5f653ce532035a0f6495663376aa18137301e425

FWIW I had used `Literal` with `assert_never` here[^1] since it gives enum-like typechecking.

[^1]: `Literal` with [`exhaustive-match`](https://mypy.readthedocs.io/en/stable/error_code_list2.html#check-that-match-statements-match-exhaustively-exhaustive-match) would also do the same thing

tests/test_taskgroups.py
agronholm: If this branch catches a cancellation exception, it gets always re-raised without ever hitting the `pytest.fail`, so I'm not sure what you intention was here.

gschaffner: Oh, oops. It was meant to behave like `pytest.raises`. Fixed now, hopefully: 95ceed03168edc55bf0df9106818f774ab2afbd3

tests/test_taskgroups.py
```diff
@@ -1549,6 +1550,130 @@ async def test_cancel_before_entering_task_group() -> None:
...
+                        with CancelScope(shield=True):
+                            await taskfunc_exited.wait()
+                            await checkpoint()
+                    case _:
+                        pytest.fail("Execution should not reach this point")
+        except BaseException as exc:
+            if isinstance(exc, get_cancelled_exc_class()):
+                raise
+
+            pytest.fail(f"Raised the wrong type of exception: {exc}")
```
agronholm: I think this is a bit more elegant. We don't want to swallow any `KeyboardInterrupt` or `SystemExit` exceptions which should terminate the test suite, right?
```suggestion
        except get_cancelled_exc_class()):
            raise
        except Exception as exc:
            pytest.fail(f"Raised the wrong type of exception: {exc.__class__.__name__}")
```

gschaffner: Sounds good. Applied: 7a0aa1d814ad3d22ce6447551622aff5d8b83c55 & d4a3233241cfec46a6fb4c7cd56a5211ac2ce0bf

This snippet was copied from https://github.com/agronholm/anyio/blob/1c403b1a7bf278455dbf0abd4d475e8de3193683/tests/test_taskgroups.py#L1336 so I've changed it there too.

gschaffner:
> I added a small change to make sure the semantics align with `asyncio.TaskGroup` too.

Okay, looks good to me.

gschaffner:
It might be possible to implement this in a different, possibly cleaner/preferable way by relying on the `CancelScope` `ExceptionGroup` semantics that are buggy right now (#1091). I will mark this as a draft for now and I'll take another look at this after #1091 is resolved.

PR #1073, opened 2026-02-20
Add `ttl` parameter to functools.lru_cache

Graeme22:
## Changes

Many kinds of cache implementations have a TTL parameter, which allows
for expiring entries after a certain period of time. This is useful for
preventing unnecessary repitition of work.

This commit adds a `ttl` parameter to the `functools.lru_cache` wrapper
implemented in a very simple way with minimal overhead compared to the
existing code. Expiration time is simply added as an additional value to
the LRU cache tuple. Expiration happens lazily (that is, at the time of
the check)--if the key's expiry has passed at the time we try and fetch
an entry, we invalidate it instead and trigger a cache miss.

## Checklist

features

### Updating the changelog

```
```

### DIFF
--- a/src/anyio/functools.py
+++ b/src/anyio/functools.py
@@ -182,7 +182,6 @@ class AsyncLRUCacheWrapper(Generic[P, T]):
 
         if lock is None:
             if expires_at is not None and current_time() >= expires_at:
-                del cache_entry[key]
                 self._currsize -= 1
                 cached_value, lock, expires_at = (
                     initial_missing,


src/anyio/functools.py
```diff
@@ -48,7 +50,11 @@
 lru_cache_items: RunVar[
     WeakKeyDictionary[
         AsyncLRUCacheWrapper[Any, Any],
-        OrderedDict[Hashable, tuple[_InitialMissingType, Lock] | tuple[Any, None]],
+        OrderedDict[
+            Hashable,
+            tuple[_InitialMissingType, Lock, float | None]
+            | tuple[Any, None, float | None],
```
Graeme22, line 55: I was going back and forth between int and float for this, so note to self this isn't consistent with the rest of the code

src/anyio/functools.py
```diff
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+from anyio._core._eventloop import current_time
```
agronholm: Use relative imports and place all possible imports after `__all__`, as per PEP 8.

src/anyio/functools.py
agronholm: Do you intend to make this consistent, or is there a reason why this is a float here?

tests/test_functools.py
```diff
@@ -14,6 +15,7 @@
     get_cancelled_exc_class,
     wait_all_tasks_blocked,
 )
+from anyio._core._eventloop import sleep
```
agronholm: `sleep()` is an exported function, no need to import it from the private module.

src/anyio/functools.py
Graeme22: Yes of course. Though it is worth discussing which type is better suited for this, I suppose float doesn't cost anything so perhaps better to give a bit more granular control

Graeme22: Actually I marked the wrong line, this has to be float regardless because this is where we store expiration time

src/anyio/functools.py
```diff
@@ -158,23 +170,34 @@ async def __call__(self, *args: P.args, **kwargs: P.kwargs) -> T:
...
         try:
-            cached_value, lock = cache_entry[key]
+            cached_value, lock, expires_at = cache_entry[key]
         except KeyError:
             # We're the first task to call this function
-            cached_value, lock = (
+            cached_value, lock, expires_at = (
                 initial_missing,
                 Lock(fast_acquire=not self._always_checkpoint),
+                current_time() + self._ttl if self._ttl is not None else None,
```
agronholm: I'm a bit uncomfortable with the number of `current_time()` calls here. Could we just call it once for consistency, as long as there are no yield points in between?

agronholm: That is, the last call to `current_time()` should be kept since it's done after visiting a yield point, but the others could use a shared timestamp.

agronholm:
Please do not force push to PRs. It makes it hard to see what has changed. I wish I could disable this somehow.

src/anyio/functools.py
Graeme22: Actually it looks like I was setting an expiration for the sentinel value which doesn't make any sense. Just two calls present now!

Graeme22:
> > > Please do not force push to PRs. It makes it hard to see what has changed. I wish I could disable this somehow.
> > 
> > 
> > Sorry, a lot of people prefer this--will try to remember in the future
> 
> I can't fathom why they would prefer that. If they wanted a clean history on merge, they can just squash on merge like I do.

My guess is it's an old habit and it used to be more work to squash before merge on github but idk

tests/test_functools.py
```diff
@@ -329,6 +334,69 @@ async def instance_method(self, x: int) -> int:
...
+            nonlocal call_count
+            call_count += 1
+            await sleep(time)
+            return time
+
+        with move_on_after(1) as scope:
+            async with create_task_group() as tg:
+                for _ in range(100):
+                    tg.start_soon(sleeper, 0.1)
+        assert not scope.cancelled_caught
```
agronholm: Add a blank line after a control block ends.
```suggestion

        assert not scope.cancelled_caught
```

tests/test_functools.py
```diff
@@ -329,6 +334,69 @@ async def instance_method(self, x: int) -> int:
...
+    async def test_ttl_sequential(self, checkpoint: bool) -> None:
+        @lru_cache(ttl=1, always_checkpoint=checkpoint)
+        async def sleeper(time: float) -> float:
+            await sleep(time)
+            return time
+
+        with move_on_after(1) as scope:
+            for _ in range(100):
+                await sleeper(0.1)
+        assert not scope.cancelled_caught
```
agronholm, line 408: Add a blank line after a control block ends.
```suggestion

        assert not scope.cancelled_caught
```

tests/test_functools.py
```diff
@@ -329,6 +334,69 @@ async def instance_method(self, x: int) -> int:
 
         await self._do_cache_asserts(Foo().instance_method)
 
+    async def test_ttl_cache_hit(self) -> None:
+        @lru_cache(ttl=1)
+        async def func() -> float:
+            return random.random()
```
agronholm: While the chances are quite low, `random.random()` can theoretically produce two identical values in succession which would make the test fail. Perhaps it's just pedantry, but I'd feel better without any involvement of randomness.

agronholm (changes requested):
I'll need to re-read the first part of the implementation to understand what changed.

src/anyio/functools.py
```diff
@@ -158,23 +170,34 @@ async def __call__(self, *args: P.args, **kwargs: P.kwargs) -> T:
...
+            cache_entry[key] = cached_value, lock, expires_at
 
         if lock is None:
-            # The value was already cached
-            self._hits += 1
-            cache_entry.move_to_end(key)
-            if self._always_checkpoint:
-                await checkpoint()
+            if expires_at is not None and current_time() >= expires_at:
+                del cache_entry[key]
```
agronholm: Is there any point in deleting the key in the cache entry if we're replacing it right after anyway?

agronholm (changes requested):
The biggest remaining issue with this now is the lack of documentation for the `ttl` parameter. I didn't document the others because they directly matched the stdlib `lru_cache`.

Graeme22:
> The biggest remaining issue with this now is the lack of documentation for the `ttl` parameter. I didn't document the others because they directly matched the stdlib `lru_cache`.

I will add that tomorrow. Would just docstrings be sufficient or is there another place needed?

src/anyio/functools.py
Graeme22: Nope!

PR #1076, opened 2026-02-20
Fix broken worker interpreter being recycled to idle pool

bysiber:
## Summary

When `worker.call()` raises `BrokenWorkerInterpreter`, the subinterpreter is in a broken state and can't handle further calls. However, the `finally` block in `run_sync()` unconditionally adds the worker back to `idle_workers`. This means a later `run_sync()` call can pop this broken worker from the pool and immediately fail.

## The Bug

In `to_interpreter.py`, the `run_sync()` function uses a `finally` block to return workers to the idle pool:

```python
try:
    return await to_thread.run_sync(worker.call, func, args, limiter=limiter)
finally:
    # ... pruning logic ...
    worker.last_used = current_time()
    idle_workers.append(worker)  # <-- always recycles, even if broken
```

When `BrokenWorkerInterpreter` is raised (e.g., an internal `ExecutionFailed` on 3.14 or a non-None `exc_info` from `_interpreters.exec` on 3.13), the interpreter is broken — but the worker still gets appended to `idle_workers`.

Any subsequent call that pops this worker will try to use the broken interpreter and fail again.

## The Fix

Track whether `BrokenWorkerInterpreter` was raised. If so, destroy the worker instead of returning it to the idle pool. Normal exceptions raised by the user's function are re-raised by `call()` without going through `BrokenWorkerInterpreter`, so those workers are still healthy and get recycled as before.

```python
broken = False
try:
    return await to_thread.run_sync(...)
except BrokenWorkerInterpreter:
    broken = True
    raise
finally:
    if broken:
        await to_thread.run_sync(worker.destroy, limiter=limiter)
    else:
        # ... normal pruning and recycling ...
```

### DIFF
--- a/src/anyio/to_interpreter.py
+++ b/src/anyio/to_interpreter.py
@@ -208,6 +208,7 @@ async def run_sync(
         except IndexError:
             worker = _Worker()
 
+    broken = False
     try:
         return await to_thread.run_sync(
             worker.call,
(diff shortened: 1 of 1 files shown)


smurfix:
```
finally:
    if broken:
        await to_thread.run_sync(worker.destroy, limiter=limiter)
```
IMHO there should be a shielded cancel scope around that `await` … just as in every other `finally:` clause out there.

agronholm:
I have no clue how to break a worker interpreter without crashing the whole process so I'm not sure there is any other way besides mocking.

PR #1082, opened 2026-03-04
fix: Consider tests which are marked for anyio programmatically

### LINKED ISSUE #422: Cannot mark tests programmatically for anyio
I used to mark all my tests for `pytest-asyncio` using this hook in my `conftest.py` file (here adopted for anyio already):

```py
import inspect

import pytest


def pytest_collection_modifyitems(session, config, items):  # noqa
    for item in items:
        # pylint: disable=protected-access
        if (
            isinstance(item, pytest.Function)
            and inspect.iscoroutinefunction(item.function)
        ):
            print(f"\nMarking {item}...")
            item.add_marker(pytest.mark.anyio)
```

If I run this test file, only the second test gets executed which I consider a bug given the documentation which states that a test is picked up from anyio by marking it.
If it's not a bug, what am I missing?

```py
import pytest


@pytest.fixture(scope="session")
def anyio_backend():
    return 'asyncio'


async def test_not_executed():
    assert True


@pytest.mark.anyio
async def test_executed():
    assert True
```

chbndrhnns:
## Changes

Fixes #422 

Added pytest_collection_finish hook that runs after pytest_collection_modifyitems. For any pytest.Function items that are coroutines, have the anyio marker, but lack anyio_backend in their fixture names, it generates new parametrized Function items one per available backend using CallSpec2, replacing the original item in session.items.

## Checklist

features

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/src/anyio/pytest_plugin.py
+++ b/src/anyio/pytest_plugin.py
@@ -179,6 +179,7 @@ def pytest_collection_finish(session: pytest.Session) -> None:
                 cs_fields = {f.name for f in dataclasses.fields(CallSpec2)}
             except TypeError:
                 cs_fields = set()
+
             for param_index, backend in enumerate(get_available_backends()):
                 if "_arg2scope" in cs_fields:  # pytest >= 8
                     callspec = CallSpec2(
@@ -197,10 +198,12 @@ def pytest_collection_finish(session: pytest.Session) -> None:
                         idlist=[backend],
                         marks=[],
                     )
+
                 fi = item._fixtureinfo
                 new_names_closure = list(fi.names_closure)
                 if "anyio_backend" not in new_names_closure:
                     new_names_closure.append("anyio_backend")
+
                 new_fixtureinfo = FuncFixtureInfo(
                     argnames=fi.argnames,
                     initialnames=fi.initialnames,
@@ -217,6 +220,7 @@ def pytest_collection_finish(session: pytest.Session) -> None:
                     originalname=item.originalname,
                 )
                 new_items.append(new_item)
+
             session.items[i : i + 1] = new_items
 
 


src/anyio/pytest_plugin.py
```diff
@@ -163,6 +166,60 @@ def pytest_pycollect_makeitem(
...
+            and iscoroutinefunction(item.function)
+            and item.get_closest_marker("anyio") is not None
+            and "anyio_backend" not in item.fixturenames
+        ):
+            new_items = []
+            try:
+                cs_fields = {f.name for f in dataclasses.fields(CallSpec2)}
+            except TypeError:
+                cs_fields = set()
+            for param_index, backend in enumerate(get_available_backends()):
```
agronholm, line 183: Leave a blank line after a control block ends.
```suggestion
                cs_fields = set()

            for param_index, backend in enumerate(get_available_backends()):
```

src/anyio/pytest_plugin.py
```diff
@@ -163,6 +166,60 @@ def pytest_pycollect_makeitem(
...
+                    item.parent,
+                    name=f"{item.originalname}[{backend}]",
+                    callspec=callspec,
+                    callobj=item.obj,
+                    fixtureinfo=new_fixtureinfo,
+                    keywords=item.keywords,
+                    originalname=item.originalname,
+                )
+                new_items.append(new_item)
+            session.items[i : i + 1] = new_items
```
agronholm, line 224: As above.
```suggestion

            session.items[i : i + 1] = new_items
```

src/anyio/pytest_plugin.py
```diff
@@ -163,6 +166,60 @@ def pytest_pycollect_makeitem(
...
+                        indices={"anyio_backend": param_index},
+                        arg2scope={"anyio_backend": Scope.Module},
+                        idlist=[backend],
+                        marks=[],
+                    )
+                fi = item._fixtureinfo
+                new_names_closure = list(fi.names_closure)
+                if "anyio_backend" not in new_names_closure:
+                    new_names_closure.append("anyio_backend")
+                new_fixtureinfo = FuncFixtureInfo(
```
agronholm, line 207: More blank spaces.
```suggestion
                    )

                fi = item._fixtureinfo
                new_names_closure = list(fi.names_closure)
                if "anyio_backend" not in new_names_closure:
                    new_names_closure.append("anyio_backend")

                new_fixtureinfo = FuncFixtureInfo(
```

agronholm:
> I think I found a way. However, pytest 7 seems really old.

Not too long ago someone still using pytest 7 complained about something that was subtly broken.

PR #1085, opened 2026-03-12
Widen _BaseFileStream to accept IO[bytes]

### LINKED ISSUE #1078: Inconsistent type hints lead to bad type checks
### Things to check first

### AnyIO version

4.12.1

### Python version

3.14.2

### What happened?

```python
import typing as T
import anyio


async def hash_stream(algo: str, stream: anyio.abc.ByteReceiveStream, **opts) -> bytes:
    ...


async def _do_hash(file: anyio.AsyncFile[bytes], algo: str, meta: object):
    # Don't context-manage the stream, it's the caller's job to close everything
    T.reveal_type(file)
    T.reveal_type(file.wrapped)
    stream = anyio.streams.file.FileReadStream(file.wrapped)
    digest = await hash_stream(algo, stream)
    setattr(meta, f"hash_{algo}", digest)
```

mypy output: 
```
src/swiss_army_upload/junk_drawer/rsync.py:63: note: Revealed type is "anyio._core._fileio.AsyncFile[builtins.bytes]"
src/swiss_army_upload/junk_drawer/rsync.py:64: note: Revealed type is "typing.IO[builtins.bytes]"
src/swiss_army_upload/junk_drawer/rsync.py:65: error: Argument 1 to "FileReadStream" has incompatible type "IO[bytes]"; expected "BinaryIO"  [arg-type]
```

This happens because `BinaryIO` is a subclass of `IO[bytes]`, not an equivalent alias. (https://docs.python.org/3/library/typing.html#abcs-and-protocols-for-working-with-i-o) So while you can use a `BinaryIO` as an `IO[bytes]`, you cannot use an `IO[bytes]` as a `BinaryIO` (in the eyes of mypy).

Type check the above code

rexkirshner:
## Summary

Fixes #1078

`AsyncFile[bytes].wrapped` is typed as `IO[bytes]`, but `FileReadStream` (via `_BaseFileStream`) required `BinaryIO`. Since `BinaryIO` is a subclass of `IO[bytes]` (not an alias), mypy rejects:

```python
stream = FileReadStream(file.wrapped)  # error: expected "BinaryIO", got "IO[bytes]"
```

Widened `_BaseFileStream.__init__` to accept `IO[bytes]` instead of `BinaryIO`. This is safe because `_BaseFileStream` only uses methods available on `IO[bytes]` (`read`, `write`, `close`, `seek`, `tell`, `fileno`).

## Test plan

- Existing tests pass (they use `open("rb")` which returns `BinaryIO`, a subtype of `IO[bytes]`)
- The code from the issue now type-checks correctly

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -20,8 +20,8 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 - Fixed the pytest plugin not running tests that had the ``anyio`` marker added
   programmatically via ``pytest_collection_modifyitems``
   (`#422 <https://github.com/agronholm/anyio/issues/422>`_; PR by @chbndrhnns)
-- Broadened the type annotation of``FileReadStream`` and ``FileWriteStream`` to accept
-  ``IO[bytes]`` instead of the narrower ``BinaryIO``
+- Broadened the type annotation of ``FileReadStream`` and ``FileWriteStream`` to
+  accept ``IO[bytes]`` instead of the narrower ``BinaryIO``
   (`#1078 <https://github.com/agronholm/anyio/issues/1078>`_; PR by @rexkirshner)
 
 **4.12.1**


agronholm:
Seems reasonable, but needs a changelog entry which Claude never adds.

agronholm:
Hi, do you intend to bring over the changes from the PR I closed as duplicate?

PR #1092, opened 2026-03-18
Fix cancellation exceptions leaking from a `CancelScope` on asyncio when they are in a group with non-cancellation exceptions

### LINKED ISSUE #1091: On asyncio, `CancelScope` doesn't filter a level-cancellation exception that it should
### Things to check first

### AnyIO version

master (d68670b)

### Python version

CPython 3.13.12

### What happened?

On the asyncio backend, `CancelScope.__exit__` doesn't filter/swallow a level-cancellation exception that it should if the cancellation exception is in a group. The cancellation exception instead propagates and causes the `run()` to crash and raise the level-cancellation exception (as part of a group).

Minimal reproducers that fail on asyncio and pass on Trio:

```python
async def test_exception_group_filtering() -> None:
    """Test that CancelledErrors are filtered out of exception groups."""

    body_exc = RuntimeError()

    def check_body_exc(exc: RuntimeError, /) -> bool:
        return exc is body_exc

    with pytest.RaisesGroup(pytest.RaisesExc(RuntimeError, check=check_body_exc)):
        with CancelScope() as cs:
            cs.cancel()
            try:
                raise body_exc
            except BaseException as exc:
                exceptions = [exc]

                try:
                    await checkpoint()
                except BaseException as exc2:
                    exceptions.append(exc2)
                else:
                    pytest.fail("Did not raise a cancellation exception")

                try:
                    raise BaseExceptionGroup("", exceptions)
                finally:
                    # Prevent reference cycles.
                    del exceptions


async def test_nested_exception_group_filtering() -> None:
    """Test that CancelledErrors are filtered out of nested exception groups."""

    body_exc = RuntimeError()

    def check_body_exc(exc: RuntimeError, /) -> bool:
        return exc is body_exc

    with pytest.RaisesGroup(
        pytest.RaisesGroup(pytest.RaisesExc(RuntimeError, check=check_body_exc))
    ):
        with CancelScope() as cs:
            cs.cancel()
            try:
                raise body_exc
            except BaseException as exc:
                exceptions = [exc]

                try:
                    await checkpoint()
                except BaseException as exc2:
                    exceptions.append(exc2)
                else:
                    pytest.fail("Did not raise a cancellation exception")

                try:
                    raise BaseExceptionGroup("", (BaseExceptionGroup("", exceptions),))
                finally:
                    # Prevent reference cycles.
                    del exceptions
```

These ^ were minimized from the following less-minimized reproducer:

```python
from anyio import CancelScope
from anyio.abc import AsyncResource
import anyio
import anyio.lowlevel
import logging


class SafeAsyncResource(AsyncResource):
    # This is an AsyncResource that always wraps its exceptions in a group. This
    # prevents exceptions from the body of the `async with` from getting mistakenly
    # swallowed & discarded. See also https://github.com/python-trio/trio/issues/455 /
    # https://github.com/python-trio/trio/issues/1559.

    async def __aexit__(
        self,
        exc_type: type[BaseException] | None,
        exc_val: BaseException | None,
        exc_tb: TracebackType | None,
    ) -> None:
        exceptions = list[BaseException]()
        try:
            if exc_val is not None:
                exceptions.append(exc_val)

            # aclose() can raise an arbitrary exception or a cancellation exception,
            # both of which may get caught later. Therefore we must not allow such an
            # exception to eat `exc_val`; we should never swallow `exc_val`. So we raise
            # both.
            try:
                await self.aclose()
            except BaseException as exc:
                exceptions.append(exc)
        finally:
            if exceptions:
                try:
                    raise BaseExceptionGroup(
                        "Exceptions from AsyncResource __aexit__", exceptions
                    )
                finally:
                    del exceptions


class SomeSafeAsyncResource(SafeAsyncResource):


async def main() -> None:
    try:
        with CancelScope() as cs:
            async with SomeSafeAsyncResource():
                cs.cancel()
                raise RuntimeError(
                    "exception raised by user code running under `async with SomeSafeAsyncResource`"
                )
    except* RuntimeError:
        # Swallow it.
        # logging.getLogger(__name__).exception("appropriate message:")
        pass



if __name__ == "__main__":
    # anyio.run(main, backend="trio")
    print("run() exited cleanly")
```

gschaffner:
## Changes

Fixes #1091.

## Checklist

features

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -29,7 +29,7 @@ repos:
           - tomli
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.15.6
+    rev: v0.15.7
     hooks:
       - id: ruff-check
         args: [--fix, --show-fixes]


docs/versionhistory.rst
```diff
@@ -20,6 +20,9 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 - Fixed the pytest plugin not running tests that had the ``anyio`` marker added
   programmatically via ``pytest_collection_modifyitems``
   (`#422 <https://github.com/agronholm/anyio/issues/422>`_; PR by @chbndrhnns)
+- Fixed cancellation exceptions leaking from a ``CancelScope`` on asyncio when they are
+  contained in an exception group alongside non-cancellation exceptions (`#1091
+  <https://github.com/agronholm/anyio/issues/1091>`_; PR by @gschaffner)
```
gschaffner, line 29: I borrowed the phrasing of this changelog entry from an old changelog entry from v1.2.1:

https://github.com/agronholm/anyio/blob/d68670b3b4e0917d4caff2de082e03220f3e05a1/docs/versionhistory.rst?plain=1#L1152-L1153

src/anyio/_backends/_asyncio.py
```diff
@@ -493,17 +492,31 @@ def __exit__(
...
-                            self._cancelled_caught = True
-                        else:
-                            cannot_swallow_exc_val = True
+                if isinstance(exc_val, BaseExceptionGroup):
+                    cancelleds_caught, remaining = exc_val.split(
+                        lambda exc: (
+                            isinstance(exc, CancelledError)
+                            and is_anyio_cancellation(exc)
+                        )
+                    )
```
gschaffner, line 501: I am not sure if calling `BaseExceptionGroup.split` unconditionally here is a performance concern, compared to the previous `iterate_exceptions`.

If it is a performance concern, we could switch to an approach where we call `iterate_exceptions` unconditionally and then call `split` if and only if `cancelled_caught and cannot_swallow_exc_val`. This approach would do duplicate work and would be slower in the case where `cancelled_caught and cannot_swallow_exc_val`, but it would _possibly_(?) be faster in some other cases.

src/anyio/_backends/_asyncio.py
```diff
@@ -493,17 +492,31 @@ def __exit__(
...
+                    if cancelleds_caught is None:
+                        return False
+
+                    self._cancelled_caught = True
 
-                return self._cancelled_caught and not cannot_swallow_exc_val
+                    if remaining is None:
+                        return True
+
+                    raise remaining
```
gschaffner: Should this be `from None`?

gschaffner: It should not, but we do need to add something here to preserve `__context__`. Added a test and a fix for it.

src/anyio/_backends/_asyncio.py
agronholm: "Hot code" shouldn't involve so many exceptions that this would become a real concern.

PR #1093, opened 2026-03-18
Fixed child task not cancelling its task group scope when a parent scope is already cancelled

### LINKED ISSUE #787: Child tasks don't cancel their group's scope correctly when they raise an exception on asyncio
### Things to check first

### AnyIO version

master (d1aea98), as well as the latest #774 (f9a1e1a)

### Python version

3.12.6, CPython

### What happened?

There are actually two separate bugs here that (I believe) require two separate fixes, but I am putting them both in this issue since they are so closely related (and to avoid spamming the issue tracker).

1.  On asyncio, a child task can finish with an unhandled exception without cancelling its group's scope.

1.  On asyncio, there is a delay between a child task finishing with an unhandled exception and its group's scope getting cancelled. This does not match the behavior on Trio. Per Trio's documentation:

    > If any task inside the nursery finishes with an unhandled exception, then the nursery **immediately** cancels all the tasks inside the nursery. (emphasis added)

Here are two tests. They all fail on asyncio and pass on trio. The first test is just a weaker version of the second test, but I have separated the two here because I think it is easier to follow that way.

The weaker case:

```python
# Easy mode: `tg.cancel_scope` gets shielded many event loop cycles after the child task
# exits with an exception.
async def test_shield_after_task_failed_weak() -> None:
    async def taskfunc() -> None:
        raise Exception("child task failed")

    with pytest.raises(BaseExceptionGroup) as exc:
        with CancelScope() as outer_scope:
                outer_scope.cancel()
                tg.start_soon(taskfunc)
                    # Wait at least one more scheduling round to ensure that taskfunc's
                    # task_done on asyncio has finished.
                    await sleep(0.1)
                tg.cancel_scope.shield = True
                assert tg.cancel_scope.cancel_called

    assert len(exc.value.exceptions) == 1
    assert str(exc.value.exceptions[0]) == "child task failed"
```

This weaker test is a regression test only for bug (1); it does not test for bug (2).

To fix this case I believe we just need to change the `.cancel` call in `task_done` to be unconditional here:

https://github.com/agronholm/anyio/blob/d1aea98c4f552d04458e49f1484b9c018f2a3ca8/src/anyio/_backends/_asyncio.py#L733-L734

The stronger case:

```python
# Hard mode: tg.cancel_scope gets shielded after the child task exits with an exception
# but before its task_done (on asyncio) runs.
async def test_shield_after_task_failed() -> None:
    taskfunc_exited = anyio.Event()

    async def taskfunc() -> None:
        try:
            raise Exception("child task failed")
        finally:
            taskfunc_exited.set()
            # (Technically we are setting this event slightly before taskfunc exits, but
            # this won't wake any waiters until the event loop's next batch of task
            # steps/callbacks, which will be the first batch to run after taskfunc has
            # exited.)

    with pytest.raises(BaseExceptionGroup) as exc:
        with CancelScope() as outer_scope:
                outer_scope.cancel()
                tg.start_soon(taskfunc)
                    # Trio documents: "If any task inside the nursery finishes with an
                    # unhandled exception, then the nursery *immediately* cancels all
                    # the tasks inside the nursery" (emphasis added). So when the event
                    # loop drives taskfunc's coro one step forward and it exits with an
                    # exception, tg.cancel_scope must get cancelled *immediately*, i.e.
                    # it must get cancelled before any other tasks/callbacks get to run
                    # a step.
                    await taskfunc_exited.wait()
                tg.cancel_scope.shield = True
                assert tg.cancel_scope.cancel_called

    assert len(exc.value.exceptions) == 1
    assert str(exc.value.exceptions[0]) == "child task failed"
```

This stronger case is a regression test for both bug (1) and bug (2).

(It's a bit hard to read because the test has to be pretty careful to be able to deterministically hit the problematic scheduling order.)

To fix this case I believe we need to move the exception-checking logic out of `task_done` and into a wrapper function:

```python
# in _spawn:
@wraps(func)
async def wrapper():
    try:
        return await func(...)
    finally:
        # task_done's exception-checking logic goes here, i.e. (simplified pseudocode)
        if exception:
            self.cancel_scope.cancel()
```

This way the `.cancel` call will happen **immediately** when the task finishes with an exception, as Trio documents, rather than the `.cancel` being postponed one scheduling batch later.

(Note, this test can also be written equivalently (I believe) using `wait_all_tasks_blocked` in order to hit the problematic scheduling order rather than using an event as I did above. I find the event-based version preferable (easier to understand), but the following is (I think) an equivalent way to implement this test:

```python
async def test_shield_after_task_failed2() -> None:
    async def taskfunc() -> None:
        raise Exception("child task failed")

    with pytest.raises(BaseExceptionGroup) as exc:
        with CancelScope() as outer_scope:
                outer_scope.cancel()
                tg.start_soon(taskfunc)
                    # Trio documents: "If any task inside the nursery finishes with an
                    # unhandled exception, then the nursery *immediately* cancels all
                    # the tasks inside the nursery" (emphasis added). So when the event
                    # loop drives taskfunc's coro one step forward and it exits with an
                    # exception, tg.cancel_scope must get cancelled *immediately*, i.e.
                    # it must get cancelled before any other tasks/callbacks get to run
                    # a step.
                tg.cancel_scope.shield = True
                assert tg.cancel_scope.cancel_called

    assert len(exc.value.exceptions) == 1
    assert str(exc.value.exceptions[0]) == "child task failed"
```

)

danieldorado-cmyk:
## Changes
On the asyncio backend, the `task_done` callback in `TaskGroup._spawn` skipped
calling `cancel()` on the task group's cancel scope when `_effectively_cancelled`
returned `True` (i.e. any ancestor scope was already cancelled). This meant a child
task could finish with an unhandled exception without cancelling the group's scope,
leaving the host task unaware of the failure.
This is the minimal fix suggested by @gschaffner during the review of #774
(https://github.com/agronholm/anyio/pull/774#discussion_r1755340156), which was
deferred at the time.

Fixes #787.

## Checklist

### DIFF
--- a/tests/test_taskgroups.py
+++ b/tests/test_taskgroups.py
@@ -1308,6 +1308,35 @@ async def test_cancelscope_exit_in_wrong_task() -> None:
     )
 
 
+@pytest.mark.parametrize("anyio_backend", asyncio_params)
+async def test_child_task_cancels_scope_when_parent_scope_cancelled() -> None:
+    """
+    Regression test for #787 (weak case).
+
+    When a child task exits with an unhandled exception, the task group's
+    cancel scope must be cancelled even if an outer scope is already cancelled.
+    Previously the ``_effectively_cancelled`` guard in ``task_done`` prevented
+    this, leaving the host task unaware that a child had failed.
+    """
+
+    async def taskfunc() -> None:
+        raise Exception("child task failed")
+
+    with pytest.raises(BaseExceptionGroup) as exc:
+        with CancelScope() as outer_scope:
+            async with create_task_group() as tg:
+                outer_scope.cancel()
+                tg.start_soon(taskfunc)
+                with CancelScope(shield=True):
+                    await wait_all_tasks_blocked()
+                    await sleep(0.1)
+                tg.cancel_scope.shield = True
+                assert tg.cancel_scope.cancel_called
+
+    assert len(exc.value.exceptions) == 1
+    assert str(exc.value.exceptions[0]) == "child task failed"
+
+
 def test_unhandled_exception_group(caplog: pytest.LogCaptureFixture) -> None:
     def crash() -> NoReturn:
         raise KeyboardInterrupt
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -5,6 +5,9 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
 **UNRELEASED**
 
+- Fixed a child task failing without cancelling its task group's scope on asyncio
+  when an outer cancel scope was already cancelled
+  (`#787 <https://github.com/agronholm/anyio/issues/787>`_)
 - Dropped support for Python 3.9
 - Fixed ``anyio.Path`` not being compatible with Python 3.15 due to the removal of
   ``pathlib.Path.is_reserved()`` and the addition of ``pathlib.Path.__vfspath__()``
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -835,8 +835,7 @@ class TaskGroup(abc.TaskGroup):
                     if not isinstance(exc, CancelledError):
                         self._exceptions.append(exc)
 
-                    if not self.cancel_scope._effectively_cancelled:
-                        self.cancel_scope.cancel()
+                    self.cancel_scope.cancel()
                 else:
                     task_status_future.set_exception(exc)
             elif task_status_future is not None and not task_status_future.done():


tests/test_taskgroups.py
```diff
@@ -1308,6 +1308,35 @@ async def exit_scope(scope: CancelScope) -> None:
     )
 
 
+@pytest.mark.parametrize("anyio_backend", asyncio_params)
```
gschaffner, line 1311: We should run this on the Trio backend too:


```suggestion
```

tests/test_taskgroups.py
```diff
@@ -1308,6 +1308,35 @@ async def exit_scope(scope: CancelScope) -> None:
...
+        raise Exception("child task failed")
+
+    with pytest.raises(BaseExceptionGroup) as exc:
+        with CancelScope() as outer_scope:
+            async with create_task_group() as tg:
+                outer_scope.cancel()
+                tg.start_soon(taskfunc)
+                with CancelScope(shield=True):
+                    await wait_all_tasks_blocked()
+                    await sleep(0.1)
```
gschaffner, line 1332: This sleep is the difference between the weak and the strong case. The reason that the sleep is here is just to work around bug (2) in #787. It might be good to have a comment here about how this test intentionally waits an extra event loop cycle for the `task_done` callback to run on the asyncio backend.

gschaffner (commented):
> Fixes #787.

This PR currently only fixes the weaker case of #787 (the stronger case is still bugged) so this probably shouldn't close #787.

gschaffner:
Disclosure about the review above: it is not an independent review—I was the author of most of this patch, so I may have some blind spots here.

tests/test_taskgroups.py
```diff
@@ -1308,6 +1308,35 @@ async def exit_scope(scope: CancelScope) -> None:
...
+    async def taskfunc() -> None:
+        raise Exception("child task failed")
+
+    with pytest.raises(BaseExceptionGroup) as exc:
+        with CancelScope() as outer_scope:
+            async with create_task_group() as tg:
+                outer_scope.cancel()
+                tg.start_soon(taskfunc)
+                with CancelScope(shield=True):
+                    await wait_all_tasks_blocked()
```
gschaffner, line 1331: Should we use the version of this test that uses an event ("`taskfunc_exited`") instead of the version that uses `wait_all_tasks_blocked`? It would give me a bit more confidence that this test is testing what it's supposed to. The asyncio version of `wait_all_tasks_blocked` is a bit less precise than the Trio version, and the point of these awaits is to control scheduling order.

I suppose that this point matters more for the strong case than the weak case. For the weak case, as long as we wait _long enough_, the test is fine and will not pass when it should fail. For the strong case, the test needs to wait until a particular event loop cycle and not longer than that.

gschaffner: I guess I am foreseeing that we may remove the sleep from this test in the future in order to change it from being the weak test into the strong test. Per #787: the weak case only tests for bug (1) in #787, but the strong case tests for bug (2) also. If/when we fix bug (2), I don't see a reason to retain a separate test that still has the sleep (the weak form), because bug (1) will clearly be completely covered by the strong test already, making the weak test redundant (a waste of 0.1 s every run).

In other words: if we remove `task_done` to fix bug (2) (i.e. we change `task_done` from a callback to a `finally`), then the line

```python
                    # Wait at least one more scheduling round to ensure that taskfunc's
                    # done callback (task_done) on asyncio has finished. This is
                    # workaround for the delay that is currently present between a task
                    # failing and cancelling its task group on asyncio (#787, bug (2)).
                    await sleep(0.1)
```

in the weak test would not make sense anymore, because bug (2) and the callback `task_done` would no longer exist :)

PR #1098, opened 2026-03-25
Added the create_task() method to TaskGroup

agronholm:

## Changes

This adds a mostly asyncio-compatible `create_task()` method to the `TaskGroup` class.
Like its asyncio counterpart, this method accepts a coroutine object and returns an awaitable task handle.

## Checklist

features

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/docs/tasks.rst
+++ b/docs/tasks.rst
@@ -49,6 +49,7 @@ method which will return a :class:`~.TaskHandle`. This handle can be used to:
 
 #. Wait for the task to finish before exiting the task group
 #. Retrieve the task's return value or exception
+#. Cancel the task
 #. Check the task's status
 
 ::


src/anyio/_core/_tasks.py
```diff
@@ -1,12 +1,37 @@
...
+    from typing_extensions import TypeVarTuple
+
+if sys.version_info >= (3, 13):
+    from typing import TypeVar
+else:
+    from typing_extensions import TypeVar
+
+T = TypeVar("T")
+TRetval = TypeVar("TRetval")
+TStartval = TypeVar("TStartval", default=None)
```
gschaffner: `TStartval` should be covariant I think. Probably `TRetval` too

agronholm: How would you type annotate the `_set_return_value()` method then? Type variables in parameters cannot be covariant.

src/anyio/_core/_tasks.py
```diff
@@ -171,3 +189,94 @@ def create_task_group() -> TaskGroup:
...
+        assert self._exception is None
+        self._return_value = val
+        self._finished_event.set()
+
+    def __await__(self) -> Generator[Any, Any, T]:
+        if not self._finished_event.is_set():
+            try:
+                yield from self._finished_event.wait().__await__()
+            except BaseException:
+                raise
```
tapetersen: I assume I'm missing something about why this try and re-raise block is needed vs. just the naked yield from (so please just remind me).

src/anyio/_core/_tasks.py
```diff
@@ -171,3 +189,94 @@ def create_task_group() -> TaskGroup:
...
+
+        if self._exception is not None:
+            raise self._exception
+
```
tapetersen, line 370: Since these handles now are "leaked" to user code they can be awaited multiple times. Raising the same exception instance will modify the shared-instance's stack-trace (which is what I understand the reason why Outcome has dissallowed unwrapping more than once and now tries to change the api to copy the exception before raising).

I'm quite sure the futures in vanilla asyncio has the same issue so status quo may very well be acceptable, especially if this is intended to be a straight alternative, and we can assume users already need to be aware.

src/anyio/_core/_tasks.py
agronholm: You're right, I had code in the `except:` block which I later removed. Good catch, updated.

src/anyio/_core/_tasks.py
agronholm: This is precisely the code I had in that try...except block which I removed. I'm unsure which way to go here.

src/anyio/_core/_tasks.py
tapetersen: I think you can work around that by the classic `self` parametrization with a method scoped type variable i.e. `def _set_return_value(self: TaskHandle[_T], val: _T) -> None:` . If T, which is scoped to class-level is covariant.

tapetersen: Code sample in [pyright playground](https://pyright-play.net/?code=GYJw9gtgBALgngBwJYDsDmB9ApgDxllAZyTCKiQgTBBigBVEsA1AQxABooBxArEJAMYAoIXSgBeeo1YgAFAHI68zgLAA3NkhYoY4uiACuWAJRCMYyQwTM2C8-NN0AyjDYwNAGwlTrMhc9caT2UoABMsYBYDD10AOVITEQEPFkJCelSAawAJbVCPLFkeFD5BAG06TgC3TwBdYwAuIShmqDaMECwYAxAUDE8jBvpW9sJAmH6WD0H6FxqpkTa2gAEEcGsaOBGwiKhO7t7J6cLCLA9gYygAWgA%2Beialx9gQOAen9-2elChT84A6DpdL5HIzbR64ARYBC0ACCMBg-AARgZ8ABREDgEBvd6PEAsJCnKAAJQMOgoWHRmNkACIYAALAmwLJQOmpKAoMC0T69LChKBwLrUy6gSBQeIlRZLcLAKAYU4Tbl9AYnM7AIZ0LK5FD5LBlcy1TieIbmS63MUJbFPX7AAGKkFYbyeIRAA)

```python
from typing_extensions import TypeVar, Generic

T = TypeVar('T', covariant=True)
_T = TypeVar('_T')
TStartval = TypeVar('TStartval', default=None)

class TaskHandle(Generic[T, TStartval]):
 
    _return_value: T
    _start_value: TStartval

    @property
    def return_value(self) -> T:
        try:
            return self._return_value
        except AttributeError:
            raise RuntimeError("this task has not returned yet") from None

    def _set_return_value(self: TaskHandle[_T], val: _T) -> None:
        self._return_value = val

```

src/anyio/_core/_tasks.py
tapetersen: As the current behavior is consistent with asyncio I would say it's completely acceptable to leave it as is.

Copying is arguably more correct but I'm really not sure if it's generally safe and how expensive it is given that the vast majority of cases will definitely be the case of it being awaited exactly once.

Purity wise you could even argue that you shoulnd't throw the exception from the task to an outside waiter at all (what if it's a leaked `StopIteration` ...) and instead wrap it in a `TaskErrored` like the aborted/cancelled cases with the caller responsible for unwrapping it but I would definitely say that's too far.

I would probably lean towards the best choice being  keeping it as it is now unless the copying costs a lot less than I think even if I myself tend to prefer api's doing it the "correct but slower" way. I'm completely fine with either.

src/anyio/_core/_tasks.py
agronholm: Updated. Thanks for the tip!

src/anyio/_core/_tasks.py
agronholm: Exception copying is not generally safe. What's on the table is disallowing users from waiting on the same task handle twice, but I'm not sure if such a thing is warranted in the first place.

gschaffner: > What's on the table is disallowing users from waiting on the same task handle twice, but I'm not sure if such a thing is warranted in the first place.

Would it be reasonable to disallow this for now and reconsider supporting it later if a compelling use-case comes up? If we support it now and want to disallow it later, it would be backwards incompatible.

src/anyio/_core/_tasks.py
```diff
@@ -171,3 +190,91 @@ def create_task_group() -> TaskGroup:
...
+        assert self._exception is None
+        self._return_value = val
+        self._finished_event.set()
+
+    def __await__(self) -> Generator[Any, Any, T_co]:
+        if not self._finished_event.is_set():
+            yield from self._finished_event.wait().__await__()
+
+        if self._exception is not None:
+            raise self._exception
```
gschaffner: This mutates `self._exception.__context__`. Should we preserve the original `__context__` here?

src/anyio/_core/_tasks.py
m-u-xyz: A compromise would be to return the "real" result when awaiting the task more than once (I can think of a usecase or two for that), but if there was an exception we subsequently raise `AlreadyUsedError`.

tapetersen: I've also definitely had times when I've used asyncio tasks as futures returnad and waited for in multiple places, if nothing else like an async cache.

I've actually began to reconsider my loose thought about unconditionally wrapping the exception like in the Aborted/Cancelled state as an actually sensible alternative. If you only want the exception raised to abort and log/inform a `TaskFailed directly caused by RuntimeException` still has the info and in the case you really want to catch specific exceptions, which I estimate is much rarer, we could offer that api as an alternative (either a flag/alternative method to `create_task` or as a method on the `TaskHandle`)

Also though about the case where user code only awaits the result once but the exception also causes the `TaskGroup` to fail and is a part of the `ExceptionGroup` and possibly shows a corrupted stack-trace in that (havn't verified just suspected).

Worst case if we get the api wrong there is room to create and recommend another method with better semantics.

src/anyio/_core/_tasks.py
agronholm: Does it? How can I reproduce that? This doesn't:
```py
import anyio


async def error_task() -> None:
    try:
        raise RuntimeError("outer error")
    except RuntimeError:
        raise RuntimeError("inner error")


async def main() -> None:
        try:
            await tg.create_task(error_task())
        except RuntimeError as inner_exc:
            print(str(inner_exc))
            print(str(inner_exc.__context__))

anyio.run(main)
```
This prints:
```
inner error
outer error
```

agronholm: Did you mean a situation where `await handle` is done when there is already an exception in flight?

agronholm: If so, what would you offer as the solution here, so that all the context is preserved?

src/anyio/_core/_tasks.py
agronholm: IIRC one case where tasks are abused this way is with `async_lru`, but this shouldn't be an issue since we have our own LRU cache so there shouldn't be any need to use `async_lru`.

src/anyio/_core/_tasks.py
gschaffner: > Did you mean a situation where `await handle` is done when there is already an exception in flight?

Yes, e.g.

```python

import anyio


async def error_task() -> None:
    try:
        raise RuntimeError("original context of inner error")
    except RuntimeError:
        try:
            raise RuntimeError("inner error")
        except RuntimeError as inner_exc:
            print(f"{inner_exc=!r}")
            print(f"{inner_exc.__context__=!r}")
            raise


async def main() -> None:
        try:
            raise RuntimeError("boom!")
        finally:
            try:
                await tg.create_task(error_task())
            except RuntimeError as inner_exc:
                print(f"{inner_exc=!r}")
                print(f"{inner_exc.__context__=!r}")
                raise


anyio.run(main)
```

prints

```
inner_exc=RuntimeError('inner error')
inner_exc.__context__=RuntimeError('original context of inner error')
inner_exc=RuntimeError('inner error')
inner_exc.__context__=RuntimeError('boom!')
```

> If so, what would you offer as the solution here, so that all the context is preserved?

If we wanted to preserve the original context as-is, we could do

```diff
diff --git a/src/anyio/_core/_tasks.py b/src/anyio/_core/_tasks.py
index 187396d..095fc6f 100644
--- a/src/anyio/_core/_tasks.py
+++ b/src/anyio/_core/_tasks.py
@@ -330,7 +330,15 @@ class TaskHandle(Generic[T_co]):
             yield from self._finished_event.wait().__await__()
 
         if self._exception is not None:
-            raise self._exception
+            context = self._exception.__context__
+            try:
+                # Preserve __cause__ and __suppress_context__ by avoiding `raise ...
+                # from ...`
+                raise self._exception
+            finally:
+                # Preserve __context__
+                self._exception.__context__ = context
+                del context
 
         return self._return_value
```

```
inner_exc=RuntimeError('inner error')
inner_exc.__context__=RuntimeError('original context of inner error')
inner_exc=RuntimeError('inner error')
inner_exc.__context__=RuntimeError('original context of inner error')
```

If we instead wanted to preserve all of the context (i.e. both contexts), we could do

```diff
diff --git a/src/anyio/_core/_tasks.py b/src/anyio/_core/_tasks.py
index 187396d..96d353c 100644
--- a/src/anyio/_core/_tasks.py
+++ b/src/anyio/_core/_tasks.py
@@ -21,6 +21,7 @@ from ._exceptions import TaskAborted, TaskCancelled
 if sys.version_info >= (3, 11):
     from typing import TypeVarTuple
 else:
+    from exceptiongroup import BaseExceptionGroup
     from typing_extensions import TypeVarTuple
 
 T = TypeVar("T")
@@ -330,7 +331,23 @@ class TaskHandle(Generic[T_co]):
             yield from self._finished_event.wait().__await__()
 
         if self._exception is not None:
-            raise self._exception
+            context = self._exception.__context__
+            try:
+                # Preserve __cause__ and __suppress_context__ by avoiding `raise ...
+                # from ...`
+                raise self._exception
+            finally:
+                # Preserve __context__
+                if (
+                    context is not None
+                    and self._exception.__context__ is not None
+                    and self._exception.__context__ is not context
+                ):
+                    self._exception.__context__ = BaseExceptionGroup(
+                        "Exception has multiple __context__s",
+                        [context, self._exception.__context__],
+                    )
+                del context
 
         return self._return_value
```

```
inner_exc=RuntimeError('inner error')
inner_exc.__context__=RuntimeError('original context of inner error')
inner_exc=RuntimeError('inner error')
inner_exc.__context__=ExceptionGroup('Exception has multiple __context__s', [RuntimeError('original context of inner error'), RuntimeError('boom!')])
```

src/anyio/_core/_tasks.py
agronholm: I've added a guard that disallows awaiting on the handle twice.

src/anyio/_core/_tasks.py
agronholm: I get what you're trying to do here, but this...just looks weird to me. Would it be okay to stack the new context on top of the stored exception's context?

gschaffner: Yeah, the `.__context__ = ExceptionGroup()` one does initially look kinda weird to me too. I thought about stacking the contexts that way, but that seems like it would be more confusing to users than using a group because the rendered context would claim/describe a control flow that does not actually match anything possible in the program.

E.g. if we stack the new context on top of the stored one:

```diff
diff --git a/src/anyio/_core/_tasks.py b/src/anyio/_core/_tasks.py
index 7b5c3dc..fea39e5 100644
--- a/src/anyio/_core/_tasks.py
+++ b/src/anyio/_core/_tasks.py
@@ -340,7 +340,20 @@ class TaskHandle(Generic[T_co]):
                 raise
 
         if self._exception is not None:
-            raise self._exception
+            context = self._exception.__context__
+            try:
+                # Preserve __cause__ and __suppress_context__ by avoiding `raise ...
+                # from ...`
+                raise self._exception
+            finally:
+                # If there is a new __context__ from the await site, put it on top of
+                # the existing __context__ stack from the task function
+                if (
+                    self._exception.__context__ is not context
+                    and self._exception.__context__ is not None
+                ):
+                    self._exception.__context__.__context__ = context
+                del context
 
         return self._return_value
```

then

```python

import sys

import anyio


def foo() -> None:
    raise RuntimeError(f"error while handling {sys.exc_info()[1]!r}")


async def error_task() -> None:
    try:
        raise ValueError("asdf")
    except ValueError:
        foo()


async def hello() -> None:
        try:
            raise RuntimeError("boom!")
        finally:
            await tg.create_task(error_task())


async def main() -> None:
    await hello()


anyio.run(main)
```

prints

```
  |   [...]
    |   File "/home/ganden/vc/anyio/eg.py", line 14, in error_task
    |     raise ValueError("asdf")
    | ValueError: asdf
    | 
    | During handling of the above exception, another exception occurred:
    | 
    |   File "/home/ganden/vc/anyio/eg.py", line 22, in hello
    |     raise RuntimeError("boom!")
    | RuntimeError: boom!
    | 
    | During handling of the above exception, another exception occurred:
    | 
    |   File "/home/ganden/vc/anyio/eg.py", line 24, in hello
    |     await tg.create_task(error_task())
    |   File "/home/ganden/vc/anyio/src/anyio/_core/_tasks.py", line 347, in __await__
    |     raise self._exception
    |   File "/home/ganden/vc/anyio/src/anyio/abc/_tasks.py", line 107, in _run_coro
    |     retval = await coro
    |              ^^^^^^^^^^
    |   File "/home/ganden/vc/anyio/eg.py", line 16, in error_task
    |     foo()
    |     ~~~^^
    |   File "/home/ganden/vc/anyio/eg.py", line 9, in foo
    |     raise RuntimeError(f"error while handling {sys.exc_info()[1]!r}")
    | RuntimeError: error while handling ValueError('asdf')
```

This traceback states to the user that a `ValueError` was raised by `error_task`, then the `ValueError` bubbled up to `main` and cleanup code in `main` raised `boom!` while handling it, then the `boom!` from `main` bubbled up to `error_task` (impossible) and cleanup code somehow back in `error_task` raised the task's actual exception while handling the exception from `main`.

gschaffner: We could also do the reverse and try slipping the new context _under_ the existing context stack, but this also causes impossible tracebacks to be reported to users:

```diff
diff --git a/src/anyio/_core/_tasks.py b/src/anyio/_core/_tasks.py
index 7b5c3dc..a3d0c03 100644
--- a/src/anyio/_core/_tasks.py
+++ b/src/anyio/_core/_tasks.py
@@ -340,7 +340,24 @@ class TaskHandle(Generic[T_co]):
                 raise
 
         if self._exception is not None:
-            raise self._exception
+            context = self._exception.__context__
+            try:
+                # Preserve __cause__ and __suppress_context__ by avoiding `raise ...
+                # from ...`
+                raise self._exception
+            finally:
+                # If there is a new __context__ from the await site, slip it *under* the
+                # existing __context__ stack from the task function.
+                if context is not None:
+                    last_exc = context
+                    while last_exc.__context__ is not None:
+                        last_exc = last_exc.__context__
+
+                    last_exc.__context__ = self._exception.__context__
+                    del last_exc
+                    self._exception.__context__ = context
+
+                del context
 
         return self._return_value
```

```python

import sys

import anyio


def foo() -> None:
    raise RuntimeError(f"error while handling {sys.exc_info()[1]!r}")


async def error_task() -> None:
    try:
        raise ValueError("asdf")
    except ValueError:
        foo()


async def hello() -> None:
        handle = tg.create_task(error_task())
        try:
            raise RuntimeError("boom!")
        finally:
            await handle


async def main() -> None:
    await hello()


anyio.run(main)
```

```
  |   [...]
    |   File "/home/ganden/vc/anyio/eg.py", line 23, in hello
    |     raise RuntimeError("boom!")
    | RuntimeError: boom!
    | 
    | During handling of the above exception, another exception occurred:
    | 
    |   File "/home/ganden/vc/anyio/eg.py", line 14, in error_task
    |     raise ValueError("asdf")
    | ValueError: asdf
    | 
    | During handling of the above exception, another exception occurred:
    | 
    |   File "/home/ganden/vc/anyio/eg.py", line 25, in hello
    |     await handle
    |   File "/home/ganden/vc/anyio/src/anyio/_core/_tasks.py", line 347, in __await__
    |     raise self._exception
    |   File "/home/ganden/vc/anyio/src/anyio/abc/_tasks.py", line 107, in _run_coro
    |     retval = await coro
    |              ^^^^^^^^^^
    |   File "/home/ganden/vc/anyio/eg.py", line 16, in error_task
    |     foo()
    |     ~~~^^
    |   File "/home/ganden/vc/anyio/eg.py", line 9, in foo
    |     raise RuntimeError(f"error while handling {sys.exc_info()[1]!r}")
    | RuntimeError: error while handling ValueError('asdf')
```

The task code raising `asdf` did _not_ run in a context where `sys.exception` is the `boom!` from main, but the traceback claims that it did.

gschaffner: With _any_ approach that doesn't set the final context to `.__context__ = stored_context` (or possibly also `None`), I think that there are user programs for which we would give the user a traceback which claims that a control flow that did not happen in their program happened.

This is true for the `.__context__ = BaseExceptionGroup` approach too, but at least the group approach can try hard to make it clear to the user what actually happened via a descriptive exception group message. The stacking approaches above have no such text field for a descriptive message and IMO just produce unclear/confusing tracebacks.

gschaffner: What if we preserved the saved `__context__` as-is and added a note with the new context?:

```diff
diff --git a/src/anyio/_core/_tasks.py b/src/anyio/_core/_tasks.py
index 7b5c3dc..1ac06d7 100644
--- a/src/anyio/_core/_tasks.py
+++ b/src/anyio/_core/_tasks.py
@@ -2,6 +2,8 @@ from __future__ import annotations
 
 import math
 import sys
+import textwrap
+import traceback
 from collections.abc import (
     Coroutine,
     Generator,
@@ -340,7 +342,37 @@ class TaskHandle(Generic[T_co]):
                 raise
 
         if self._exception is not None:
-            raise self._exception
+            context = self._exception.__context__
+            try:
+                # Preserve __cause__ and __suppress_context__ by avoiding `raise ...
+                # from ...`
+                raise self._exception
+            finally:
+                # Preserve __context__. If there is a second context at the call site
+                # that is awaiting this, add it as a note. (Note that we cannot modify
+                # the existing context to somehow add the new context into it without
+                # causing some programs to report tracebacks that describe control flow
+                # that did not actually happen in the program.)
+                if (
+                    self._exception.__context__ is not context
+                    and self._exception.__context__ is not None
+                ):
+                    if sys.version_info >= (3, 11):
+                        self._exception.add_note(
+                            "The above exception was propagated by `await TaskHandle()` during the handling of the following exception:\n"
+                            + textwrap.indent(
+                                (
+                                    "".join(
+                                        traceback.format_exception(
+                                            self._exception.__context__
+                                        )
+                                    )
+                                ).rstrip(),
+                                "  ",
+                            )
+                        )
+                self._exception.__context__ = context
+                del context
 
         return self._return_value
```

Then e.g.

```python

import anyio


async def error_task() -> None:
    try:
        raise RuntimeError("original context of inner error")
    except RuntimeError:
        try:
            raise RuntimeError("inner error")
        except RuntimeError as inner_exc:
            print(f"{inner_exc=!r}")
            print(f"{inner_exc.__context__=!r}")
            raise


async def main() -> None:
        try:
            raise RuntimeError("boom!")
        finally:
            try:
                await tg.create_task(error_task())
            except RuntimeError as inner_exc:
                print(f"{inner_exc=!r}")
                print(f"{inner_exc.__context__=!r}")
                raise


anyio.run(main)
```

gives

```
inner_exc=RuntimeError('inner error')
inner_exc.__context__=RuntimeError('original context of inner error')
inner_exc=RuntimeError('inner error')
inner_exc.__context__=RuntimeError('original context of inner error')
  |   [...]
    |   File "/home/ganden/vc/anyio/eg.py", line 8, in error_task
    |     raise RuntimeError("original context of inner error")
    | RuntimeError: original context of inner error
    | 
    | During handling of the above exception, another exception occurred:
    | 
    |   File "/home/ganden/vc/anyio/eg.py", line 24, in main
    |     await tg.create_task(error_task())
    |   File "/home/ganden/vc/anyio/src/anyio/_core/_tasks.py", line 349, in __await__
    |     raise self._exception
    |   File "/home/ganden/vc/anyio/src/anyio/abc/_tasks.py", line 107, in _run_coro
    |     retval = await coro
    |              ^^^^^^^^^^
    |   File "/home/ganden/vc/anyio/eg.py", line 11, in error_task
    |     raise RuntimeError("inner error")
    | RuntimeError: inner error
    | The above exception was propagated by `await TaskHandle()` during the handling of the following exception:
    |   Traceback (most recent call last):
    |     File "/home/ganden/vc/anyio/eg.py", line 21, in main
    |       raise RuntimeError("boom!")
    |   RuntimeError: boom!
```

With this approach, all tracebacks will have the task function's correct stack trace in it (including correctly adhering to `__suppress_context__` if the task function used `raise ... from ...`), but will still provide the `__context__` of the await-site for debugging purposes.

src/anyio/_core/_tasks.py
gschaffner, line 1: This seems like a footgun:

```python
import anyio


async def taskfunc() -> None:
    raise RuntimeError("boom!")


async def main() -> None:
        tg.create_task(taskfunc())
        # forgot to await handle


anyio.run(main)
```

Output: nothing; exit code 0, success.

I think that we should not swallow the exception if a `TaskHandle` is abandoned like this. If we silently swallow exceptions like this, we are handing folks a footgun that is really easy to accidentally shoot themselves with. `asyncio.create_task` has this same footgun. `asyncio.TaskGroup.create_task` does not (IIRC).

(I also suspect that abandoning a `TaskHandle` like this is often a programming error: when someone does this, I think that they will often have assumed that `create_task`, used like this, behaves like `start_soon` and does not silently eat exceptions.

In such a case where a `TaskHandle` is abandoned, the only semantics that we _know_ are safe to give to the user are `start_soon` semantics. Anything else could IMO easily result in users accidentally suppressing exceptions.)

Another way to state this is: I think that `create_task`/`TaskHandle` should have the property that a task that exits with an exception will have that exception raised exactly once: certainly not less, and (probably) not more. So:

*   If it's a `BaseException`, raise it where it is since we can't safely teleport it elsewhere, and have any `TaskHandle` waiter raise `TaskAborted`.

*   If it's an `Exception`, either:

    *   If the task group exits without a `TaskHandle` waiter having raised the `create_task` task's exception, then raise the task's exception from the `TaskGroup`, just as we would have if the user used `tg.start_soon` instead.

    *   Else there was a `TaskHandle` waiter that we teleported the task's exception to, raising it there.

If we go with that route, this "exactly once" thing would also IMO be a fairly intuitive way to explain the exception semantics of `anyio.abc.TaskGroup.create_task` to users.

smurfix:
> `tg.create_task(taskfunc())`

We intentionally don't do that anywhere else in the anyio code base.

I'm mostly-OK with supporting a coroutine-based task creation method for more convenient migration (though I'd never use it), but what I'm not really OK with is a `TaskHandle` that's *only* available via asyncio's IMHO-broken calling conventions.

src/anyio/_core/_exceptions.py
```diff
@@ -154,3 +154,31 @@ def __init__(self) -> None:
...
+class TaskAborted(Exception):
+    """
+    Raised when awaiting on a :class:`.TaskHandle` which was terminated by a
+    :exc:`BaseException`.
+
+    This exception class exists because a :exc:`BaseException` (i.e. one
+    that is not an :exc:`Exception`) should not be suppressed nor
+    forwarded to another scope.
+
+    .. seealso:: :exc:`TaskCancelled`
```
tapetersen: The docstring here would need to be updated if we raise this for all exceptions.

I'm contemplating if there is a reason to have a separate subclass for non-base exceptions but that's probably YAGNI as long as the docstring is clear.

agronholm: Updated. One minor detail is whether `TaskCancelled` should be a subclass of `TaskAborted` or not. I have no strong opinion on this.

tapetersen: Otherwise I give the bikeshed-colors a green light for both exceptions and status names.

tapetersen: Instinct is that a common baseclass would be useful for handling errors from an awaited task vs. those in directly called functions in a block so if we don't want `TaskCancelled` to be a `TaskAborted` I would want a, possibly abstract `TaskError` as base for both.

agronholm: I contemplated renaming `TaskAborted` to `TaskError` now that we no longer make a distinction between `BaseException` and `Exception`, but I'm -1 on having a common base class just for this.

tapetersen: Another thought is if providing a standalone `Future` is something we ever want to support and if so the consumer-side interface of that would probably overlap heavily with this in which case the `Task` prefix might be a slightly bad match worth considering.

tapetersen: Given you can catch multiple exception types in the same except block without them being related it's not that onerous to do explicitly so I don't care that much. 

Regarding `TaskAborted` vs `TaskError` I prefer `TaskError` but there's a reasonable argument that `TaskCancelled` doesn't match there. Both will work.

src/anyio/_core/_tasks.py
agronholm: (And now replaced it with exception wrapping...)

src/anyio/_core/_exceptions.py
agronholm: I renamed it to `TaskError` now.

gschaffner:
> I agree that `create_task()` shouldn't swallow exceptions. Even `asyncio.TaskGroup` would raise an exception group here. But at the same time we can't hold on to the principle of awaiting only once since the exception goes to the task group first.

I don't think that that would prevent us from implementing "exactly once" exception semantics. Here is a quick-and-dirty sketch of an implementation: cef88f6ee9972489a1863fab55c9b0b3bc31c844...b6729cce6f6f84b8661c52a5d5a138b86aa191b8

> OK, so where do we stand on the design now? Right now the implementation raises the exceptions directly from the task group, but if the handle is waited on, the exception is wrapped so as not to pollute the original `__context__`.

An observation: this breaks the symmetry between `TaskGroup.create_exception` and `BlockingPortal.start_task_soon`. Not to imply that that's necessarily a problem—just something to note/consider.

Overall the new "never move an exception" exception semantics seem good to me. I think that the "can move an exception"/"exactly once" exception semantics would be more flexible and would enable more use cases. However, on balance, the "never-move" semantics are IMO likely the better decision here if the primary objective of this API is to make it easier for `asyncio.TaskGroup.create_task` users (and to a lesser extent, `asyncio.create_task` users) to switch to AnyIO.

<details>

<summary>Analysis that motivated ^ (click to expand)</summary>

Different options for `TaskGroup.create_task` semantics will affect users differently. Comparing the following two options that have been proposed:

*   An `Exception` from a `create_task` task is never moved to the task's `TaskHandle` waiter, but instead the waiter will wake up with `raise TaskError`. An `Exception` from a `create_task` task immediately cancels the task group. Call this "never-move"/"can-cancel" semantics.

*   A `create_task` task can have its `Exception` moved to its `TaskHandle` waiter, but if it isn't moved then the exception will still be raised exactly once. An `Exception` from a `create_task` task does not cancel the task group. Call this "can-move"/"never-cancel" semantics.

Both of these options prevent the user from accidentally suppressing an exception.

The two different options will place different restrictions on the use cases for `TaskGroup.create_task`. Here is my current thinking:

*   "Can-move" semantics (but not "never-move" semantics) allow the user to use a `TaskGroup` to capture a function's result. I.e. if the user awaits the handle (and even if the handle wait gets cancelled), then `create_task` behaves like this:

    ```python
    T = TypeVar("T")
    P = ParamSpec("P")
    BaseExceptionT = TypeVar("BaseExceptionT", bound=BaseException)


    @dataclass(frozen=True, slots=True)
    class Result(Generic[T]):
        value: T

        def unwrap(self) -> T:
            return self.value


    @dataclass(frozen=True, slots=True)
    class Error(Generic[BaseExceptionT]):
        exception: BaseExceptionT

        def unwrap(self) -> Never:
            context = self.exception.__context__
            try:
                # Preserve __cause__ and __suppress_context__ by avoiding `raise ...
                # from ...`
                raise self.exception
            finally:
                # Preserve __context__. TODO: Consider doing: If there is a second context
                # at the call site that is awaiting this, add it as a note. (Note that we
                # cannot modify the existing context to somehow add the new context into it
                # without causing some programs to report tracebacks that describe control
                # flow that did not actually happen in the program.)
                self.exception.__context__ = context
                del context


    async def func() -> int:
        raise RuntimeError("i'm func")


    async def main() -> None:
        container = list[Result[int] | Error[BaseException]]()
        event = Event()

        async def acapture(
            func: Callable[P, Awaitable[T]], *args: P.args, **kwargs: P.kwargs
        ) -> None:
            try:
                ret_val = await func(*args, **kwargs)
            except BaseException as exc:
                container.append(Error(exc))
                event.set()
            else:
                container.append(Result(ret_val))
                event.set()

        try:
                tg.start_soon(acapture, func)
                await event.wait()
                container.pop().unwrap()
        except BaseExceptionGroup as exc_grp:
            if container and isinstance(result := container.pop(), Error):
                # event.wait() was cancelled.
                raise exc_grp.derive((*exc_grp.exceptions, result.exception))  # noqa: B904
            else:
                raise
        except BaseException as exc:
            if container and isinstance(result := container.pop(), Error):
                # event.wait() was cancelled.
                raise BaseExceptionGroup("some msg", (exc, result.exception))  # noqa: B904
            else:
                raise
        else:
            if container and isinstance(result := container.pop(), Error):
                # event.wait() was cancelled.
                raise BaseExceptionGroup("some msg", (result.exception,))
     ```

     It is...challenging...for a user to implement this correctly in a way that (a) won't accidentally swallow the exception and (b) won't give them and their users tracebacks that describe control flows that didn't actually happen. (Actually, I'm not completely certain that the above will not produce incorrect tracebacks; I'd need to write some tests for it.)

     So, do users have use cases for that `outcome.acapture`-like behavior? My `n=1`: yes.

     *   I have spent a lot of time working on de/multiplexers and this would have been a very useful feature in many of those cases. (Details: You often end up wanting something like `(await task_portal.create_task(outcome.acapture, func)).unwrap()` when `func` includes some stateful operation that you can't safely cancel without unrecoverably corrupting the state of something. E.g. if `func` contains a `TLSStream` operation, or if `func` contains multiple subsequent `ReliableStream.send` calls that together compose a larger protocol message that must remain atomic (even in the face of cancellation of a single protocol-level operation). Without an async-to-async portal, you can also do this with a service task (`tricycle.open_service_nursery`) and some actor-like inter-task messaging (`MemoryObjectStream`), but portals are IMO often more convenient and more familiar to Python developers than actors. A portal is just like `concurrent.futures.Executor`!)

     *   A _potential_/future use case: as nogil increasingly becomes a thing, there will be more people running multiple event loops in one interpreter, and use cases for async-thread-to-async-thread portal calls will probably increase. If we wanted to add cross-threaded-`__await__ ` support to `create_task` in the future, I suspect it's mostly just a matter of adding some `loop_token.run_sync_soon` calls and a threading mutex for the shared state.

     (Distantly related request/issue: sync-async-thread-to-async-async-thread portal calls: #341/#525/https://github.com/python-trio/trio/issues/2534)

     Counterargument 1: The difficult part of implementing this as a user is the "raise the exception out of the group if it was never raised by a `TaskHandle`" part. A user that can shield their `event.wait()` call does not have this problem.

     Response to 1: Shielding that is not suitable in the multiplexer cases that I've encountered. When I've been in a situation where I could shield the `await task_handle` call, I didn't need a `TaskHandle` in the first place, because I could just replace `with shield: (await task_portal.create_task(outcome.acapture, func)).unwrap()` with `with shield: await func()`.

     Counterargument 2: It may be surprising to users that the exception of a `TaskHandle` they abandoned some time ago will wait around until the body of their `async with tg` exits, and _then_ will get raised. Without automatic cancellation of the entire task group, there is no mechanism to hasten the `create_task` task's exception being raised somewhere.

     Response to 2: The mechanism to hasten the exception being raised somewhere is `await task_handle`.

     Response to response to 2: Yes, but it's not difficult for a user to accidentally forget `await task_handle`, or to `await task_handle` to get cancelled when they meant to shield it.

     Counterargument 3: `TaskGroup.create_task` is not intended to be like `concurrent.futures.Executor`/`BlockingPortal`. It's intended to be like `TaskGroup.start_soon` and `asyncio.TaskGroup.create_task`, which cancel the entire group immediately if the spawned task fails. If someone calls `create_task` and discards the handle, the behavior should gracefully degrade to _exactly_ the `start_soon` behavior.

*   "Never-move" semantics (but not "can-move" semantics) make an `Exception` in a `create_task` task cancel the other tasks in the task group.

     So, do users have use cases where they want that to happen? Probably. No specific user-use cases have immediately come to my mind, but semantics like this are possibly(?) useful if someone wants to write an API like `asyncio.wait(task_handles, return_when=FIRST_EXCEPTION)`. Take that with a grain of salt though; I haven't actually tried to implement that function.

     Counterargument 4: If someone wanted this behavior but `create_task` had the "can-move" semantics, it seems pretty straightforward to implement this behavior.

    ```python
    async def main2() -> None:
        async def cancel_group_on_error(
            task_group: TaskGroup,
            func: Callable[P, Awaitable[T]],
            *args: P.args,
            **kwargs: P.kwargs,
        ) -> T:
            try:
                return await func(*args, **kwargs)
            except Exception:
                raise

            handle = tg.create_task(cancel_group_on_error(tg, func))

            # The tg gets cancelled immediately when func raises:
            await anyio.sleep(10)

            # If the user does call await handle, note that it will now raise the exception,
            # whereas with pure never-move semantics it would raise TaskError instead. Is
            # that difference important? If so, we could expose a `handle.wait` that only
            # waits for the event but does not return/raise the captured result or a
            # TaskError.
            await handle
     ```

So, my preference here is somewhat mixed:

*   The "can-move" option seems more liberal, i.e. it seems to be suitable for more use cases that the alternative does not support. (At least some of them are use cases that I've spent a lot of time with and am pretty familiar with, but I'm also probably a bit biased because of how such an API would have been quite convenient for me.)

*   The "never-move" option seems to be suitable for (strictly?) fewer use cases, but the behavior in "counterargument 2" could be confusing to users, and "counterargument 3" is convincing to me as well.

Some questions for you all:

*   Are there use cases for "never-move" semantics that are challenging/complicated/impossible to do with "can-move" semantics?

*   Are there ways in which "can-move" semantics are problematic?

     The most compelling thing that I am aware of here is "counterargument 3" above.

</details>

src/anyio/_core/_tasks.py
```diff
@@ -171,3 +193,135 @@ def create_task_group() -> TaskGroup:
...
+    def return_value(self) -> T_co:
+        match self._status:
+            case TaskHandle.Status.PENDING:
+                raise RuntimeError("the task has not returned yet")
+            case TaskHandle.Status.FINISHED:
+                return self._return_value
+            case TaskHandle.Status.CANCELLED:
+                raise TaskCancelled("the task was cancelled")
+            case TaskHandle.Status.ERRORED:
+                raise TaskError("the task raised an exception") from self._exception
```
gschaffner: This `from` doesn't seem to be working:

```python

import traceback

import anyio


async def raiser():
    raise RuntimeError("boom")


async def main() -> None:
        handle = tg.create_task(raiser())
            try:
                await handle
            except BaseException as exc:
                print("".join(traceback.format_exception(exc)).rstrip())
                print()
                raise


anyio.run(main)
```

gives (CPython 3.13.12)

```
  File "/home/ganden/vc/anyio/src/anyio/_core/_tasks.py", line 262, in _run_coro
    retval = await self._coro
             ^^^^^^^^^^^^^^^^
  File "/home/ganden/vc/anyio/eg.py", line 9, in raiser
    raise RuntimeError("boom")
RuntimeError: boom

  File "/home/ganden/vc/anyio/eg.py", line 17, in main
    await handle
  File "/home/ganden/vc/anyio/src/anyio/_core/_tasks.py", line 321, in __await__
    return self.return_value
           ^^^^^^^^^^^^^^^^^
  File "/home/ganden/vc/anyio/src/anyio/_core/_tasks.py", line 315, in return_value
    raise TaskError("the task raised an exception") from self._exception
anyio.TaskError: the task raised an exception

  |   File "/home/ganden/vc/anyio/eg.py", line 24, in <module>
  |     anyio.run(main)
  |     ~~~~~~~~~^^^^^^
  |   File "/home/ganden/vc/anyio/src/anyio/_core/_eventloop.py", line 77, in run
  |     return async_backend.run(func, args, {}, backend_options)
  |            ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |   File "/home/ganden/vc/anyio/src/anyio/_backends/_asyncio.py", line 2358, in run
  |     return runner.run(wrapper())
  |            ~~~~~~~~~~^^^^^^^^^^^
  |   File "/home/ganden/.pyenv/versions/3.13/lib/python3.13/asyncio/runners.py", line 118, in run
  |            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^
  |   File "/home/ganden/.pyenv/versions/3.13/lib/python3.13/asyncio/base_events.py", line 725, in run_until_complete
  |     return future.result()
  |            ~~~~~~~~~~~~~^^
  |   File "/home/ganden/vc/anyio/src/anyio/_backends/_asyncio.py", line 2341, in wrapper
  |     return await func(*args)
  |            ^^^^^^^^^^^^^^^^^
  |   File "/home/ganden/vc/anyio/eg.py", line 13, in main
  |     async with anyio.create_task_group() as tg:
  |                ~~~~~~~~~~~~~~~~~~~~~~~^^
  |   File "/home/ganden/vc/anyio/src/anyio/_backends/_asyncio.py", line 799, in __aexit__
  |         "unhandled errors in a TaskGroup", self._exceptions
  |     ) from None
  | ExceptionGroup: unhandled errors in a TaskGroup (2 sub-exceptions)
    |   File "/home/ganden/vc/anyio/eg.py", line 17, in main
    |     await handle
    |   File "/home/ganden/vc/anyio/src/anyio/_core/_tasks.py", line 321, in __await__
    |     return self.return_value
    |            ^^^^^^^^^^^^^^^^^
    |   File "/home/ganden/vc/anyio/src/anyio/_core/_tasks.py", line 315, in return_value
    |     raise TaskError("the task raised an exception") from self._exception
    | anyio.TaskError: the task raised an exception
    +---------------- 2 ----------------
    |   File "/home/ganden/vc/anyio/src/anyio/_core/_tasks.py", line 262, in _run_coro
    |     retval = await self._coro
    |              ^^^^^^^^^^^^^^^^
    |   File "/home/ganden/vc/anyio/eg.py", line 9, in raiser
    |     raise RuntimeError("boom")
    | RuntimeError: boom
```

The `__context__` that we are setting with this `from` is missing from the exception group rendering.

src/anyio/_core/_tasks.py
gschaffner: I am uneasy about exposing the property `TaskHandle.exception`. `raise task_handle.exception` seems like a footgun that is too easy to shoot oneself with; it has the same issues that Tobias brought up in this thread.

What is the motivation for exposing `TaskHandle.exception` publicly?

src/anyio/_core/_tasks.py
```diff
@@ -171,3 +193,135 @@ def create_task_group() -> TaskGroup:
...
+
+    def __await__(self) -> Generator[Any, Any, T_co]:
```
gschaffner: It feels somewhat strange to add an `__await__` (zero of these are in AnyIO or Trio currently) when AnyIO's task spawning APIs have a longstanding lack of support for `Callable[..., Awaitable]`[^1], let alone `Awaitable`.

[^1]: Related: 6ba4245e967d26845d92c9091ca448ce98f41e95, #675, https://github.com/python-trio/trio-typing/pull/66

I'm also slightly concerned that supporting `__await__` may accidentally encourage users to write their own `def __await__`s. IMO everyone should avoid doing `def __await__` and should do `async def` instead whenever possible.

Changing `def task_handle.__await__` to `async def task_handle.wait` would make sense to me. E.g. `Event` doesn't have `Event.__await__`, it has `Event.wait`. If `TaskHandle.__await__` is considered very important in order to minimize the number of changes that asyncio users porting to AnyIO need to do, I think that we should at least expose `.wait` too.

src/anyio/_core/_tasks.py
gschaffner: It works on CPython 3.11 and 3.12, but not 3.10, 3.13, and 3.14. Possibly a CPython regression? I will take a closer look tomorrow.

src/anyio/_core/_tasks.py
agronholm: > It feels somewhat strange to add an __await__ (zero of these are in AnyIO or Trio currently) when AnyIO's task spawning APIs have a longstanding lack of support for Callable[..., Awaitable][1](https://github.com/agronholm/anyio/pull/1098#user-content-fn-1-d45dd1163856e86ecc919c3a562cfc14), let alone Awaitable.

Sure, we're kind of breaking new ground here, to serve the stated goals.

> I'm also slightly concerned that supporting __await__ may accidentally encourage users to write their own def __await__s. IMO everyone should avoid doing def __await__ and should do async def instead whenever possible.

I feel you're overthinking here. `__await__` is kind of that "Here be dragons" on the map for 99.9% of users.

> Changing def task_handle.__await__ to async def task_handle.wait would make sense to me. E.g. Event doesn't have Event.__await__, it has Event.wait. If TaskHandle.__await__ is considered very important in order to minimize the number of changes that asyncio users porting to AnyIO need to do, I think that we should at least expose .wait too.

Indeed, the express goals of this were to provide:

1. An easy way to retrieve an individual task's return value
2. An easy way to cancel an individual task
3. A familiar interface for users migrating from asyncio

Providing an `__await__` serves goal number 3. What would be actually gain by duplicating this functionality in a `wait()` method? Further, on asyncio, the free function `wait()` is used to wait for completion without actually get the return value.

src/anyio/_core/_tasks.py
agronholm: > What is the motivation for exposing TaskHandle.exception publicly?

Parity with `asyncio.Task`.

agronholm: There is a huge number of ways people could shoot their foot, and surprisingly many will go the great lengths to find such ways. If I removed the `exception` property, rest assured some enterprising coder would find the `_exception` attribute and use that to re-raise the exception.

src/anyio/abc/_tasks.py
```diff
@@ -55,6 +58,37 @@ class TaskGroup(metaclass=ABCMeta):
 
     cancel_scope: CancelScope
 
+    def create_task(
+        self,
+        coro: Coroutine[Any, Any, T_Retval],
```
Graeme22, line 63: I get that part of the idea is to have an API that's more similar to asyncio, but it does feel weird that you *must* pass a coroutine object to `create_task` and you *must* pass arguments to `start_soon`. Very inconsistent since many users will likely end up using both, I'm envisioning a pattern like:

```python
    tg.start_soon(task_we_dont_need_result_for)
    handle = tg.create_task(task_we_need_result_for())
res = await handle
```

My impression was that the primary purpose of this PR is to facilitate accessing task results easily, and that compatibility with asyncio's API was a secondary or tertiary issue.

agronholm: What's the problem with this then? Do you have a better alternative?

agronholm: Note that passing keyword arguments is also more convenient compared to `start_soon()`.

Graeme22: I would propose removing the context arg and just copying the `start_soon` signature.
Re kwargs: that is nice, but would also be nice on `start_soon`. Not sure why we'd have it in just one place and not both.

agronholm: You can pass positional args but not kwargs with `start_soon`, and we can't start accepting arbitrary kwargs. And the name `create_task()` doesn't make sense unless it's very close to the asyncio API.

Graeme22: Right which is why I propose maintaining same style as start_soon. The point about the name is valid--did the overload start_soon with a return_handle arg idea get rejected? I liked that better (though I suppose both could be done technically?)

src/anyio/_core/_tasks.py
m-u-xyz: While I do agree (somewhat) with your motivation here, it's quite unfortunate that sub-optimal asyncio design decisions that somewhat-worked a decade ago now force us to replicate these misteaks.

On the plus side, we can always replace that `exception` attribute with a property that issues a deprecation warning and tells people to please call `reraise()` instead (or whichever other reasonable interface we can think of).

src/anyio/abc/_tasks.py
agronholm: No, it wasn't rejected, but I've made several PRs that just got too complex for their own good, so this time I wanted to take it in small steps. But a new task handle class is a required part in any implementation, however minimal.

tapetersen: I wasn't involved in the design discussions so I'm not clear on what's main goals and secondary here. If it's mainly meant to be a temporary help for users migrating from asyncio I agree that taking the current name and accepting coroutines rather than callables are sensible.

If we at all want the concept of task-handles to be used in anyio for "cleaner" projects it would be somewhat unfortunate if we were forced to use the coroutine calling style for this one useful concept.

If access to the same taskhandle was later possible through `start_soon` or another call that would be ok though.

agronholm: Not sure what you mean by "temporary" – do you see this method being removed from `TaskGroup` at some point in time? I don't see that as likely. But yes, ideally all the task-spawning methods would return task handles. I think that was the consensus in the design meeting.

tapetersen: I meant "temporary" as in that we only expect task handles to be a band-aid while porting but you answered that. My thoughts would then be that it's ok for create_task to be an odd API if it's basically just a familiar wrapper over functionality start_soon provides and perhaps marked as such.

agronholm: If task handles were band-aid, what would the "final" solution be, then?

tapetersen: I just wasn't sure there would be any at all since Trio doesn't provide TaskHandles for a reason and thought this could be a limited compromise only for create_task and not expanded to the general task-handling in AnyIO. I'm fully satisfied by the answers.

(If I remember the reasons for not having TaskHandles in Trio a lot of it had todo with both the double unwrap problem as well as the issue of cancellation propagating through an await of any task which are both solved by this solution).

Graeme22: If the intention is for this change to be present alongside changes to start and start_soon and not instead of them, then consider my objection resolved.

As for Trio disallowing this--I think they're just wrong. Why force basically every user of your library to write the same utilities to get task results when we can easily add that to the library?

src/anyio/_core/_tasks.py
Graeme22: Do we want it to be an anti-pattern to await a task result before its parent TaskGroup finishes? If so we could potentially add some kind of guard with a helpful error/warning message

agronholm: I see no reason to prohibit users from doing that.

Graeme22: I suppose if we had that rule we wouldn't need `__await__` at all! But I can envision scenarios where you'd want to get a result before other tasks have finished, just wasn't sure how strict we want to be.

src/anyio/abc/_tasks.py
m-u-xyz: > Why force basically every user of your library to write the same utilities to get task results when we can easily add that to the library?

Because

(a) when trio was created there was no consensus WRT what that should look like. For example, if task A (which waits on task B) is cancelled, is task B auto-cancelled along with it?

(b) Trio intentionally disallowed returning a value from a task because it's far too easy to create code full of anti-patterns when you do this. "Create a global taskgroup and then it's business as usual" is not a goal of Trio.

Granted that there _are_ valid usecases for actually returning a value from a task, but I'd wager that 99% of *actual* uses of this interface are going to be 1:1 ports from asyncio with, um, otherwise zero consideration for "real" Structured Concurrency.

agronholm: FWIW, you can still have a design with just one root task group in Trio. I don't see what that has to do with not returning task handles.

Graeme22: Honestly it hadn't even occurred to me that users could abuse this to create a global task group and basically ignore structured concurrency--but that could already be done before, just with more steps. I think your 99% estimate is wrong, it's completely reasonable for people who already use SC to want to collect results easily. Since `create_task` is part of the `TaskGroup` API, there's no violation of SC here unless users intentionally work around it. However, I do see how the copying of asyncio's API specifically could promote a mindset of lazily migrating without changing the mindset. Perhaps forcing users to only use `start` and `start_soon` would help with that since they *have* to rewrite their code.

src/anyio/_core/_tasks.py
Graeme22: I see we can just use `TaskHandle.return_value` outside of the task group block and it will just work, no awaiting needed, which is great.

src/anyio/abc/_tasks.py
agronholm: I'm not sure what repercussions such a lazy migration would have. If it causes errors, a test suite would catch it. If a project has no test suite, it's a lost cause anyway.

src/anyio/_core/_tasks.py
agronholm: First, I'd like to point out that `from` doesn't set `__context__`, it sets `__cause__`.

This has everything to do with how exception groups are formatted. Consider this:
```py
try:
    raise RuntimeError("boom")
except RuntimeError as exc:
    try:
        raise RuntimeError("another boom") from exc
    except RuntimeError as exc2:
        excgrp = ExceptionGroup("group", (exc, exc2))

print("".join(traceback.format_exception(excgrp)).rstrip())
print("".join(traceback.format_exception(excgrp.exceptions[1])).rstrip())
```
This results in:
```
  | exceptiongroup.ExceptionGroup: group (2 sub-exceptions)
    |   File "/home/alex/workspace/anyio/bug-trio.py", line 30, in <module>
    |     raise RuntimeError("boom")
    | RuntimeError: boom
    +---------------- 2 ----------------
    |   File "/home/alex/workspace/anyio/bug-trio.py", line 33, in <module>
    |     raise RuntimeError("another boom") from exc
    | RuntimeError: another boom
  File "/home/alex/workspace/anyio/bug-trio.py", line 30, in <module>
    raise RuntimeError("boom")
RuntimeError: boom

  File "/home/alex/workspace/anyio/bug-trio.py", line 33, in <module>
    raise RuntimeError("another boom") from exc
RuntimeError: another boom
```
The `__cause__` is indeed preserved but not rendered in the exception group's traceback when the `__cause__` is rendered in the traceback.
If we leave out the `__cause__` (`RuntimeError("boom")`) from the exception group, you'll find that the exception is indeed rendered as the cause of the second one:
```py
try:
    raise RuntimeError("boom")
except RuntimeError as exc:
    try:
        raise RuntimeError("another boom") from exc
    except RuntimeError as exc2:
        excgrp = ExceptionGroup("group", (exc2,))

print("".join(traceback.format_exception(excgrp)).rstrip())
```
Results in:
```
  | exceptiongroup.ExceptionGroup: group (1 sub-exception)
    |   File "/home/alex/workspace/anyio/bug-trio.py", line 30, in <module>
    |     raise RuntimeError("boom")
    | RuntimeError: boom
    | 
    | The above exception was the direct cause of the following exception:
    | 
    |   File "/home/alex/workspace/anyio/bug-trio.py", line 33, in <module>
    |     raise RuntimeError("another boom") from exc
    | RuntimeError: another boom
```
To summarize, this behavior has nothing whatsoever to do with the task handle's exception handling logic.

src/anyio/_core/_tasks.py
agronholm: What mistakes are you seeing here? And no, I'm not going slap a deprecation on the `exception` attribute. It's useful for introspection and monitoring tools. Otherwise you'll have to `await` on the handle and get the exception re-raised, and even then you'd have to dig it up from `__cause__`. Is that what you want?

agronholm:
I realized that asyncio only has 3 states in a task: pending, cancelled and finished. Do we actually need that extra "errored" state?

smurfix:
Is there any actual difference (except the observable state)? I'd assume that there is not …

agronholm:
The only difference is that with `asyncio.Task` where the state is "finished", you have to call the `exception()` method to see if the task completed successfully or not.

agronholm:
I was also considering adding a `wait()` method to ensure that the task has finished one way or another. It would be a simple wrapper for `await self._finished_event.wait()`. Otherwise users would have to `await` on the task directly and handle any potential exceptions which makes no sense if they just want to ensure that the task is no longer running.

smurfix:
> I was also considering adding a wait() method

+1; would allow me to ditch the "task_status.started(cancelscope) plus a 'done' event" pattern that I otherwise need way too often.

agronholm:
I added the `wait()` method, and a `CANCELLING` state to differentiate between `PENDING` and a state where the task is still running but in a cancelled state. I also tweaked the `__await__` to always go through a checkpoint.

agronholm:
Some further thoughts:

- The `_status` field might be redundant, as the status could be inferred from `self._finished_event`, `self.cancel_scope.cancel_called` and `self._exception`
- Do we want to store the cancellation exception in `self._exception`?
- Do we want a dedicated exception in place of `RuntimeError("the task has not returned yet")`?

davidbrochart:
> * The `_status` field might be redundant, as the status could be inferred from `self._finished_event`, `self.cancel_scope.cancel_called` and `self._exception`

Yes I think it's better to have only one source of truth.

davidbrochart:
> * Do we want to store the cancellation exception in `self._exception`?

On one hand it would be redundant with the `CANCELLED` status, but on the other hand it would carry more information (the reason for cancellation IIRC?).

agronholm:
> > * Do we want to store the cancellation exception in `self._exception`?
> 
> On one hand it would be redundant with the `CANCELLED` status, but on the other hand it would carry more information (the reason for cancellation IIRC?).

And if the cancellation comes from outside, the cancel scope will return `False` for `cancel_called`, leaving us no way to tell if the task was cancelled unless we store the cancellation exception.

davidbrochart:
> * Do we want a dedicated exception in place of `RuntimeError("the task has not returned yet")`?

I think it doesn't hurt to have a dedicated exception, as with `TaskCancelled`.

agronholm:
> > * Do we want a dedicated exception in place of `RuntimeError("the task has not returned yet")`?
> 
> I think it doesn't hurt to have a dedicated exception, as with `TaskCancelled`.

Done.

agronholm:
I removed the `_status` field and stored the cancellation exception now. But I think I need to chain that exception to TaskCancelled now.

agronholm:
Also bikeshedding the exception names: `TaskFailed`, rather than `TaskError` would be a name consistent with the others, yes?

docs/tasks.rst
```diff
@@ -41,6 +41,47 @@ Here's a demonstration::
 .. _Trio: https://trio.readthedocs.io/en/latest/reference-core.html
    #tasks-let-you-do-multiple-things-at-once
 
+Working with task handles
+-------------------------
+
+As an alternative way to spawn tasks, you can use the :meth:`~.TaskGroup.create_task`
+method which will return a :class:`~.TaskHandle`. This handle can be used to:
+
+#. Wait for the task to finish before exiting the task group
```
davidbrochart, line 50: This confuses me a little bit, as it could suggest that if we don't wait for the task to finish (using its handle) the task group could exit.

docs/tasks.rst
```diff
@@ -41,6 +41,47 @@ Here's a demonstration::
...
+#. Retrieve the task's return value or exception
+#. Check the task's status
+
+::
+
+    from anyio import sleep, create_task_group, run, TaskHandle
```
davidbrochart: Just some linting:
```suggestion
    from anyio import TaskHandle, create_task_group, run, sleep
```

docs/tasks.rst
```diff
@@ -41,6 +41,47 @@ Here's a demonstration::
...
+        # Should print:
+        #   2
+        #   Tasks finished: 0, 2, 3, 4
+        print(
+            'Tasks finished:',
+            ', '.join(
+                handle.return_value for handle in handles
+                if handle.status is TaskHandle.Status.FINISHED
+            )
+        )
```
davidbrochart: Maybe not put the comment before the last print statement?
```suggestion
        print(
            'Tasks finished:',
            ', '.join(
                handle.return_value for handle in handles
                if handle.status is TaskHandle.Status.FINISHED
            )
        )
        # Should print:
        #   2
        #   Tasks finished: 0, 2, 3, 4
```

src/anyio/_core/_tasks.py
```diff
@@ -171,3 +192,188 @@ def create_task_group() -> TaskGroup:
...
     """
     return get_async_backend().create_task_group()
+
+
+@final
+class TaskHandle(Generic[T_co]):
+    """
+    Returned from :meth:`TaskGroup.create_task() <.abc.TaskGroup.create_task>`.
+    Can be awaited on to get the return value of the task (or the raised exception).
+    If the task was terminated by a :exc:`BaseException`, :exc:`TaskAborted` will be
```
davidbrochart:
```suggestion
    If the task was terminated by a :exc:`BaseException`, :exc:`TaskFailed` will be
```

src/anyio/_core/_tasks.py
```diff
@@ -171,3 +192,188 @@ def create_task_group() -> TaskGroup:
...
+        .. attribute:: FINISHED
+
+            The task has finished with a return value.
+        .. attribute:: CANCELLING
+
+            The task has been cancelled but has not finished yet.
+        .. attribute:: CANCELLED
+
+            The task was cancelled and has finished since.
+        .. attribute:: ERRORED
```
davidbrochart: To be consistent with `TaskFailed`?
```suggestion
        .. attribute:: FAILED
```

src/anyio/_core/_tasks.py
```diff
@@ -171,3 +192,188 @@ def create_task_group() -> TaskGroup:
...
+    def cancel(self) -> None:
+        """
+        Set the task to a cancelled state.
+
+        This will interrupt any interruptible asynchronous operation, and will cause
+        any further awaits on this task to get immediately cancelled, unless done in
+        a shielded cancel scope.
+
+        If the task has already finished, this method has no effect.
+
```
davidbrochart:
```suggestion
```

src/anyio/_core/_tasks.py
```diff
@@ -171,3 +192,188 @@ def create_task_group() -> TaskGroup:
...
+        return self._coro
+
+    @property
+    def status(self) -> TaskHandle.Status:
+        """
+        The current status of the task.
+
+        Every task starts in the :attr:`~TaskHandle.Status.PENDING` state, and then
+        transitions to one of the other states later. A pending task will transition
+        from :attr:`~TaskHandle.Status.CANCELLING` to one of the final statuses, but no
```
davidbrochart: Typo? I'm a bit confused by the sentence too.
```suggestion
        from :attr:`~TaskHandle.Status.PENDING` to one of the final statuses, but no
```

src/anyio/_core/_tasks.py
```diff
@@ -171,3 +192,188 @@ def create_task_group() -> TaskGroup:
...
+            case TaskHandle.Status.ERRORED:
+                raise TaskFailed("the task raised an exception") from self._exception
+
+    async def wait(self) -> None:
+        """
+        Wait for the task to finish.
+
+        This method will return as soon as the task has finished, no matter how it
+        happened.
+
```
davidbrochart:
```suggestion
```

src/anyio/_core/_tasks.py
agronholm: I was actually being consistent as I recall a style guide recommending the use of blank lines at the end of docstrings, but now I can't find anything of the sort on either PEP 8 or PEP 257.

src/anyio/_core/_tasks.py
agronholm: I think I was distracted while writing this passage. The suggested edit is not sufficient to fix the language here.

agronholm: I rewrote the docstring. It should be fine now.

davidbrochart: LGTM.

docs/tasks.rst
agronholm: I read this comment several times and I cannot comprehend how one would arrive to that conclusion. No task group implementation anywhere works that way.

davidbrochart: Reading it again, I understand now that `create_task` allows to wait for the task to finish before all the tasks in this task group have finished, at which point the task group exits. I agree this is equivalent :+1:

agronholm:
> Should we use `Done` instead of `Finished` everywhere? It seems that asyncio does so.

What do you mean by that? The repr of a `Task` says `finished` when it's done.

davidbrochart:
But: `Task.add_done_callback()`, `Task.done()`...

agronholm:
> But: `Task.add_done_callback()`, `Task.done()`...

We don't support adding callbacks, and `Task.done()` is replaced with the `status` property.

src/anyio/_core/_tasks.py
davidbrochart: Still need to replace `Status.ERRORED` with `Status.FAILED`?

davidbrochart:
Sure, I'm just saying asyncio uses "done", but it's fine.

agronholm:
```pycon
>>> asyncio.run(main())
<Task finished name='Task-2' coro=<sleep() done, defined at /usr/lib64/python3.10/asyncio/tasks.py:593> result=None>
```
This is where I originally got the `FINISHED` state from.

docs/tasks.rst
agronholm: How would you change the description to reduce confusion?

agronholm:
I'd like to get at least two accepting reviews to ensure some consensus. There's one already, so one more is needed.

agronholm:
> Well … I'm +.5 on this, because I really dislike asyncio's calling conventions.
> 
> But I suppose we can (should, IMHO) add a `start_task` method (or whatever) later.

WDYM? We already have a `start()` method, and it will be modified later to optionally return a task handle (in a backwards-compatible API change).

smurfix:
> it will be modified later to optionally return a task handle

Yeah, that was my "or whatever" part.

Sorry if that was too terse / didn't consider past discussion; I'm a bit stressed out at the moment. 😔

docs/tasks.rst
```diff
@@ -41,6 +41,47 @@ Here's a demonstration::
...
+Working with task handles
+-------------------------
+
+As an alternative way to spawn tasks, you can use the :meth:`~.TaskGroup.create_task`
+method which will return a :class:`~.TaskHandle`. This handle can be used to:
+
+#. Wait for the task to finish before exiting the task group
+#. Retrieve the task's return value or exception
+#. Check the task's status
+
```
Graeme22, line 54: Maybe mention cancel here too?

agronholm: Good point. I'll add that.

agronholm:
Thanks, now I can continue to the next steps.

PR #1100, opened 2026-03-28
Added support for custom limiters in file I/O

agronholm:

## Changes

This makes it possible to use custom capacity limiters with the `AsyncFile` and `Path` classes.

## Checklist

features

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/src/anyio/_core/_fileio.py
+++ b/src/anyio/_core/_fileio.py
@@ -74,26 +74,22 @@ class AsyncFile(AsyncResource, Generic[AnyStr]):
     def __init__(
         self, fp: IO[AnyStr], *, limiter: CapacityLimiter | None = None
     ) -> None:
-        self._fp: Any = fp
-        self.limiter = limiter
-
-    def __getattr__(self, name: str) -> object:
-        return getattr(self._fp, name)
-
-    @property
-    def limiter(self) -> CapacityLimiter | None:
-        """The capacity limiter used by this file object."""
-        return self._limiter
-
-    @limiter.setter
-    def limiter(self, limiter: CapacityLimiter | None) -> None:
         if limiter is not None and not isinstance(limiter, CapacityLimiter):
             raise TypeError(
                 f"limiter must be a CapacityLimiter or None, not "
                 f"{limiter.__class__.__name__}"
             )
 
+        self._fp: Any = fp
         self._limiter = limiter
+
+    def __getattr__(self, name: str) -> object:
+        return getattr(self._fp, name)
+
+    @property
+    def limiter(self) -> CapacityLimiter | None:
+        """The capacity limiter used by this file object, if not the global limiter."""
+        return self._limiter
 
     @property
     def wrapped(self) -> IO[AnyStr]:
@@ -355,8 +351,14 @@ class Path:
     def __init__(
         self, *args: str | PathLike[str], limiter: CapacityLimiter | None = None
     ) -> None:
+        if limiter is not None and not isinstance(limiter, CapacityLimiter):
+            raise TypeError(
+                f"limiter must be a CapacityLimiter or None, not "
+                f"{limiter.__class__.__name__}"
+            )
+
         self._path: Final[pathlib.Path] = pathlib.Path(*args)
-        self.limiter = limiter
+        self._limiter = limiter
 
     def __fspath__(self) -> str:
         return self._path.__fspath__()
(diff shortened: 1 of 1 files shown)


src/anyio/_core/_fileio.py
```diff
@@ -516,54 +615,64 @@ def glob(  # type: ignore[misc] # mypy doesn't allow for differing signatures in
...
             target = target._path
 
-        await to_thread.run_sync(os.link, target, self)
+        await to_thread.run_sync(os.link, target, self, limiter=self._limiter)
 
     @classmethod
-    async def home(cls) -> Path:
-        home_path = await to_thread.run_sync(pathlib.Path.home)
-        return cls(home_path)
+    async def home(cls, limiter: CapacityLimiter | None = None) -> Path:
```
graingert:
```suggestion
    async def home(cls, *, limiter: CapacityLimiter | None = None) -> Path:
```

tests/test_fileio.py
```diff
@@ -9,14 +9,20 @@
...
+
+@pytest.fixture(params=[False, True])
+async def limiter(request: FixtureRequest) -> CapacityLimiter | None:
+    return CapacityLimiter(10) if request.param else None
 
 
 class TestAsyncFile:
     @pytest.fixture(scope="class")
-    def testdata(cls) -> bytes:
+    def testdata(self) -> bytes:
```
graingert: If it's scope class don't you get a cls not a self?

agronholm: My bad. I was confused by pycharm's erroneous warning.

agronholm: Fixed.

src/anyio/_core/_fileio.py
```diff
@@ -67,12 +71,30 @@ class AsyncFile(AsyncResource, Generic[AnyStr]):
...
+    def __init__(
+        self, fp: IO[AnyStr], *, limiter: CapacityLimiter | None = None
+    ) -> None:
         self._fp: Any = fp
+        self.limiter = limiter
 
     def __getattr__(self, name: str) -> object:
         return getattr(self._fp, name)
 
+    @property
```
graingert, line 89: I don't think `limiter` should be writable. Swapping it on a live `AsyncFile` or `Path` mid-operation (e.g. between two `to_thread.run_sync` calls in a loop) would be a footgun with no obvious use case.

I'd drop the setter and validate in `__init__` instead:

```python
def __init__(self, fp: IO[AnyStr], *, limiter: CapacityLimiter | None = None) -> None:
    self._fp: Any = fp
    if limiter is not None and not isinstance(limiter, CapacityLimiter):
        raise TypeError(
            f"limiter must be a CapacityLimiter or None, not {limiter.__class__.__name__}"
        )
    self._limiter = limiter

@property
def limiter(self) -> CapacityLimiter | None:
    """The capacity limiter used by this file object."""
    return self._limiter
```

Same applies to `Path`.

graingert: But up to you on this one I'm not too bothered either way

graingert: If the user wants to change the limiter they should close the file and open a new file with a new limiter imho

graingert: CapacityLimiter.total_tokens is mutable so I'd rather users update that than update the limiter

graingert: I also don't see any test mutating the limiter on an open file or Path

agronholm: The reason I made it writable was because derived Paths inherit the limiter from the source. If we make it read-only, the only way to change the limiter is to construct a new Path from the old one with a different `limiter` argument. I don't quite see the problem with changing the `limiter` mid-operation. Would you like to elaborate?

graingert: Does limiter even need to be public? If a user wants to change the total_tokens on the fly they should keep hold of a reference to the limiter they pass in and mutate that

graingert: Fair point on derived paths: exposing `limiter` as a readable property makes sense so users can inspect what was inherited. But I'd still argue it shouldn't be *writable*. If someone wants to repath with a different limiter they can do `Path(old_path, limiter=new_limiter)` explicitly, which makes the intent clear.

The mid-operation concern is: if you do `path.limiter = x` between two `await` points inside e.g. `iterdir()` or `walk()`, the in-flight `_PathIterator` still holds a reference to the old `_limiter` value captured at construction. So the setter gives the illusion of affecting ongoing operations when it doesn't. That's the footgun.

Making it read-only sidesteps all of this cleanly.

But again I'm not overly concerned I imagine anyone mutating limiters can accept the caveat emptor

agronholm: OK, we can make it writable again later if there is a strong use case for it.

agronholm: Updated.

graingert (approved):

agronholm:
Thanks for the review!

PR #1102, opened 2026-03-31
Validate max_bytes >= 1 in ByteReceiveStream.receive() implementations

### LINKED ISSUE #1081: `ByteReceiveStream.receive(max_bytes < 1)` behaves differently between backends (and between stream types)
### Things to check first

### AnyIO version

master (fcf6c5b83c94a2f442c1d1eab5da8b9ee81f070f)

### Python version

CPython 3.13.12

### What happened?

**Summary at the top**: In various implementations of `ByteReceiveStream` that are built-in to AnyIO, `receive_stream.receive` with `max_bytes=negative` and/or `max_bytes=0` is buggy. The way that it is buggy depends on both the stream type and on which backend is being used. These issues are all related in the sense that adding `if max_bytes < 1: raise ValueError` checks would fix all of these. I'm reporting them together because I don't want to spam with many very similar/near-identical reports.

`SocketStream.receive(max_bytes=negative)` doesn't raise on asyncio, but it raises `ValueError` on Trio. I expected it to behave consistently between backends, i.e. I expected it to either `ValueError` on both backends or `checkpoint(); return b""` on both backends. The docs state that implementers of `ByteReceiveStream.receive` should not return `b""`, so it seems like `receive(negative)` should raise `ValueError`.

While writing a reproducer and report for the above `max_bytes < 0` case, I also noticed closely-related bugs in the `SocketStream.receive(max_bytes=0)` case:

*   `SocketStream.receive(max_bytes=0)` doesn't raise on asyncio. As above, I think it should raise `ValueError`.

*   `SocketStream.receive(max_bytes=0)` raises `EndOfStream` on Trio (even though the peer did not close or send EoF). As above, I think that it should raise `ValueError`.

On asyncio, instead of raising `ValueError`, `SocketStream.receive` currently (in both the negative and zero cases)

*   returns data with length `len(buffer_chunk) - abs(max_bytes)` if `abs(max_bytes) < len(buffer_chunk)`

*   else returns `b""`

On asyncio it also waits for data to be received into the buffer, even though it doesn't need any data.

I checked to see what Trio does; `trio.abc.ReceiveStream.receive` raises `ValueError` in both the negative and zero cases: https://github.com/python-trio/trio/blob/v0.33.0/src/trio/_abc.py#L447-L448

I checked the other `ByteReceiveStream.receive` implementations that are built-in to AnyIO to see if they are similarly broken for `max_bytes < 0` and/or `max_bytes == 0`:

*   Subprocess std{out,err} streams:

    *   negative case: `ValueError` on Trio, as expected. No `ValueError` on asyncio; instead the stream is read until EoF and all read bytes are returned.

    *   zero case: `ValueError` on Trio, as expected. No `ValueError` on asyncio; instead we get `EndOfStream` (even though the peer did not close or send EoF).

*   Unix SOCK_STREAM streams:

    *   negative case: works. `ValueError` on Trio and asyncio. It bubbles up from https://github.com/python/cpython/blob/v3.13.12/Modules/socketmodule.c#L3771.

    *   zero case: No `ValueError` on either backend; instead we get a false `EndOfStream`.

        Note: We are calling `socket.socket.recv(0)` here, and it returns `b""`. Our code that converts `b""` to `EndOfStream` is actually correct, because `recv` documents that "A returned empty bytes object indicates that the client has disconnected." However, `socket.socket.recv` appears to violate this documentation when we pass `bufsize=0` to it, returning `b""` instead.

*   `BufferedByteReceiveStream`:

    *   case `receive`:

        *   cases (negative | zero, buffer empty, underlying stream is a `ByteReceiveStream`): Behaves the same as `underlying_stream.receive(negative)`.

        *   cases (negative, buffer not empty) | (negative, buffer empty, underlying stream is an `ObjectReceiveStream[bytes]`): No `ValueError`; instead it

            *   returns data with length `len(buffer) - abs(max_bytes)` if `abs(max_bytes) < len(buffer)`

            *   else returns `b""`

        *   cases (zero, buffer not empty) | (zero, buffer empty, underlying stream is an `ObjectReceiveStream[bytes]`): No `ValueError`; instead it returns `b""`.

    *   case `receive_exactly`:

        *   case (`nbytes` negative, buffer not empty): Behaves the same as case (`receive`, negative, buffer not empty).

        *   cases (`nbytes` negative, buffer empty) | (`nbytes` zero): No `ValueError`; instead it returns `b""`.

    *   case `receive_until`:

        *   case (negative, delimiter in buffer): Behaves the same as case (`receive`, negative, buffer not empty).

        *   case (negative, delimiter not in buffer): `DelimiterNotFound: The delimiter was not found among the first -{abs(max_bytes)} bytes`.

*   `FileReadStream`:

    *   negative case: No `ValueError`. It does whatever `UnderlyingFileType.read(negative)` does, which is usually to read until EoF or to read "as much as possible".

    *   zero case: No `ValueError`; instead we get a false `EndOfStream`.

        Note: We are calling `file.read(0)` here, and it returns `b""`. Compare this to the `socket.socket.recv(0)` note above. Unlike the `socket.socket.recv(0)` docs, the docs here are correct:

        > If 0 bytes are returned, and size was not 0, this indicates end of file.

*   `StapledByteStream`: Behaves the same as underlying `ByteReceiveStream.receive`.

*   `TLSStream`:

    *   negative case: `ValueError`. It bubbles up from https://github.com/python/cpython/blob/v3.13.12/Modules/_ssl.c#L2567.

    *   zero case: No `ValueError`; instead, data is returned.

        (I didn't look into the semantics of how much data is returned, or whether data will be returned every time, or only in some state machine states. However, `SSLObject.read(0)` is definitely not reading until EoF; it is probably just reading one chunk, for some definition of chunk.)

These `ByteReceiveStream` reproducers correspond to the same order of streams above.

```python
class TestTCPStream:
    ...

    @pytest.mark.parametrize("max_bytes", (-1, -3, -4, -5))
    async def test_receive_negative_with_data_to_receive(
        self, max_bytes: int, server_sock: socket.socket, server_addr: tuple[str, int]
    ) -> None:
        async with await connect_tcp(*server_addr) as stream:
            client, _ = server_sock.accept()
            with client:
                client.sendall(b"blah")
                with pytest.raises(ValueError):
                    # On asyncio, this doesn't raise. On Trio, this raises ValueError.
                    result = await stream.receive(max_bytes)
                    # We should probably not get here. But on asyncio we do. It
                    # definitely shouldn't have returned data, though:
                    assert len(result) <= max(max_bytes, 0)
                    # On asyncio, this ^ assertion fails in the -1 and -3 cases.

    async def test_receive_negative_without_data_to_receive(
        self, server_sock: socket.socket, server_addr: tuple[str, int]
    ) -> None:
        async with await connect_tcp(*server_addr) as stream:
            client, _ = server_sock.accept()
            with client:
                with fail_after(0.1):
                    with pytest.raises(ValueError):
                        # On Trio, this raises ValueError. On asyncio, this gets stuck
                        # waiting to receive data even though it doesn't need any data.
                        await stream.receive(-1)

    async def test_receive_zero_with_data_to_receive(
        self, server_sock: socket.socket, server_addr: tuple[str, int]
    ) -> None:
        async with await connect_tcp(*server_addr) as stream:
            client, _ = server_sock.accept()
            with client:
                client.sendall(b"blah")
                with pytest.raises(ValueError):
                    # On asyncio, this doesn't raise ValueError. On Trio, this raises
                    # EndOfStream.
                    result = await stream.receive(0)
                    # We should probably not get here. But on asyncio we do. It
                    # definitely shouldn't have returned data, though:
                    assert len(result) <= 0

    async def test_receive_zero_without_data_to_receive(
        self, server_sock: socket.socket, server_addr: tuple[str, int]
    ) -> None:
        async with await connect_tcp(*server_addr) as stream:
            client, _ = server_sock.accept()
            with client:
                with fail_after(0.1):
                    with pytest.raises(ValueError):
                        # On Trio, this raises EndOfStream. On asyncio, this gets stuck
                        # waiting to receive data even though it doesn't need any data.
                        await stream.receive(0)
```

```python
# test_subprocesses.py

@pytest.mark.parametrize("max_bytes", (-1, -3, -4, -5))
async def test_receive_negative(max_bytes: int) -> None:
        [sys.executable, "-c", "import sys; print('blah', end='')"]
    ) as process:
        with pytest.raises(ValueError):
            # On asyncio, this doesn't raise. On Trio, this raises ValueError.
            result = await process.stdout.receive(-1)
            # We should probably not get here. But on asyncio we do. It
            # definitely shouldn't have returned data, though:
            assert len(result) == 0
            # On asyncio, this ^ assertion fails.


async def test_receive_zero() -> None:
        [sys.executable, "-c", "import sys; print('blah', end='')"]
    ) as process:
        with pytest.raises(ValueError):
            # On Trio, this raises ValueError. On asyncio, this raises EndOfStream.
            result = await process.stdout.receive(0)
            # We should probably not get here. But on asyncio we do. It definitely
            # shouldn't have returned data, though:
            assert len(result) == 0
```

```python
class TestUNIXStream:
    ...

    async def test_receive_zero(
        self, server_sock: socket.socket, socket_path_or_str: Path | str
    ) -> None:
        async with await connect_unix(socket_path_or_str) as stream:
            client, _ = server_sock.accept()
            with client:
                client.sendall(b"blah")
                with pytest.raises(ValueError):
                    # On both backends, this raises EndOfStream.
                    await stream.receive(0)
```

```python
# test_buffered.py

@pytest.mark.parametrize("max_bytes", (-1, -3, -4, -5))
async def test_receive_negative_with_buffer(max_bytes: int) -> None:
    send_stream, receive_stream = create_memory_object_stream[bytes](2)
    async with (
        send_stream,
        BufferedByteReceiveStream(receive_stream) as buffered_stream,
    ):
        await send_stream.send(b"ablah")
        await buffered_stream.receive_exactly(1)
        assert buffered_stream._buffer
        with pytest.raises(ValueError):
            # This doesn't raise.
            result = await buffered_stream.receive(max_bytes)
            # We should probably not get here. But we do. It definitely shouldn't have
            # returned data, though:
            assert len(result) == 0
            # This ^ assertion fails.


@pytest.mark.parametrize("max_bytes", (-1, -3, -4, -5))
async def test_receive_negative_without_buffer_object_receive_stream(
    max_bytes: int,
) -> None:
    send_stream, receive_stream = create_memory_object_stream[bytes](2)
    async with (
        send_stream,
        BufferedByteReceiveStream(receive_stream) as buffered_stream,
    ):
        await send_stream.send(b"blah")
        assert not buffered_stream._buffer
        with pytest.raises(ValueError):
            # This doesn't raise.
            result = await buffered_stream.receive(max_bytes)
            # We should probably not get here. But we do. It definitely shouldn't have
            # returned data, though:
            assert len(result) == 0
            # This ^ assertion fails.


async def test_receive_zero_with_buffer() -> None:
    send_stream, receive_stream = create_memory_object_stream[bytes](2)
    async with (
        send_stream,
        BufferedByteReceiveStream(receive_stream) as buffered_stream,
    ):
        await send_stream.send(b"ablah")
        await buffered_stream.receive_exactly(1)
        assert buffered_stream._buffer
        with pytest.raises(ValueError):
            # This doesn't raise.
            result = await buffered_stream.receive(0)
            # We should probably not get here. But we do. It definitely shouldn't have
            # returned data, though:
            assert len(result) == 0


async def test_receive_zero_without_buffer_object_receive_stream() -> None:
    send_stream, receive_stream = create_memory_object_stream[bytes](2)
    async with (
        send_stream,
        BufferedByteReceiveStream(receive_stream) as buffered_stream,
    ):
        await send_stream.send(b"blah")
        assert not buffered_stream._buffer
        with pytest.raises(ValueError):
            # This doesn't raise.
            result = await buffered_stream.receive(0)
            # We should probably not get here. But we do. It definitely shouldn't have
            # returned data, though:
            assert len(result) == 0


@pytest.mark.parametrize("nbytes", (-1, -3, -4, -5))
async def test_receive_exactly_negative_with_buffer(nbytes: int) -> None:
    send_stream, receive_stream = create_memory_object_stream[bytes](2)
    async with (
        send_stream,
        BufferedByteReceiveStream(receive_stream) as buffered_stream,
    ):
        await send_stream.send(b"ablah")
        await buffered_stream.receive_exactly(1)
        assert buffered_stream._buffer
        with pytest.raises(ValueError):
            # This doesn't raise.
            result = await buffered_stream.receive_exactly(nbytes)
            # We should probably not get here. But we do. It definitely shouldn't have
            # returned data, though:
            assert len(result) == 0
            # This ^ assertion fails.


@pytest.mark.parametrize("nbytes", (-1, -3, -4, -5))
async def test_receive_exactly_negative_without_buffer(nbytes: int) -> None:
    send_stream, receive_stream = create_memory_object_stream[bytes](2)
    async with (
        send_stream,
        BufferedByteReceiveStream(receive_stream) as buffered_stream,
    ):
        await send_stream.send(b"blah")
        assert not buffered_stream._buffer
        with pytest.raises(ValueError):
            # This doesn't raise.
            result = await buffered_stream.receive_exactly(nbytes)
            # We should probably not get here. But we do. It definitely shouldn't have
            # returned data, though:
            assert len(result) == 0


async def test_receive_exactly_zero_with_buffer() -> None:
    send_stream, receive_stream = create_memory_object_stream[bytes](2)
    async with (
        send_stream,
        BufferedByteReceiveStream(receive_stream) as buffered_stream,
    ):
        await send_stream.send(b"ablah")
        await buffered_stream.receive_exactly(1)
        assert buffered_stream._buffer
        with pytest.raises(ValueError):
            # This doesn't raise.
            result = await buffered_stream.receive_exactly(0)
            # We should probably not get here. But we do. It definitely shouldn't have
            # returned data, though:
            assert len(result) == 0


async def test_receive_exactly_zero_without_buffer() -> None:
    send_stream, receive_stream = create_memory_object_stream[bytes](2)
    async with (
        send_stream,
        BufferedByteReceiveStream(receive_stream) as buffered_stream,
    ):
        await send_stream.send(b"blah")
        assert not buffered_stream._buffer
        with pytest.raises(ValueError):
            # This doesn't raise.
            result = await buffered_stream.receive_exactly(0)
            # We should probably not get here. But we do. It definitely shouldn't have
            # returned data, though:
            assert len(result) == 0


@pytest.mark.parametrize("max_bytes", (-1, -3, -4, -5))
async def test_receive_until_negative_with_buffer(max_bytes: int) -> None:
    send_stream, receive_stream = create_memory_object_stream[bytes](2)
    async with (
        send_stream,
        BufferedByteReceiveStream(receive_stream) as buffered_stream,
    ):
        await send_stream.send(b"ablah\n")
        await buffered_stream.receive_exactly(1)
        assert buffered_stream._buffer
        with pytest.raises(ValueError):
            # This doesn't raise.
            result = await buffered_stream.receive_until(b"\n", max_bytes)
            # We should probably not get here. But we do. It definitely shouldn't have
            # returned data, though:
            assert len(result) == 0
            # This ^ assertion fails.


@pytest.mark.parametrize("max_bytes", (-1, -3, -4, -5))
async def test_receive_until_negative_without_buffer(max_bytes: int) -> None:
    send_stream, receive_stream = create_memory_object_stream[bytes](2)
    async with (
        send_stream,
        BufferedByteReceiveStream(receive_stream) as buffered_stream,
    ):
        await send_stream.send(b"blah\n")
        assert not buffered_stream._buffer
        with pytest.raises(ValueError):
            # This raises DelimiterNotFound: The delimiter was not found among the first
            # -{abs(max_bytes)} bytes.
            await buffered_stream.receive_until(b"\n", max_bytes)
```

```python
class TestFileReadStream:
    ...

    async def test_receive_negative(self, file_path: Path) -> None:
        async with await FileReadStream.from_path(file_path) as stream:
            with pytest.raises(ValueError):
                # This doesn't raise.
                result = await stream.receive(-1)
                # We should probably not get here. But we do. It definitely shouldn't
                # have returned data, though:
                assert len(result) == 0

    async def test_receive_zero(self, file_path: Path) -> None:
        async with await FileReadStream.from_path(file_path) as stream:
            with pytest.raises(ValueError):
                # This raises EndOfStream.
                await stream.receive(0)
```

```python
class TestTLSStream:
    ...

    async def test_receive_zero(
        self,
        server_context: ssl.SSLContext,
        client_context: ssl.SSLContext,
    ) -> None:
        sent = threading.Event()

        def serve_sync() -> None:
            conn, addr = server_sock.accept()
            conn.settimeout(1)
            conn.send(b"blah")
            sent.set()
            conn.close()

        with server_context.wrap_socket(
            socket.socket(), server_side=True, suppress_ragged_eofs=False
        ) as server_sock:
            server_sock.settimeout(1)
            server_sock.bind(("127.0.0.1", 0))
            server_sock.listen()
            with ThreadPoolExecutor(1) as executor:
                future = executor.submit(serve_sync)
                try:
                    async with await connect_tcp(*server_sock.getsockname()) as stream:
                        tls_stream = await TLSStream.wrap(
                            stream, hostname="localhost", ssl_context=client_context
                        )
                        await to_thread.run_sync(sent.wait)
                        with pytest.raises(ValueError):
                            # This doesn't raise.
                            result = await tls_stream.receive()
                            # We should probably not get here. But we do. It definitely
                            # shouldn't have returned data, though:
                            assert len(result) == 0
                            # This ^ assertion fails.
                finally:
                    future.result()
```

joaquinhuigomez:

## Changes

Fixes #1081.

Add `if max_bytes < 1: raise ValueError` guards across all `ByteReceiveStream.receive()` implementations and `BufferedByteReceiveStream` helper methods (`receive_exactly`, `receive_until`). This makes behavior consistent between asyncio and trio backends, and across all stream types.

Affected implementations:
- **asyncio backend**: `SocketStream.receive`, `StreamReaderWrapper.receive` (subprocess), `UNIXSocketStream.receive`
- **trio backend**: `ReceiveStreamWrapper.receive` (subprocess), `SocketStream.receive`
- **Streams**: `TLSStream.receive`, `FileReadStream.receive`, `BufferedByteReceiveStream.receive` / `receive_exactly` / `receive_until`
- **ABC**: Updated `ByteReceiveStream.receive` docstring to document the `ValueError` contract

## Checklist

features

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/tests/streams/test_tls.py
+++ b/tests/streams/test_tls.py
@@ -32,6 +32,38 @@ from anyio.streams.tls import TLSAttribute, TLSConnectable, TLSListener, TLSStre
 
 
 class TestTLSStream:
+    @pytest.mark.parametrize("max_bytes", [0, -1, -5])
+    async def test_receive_max_bytes_validation(
+        self,
+        max_bytes: int,
+        server_context: ssl.SSLContext,
+        client_context: ssl.SSLContext,
+    ) -> None:
+        def serve_sync() -> None:
+            conn, addr = server_sock.accept()
+            conn.settimeout(1)
+            conn.send(b"blah")
+            conn.close()
+
+        server_sock = server_context.wrap_socket(
+            socket.socket(), server_side=True, suppress_ragged_eofs=False
+        )
+        server_sock.settimeout(1)
+        server_sock.bind(("127.0.0.1", 0))
+        server_sock.listen()
+        server_thread = Thread(target=serve_sync)
+        server_thread.start()
+
+        async with await connect_tcp(*server_sock.getsockname()) as stream:
+            wrapper = await TLSStream.wrap(
+                stream, hostname="localhost", ssl_context=client_context
+            )
+            with pytest.raises(ValueError, match="max_bytes must be a positive integer"):
+                await wrapper.receive(max_bytes)
+
+        server_thread.join()
+        server_sock.close()
+
     async def test_send_receive(
         self, server_context: ssl.SSLContext, client_context: ssl.SSLContext
     ) -> None:
--- a/tests/streams/test_tls.py
+++ b/tests/streams/test_tls.py
@@ -58,7 +58,9 @@ class TestTLSStream:
             wrapper = await TLSStream.wrap(
                 stream, hostname="localhost", ssl_context=client_context
             )
-            with pytest.raises(ValueError, match="max_bytes must be a positive integer"):
+            with pytest.raises(
+                ValueError, match="max_bytes must be a positive integer"
+            ):
                 await wrapper.receive(max_bytes)
 
         server_thread.join()
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -7,6 +7,11 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
 - Added support for custom capacity limiters in async path and file I/O
   functions and classes
+- Fixed ``ByteReceiveStream.receive()`` accepting ``max_bytes < 1`` with
+  inconsistent behavior across backends and stream types. All implementations
+  now raise ``ValueError`` when ``max_bytes < 1``. Also fixed ``receive_exactly()``
+  and ``receive_until()`` on ``BufferedByteReceiveStream``
+  (`#1081 <https://github.com/agronholm/anyio/issues/1081>`_; PR by @joaquinhuigomez)
 
 **4.13.0**
 
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -1041,6 +1041,9 @@ class StreamReaderWrapper(abc.ByteReceiveStream):
     _stream: asyncio.StreamReader
 
     async def receive(self, max_bytes: int = 65536) -> bytes:
+        if max_bytes < 1:
+            raise ValueError("max_bytes must be a positive integer")
+
         data = await self._stream.read(max_bytes)
         if data:
             return data
@@ -1274,6 +1277,9 @@ class SocketStream(abc.SocketStream):
         return self._transport.get_extra_info("socket")
 
     async def receive(self, max_bytes: int = 65536) -> bytes:
+        if max_bytes < 1:
+            raise ValueError("max_bytes must be a positive integer")
+
         with self._receive_guard:
             if (
                 not self._protocol.read_event.is_set()
@@ -1398,6 +1404,9 @@ class UNIXSocketStream(_RawSocketMixin, abc.UNIXSocketStream):
             self._raw_socket.shutdown(socket.SHUT_WR)
 
     async def receive(self, max_bytes: int = 65536) -> bytes:
+        if max_bytes < 1:
+            raise ValueError("max_bytes must be a positive integer")
+
         loop = get_running_loop()
         await AsyncIOBackend.checkpoint()
         with self._receive_guard:
--- a/src/anyio/_backends/_trio.py
+++ b/src/anyio/_backends/_trio.py
@@ -228,6 +228,9 @@ class ReceiveStreamWrapper(abc.ByteReceiveStream):
     _stream: trio.abc.ReceiveStream
 
     async def receive(self, max_bytes: int | None = None) -> bytes:
+        if max_bytes is not None and max_bytes < 1:
+            raise ValueError("max_bytes must be a positive integer")
+
         try:
             data = await self._stream.receive_some(max_bytes)
         except trio.ClosedResourceError as exc:
@@ -383,6 +386,9 @@ class SocketStream(_TrioSocketMixin, abc.SocketStream):
         self._send_guard = ResourceGuard("writing to")
 
     async def receive(self, max_bytes: int = 65536) -> bytes:
+        if max_bytes < 1:
+            raise ValueError("max_bytes must be a positive integer")
+
         with self._receive_guard:
             try:
                 data = await self._trio_socket.recv(max_bytes)
--- a/src/anyio/abc/_streams.py
+++ b/src/anyio/abc/_streams.py
@@ -140,9 +140,10 @@ class ByteReceiveStream(AsyncResource, TypedAttributeProvider):
         .. note:: Implementers of this interface should not return an empty
             :class:`bytes` object, and users should ignore them.
 
-        :param max_bytes: maximum number of bytes to receive
+        :param max_bytes: maximum number of bytes to receive (must be >= 1)
         :return: the received bytes
         :raises ~anyio.EndOfStream: if this stream has been closed from the other end
+        :raises ValueError: if ``max_bytes`` is less than 1
         """
 
 
--- a/src/anyio/streams/buffered.py
+++ b/src/anyio/streams/buffered.py
@@ -65,6 +65,9 @@ class BufferedByteReceiveStream(ByteReceiveStream):
         self._buffer.extend(data)
 
     async def receive(self, max_bytes: int = 65536) -> bytes:
+        if max_bytes < 1:
+            raise ValueError("max_bytes must be a positive integer")
+
         if self._closed:
             raise ClosedResourceError
 
@@ -95,6 +98,9 @@ class BufferedByteReceiveStream(ByteReceiveStream):
             amount of bytes could be read from the stream
 
         """
+        if nbytes < 1:
+            raise ValueError("nbytes must be a positive integer")
+
         while True:
             remaining = nbytes - len(self._buffer)
             if remaining <= 0:
@@ -126,6 +132,9 @@ class BufferedByteReceiveStream(ByteReceiveStream):
             bytes read up to the maximum allowed
 
         """
+        if max_bytes < 1:
+            raise ValueError("max_bytes must be a positive integer")
+
         delimiter_size = len(delimiter)
         offset = 0
         while True:
--- a/src/anyio/streams/file.py
+++ b/src/anyio/streams/file.py
@@ -79,6 +79,9 @@ class FileReadStream(_BaseFileStream, ByteReceiveStream):
         return cls(file)
 
     async def receive(self, max_bytes: int = 65536) -> bytes:
+        if max_bytes < 1:
+            raise ValueError("max_bytes must be a positive integer")
+
         try:
             data = await to_thread.run_sync(self._file.read, max_bytes)
         except ValueError:
--- a/src/anyio/streams/tls.py
+++ b/src/anyio/streams/tls.py
@@ -236,6 +236,9 @@ class TLSStream(ByteStream):
         await self.transport_stream.aclose()
 
     async def receive(self, max_bytes: int = 65536) -> bytes:
+        if max_bytes < 1:
+            raise ValueError("max_bytes must be a positive integer")
+
         data = await self._call_sslobject_method(self._ssl_object.read, max_bytes)
         if not data:
             raise EndOfStream
--- a/tests/streams/test_buffered.py
+++ b/tests/streams/test_buffered.py
@@ -148,9 +148,7 @@ class TestMaxBytesValidation:
             BufferedByteReceiveStream(receive_stream) as buffered_stream,
         ):
             await send_stream.send(b"blah")
-            with pytest.raises(
-                ValueError, match="nbytes must be a positive integer"
-            ):
+            with pytest.raises(ValueError, match="nbytes must be a positive integer"):
                 await buffered_stream.receive_exactly(nbytes)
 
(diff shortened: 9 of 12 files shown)


tests/streams/test_tls.py
```diff
@@ -32,6 +32,40 @@
...
+            wrapper = await TLSStream.wrap(
+                stream, hostname="localhost", ssl_context=client_context
+            )
+            with pytest.raises(
+                ValueError, match="max_bytes must be a positive integer"
+            ):
+                await wrapper.receive(max_bytes)
+
+        server_thread.join()
+        server_sock.close()
```
gschaffner, line 67: > There's something wrong with the code, with test failures being this consistent.

There's a race between the two threads in this test, because the client thread doesn't wait to receive `b"blah"` before closing the client socket. Removing the `send` call should fix it, I think.

joaquinhuigomez:
Thanks for looking at this. The tests are flagging real issues — closing for now.

PR #1104, opened 2026-04-02
Fix lost type information when using lru_cache with kwargs

Graeme22:
## Changes

Due to incorrect usage of generics in the `lru_cache` decorator, type information was unnecessarily discarded when keyword arguments are passed (eg `ttl`, `always_checkpoint`).

This commit addresses the issue by removing the (unused) generic parameter from the `_LRUCacheWrapper` class. Note that in the class, `T` is only defined in `__call__`, making the parametrization useless. To make things worse, `Any` was being injected as `T`, hiding existing type information.

Example code to demonstrate:

```python
from anyio.functools import lru_cache

@lru_cache(always_checkpoint=True)  # any kwarg will reproduce
async def foo() -> str: ...

async def main():
    reveal_type(await foo())  # type is `Any`
```

With this commit, the revealed type will be correctly inferred as `str`.

## Checklist

features

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -18,6 +18,8 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 - Fixed ``TaskGroup`` raising ``AttributeError`` instead of a clear error when entered
   more than once
   (`#1109 <https://github.com/agronholm/anyio/issues/1109>`_; PR by @bahtya)
+- Fixed lost type information when passing arguments to ``lru_cache``
+  (`#1104 <https://github.com/agronholm/anyio/pull/1104>`_; PR by @Graeme22)
 
 **4.13.0**
 


Graeme22:
> Creating a test that expects a static type checking failure isn't trivial. I have another project that does that, but it involved manually running mypy. But I need some kind of proof that the PR fixes what it says it fixes.

Could you link the other project? I'm having trouble conceptualizing what you're asking for.

agronholm:
> > Creating a test that expects a static type checking failure isn't trivial. I have another project that does that, but it involved manually running mypy. But I need some kind of proof that the PR fixes what it says it fixes.
> 
> Could you link the other project? I'm having trouble conceptualizing what you're asking for.

Here: https://github.com/agronholm/typeguard

The gist is that I need a regression test of some kind to make sure the issue stays fixed and won't break without us noticing when further changes are made. The nature of this particular fix makes that tricky to accomplish, though.

Graeme22:
> The gist is that I need a regression test of some kind to make sure the issue stays fixed and won't break without us noticing when further changes are made. The nature of this particular fix makes that tricky to accomplish, though.

Hm. I mean all I'm doing in this PR is removing generic type information that's not even being used/specified. Are these kind of typing tests (beyond just what's picked up by static typing) used elsewhere in anyio? If not, why is this the first place we need them?

agronholm:
The problem was that if kwargs are added to `@anyio.lru_cache()`, it will erase the return type annotation, right? I would like to add some kind of test that makes sure that the type won't be erased. There are other tests in the test suite purely for the benefit of static type checking, but usually they check that no type checking errors occur. With a return type of `Any`, this gets more complicated as we need to check that an error _does_ occur if the return type annotation is not exactly what we expect.

I'm sorry I'm making this difficult, but I need to explore all avenues to ensure that a problem **stays** fixed.

Graeme22:
> With a return type of `Any`, this gets more complicated as we need to check that an error _does_ occur if the return type annotation is not exactly what we expect.

So `Any` isn't actually used, notice that the places where `Any` is returned there are overloads, so `Any` won't ever actually be inferred. Take this example:
```python
@overload
def foo(bar: str) -> int: ...

@overload
def foo(bar: int) -> str: ...
```

For the actual implementation we could write:
```python
def foo(bar: str | int) -> str | int:
    pass
```

But since the overloads will be used and never the actual function's signature, we could just as well write:
```python
def foo(bar: str | int) -> Any:
    pass
```

agronholm:
I'm not sure I follow. You original post showed a clear failure to narrow down the return type when decorator arguments were used. Did I misunderstand something?

agronholm:
You need to make the import conditional on `sys.version_info`.

Graeme22:
> I'm not sure I follow. You original post showed a clear failure to narrow down the return type when decorator arguments were used. Did I misunderstand something?

I thought you believed `Any` was still being used, when in reality you're talking about making sure the old behavior (which actually _did_ use `Any`) doesn't resurface. I added a simple test and checked that it fails on current master.

tests/test_functools.py
```diff
@@ -26,6 +27,11 @@
 )
 from anyio.lowlevel import checkpoint
 
+if sys.version_info < (3, 13):
+    from typing_extensions import assert_type
+else:
+    from typing import assert_type
```
agronholm: `assert_type()` was introduced in Python 3.11.
```suggestion
if sys.version_info >= (3, 11):
    from typing import assert_type
else:
    from typing_extensions import assert_type
```

agronholm:
Now if you'll just add a changelog entry, this should be fine.

PR #1107, opened 2026-04-06
Feature/async itertools

### LINKED ISSUE #998: Add asynchronous itertools implementation
### Things to check first

### Feature description

This would add an async counterpart for the [itertools](https://docs.python.org/3/library/itertools.html) standard library module.

### Use case

The need to iterate over async iterables is a fairly common one, and this is why [aioitertools](https://github.com/omnilib/aioitertools) exists.

11kkw:
## Changes

Fixes #998

Adds async counterparts to several itertools helpers for working with async iterables.

Documentation and changelog entries will be added after the initial review.

## Checklist

- [x] You've updated the documentation (in `docs/`) – will add after initial review
- [x] You've added a new changelog entry (in `docs/versionhistory.rst`) – will add after initial review


src/anyio/itertools.py
```diff
@@ -0,0 +1,604 @@
+from __future__ import annotations
+
```
agronholm, line 2: This module needs an `__all__` declaration. See the functools module for an equivalent example.

docs/versionhistory.rst
```diff
@@ -5,6 +5,7 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
 **UNRELEASED**
 
+- Added an asynchronous implementation of the ``itertools`` module
```
```suggestion
- Added an asynchronous implementation of the ``itertools`` module
  (`#998 <https://github.com/agronholm/anyio/issues/998>`_; PR by @11kkw)
```

README.rst
```diff
@@ -52,6 +52,7 @@ AnyIO offers the following functionality:
 * Asynchronous file I/O (using worker threads)
 * Signal handling
 * Asynchronous version of the functools_ module
+* Asynchronous version of the itertools_ module
```
agronholm:
```suggestion
* Asynchronous versions of the itertools_ and functools_ modules
```

agronholm:
It keeps timing out but only in this branch. I suspect there's an issue with one of your new tests.

11kkw:
> It keeps timing out but only in this branch. I suspect there's an issue with one of your new tests.

Thanks for checking. I'll take a look at the new tests in this branch and see which one might be causing the timeout.

11kkw:
> It keeps timing out but only in this branch. I suspect there's an issue with one of your new tests.

I consolidated repeated test cases in test_itertools.py to reduce the test count and runtime while keeping the coverage unchanged.

I'm not entirely sure if this is the right approach for this repository, so I'd appreciate your feedback.

agronholm:
I'll try to get around to reviewing this soon. It will be in the next minor release.

docs/why.rst
```diff
@@ -547,3 +547,9 @@ Asynchronous functools
 
 The :mod:`functools` module does not support coroutine functions, so AnyIO offers its
 own version that does. See :ref:`async-functools` for the list of implemented functions.
+
+Asynchronous itertools
+----------------------
+
+The :mod:`itertools` module does not support asynchronous iterables, so AnyIO offers its
+own version that does. See :ref:`async-itertools` for the list of implemented functions.
```
agronholm: These could be merged together.
```suggestion
Asynchronous functools and itertools
------------------------------------

The :mod:`functools` and :mod:`itertools` modules do not support coroutine functions,
so AnyIO offers its own versions that do. See :ref:`async-functools` and
:ref:`async-itertools`, respectively, for the list of implemented functions.
```

README.rst
```diff
@@ -51,7 +51,7 @@ AnyIO offers the following functionality:
 * Subinterpreter support for code parallelization (on Python 3.13 and later)
 * Asynchronous file I/O (using worker threads)
 * Signal handling
-* Asynchronous version of the functools_ module
+* Asynchronous versions of the itertools_ and functools_ modules
```
agronholm: This is consistent with the docs.
```suggestion
* Asynchronous versions of the functools_ and itertools_ modules
```

src/anyio/itertools.py
```diff
@@ -0,0 +1,629 @@
...
+@dataclass(eq=False)
+class _IterableAsyncIterator(AsyncIterator[T]):
+    iterator: Iterator[T]
+
+    async def __anext__(self) -> T:
+        try:
+            return next(self.iterator)
+        except StopIteration:
+            raise StopAsyncIteration from None
```
agronholm, line 57: This needs cancellation checking and a yield point. Otherwise iterating an indefinite iterator will hog the event loop all to itself.

src/anyio/itertools.py
```diff
@@ -0,0 +1,629 @@
...
+    function: Callable[[T, T], Awaitable[T]] = _operator_add,
+    *,
+    initial: T | None = None,
+) -> AsyncIterator[T]:
+    iterator = _iterate(iterable)
+    if initial is None:
+        try:
+            total = await iterator.__anext__()
+        except StopAsyncIteration:
+            await checkpoint()
```
agronholm, line 148: What's the checkpoint doing here?

agronholm (changes requested):
The code needs improvements to its cancellation and yielding behavior. In a nutshell, before a mutating operation is done, `await checkpoint_if_cancelled()` should be called. Afterwards, if no other yieldpoint has been hit, `await cancel_shielded_checkpoint()` should be called to ensure that the task yields control at least once. Calling `checkpoint()` essentially does both at once which is not ideal. LMK if you need further help with that.

src/anyio/itertools.py
agronholm: Any function that returns an async generator should be annotated as such, so that we can type-safely use them with `async with aclosing()`.

src/anyio/itertools.py
```diff
@@ -477,28 +495,8 @@ async def permutations(
...
-                cycles[i] = n - i
-            else:
-                j = cycles[i]
-                indices[i], indices[-j] = indices[-j], indices[i]
-                yield tuple(pool[index] for index in indices[:r])
-                break
-        else:
-            return
+    async for permutation in _iterate(itertools.permutations(pool, r)):
+        yield permutation
```
agronholm, line 500: Couldn't you just return the async iterator directly here?

11kkw: Do you mean refactoring this into a regular function that returns a lazily initialized async iterator?

agronholm: Ah, sorry, I didn't realize it had other yield points. Maybe not a good idea for this function at least.

src/anyio/itertools.py
```diff
@@ -0,0 +1,625 @@
...
+    async def fill(self, link: _TeeLink[T]) -> bool:
+        if link.filled:
+            return False
+
+        async with self.lock:
+            if link.filled:
+                return True
+
+            try:
+                link.value = await self.iterator.__anext__()
```
agronholm: Perhaps we should start using `anext()`? It's available on all supported Python versions.

11kkw:
I'm repeatedly encountering the following error in the workflow: Error: The action 'Test with pytest' has timed out after 5 minutes. When running the tests locally, none of them seem particularly slow, so I'm not sure what is causing the timeout. If you have a moment to take a look or have any suggestions, I would really appreciate it.

agronholm:
We could look at which tests did complete successfully, and figure out which ones were next to find clues. Also, there are some plug-ins that enforce timeouts which could be useful.

agronholm:
If you add `pytest-timeout` as a test dependency and `--timeout=30` into the pytest args, that should give us an idea which tests are causing trouble. It's very likely to be caused by one or more existing tests.

11kkw:
Thanks for the suggestion. 
I tried adding `pytest-timeout` and running the suite with `--timeout=30`, but **none of the tests actually triggered the timeout**.

I also measured `tests/test_itertools.py` separately in my local environment using the same CI-style command:

`PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 coverage run -m pytest -v tests/test_itertools.py`

Locally it completed in **8.64s (`396 passed`)**. I also checked per-test timings with 
`pytest --durations=0 tests/test_itertools.py`, and the slowest cases were only around **`0.04s`**.

For comparison, the full CI-like test command takes about **`245s` locally on my machine**, while `tests/test_itertools.py` by itself takes about **`9s` clock**. This makes me think that **this file is unlikely to be the main cause of a 5-minute timeout**.

At this point I'm not sure where the slowdown is coming from. Since `pytest-timeout` didn't identify any individual test exceeding 30 seconds, I suspect **this may be a cumulative slowdown rather than a single pathological test**.

agronholm:
That's a really good point which I didn't actually consider! Seems like we're reaching the 5 minute mark naturally in our expanding test suite. It would make sense, then, to extend the time allowance. That said, permanently adding `pytest-timeout` wouldn't be such a bad idea either. But I can do that outside of this PR.

I'll consider your research as proof that your tests aren't causing trouble on their own.

agronholm:
I doubled the timeout value which should eliminate the problem unless tests are actually hanging.

agronholm:
Yup, in the last run, the slowest test job took over 5 minutes to complete.

agronholm (approved):
I haven't found any new issues, and the AI review came up clean too. Thanks!

PR #1112, opened 2026-04-10
asyncio: skip done tasks in _deliver_cancellation to prevent CPU spin

### LINKED ISSUE #1111: 100% CPU spin in asyncio _deliver_cancellation due to missing task.done() check
### Things to check first

### AnyIO version

master

### Python version

python 3.13

### What happened?

There is a bug in the asyncio backend where `_deliver_cancellation` can enter an infinite loop, causing 100% CPU usage. 

The root cause is that `_deliver_cancellation` lacks a `task.done()` check. If a completed task remains in `CancelScope._tasks` for any reason, it triggers an infinite spin.

The Spin Mechanism

1. A done task is left in `CancelScope._tasks`.
3. `_deliver_cancellation` iterates over it and calls `task.cancel()` (a no-op for completed tasks in asyncio).
4. Because the task is already done, the internal state (`_must_cancel`) never resolves.
5. The method sets `should_retry = True` and reschedules itself via `call_soon` forever, locking the event loop.

While leaving a completed task in `_tasks` usually implies a user misuse, we think a 100% CPU lockup is a mode that should be prevented. 

Proposed Fix

Skip completed tasks in _deliver_cancellation:

```python
# anyio/_backends/_asyncio.py -> CancelScope._deliver_cancellation
    for task in self._tasks:
        if task.done():
            continue
        should_retry = True
        ...
```

We can submit a PR with this fix and a test case if the maintainers agree with this approach.

```python

import asyncio
import gc
from contextlib import AsyncExitStack

import anyio


_deliver_call_count: int = 0
_original_deliver_cancellation = CancelScope._deliver_cancellation


def _counting_deliver_cancellation(
    self: CancelScope,
    origin: CancelScope,
) -> bool:
    global _deliver_call_count
    _deliver_call_count += 1
    return _original_deliver_cancellation(self, origin)


CancelScope._deliver_cancellation = _counting_deliver_cancellation  # type: ignore[assignment]


@asynccontextmanager
async def open_connection() -> AsyncGenerator[str, None]:
    """Open a connection backed by an anyio TaskGroup."""

        async def _reader() -> None:
            try:
                await asyncio.Event().wait()
                pass

        tg.start_soon(_reader)
        try:
            yield "connection"
        finally:
            tg.cancel_scope.cancel()


async def create_client() -> str:
    """Open a connection and perform a handshake.

    If the handshake fails, the ``AsyncExitStack`` is never closed and
    the ``open_connection`` generator is abandoned.
    """
    stack = AsyncExitStack()
    await stack.enter_async_context(open_connection())

    # Handshake fails — e.g. server returns 401 Unauthorized.
    # stack.aclose() is never called.
    raise RuntimeError("handshake failed")


_SPIN_THRESHOLD = 50


async def main() -> None:
    global _deliver_call_count

    async def _worker() -> None:
        try:
            await create_client()
        except Exception:
            pass  # error logged; move on

    task = asyncio.create_task(_worker())
    await task

    # task is done.  Force GC — the abandoned generator's finalizer
    # will run in a *new* asyncio.Task (cross-task CancelScope exit).
    del task
    gc.collect()

    # Give the finalizer task a chance to run and the initial (legitimate)
    # cancellation deliveries to complete.
    await asyncio.sleep(0.1)

    # Reset counter and measure over a short idle window.
    _deliver_call_count = 0
    await asyncio.sleep(0.5)
    count = _deliver_call_count

    print(f"_deliver_cancellation called {count} times during 0.5 s idle window")
    if count > _SPIN_THRESHOLD:
        print("BUG: _deliver_cancellation is spinning!")
        raise SystemExit(1)
    else:
        print("OK: _deliver_cancellation is not spinning.")


if __name__ == "__main__":
    asyncio.run(main())
```

joaquinhuigomez:
`_deliver_cancellation` sets `should_retry=True` for every task in `self._tasks`, including completed tasks. A done task cannot be cancelled, but the retry flag still triggers a `call_soon` callback, which re-iterates, finds the same done task, and schedules again — infinite loop at 100% CPU.

Add a `task.done()` guard at the top of the loop to skip completed tasks entirely.

Fixes #1111

### DIFF
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -583,6 +583,9 @@ class CancelScope(BaseCancelScope):
         should_retry = False
         current = current_task()
         for task in self._tasks:
+            if task.done():
+                continue
+
             should_retry = True
             if task._must_cancel:  # type: ignore[attr-defined]
                 continue


agronholm:
Did you not read the note? Is this an AI generated PR?

joaquinhuigomez:
You're right — closing. Apologies for the noise.

PR #1114, opened 2026-04-12
Updated the cancel message in check_cancelled()

### LINKED ISSUE #1113: CancelledError leaks from from_thread.check_cancelled on Asyncio backend
### Things to check first

### AnyIO version

master (after #1092)

### Python version

3.10

### What happened?

In a slightly contrived setting, the `CancelledError` from `from_thread.check_cancelled` escapes the relevant cancel scope on asyncio backend.

This test passes on trio but not asyncio:

```python    
import threading
import anyio
import pytest

pytestmark = pytest.mark.anyio


async def test_anyio_cancel_thread():
    tev = threading.Event()
    ev = anyio.Event()

    async def cancel_detector() -> None:
        try:
            await ev.wait()
            tev.set()
            # we do NOT re-raise the cancellation so that the task group
            # does not swallow our retval. If a Cancelled is to propagate,
            # it should come out of to_thread.run_sync

    def get_cancelled():
        tev.wait(10)  # fake interruptible work for test
        if tev.is_set:
            from_thread.check_cancelled()
        # in application, return some computed/retrieved value
        # return precious_retval

    async def do_cancel():
        await anyio.sleep(0.01)
        g.cancel_scope.cancel()

    async with anyio.create_task_group() as g:
        g.start_soon(cancel_detector)
        g.start_soon(do_cancel)
        precious_retval = await anyio.to_thread.run_sync(get_cancelled)
        ev.set()

### PR DESCRIPTION
**Pull request by agronholm**

## Changes

The wrong cancel message caused is_anyio_cancellation() to falsely return `False` for cancellation exceptions originating from check_cancelled().

Fixes #1113.

## Checklist

features

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -7,6 +7,9 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
 - Added support for custom capacity limiters in async path and file I/O
   functions and classes
+- Fixed cancellation exception escaping a cancel scope when triggered via
+  ``check_cancelled()`` in a worker thread
+  (`#1113 <https://github.com/agronholm/anyio/issues/1113>`_)
 
 **4.13.0**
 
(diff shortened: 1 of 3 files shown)


agronholm:
@richardsheridan LMK if this fix is acceptable.

richardsheridan:
Yes, the simplified test makes sense and this PR also fixes the original application.

gschaffner (approved):

PR #1115, opened 2026-04-13
Improve missing backend error

EmmanuelNiyonshuti:

## Changes

Fixes #. 

Right now when you have a backend like `trio` in `BACKENDS` but it is not installed, the error says:
```
LookupError: No such backend: trio
```

This is a bit confusing because `trio` is a valid backend, it is just not installed.  I changed it to tell the user what to do instead:

```
LookupError: Backend 'trio' is not available. Install it with: pip install anyio[trio]
```

The fix uses the already existing `BACKENDS` constant to check if the backend is known but just not installed, and gives a different message for that case.

## Checklist

features

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -9,7 +9,8 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
   functions and classes
 
 **4.13.0**
-
+- Improved the error message when a known backend is not installed to suggest the install command
+  (`<https://github.com/agronholm/anyio/pull/1115>`_; PR by @EmmanuelNiyonshuti)
 - Dropped support for Python 3.9
 - Added a ``ttl`` parameter to the ``anyio.functools.lru_cache`` wrapper
   (`#1073 <https://github.com/agronholm/anyio/pull/1073>`_; PR by @Graeme22)
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -7,11 +7,11 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
 - Added support for custom capacity limiters in async path and file I/O
   functions and classes
+- Improved the error message when a known backend is not installed to suggest the install command
+  (`#1115 <https://github.com/agronholm/anyio/pull/1115>`_; PR by @EmmanuelNiyonshuti)
 
 **4.13.0**
 
-- Improved the error message when a known backend is not installed to suggest the install command
-  (`#1115 <https://github.com/agronholm/anyio/pull/1115>`_; PR by @EmmanuelNiyonshuti)
 - Dropped support for Python 3.9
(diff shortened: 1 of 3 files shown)


docs/versionhistory.rst
```diff
@@ -9,7 +9,8 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
   functions and classes
 
 **4.13.0**
-
+- Improved the error message when a known backend is not installed to suggest the install command
+  (`<https://github.com/agronholm/anyio/pull/1115>`_; PR by @EmmanuelNiyonshuti)
```
agronholm: 4.13.0 has shipped already.
```suggestion
- Added support for custom capacity limiters in async path and file I/O
functions and classes
- Improved the error message when a known backend is not installed to suggest the install command
  (`<https://github.com/agronholm/anyio/pull/1115>`_; PR by @EmmanuelNiyonshuti)

**4.13.0**
```

src/anyio/_core/_eventloop.py
```diff
@@ -64,6 +64,11 @@ def run(
     try:
         async_backend = get_async_backend(backend)
     except ImportError as exc:
+        if backend in BACKENDS:
+            raise LookupError(
+                f"backend '{backend}' is not available. "
+                f"Install it with: pip install anyio[{backend}]"
+            ) from exc
```
agronholm, line 71: Add a blank line after a control block ends.
```suggestion
            ) from exc

```

tests/test_eventloop.py
```diff
@@ -51,6 +51,21 @@ async def async_add(x: int, y: int) -> int:
     assert result == 3
 
 
+async def main() -> None:
+    pass
+
+
+def test_run_unknown_backend() -> None:
+    with pytest.raises(LookupError, match="No such backend: asncio"):
+        run(main, backend="asncio")
```
agronholm: `asncio` looks too much like a typo. I suggest changing it to something more obvious.
```suggestion
    with pytest.raises(LookupError, match="No such backend: somebackend"):
        run(main, backend="somebackend")
```

tests/test_eventloop.py
```diff
@@ -51,6 +51,21 @@ async def async_add(x: int, y: int) -> int:
     assert result == 3
 
 
+async def main() -> None:
+    pass
```
agronholm: Tests should be self-contained whenever possible. Move this to both the tests that use it.

agronholm (changes requested):
Looks good. I just need a number of minor changes first.

tests/test_eventloop.py
EmmanuelNiyonshuti: Thanks for the feedback! let me do that.

EmmanuelNiyonshuti: done! I updated the tests as requested.  looks like the coveralls upload is not working tho.

agronholm (approved):
LGTM now. Thanks!

PR #1119, opened 2026-04-17
Add local_port parameter to connect_tcp

### LINKED ISSUE #1067: Support ``IPSockAddrType`` as ``local_host`` in highlevel ``connect_tcp``
### Things to check first

### Feature description

Hello hello!

While implementing a controlplane service that needs to query the cloud-init process for instance metadata on a VPS, I [came across the following requirement](https://www.scaleway.com/en/docs/instances/how-to/use-cloud-init/):

> Alternatively, you can send a raw HTTP request to http://169.254.42.42/user_data (or http://[fd00:42::42]/user_data for IPv6 connectivity). The request must originate from a port below 1024

This put me down a pretty long rabbit hole, eventually ending with discovering a discrepancy between the high-level ``connect_tcp`` and the individual event loop backend implementations. The former accepts only a [``local_host: IPAddressType | None = None,`` parameter](https://github.com/agronholm/anyio/blob/master/src/anyio/_core/_sockets.py#L135), while the latter accepts a [``local_address: IPSockAddrType | None = None`` parameter](https://github.com/agronholm/anyio/blob/a2a07c7c7af5d0f154b770f280d8772299a20a38/src/anyio/abc/_eventloop.py#L249).

My feature request would be to add support for the explicit ``local_address`` as an additional overload on the high-level ``connect_tcp``, which would allow clients to specify an explicit local port, as you can with eg cURL (ex ``curl --local-port 1-1023 http://169.254.42.42/user_data?format=json``).

### Use case

Security-based restrictions on origin port for requests when querying instance metadata from cloud-init

nullwiz:
## Changes

Fixes #1067
Adds a `local_port` parameter to the high-level `connect_tcp()` function. 

## Checklist

features

I feel the change is trivial! Ill see if I can contribute to other issues.

### DIFF
--- a/tests/test_sockets.py
+++ b/tests/test_sockets.py
@@ -397,6 +397,18 @@ class TestTCPStream:
         server_sock.close()
         assert client_addr[0] == expected_client_addr
 
+    async def test_connect_tcp_with_local_port(
+        self,
+        server_sock: socket.socket,
+        server_addr: tuple[str, int],
+        family: AnyIPAddressFamily,
+    ) -> None:
+        local_addr = "127.0.0.1" if family == AddressFamily.AF_INET else "::1"
+        async with await connect_tcp(
+            *server_addr, local_host=local_addr, local_port=1234
+        ) as stream:
+            assert stream.extra(SocketAttribute.local_port) == 1234
+
     @pytest.mark.skipif(
         sys.implementation.name == "pypy",
         reason=(
--- a/tests/test_sockets.py
+++ b/tests/test_sockets.py
@@ -402,12 +402,13 @@ class TestTCPStream:
         server_sock: socket.socket,
         server_addr: tuple[str, int],
         family: AnyIPAddressFamily,
+        free_tcp_port: int,
     ) -> None:
         local_addr = "127.0.0.1" if family == AddressFamily.AF_INET else "::1"
         async with await connect_tcp(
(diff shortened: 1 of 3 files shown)


tests/test_sockets.py
```diff
@@ -397,6 +397,18 @@ def serve() -> None:
...
+        self,
+        server_sock: socket.socket,
+        server_addr: tuple[str, int],
+        family: AnyIPAddressFamily,
+    ) -> None:
+        local_addr = "127.0.0.1" if family == AddressFamily.AF_INET else "::1"
+        async with await connect_tcp(
+            *server_addr, local_host=local_addr, local_port=1234
+        ) as stream:
+            assert stream.extra(SocketAttribute.local_port) == 1234
```
agronholm: Using a fixed port number prevents the parallel running of the test suite against multiple Python versions.
```suggestion
        family: AnyIPAddressFamily,
        free_tcp_port: int,
    ) -> None:
        local_addr = "127.0.0.1" if family == AddressFamily.AF_INET else "::1"
        async with await connect_tcp(
            *server_addr, local_host=local_addr, local_port=free_tcp_port
        ) as stream:
            assert stream.extra(SocketAttribute.local_port) == free_tcp_port
```

agronholm (changes requested):
Just one change I'd like. Otherwise this looks solid!

agronholm (approved):
Thanks!

PR #1126, opened 2026-04-24
Use eager_start=False on Python 3.14+ instead of __code__ hacks

### LINKED ISSUE #1125: use create_task=False on 3.14+
### Things to check first

### Feature description

python 3.14 eager_start=False was added to loop.create_task and asyncio.create_task we should use it rather than the __code__ hacks

### Use case

This would simplify the codebase in 3.14+

graingert:
On Python 3.14+, `asyncio.create_task()` accepts `eager_start=False` directly, avoiding the need to inspect the task factory's `__code__` and `__closure__`.

## Changes

Fixes #1125

## Checklist

features

these instructions.

### Updating the changelog

```
```


graingert:
curses to uvloop opened https://github.com/MagicStack/uvloop/issues/746

agronholm:
This didn't actually reduce complexity. Only when we can drop earlier versions would this improve things.

PR #1127, opened 2026-04-25
Minor code improvements

albertedwardson:

## Changes

Fixes #. 

- replaced `list.pop(0)` to `deque.popleft` in `create_tcp_listener` (O(n) -> O(1))
- replaced set comp to a dict comp  in `create_tcp_listener` since dicts preserve the order of insertion (in cpython impl)
- replaced tuple lookups to set lookups
- replaced `str.split` to `str.partition`
- removed creation of new list for callbacks in `find_root_task`

## Checklist

features

these instructions.

### Updating the changelog

```
```


agronholm:
I don't think any of these have any measurable effect on performance, nor will they improve readability or reliability. Do you disagree?

albertedwardson:
yes, my changes are unimportant, feel free to close

I just opened a PR with changes I personally like while just studying the library itself

PR #1128, opened 2026-04-26
Added guidelines for AI agents

agronholm:

## Changes

This adds the de facto standard `AGENTS.md` file for helping coding agents make better pull requests.

## Checklist

features

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,59 @@
+# AGENTS.md — Guidelines for AI Agents working on anyio
+
+## Project Overview
+
+AnyIO is a high-level asynchronous concurrency and networking library for Python that works on top of both **asyncio** and **Trio**. Source code lives in `src/`, tests in `tests/`.
+
+## Building & Testing
+
+- **Python**: Requires Python >= 3.10.
+- **Install the project and test dependencies**: `pip install --group test -e .`.
+- **Run tests**: `pytest` (runs from the `tests/` directory automatically via `pyproject.toml` config). Tests run against all supported backends by default.
+- **Linting**: The project uses **pre-commit** for linting, formatting and static type checking. Run `pre-commit run -a` to check everything.
+- **All source files must include** `from __future__ import annotations` as the first import (enforced by Ruff/isort).
+
+## Code Style
+
+- Follow the existing code style — the project enforces it via Ruff (see `[tool.ruff]` in `pyproject.toml`).
+- Import order is managed by Ruff's isort integration; always add `from __future__ import annotations` as the first import.
+- Target Python 3.10+ idioms (use `X | Y` union syntax in annotations, etc.).
+- Always add a blank line after a control block ends and there is more code to follow. Do not add blank lines between related parts of the same control block (e.g. `if...elif...else`).
+
+## Pull Request Guidelines
+
+Every pull request **must** follow the PR template in `.github/pull_request_template.md`. Do **not** erase or replace the template contents — PRs that do so will be closed without review.
+
+A properly filled-out PR contains:
+
+### 1. Changes section
+- Reference the related issue number, if applicable (e.g., `Fixes #123.`).
+- Provide a short description of what the PR changes and why.
+
+### 2. Checklist
+Complete the checklist where applicable:
+- **Tests** — Add or update tests in `tests/` that would fail without the patch.
+- **Documentation** — Update docs in `docs/` if behavior changes or new features are introduced.
+- **Changelog** — Add a new entry in `docs/versionhistory.rst`.
+
+### 3. Changelog entry format
+If there are no entries after the last release, use `**UNRELEASED**` as the version heading. An entry should look like:
+
+```
+- Fix big bad boo-boo in task groups
+  (`#123 <https://github.com/agronholm/anyio/issues/123>`_; PR by @yourgithubaccount)
+```
+
+If there is no linked issue, link to the pull request itself instead (update the changelog after the PR is created to get the PR number).
+
+Trivial changes (typo fixes, code reformatting) may skip the checklist items.
+
+## Repository Layout
+
+| Path | Description |
+|---|---|
+| `src/anyio/` | Main library source |
+| `src/anyio/_backends/` | asyncio and Trio backend implementations |
+| `tests/` | Test suite (pytest + anyio plugin) |
+| `docs/` | Sphinx documentation |
+| `docs/versionhistory.rst` | Changelog |
+| `.github/pull_request_template.md` | PR template (must be respected) |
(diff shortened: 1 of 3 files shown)


AGENTS.md
```diff
@@ -0,0 +1,59 @@
+# AGENTS.md — Guidelines for AI Agents working on anyio
+
+## Project Overview
+
+AnyIO is a high-level asynchronous concurrency and networking library for Python that works on top of both **asyncio** and **Trio**. Source code lives in `src/`, tests in `tests/`.
+
+## Building & Testing
+
+- **Python**: Requires Python >= 3.10.
```
webknjaz: I wonder if this could be replaced with instructions to infer this from `pyproject.toml`

AGENTS.md
```diff
@@ -0,0 +1,59 @@
...
+- **Install the project and test dependencies**: `pip install --group test -e .`.
+- **Run tests**: `pytest` (runs from the `tests/` directory automatically via `pyproject.toml` config). Tests run against all supported backends by default.
+- **Linting**: The project uses **pre-commit** for linting, formatting and static type checking. Run `pre-commit run -a` to check everything.
+- **All source files must include** `from __future__ import annotations` as the first import (enforced by Ruff/isort).
+
+## Code Style
+
+- Follow the existing code style — the project enforces it via Ruff (see `[tool.ruff]` in `pyproject.toml`).
+- Import order is managed by Ruff's isort integration; always add `from __future__ import annotations` as the first import.
+- Target Python 3.10+ idioms (use `X | Y` union syntax in annotations, etc.).
```
webknjaz: (same as above)

webknjaz (commented):
How about also including links to a good PR example and a bad PR example?

AGENTS.md
agronholm: It's hard to know what clues agents will pick up on, but I did my best in the updated version.

AGENTS.md
agronholm: Done

webknjaz (approved):
This is probably a good start.

Here's another example from a big project you might want to lurk in: https://github.com/astral-sh/ruff/blob/main/AGENTS.md

I additionally recommend adding an adjusted copy of https://github.com/ansible/ansible/blame/devel/CLAUDE.md

agronholm:
It's weird how every agent under the sun supports `AGENTS.md` – except Claude Code. I understand the `@AGENTS.md` line there, but are the other two actually necessary?

webknjaz:
> It's weird how every agent under the sun supports `AGENTS.md` – except Claude Code.

I think some of the others (like Copilot, IIRC) might also ignore it by default.
Note that the standard website overstates the number of the repos actually using it: https://github.com/agentsmd/agents.md/pull/30 — it originally used to match any file with "agent" in its path, including nested and existing for decades. They improved the search query since, but it still mismatches what's stated.

Note that the website is mostly talking about integrations with specific editors, not agents.

Claude mentions it (https://code.claude.com/docs/en/claude_code_docs_map.md / https://code.claude.com/docs/en/memory.md) but I guess there are reasons behind not loading it blindly.

> I understand the `@AGENTS.md` line there, but are the other two actually necessary?

`CLAUDE.local.md` is nice to have in `.gitignore` so when you have multiple Git worktrees in multiple separate folders on disk (like when working with several PRs at a time). It can be used to record temporary per-PR context in addition to the project-global instructions.

As for the user-global file, it may be nice to have for groups of projects. It'd be something personalized for the contributor, while global for multiple Git worktrees but not committed into the project itself.

agronholm:
The Copilot coding agent is explicitly listed as supporting `AGENTS.md`. What I don't understand is why I would need to add those paths which should be supported by Claude Code by default anyway. According to Claude Code [docs](https://code.claude.com/docs/en/memory#how-claude-md-files-load), it loads all the discovered `CLAUDE.md` files, rather than having them override each other, so explicitly including those other paths makes no sense to me. In fact, they have an example that only reads in `AGENTS.md`, and that's it.

agronholm:
I think we're good to go. LMK if you see more room for improvement.

webknjaz:
> The Copilot coding agent is explicitly listed as supporting `AGENTS.md`.

I haven't tried it yet, but I've heard people complaining that Copilot doesn't pick it up just last week. Maybe, it doesn't work w/ some of the interfaces like cloud but would work locally via CLI. This is something that needs to be checked in practice, I think.

> What I don't understand is why I would need to add those paths which should be supported by Claude Code by default anyway. According to Claude Code [docs](https://code.claude.com/docs/en/memory#how-claude-md-files-load), it loads all the discovered `CLAUDE.md` files, rather than having them override each other, so explicitly including those other paths makes no sense to me. In fact, they have an example that only reads in `AGENTS.md`, and that's it.

Interesting.. There was time where it was just `CLAUDE.local.md`, then they deprecated it replacing w/ `CLAUDE.md` (and the local one would need to be loaded). Perhaps, they walked back on the deprecation since I last checked that doc as I don't see that info anymore.


---

Anyway, I think this is good to merge as is. Any improvements could be done in follow-ups.

webknjaz:
By the way, in case you didn't see, you can grab a Claude Code subscription for half a year @ https://claude.com/contact-sales/claude-for-oss.

m-u-xyz:
I just checked with Claude Code v2.1.109. It tries to find `CLAUDE.md` in multiple locations but ignores `AGENTS.md`.

The workaround, of course, is to `echo "read AGENTS.md if it exists" > ~/.claude/CLAUDE.md`.

webknjaz:
Been doing some extra googling on the topic: https://www.humanlayer.dev/blog/writing-a-good-claude-md

agronholm:
> I just checked with Claude Code v2.1.109. It tries to find `CLAUDE.md` in multiple locations but ignores `AGENTS.md`.
> 
> The workaround, of course, is to `echo "read AGENTS.md if it exists" > ~/.claude/CLAUDE.md`.

The `@` include syntax is officially supported. I added that file before you commented.

webknjaz:
> > I just checked with Claude Code v2.1.109. It tries to find `CLAUDE.md` in multiple locations but ignores `AGENTS.md`.
> > The workaround, of course, is to `echo "read AGENTS.md if it exists" > ~/.claude/CLAUDE.md`.
> 
> The `@` include syntax is officially supported. I added that file before you commented.

I suspect that his suggestion would work for all Claude Code sessions on one's computer and would apply to any project that doesn't have a `CLAUDE.md` in it. Though I'd be careful with the security aspect of loading things that aren't guaranteed to be trusted universally.

m-u-xyz:
> I suspect that his suggestion would work for all Claude Code sessions on one's computer

It does. (I checked.) Though on second thought I'd add "if no `CLAUDE.md` is present" to it.

> Though I'd be careful with the security aspect of loading things that aren't guaranteed to be trusted universally.

That also applies to `CLAUDE.md` files in other repos, or `AGENTS.md` read by other orchestrators, so doesn't introduce any new risks IMHO.

PR #1131, opened 2026-04-29
Use Self to preserve subclass types in anyio.Path path-returning methods

### LINKED ISSUE #1130: `Self` as `Path` methods' return type annotation & corresponding implementation change
### Things to check first

### Feature description

The type annotations for Python's own `pathlib.Path` specify a return type of [`typing.Self`](https://docs.python.org/3/library/typing.html#typing.Self) for methods that return another path object, such as [`__truediv__`](https://github.com/python/typeshed/blob/68517355a3269be407bde20fea8fd66af2dc4241/stdlib/pathlib/__init__.pyi#L89) (`path / ...`):

```python
class PurePath(PathLike[str]):
    ...
    def __truediv__(self, key: StrPath) -> Self: ...
```

`anyio.Path`, on the other hand, has such method annotated as returning `anyio.Path` specifically:

https://github.com/agronholm/anyio/blob/fb4bb27279057d4cfa68707afc1ab4efe38bb13d/src/anyio/_core/_fileio.py#L403-L404

That is technically accurate because the method implementation really does construct and return an `anyio.Path` specifically, but there are some benefits to returning `Self` instead, as will be shown below. So I hope it's possible to change these method type signatures to return `Self`, ideally while changing the implementations to match ([`pathlib` uses `type(self)`](https://github.com/python/cpython/blob/005555a3f0ae20ee8154eb4ee172e1e355144c8c/Lib/pathlib/__init__.py#L159-L164) to construct a new path instance instead of hardcoding specific path classes).

### Use case

Annotating return types as `Self` instead of the class's specific type allows users to define their own subtypes and have them "just work".

#### Static (type checking only) subtypes

For example, if we're only interested in having type membership checked statically but not at runtime, we could use [`typing.NewType`](https://docs.python.org/3/library/typing.html#newtype):

```python
import anyio
from typing import NewType 

AbsolutePath = NewType("AbsolutePath", anyio.Path)

async def make_absolute(path: anyio.Path) -> AbsolutePath:
  return AbsolutePath(await path.absolute())
```

While this is currently possible, it's useless because as soon as you e.g. try to append to the path, type checkers conclude that you get back an `anyio.Path` again instead of another instance of the subtype:

```python
# Run e.g. Mypy over this:
reveal_type(AbsolutePath(anyio.Path("/")) / "foo")  # anyio.Path instead of AbsolutePath
```

#### Runtime subtypes / subclasses

Another option, if someone needs to distinguish between `anyio.Path` and the subtype _at runtime_ as well, would of course be to subclass `anyio.Path`:

```python
class AbsolutePath(anyio.Path):
  pass
```

This runs into the same issue, this time not only at the static type checking level but also at runtime:

```python
print(type(AbsolutePath("/") / "foo"))  # anyio.Path instead of AbsolutePath
```

#### Comparison to Python's pathlib

For Python's own `pathlib`, both of these work as you'd expect and you always get back an instance of your subtype.

EmmanuelNiyonshuti:

## Changes

Fixes #1130

`anyio.Path` methods returning `anyio.Path` break subclass type preservation when the class is subclassed.
Updated to return `Self` and use type(self) for construction so subclass types are preserved correctly.

This aligns the behavior with `pathlib.Path`, where path-returning methods preserve Self instead of the concrete class type.

## Checklist

features

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -5,6 +5,8 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
 **UNRELEASED**
 
+- Improved ``anyio.Path.__truediv__`` to preserve subclass types by returning ``Self``
+  (`#1130 <https://github.com/agronholm/anyio/issues/1130>`_; PR by @EmmanuelNiyonshuti)
 - Added the ``local_port`` parameter to :func:`connect_tcp` to allow binding to a
   specific local port before connecting
   (`#1067 <https://github.com/agronholm/anyio/issues/1067>`_; PR by @nullwiz)
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -16,7 +16,7 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
   install command
   (`#1115 <https://github.com/agronholm/anyio/pull/1115>`_; PR by @EmmanuelNiyonshuti)
 - Improved ``anyio.Path`` to preserve subclass types by returning ``Self`` in methods
-  that return path objects.
+  that return path objects
   (`#1130 <https://github.com/agronholm/anyio/issues/1130>`_; PR by @EmmanuelNiyonshuti)
 - Fixed cancellation exception escaping a cancel scope when triggered via
   ``check_cancelled()`` in a worker thread
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -5,8 +5,6 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
 **UNRELEASED**
 
-- Improved ``anyio.Path.__truediv__`` to preserve subclass types by returning ``Self``
-  (`#1130 <https://github.com/agronholm/anyio/issues/1130>`_; PR by @EmmanuelNiyonshuti)
 - Added the ``local_port`` parameter to :func:`connect_tcp` to allow binding to a
   specific local port before connecting
   (`#1067 <https://github.com/agronholm/anyio/issues/1067>`_; PR by @nullwiz)
@@ -17,6 +15,8 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 - Improved the error message when a known backend is not installed to suggest the
   install command
   (`#1115 <https://github.com/agronholm/anyio/pull/1115>`_; PR by @EmmanuelNiyonshuti)
+- Improved ``anyio.Path.__truediv__`` to preserve subclass types by returning ``Self``
+  (`#1130 <https://github.com/agronholm/anyio/issues/1130>`_; PR by @EmmanuelNiyonshuti)
 - Fixed cancellation exception escaping a cancel scope when triggered via
   ``check_cancelled()`` in a worker thread
   (`#1113 <https://github.com/agronholm/anyio/issues/1113>`_)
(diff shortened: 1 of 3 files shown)


agronholm:
The changelog entry seems to be missing.

docs/versionhistory.rst
agronholm: New features should be first in the changelog, then changes and then fixes.
```suggestion
- Added the ``local_port`` parameter to :func:`connect_tcp` to allow binding to a
specific local port before connecting
(`#1067 <https://github.com/agronholm/anyio/issues/1067>`_; PR by @nullwiz)
- Added support for custom capacity limiters in async path and file I/O
functions and classes
- Added the ``create_task()`` task group method for easier asyncio migration
and to allow retrieving task return values more easily
- Improved the error message when a known backend is not installed to suggest the
install command
(`#1115 <https://github.com/agronholm/anyio/pull/1115>`_; PR by @EmmanuelNiyonshuti)
- Improved ``anyio.Path.__truediv__`` to preserve subclass types by returning ``Self``
  (`#1130 <https://github.com/agronholm/anyio/issues/1130>`_; PR by @EmmanuelNiyonshuti)
```

agronholm (changes requested):
If this change is done, it should be done consistently. Right now you're special casing `__truediv__`, leaving `__rtruediv__` (among others) untouched. Why?

EmmanuelNiyonshuti:
> If this change is done, it should be done consistently. Right now you're special casing `__truediv__`, leaving `__rtruediv__` (among others) untouched. Why?

I focused on `__truediv__` first and didn’t actually think of the reverse case yet. There are many things in Path I am still trying to process. good point, I’ll update the reverse too and double-check. let me know if I miss anything.

agronholm:
What about all the other methods annotated to return `Path`?

EmmanuelNiyonshuti:
> What about all the other methods annotated to return `Path`?

I’m not fully sure about the intended scope. It seems like the same issue could apply to all methods returning Path for subclassing consistency, BUT normal anyio.Path usage works fine. Maybe this should be applied broadly, or only in specific cases like path construction?

agronholm:
> > What about all the other methods annotated to return `Path`?
> 
> I’m not fully sure about the intended scope. It seems like the same issue could apply to all methods returning Path for subclassing consistency, BUT normal anyio.Path usage works fine. Maybe this should be applied broadly, or only in specific cases like path construction?

I'm not sure what the difference is. All methods returning `Path` should be made to return `Self`, unless you can find some reason why this would not be appropriate.

EmmanuelNiyonshuti:
I think I covered most of the Path-returning methods now. I didn’t add more tests since I think it would basically repeat the same subclass-preservation check for each path returning method and be redundant.

agronholm:
> I think I covered most of the Path-returning methods now. I didn’t add more tests since I think it would basically repeat the same subclass-preservation check for each path returning method and be redundant.

Fair enough.

docs/versionhistory.rst
```diff
@@ -15,6 +15,8 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 - Improved the error message when a known backend is not installed to suggest the
   install command
   (`#1115 <https://github.com/agronholm/anyio/pull/1115>`_; PR by @EmmanuelNiyonshuti)
+- Improved ``anyio.Path.__truediv__`` to preserve subclass types by returning ``Self``
+  (`#1130 <https://github.com/agronholm/anyio/issues/1130>`_; PR by @EmmanuelNiyonshuti)
```
agronholm: This needs an update now.

agronholm (changes requested):
Almost there now!

agronholm (approved):
Thanks!

PR #1133, opened 2026-05-05
Fixed SocketListener.from_socket() for AF_UNIX listening sockets

### LINKED ISSUE #1132: `SocketListener.from_socket()` is broken for AF_UNIX sockets on both backends
### Things to check first

### AnyIO version

4.13.0 (also reproduces on `master` at `f821c37 `)

### Python version

3.13.13 (CPython, macOS 26.4.1 / Darwin 25.4.0)

### What happened?

`SocketListener.from_socket()` returns a `TCPSocketListener` when given an `AF_UNIX` listening socket. The first `accept()` then fails with `OSError: [Errno 102] Operation not supported on socket` because `TCPSocketListener.accept()` calls `setsockopt(IPPROTO_TCP, TCP_NODELAY, 1)` on the accepted UDS client.

Both backends are affected: `wrap_listener_socket` returns `TCPSocketListener` regardless of `sock.family`.

Expected: `SocketListener.from_socket(unix_sock)` returns a `UNIXSocketListener` and `accept()` succeeds.

Observed output from the reproducer:

```
listener type: TCPSocketListener
accept FAILED: [Errno 102] Operation not supported on socket
```

```python
import asyncio
import os
import socket
import tempfile

from anyio.abc import SocketListener


async def main() -> None:
    path = os.path.join(tempfile.mkdtemp(), "sock")

    sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
    sock.bind(path)
    sock.listen()

    listener = await SocketListener.from_socket(sock)
    print("listener type:", type(listener).__name__)

    client = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
    client.setblocking(False)
    try:
        client.connect(path)
    except BlockingIOError:
        pass

    try:
        await listener.accept()
        print("accept OK")
    except OSError as e:
        print(f"accept FAILED: {e}")


asyncio.run(main())
```

To reproduce on the trio backend, replace the `asyncio` import and the final line with:

```python
import trio

trio.run(main)
```

kudato:
## Changes

Fixes #1132.

`SocketListener.from_socket()` unconditionally constructed a `TCPSocketListener`
in both backends, so wrapping an `AF_UNIX` listening socket failed on the first
`accept()` call with `ENOTSUP` from `setsockopt(IPPROTO_TCP, TCP_NODELAY)`. This
dispatches on `sock.family` in `wrap_listener_socket()` and returns
`UNIXSocketListener` for `AF_UNIX` sockets — the same class that
`create_unix_listener()` already uses.

## Checklist

- [ ] You've updated the documentation (in `docs/`) — bug fix only, behavior aligns with existing docs

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -29,6 +29,9 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 - Fixed test resumption after ``KeyboardInterrupt`` in async generator fixtures on the
   asyncio backend
   (`#1060 <https://github.com/agronholm/anyio/issues/1060>`_; PR by @EmmanuelNiyonshuti)
+- Fixed ``SocketListener.from_socket()`` returning a TCP listener for ``AF_UNIX``
+  listening sockets, causing ``accept()`` to fail with ``ENOTSUP``
+  (`#1132 <https://github.com/agronholm/anyio/issues/1132>`_; PR by @kudato)
 
 **4.13.0**
 
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -2928,7 +2928,7 @@ class AsyncIOBackend(AsyncBackend):
 
     @classmethod
     async def wrap_listener_socket(cls, sock: socket.socket) -> SocketListener:
-        if sock.family == socket.AF_UNIX:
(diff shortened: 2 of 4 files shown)


agronholm (approved):
Thanks.

PR #1136, opened 2026-05-07
Widen Path.write_bytes data param to ReadableBuffer

### LINKED ISSUE #1135: widen Path.write_bytes typing
### Things to check first

### AnyIO version

4.12

### Python version

3.12

### What happened?

With mypy v2, we got flagged for passing a `bytesarray` into `Path.write_bytes`

It's currently typed as:

```
async def write_bytes(self, data: bytes) -> int:
```

`pathlib.Path` types this as:

```
def write_bytes(self, data: ReadableBuffer) -> int: ...
```

I guess all it needs to be is a buffer type per https://peps.python.org/pep-0688/

I haven't checked for other areas that may be accidentally constraining types now that mypy doesn't promote these types by default.

https://mypy.readthedocs.io/en/latest/changelog.html#enable-strict-bytes-by-default
https://mypy.readthedocs.io/en/latest/command_line.html#cmdoption-mypy-no-strict-bytes

```python
import anyio

async def fn() -> None:
    x = anyio.Path("test.bytes")
    view = memoryview(b"")
    await x.write_bytes(view)

anyio.run(fn)
```

SAY-5:
## Changes

Fixes #1135.

Widens the type of the `data` parameter on `anyio.Path.write_bytes()` from `bytes` to `ReadableBuffer`, matching the signature of `pathlib.Path.write_bytes`. mypy v2's `--strict-bytes` (now default) flagged passing `bytearray` / `memoryview` as a type error even though the underlying call accepts any buffer.

## Checklist

- [x] You've updated the documentation (in `docs/`), in case of behavior changes or new features

### DIFF
--- a/tests/test_fileio.py
+++ b/tests/test_fileio.py
@@ -854,18 +854,20 @@ class TestPath:
     def test_with_suffix(self) -> None:
         assert Path("/xyz/foo.txt.gz").with_suffix(".zip").name == "foo.txt.zip"
 
-    async def test_write_bytes(self, tmp_path: pathlib.Path) -> None:
+    @pytest.mark.parametrize(
+        "data",
+        [
+            pytest.param(b"bibbitibobbitiboo", id="bytes"),
+            pytest.param(bytearray(b"bibbitibobbitiboo"), id="bytearray"),
+            pytest.param(memoryview(b"bibbitibobbitiboo"), id="memoryview"),
+        ],
+    )
+    async def test_write_bytes(
+        self, tmp_path: pathlib.Path, data: bytes | bytearray | memoryview
+    ) -> None:
         path = tmp_path / "testfile"
-        await Path(path).write_bytes(b"bibbitibobbitiboo")
+        await Path(path).write_bytes(data)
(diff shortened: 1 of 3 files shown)


tests/test_fileio.py
```diff
@@ -859,6 +859,14 @@ async def test_write_bytes(self, tmp_path: pathlib.Path) -> None:
         await Path(path).write_bytes(b"bibbitibobbitiboo")
         assert path.read_bytes() == b"bibbitibobbitiboo"
 
+    async def test_write_bytes_buffer(self, tmp_path: pathlib.Path) -> None:
+        path = tmp_path / "testfile"
+        await Path(path).write_bytes(bytearray(b"bibbitibobbitiboo"))
+        assert path.read_bytes() == b"bibbitibobbitiboo"
+
+        await Path(path).write_bytes(memoryview(b"second"))
+        assert path.read_bytes() == b"second"
```
agronholm: The bytearray/memoryview could be refactored into test parameters.

agronholm (changes requested):

agronholm (approved):

PR #1145, opened 2026-05-11
Fix asyncio lock waiter deadlocks after cancellation

rasmusfaber:

## Changes

Cancelled asyncio lock/semaphore waiters can be observed by `release()` before their `acquire()` cleanup has removed them from the wait queue. In that window, `release()` skipped the cancelled waiter but left it queued, allowing a later live waiter to become stranded with no future release to wake it.

This PR prunes cancelled waiters during asyncio `Lock` and `Semaphore` release, while preserving the existing cancellation-after-handoff behavior. It also adds regression coverage for the interleaving where a cancelled head waiter would otherwise orphan a later waiter.

## Checklist

features

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -1880,12 +1880,12 @@ class Lock(BaseLock):
         try:
             await fut
         except CancelledError:
-            try:
-                self._waiters.remove(item)
-            except ValueError:
-                pass
-
-            if self._owner_task is task:
+            if fut.cancelled():
+                try:
+                    self._waiters.remove(item)
+                except ValueError:
+                    pass
+            else:
                 self.release()
 
             raise
@@ -1973,11 +1973,13 @@ class Semaphore(BaseSemaphore):
         try:
             await fut
         except CancelledError:
-            try:
-                self._waiters.remove(fut)
-            except ValueError:
-                if not fut.cancelled():
-                    self.release()
+            if fut.cancelled():
+                try:
+                    self._waiters.remove(fut)
+                except ValueError:
+                    pass
+            else:
+                self.release()
 
             raise
 
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -1845,7 +1845,11 @@ class Lock(BaseLock):
         try:
             await fut
         except CancelledError:
-            self._waiters.remove(item)
+            try:
+                self._waiters.remove(item)
+            except ValueError:
+                pass
+
             if self._owner_task is task:
                 self.release()
 
@@ -1871,11 +1875,18 @@ class Lock(BaseLock):
         if self._owner_task != current_task():
             raise RuntimeError("The current task is not holding this lock")
 
-        for task, fut in self._waiters:
-            if not fut.cancelled():
-                self._owner_task = task
-                fut.set_result(None)
-                return
+        # A cancelled waiter that already received ownership removes itself from
+        # _waiters before calling release(); any cancelled waiter still queued here
+        # was cancelled before being woken, so drop it.
+        while self._waiters:
+            task, fut = self._waiters[0]
+            if fut.cancelled():
+                self._waiters.popleft()
+                continue
+
+            self._owner_task = task
+            fut.set_result(None)
+            return
 
         self._owner_task = None
 
@@ -1931,7 +1942,8 @@ class Semaphore(BaseSemaphore):
             try:
                 self._waiters.remove(fut)
             except ValueError:
-                self.release()
+                if not fut.cancelled():
+                    self.release()
 
             raise
 
@@ -1945,11 +1957,15 @@ class Semaphore(BaseSemaphore):
         if self._max_value is not None and self._value == self._max_value:
             raise ValueError("semaphore released too many times")
 
-        for fut in self._waiters:
-            if not fut.cancelled():
-                fut.set_result(None)
-                self._waiters.remove(fut)
-                return
+        while self._waiters:
+            fut = self._waiters[0]
+            if fut.cancelled():
+                self._waiters.popleft()
+                continue
+
+            fut.set_result(None)
+            self._waiters.popleft()
+            return
 
         self._value += 1
 
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -1890,8 +1890,6 @@ class Lock(BaseLock):
 
             raise
 
-        self._waiters.remove(item)
-
     def acquire_nowait(self) -> None:
         task = cast(asyncio.Task, current_task())
         if self._owner_task is None and not self._waiters:
@@ -1914,9 +1912,8 @@ class Lock(BaseLock):
         # _waiters before calling release(); any cancelled waiter still queued here
         # was cancelled before being woken, so drop it.
         while self._waiters:
-            task, fut = self._waiters[0]
+            task, fut = self._waiters.popleft()
             if fut.cancelled():
-                self._waiters.popleft()
                 continue
 
             self._owner_task = task
@@ -1995,13 +1992,11 @@ class Semaphore(BaseSemaphore):
             raise ValueError("semaphore released too many times")
 
         while self._waiters:
-            fut = self._waiters[0]
+            fut = self._waiters.popleft()
             if fut.cancelled():
-                self._waiters.popleft()
                 continue
 
             fut.set_result(None)
-            self._waiters.popleft()
             return
 
         self._value += 1
--- a/tests/test_synchronization.py
+++ b/tests/test_synchronization.py
@@ -13,6 +13,7 @@ from anyio import (
     Event,
     Lock,
     Semaphore,
+    TaskCancelled,
     WouldBlock,
     create_task_group,
     fail_after,
@@ -213,6 +214,47 @@ class TestLock:
         assert statistics.owner is None
         assert statistics.tasks_waiting == 0
         lock.acquire_nowait()
+
+    async def test_cancelled_head_waiter_does_not_orphan_later_waiter(
+        self,
+    ) -> None:
+        lock = Lock()
+        lock.acquire_nowait()
+        gate = Event()
+        last_task_acquired = False
+
+        async def acquire_after_gate() -> None:
+            nonlocal last_task_acquired
+            await gate.wait()
+            await lock.acquire()
+            try:
+                last_task_acquired = True
+            finally:
+                lock.release()
+
+        async with create_task_group() as tg:
+            task1 = tg.create_task(lock.acquire())
+            task2 = tg.create_task(lock.acquire())
+            last_task = tg.create_task(acquire_after_gate())
+            await wait_all_tasks_blocked()
+            assert lock.statistics().tasks_waiting == 2
+
+            gate.set()
+            task1.cancel()
+            task2.cancel()
+            lock.release()
+            with fail_after(1):
+                await last_task
+
+            with pytest.raises(TaskCancelled):
+                await task1
+
+            with pytest.raises(TaskCancelled):
+                await task2
+
+        assert last_task_acquired
+        assert not lock.locked()
+        assert lock.statistics().tasks_waiting == 0
 
     def test_instantiate_outside_event_loop(
         self, anyio_backend_name: str, anyio_backend_options: dict[str, Any]
@@ -660,6 +702,63 @@ class TestSemaphore:
         assert semaphore.value == 1
         assert semaphore.statistics().tasks_waiting == 0
         semaphore.acquire_nowait()
+
+    async def test_cancelled_head_waiter_does_not_orphan_later_waiter(
+        self,
+    ) -> None:
+        semaphore = Semaphore(0, max_value=1)
+        gate = Event()
+        last_task_acquired = False
+
+        async def acquire_after_gate() -> None:
+            nonlocal last_task_acquired
+            await gate.wait()
+            await semaphore.acquire()
+            try:
+                last_task_acquired = True
+            finally:
+                semaphore.release()
+
+        async with create_task_group() as tg:
+            task1 = tg.create_task(semaphore.acquire())
+            task2 = tg.create_task(semaphore.acquire())
+            last_task = tg.create_task(acquire_after_gate())
+            await wait_all_tasks_blocked()
+            assert semaphore.statistics().tasks_waiting == 2
+
+            gate.set()
+            task1.cancel()
+            task2.cancel()
+            semaphore.release()
+            with fail_after(1):
+                await last_task
+
+            with pytest.raises(TaskCancelled):
+                await task1
+
+            with pytest.raises(TaskCancelled):
+                await task2
+
+        assert last_task_acquired
+        assert semaphore.value == 1
+        assert semaphore.statistics().tasks_waiting == 0
+
+    async def test_cancelled_head_waiter_does_not_overrelease(self) -> None:
+        semaphore = Semaphore(0)
+
+        async with create_task_group() as tg:
+            task = tg.create_task(semaphore.acquire())
+            await wait_all_tasks_blocked()
+            assert semaphore.statistics().tasks_waiting == 1
+
+            task.cancel()
+            semaphore.release()
+
+            with pytest.raises(TaskCancelled):
+                await task
+
+        assert semaphore.value == 1
+        assert semaphore.statistics().tasks_waiting == 0
 
     def test_instantiate_outside_event_loop(
         self, anyio_backend_name: str, anyio_backend_options: dict[str, Any]
--- a/tests/test_synchronization.py
+++ b/tests/test_synchronization.py
@@ -13,7 +13,6 @@ from anyio import (
     Event,
     Lock,
     Semaphore,
-    TaskCancelled,
     WouldBlock,
     create_task_group,
     fail_after,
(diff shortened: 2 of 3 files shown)


agronholm:
Are these super complicated tests really required to reproduce the issue?

agronholm:
Hm, I'm starting to see the logic behind the tests. Could they not use AnyIO APIs though, and also run on Trio?

tests/test_synchronization.py
```diff
@@ -214,6 +214,47 @@ async def test_cancel_after_release(self) -> None:
...
+        await asyncio.wait_for(last_task, 1)
+        with pytest.raises(asyncio.CancelledError):
+            await task1
+
+        with pytest.raises(asyncio.CancelledError):
+            await task2
+
+        assert last_task_acquired
+        assert not lock.locked()
+        assert lock.statistics().tasks_waiting == 0
```
agronholm: Converted to AnyIO APIs and enabled Trio.
```suggestion
    async def test_cancelled_head_waiter_does_not_orphan_later_waiter(
        self,
    ) -> None:
        lock = Lock()
        lock.acquire_nowait()
        gate = Event()
        last_task_acquired = False

        async def acquire_after_gate() -> None:
            nonlocal last_task_acquired
            await gate.wait()
            await lock.acquire()
            try:
                last_task_acquired = True
            finally:
                lock.release()

        async with create_task_group() as tg:
            task1 = tg.create_task(lock.acquire())
            task2 = tg.create_task(lock.acquire())
            last_task = tg.create_task(acquire_after_gate())
            await wait_all_tasks_blocked()
            assert lock.statistics().tasks_waiting == 2

            gate.set()
            task1.cancel()
            task2.cancel()
            lock.release()
            with fail_after(1):
                await last_task

            with pytest.raises(TaskCancelled):
                await task1

            with pytest.raises(TaskCancelled):
                await task2

        assert last_task_acquired
        assert not lock.locked()
        assert lock.statistics().tasks_waiting == 0
```

agronholm: Needs the `TaskCancelled` import too.

rasmusfaber: Done. Also updated the third test.

agronholm:
Please don't force-push changes. That way I can't see what you've changed.

agronholm:
Thanks, I'll review when I get the time.

rasmusfaber:
> Are these super complicated tests really required to reproduce the issue?

It is the minimal tests I could come up with that reproduces the race: a cancelled head waiter must still be queued when release runs, and a later waiter must queue before the cancelled task's cleanup removes itself.

We could drop `test_cancelled_head_waiter_does_not_overrelease()`. It just guards against an incomplete semaphore fix.

rasmusfaber:
> Please don't force-push changes. That way I can't see what you've changed.

Ah, sorry. I will stop doing that.

agronholm:
I'm currently spending some time trying to understand the conditions under which this problem can occur. My own attempts to reproduce the problem have all failed. I can move forward with this PR once I fully understand what's going on.

x42005e1f:
> Are these super complicated tests really required to reproduce the issue?

I agree; the tests do seem overly complicated. In fact, the issue can be reproduced with just two tasks:

```python
#!/usr/bin/env python3

import anyio


async def main():
    lock = anyio.Lock()
    lock_acquire = lock.acquire
    lock_release = lock.release
    await lock_acquire()

        task1 = tg.create_task(lock_acquire())

        task2 = tg.create_task(lock_acquire())
        task1.cancel()
        lock_release()

        with anyio.fail_after(1):
            await task2


if __name__ == "__main__":
    anyio.run(main)
```

I tested `asyncio.Lock`, `anyio.Lock`, and `aiologic.Lock`, and only `anyio.Lock` proved to be affected. As I see it, the problem is that the asyncio future object is cancelled immediately, but the task will not resume until its callback is invoked:

```python
#!/usr/bin/env python3

import asyncio


async def main():
    future = asyncio.get_running_loop().create_future()

    async def work():
        try:
            await future
            print("task: cancelled")

    task = asyncio.create_task(work())
    await asyncio.sleep(0)  # switch to the task

    task.cancel()
    print(f"future: {'cancelled' if future.cancelled() else 'pending'}")

    await task


if __name__ == "__main__":
    asyncio.run(main())
```

```lang-None
future: cancelled
task: cancelled
```


As a result:

1. The second task is created before `task1.cancel()` is called, so the second acquire will precede the finalization of the cancelled first task (callbacks are invoked by the asyncio event loop in FIFO order).
2. At the moment `lock_release()` is called, the second task has not yet started execution, and the future object of the first task has already been cancelled, so `self._owner_task = None`, but the queue still contains the future object of the cancelled task.
3. The second task enters `lock_acquire()` and sees that the `self.owner_task is None and not self.waiters` condition (fast path) is false (the queue is not empty!), so it joins the wait queue.
4. The first task resumes and sees that it has been cancelled, but the `self.owner_task is task` condition is false, so it does nothing except remove its future object from the queue. No `self.release()` is called.
5. The second task remains in the queue forever.

---

The solution in this PR is similar to how my primitives work, except that my `popleft()` is unconditional (and there are other differences with state passing, since I have to support threads). Is there a strong reason to leave the queue untouched when ownership is successfully transferred? If we make it unconditional, we can eliminate the second `self._waiters.remove(item)` in the `acquire()` method.

---

Note that asyncio takes a different approach and [modifies the fast-path condition](https://github.com/python/cpython/blob/1097b22fd92afa553d738983e078687441a1d7c4/Lib/asyncio/locks.py#L98-L99). This is also a possible solution, but it seems far less optimal to me (it is effectively O(n) just to check for the possibility of acquiring without waiting!).

However, the real solution lies not in that, but in the fact that [a cancelled task always wakes up the first one in the queue if the lock has not been acquired](https://github.com/python/cpython/blob/1097b22fd92afa553d738983e078687441a1d7c4/Lib/asyncio/locks.py#L119-L120). In theory, we could repeat that behavior by moving the main part of the `release()` method into a separate private method and adding the `elif self._owner_task is None` branch that calls it. But I still think that scanning the entire queue every time to find an uncancelled task is not a very good idea from a performance standpoint. If we have n tasks and they are all cancelled, the total number of iterations will be O(n²).

---

...the same example, but with just one task:

```python
#!/usr/bin/env python3

import anyio


async def main():
    lock = anyio.Lock()
    lock_acquire = lock.acquire
    lock_release = lock.release

        await lock_acquire()

        task = tg.create_task(lock_acquire())

        task.cancel()
        lock_release()

        with anyio.fail_after(1):
            await lock_acquire()


if __name__ == "__main__":
    anyio.run(main)
```

The explanation is the same as well.

src/anyio/_backends/_asyncio.py
```diff
@@ -1880,7 +1880,11 @@ async def acquire(self) -> None:
         try:
             await fut
         except CancelledError:
-            self._waiters.remove(item)
+            try:
+                self._waiters.remove(item)
+            except ValueError:
+                pass
```
x42005e1f: This would also result in O(n²) iterations, since `self._waiters.remove(item)` takes O(n) time for a missing item. Hypothetically, we could mark visited items in the `release()` method and only perform the removal when there is no such marker, but I am not sure how to do this efficiently in terms of memory for asyncio futures.

It would likely be worth changing the items (queue tokens) slightly so that they include one mutable value (for example, replacing a two-element tuple with a three-element list).

```python
>>> from sys import getsizeof
>>> getsizeof((1, 2))  # immutable
64
>>> getsizeof([1, 2, 3])  # mutable
88
>>> class LockWaiter:
...     __slots__ = ("task", "future", "removed")
>>> getsizeof(LockWaiter())  # also mutable
56  # without slots: 328 == getsizeof(LockWaiter()) + getsizeof(LockWaiter().__dict__)
```

x42005e1f:
Perhaps understanding the conditions under which the problem occurs will help in understanding it.

For the issue to be reproduced, four conditions must be met:

1. The lock must be acquired.
2. The wait queue must contain only items of cancelled tasks, with at least one such item present. If there is even one uncancelled task or the queue is empty, this condition is not met.
3. The lock must be released before the last cancelled task is resumed. That is, the queue must not be empty when the lock is released.
4. There must be at least one attempt to acquire the lock before the last cancelled task is resumed, and it must not precede the release. That is, the queue must not be empty when the lock is acquired.

agronholm:
It took me a while, but I now understand that it's here where things go wrong:
```py
        if self._owner_task is None and not self._waiters:
```
What is there is no current owner but there are waiters, all of which have been cancelled? The latest task would be put in a waiting state, waiting for a release that never comes, as the cancelled waiters only do that if they've received the lock.

agronholm:
@x42005e1f I pushed some simplifications to this branch, including the unconditional `popleft()` you suggested.

agronholm:
I removed the other semaphore test which was already passing without these changes. I hope that was okay. In any case, I'm ready to merge unless anyone watching this has objections.

x42005e1f:
I also took another look at this block in the lock and semaphore:

```python
self._waiters.append(item)
try:
    await fut
except CancelledError:
    ...
```

It looks inconsistent, and in the case of the semaphore, it could be optimized. I suggest using the following implementation for both:

```python
self._waiters.append(item)
try:
    await fut
except CancelledError:
    if fut.cancelled():
        try:
            self._waiters.remove(fut)
        except ValueError:  # already removed (in release())
            pass
    else:  # notified
        self.release()

    raise
```

It can be further improved by adding markers, but this way it will no longer attempt to remove a non-existent item when the notification has already happened (and thus perform the expensive O(n) operation on every semaphore release!). After all, if the notification has happened, then due to the unconditional `popleft()`, the item is definitely no longer in the queue. Also, in the case of locks, the notification always happens with ownership transfer, so this implementation is valid for locks as well.

agronholm:
> I also took another look at this block in the lock and semaphore:
> 
> ```python
> self._waiters.append(item)
> try:
>     await fut
> except CancelledError:
>     ...
> ```
> 
> It looks inconsistent, and in the case of the semaphore, it could be optimized. I suggest using the following implementation for both:
> 
> ```python
> self._waiters.append(item)
> try:
>     await fut
> except CancelledError:
>     if fut.cancelled():
>         try:
>             self._waiters.remove(fut)
>         except ValueError:  # already removed (in release())
>             pass
>     else:  # notified
>         self.release()
> 
>     raise
> ```
> 
> It can be further improved by adding markers, but this way it will no longer attempt to remove a non-existent item when the notification has already happened (and thus perform the expensive O(n) operation on every semaphore release!). After all, if the notification has happened, then due to the unconditional `popleft()`, the item is definitely no longer in the queue. Also, in the case of locks, the notification always happens with ownership transfer, so this implementation is valid for locks as well.

Under what circumstances would `await fut` get a `CancelledError` without `fut` being cancelled? That makes no sense.

x42005e1f:
```python
#!/usr/bin/env python3

import asyncio


async def main():
    future = asyncio.get_running_loop().create_future()

    async def work():
        try:
            await future
            print("cancelled")
            print(future)

    task = asyncio.create_task(work())
    await asyncio.sleep(0)  # switch to the task

    future.set_result(None)
    task.cancel()

    await task


if __name__ == "__main__":
    asyncio.run(main())
```

```lang-None
cancelled
<Future finished result=None>
```

agronholm:
> ```python
> #!/usr/bin/env python3
> 
> import asyncio
> 
> 
> async def main():
>     future = asyncio.get_running_loop().create_future()
> 
>     async def work():
>         try:
>             await future
>         except asyncio.CancelledError:
>             print("cancelled")
>             print(future)
> 
>     task = asyncio.create_task(work())
>     await asyncio.sleep(0)  # switch to the task
> 
>     future.set_result(None)
>     task.cancel()
> 
>     await task
> 
> 
> if __name__ == "__main__":
>     asyncio.run(main())
> ```
> 
> ```
> cancelled
> <Future finished result=None>
> ```

In our case, what would cancel the future? We don't do that manually, and otherwise the only way that happens is by the awaiting task getting cancelled.

x42005e1f:
> In our case, what would cancel the future? We don't do that manually, and otherwise the only way that happens is by the awaiting task getting cancelled.

But in this example, I am not cancelling the future object. I am only cancelling the task. The future object is set as if the `release()` method had done so. This does not even have to be done within the same task; it just needs to happen before the notified task resumes.

agronholm:
> > In our case, what would cancel the future? We don't do that manually, and otherwise the only way that happens is by the awaiting task getting cancelled.
> 
> But in this example, I am not cancelling the future object. I am only cancelling the task. The future object is set as if the `release()` method had done so. This does not even have to be done within the same task; it just needs to happen before the notified task resumes.

OK, I see what you mean. This would require a native cancellation, as AnyIO never cancels a task that is scheduled to resume with a value.

x42005e1f:
I usually use the suggested implementation not with `except CancelledError`, but with `except BaseException`, since my library treats any exception as a form of cancellation (for example, `KeyboardInterrupt` when Control-C is pressed). I do not mind if you simplify the code by referring to the AnyIO architecture. I just think my approach is safer.

x42005e1f:
However, I also have a reason to ask the following: why then is `self.release()` under `except CancelledError` in the original implementation?

agronholm:
> However, I also have a reason to ask the following: why then is `self.release()` under `except CancelledError` in the original implementation?

Do you mean, with the assumption that the future and the task would always be cancelled, or not, in lockstep? I most likely forgot about this discrepancy here, but it was baked into the original design. Cancellation semantics can be hard to wrap one's head around :frowning:

agronholm:
I'm not super favorable of adding this latest batch of changes, out of fear they could have subtle unintended effects.

x42005e1f:
And a bit off-topic, but I would also like to ask this: why is there `await AsyncIOBackend.cancel_shielded_checkpoint()` with `except CancelledError` at the beginning of `acquire()`? And why is there no `checkpoint_if_cancelled()` there either? If you cancel the `acquire()` method on the fast path (at a checkpoint), `CancelledError` will not be propagated (which violates expectations regarding checkpoints), but for some reason it is still handled.

agronholm:
> And a bit off-topic, but I would also like to ask this: why is there `await AsyncIOBackend.cancel_shielded_checkpoint()` with `except CancelledError` at the beginning of `acquire()`? And why is there no `checkpoint_if_cancelled()` there either? If you cancel the `acquire()` method on the fast path (at a checkpoint), `CancelledError` will not be propagated (which violates expectations regarding checkpoints), but for some reason it is still handled.

Are you talking about Lock or Semaphore now? You make it sound like we're swallowing cancellation exceptions somewhere.

x42005e1f:
> I'm not super favorable of adding this latest batch of changes, out of fear they could have subtle unintended effects.

I understand your point, but I do not agree that the current implementation is done correctly.

> Are you talking about Lock or Semaphore now? You make it sound like we're swallowing cancellation exceptions somewhere.

Lock:

https://github.com/agronholm/anyio/blob/d517ee1d41080073cc244d8adebf822a08b762de/src/anyio/_backends/_asyncio.py#L1859-L1872

---

Oh, excuse me, I missed that. There is the `checkpoint_if_cancelled()` here.

agronholm:
And all the paths cross some kind of cancellation check. The fast past skips yielding back control to the event loop (unless the task was cancelled on arrival or it needs to wait for the lock to be released).

agronholm:
Does the concern about correctness have to do only with the fact that we only handle `CancelledError`?

x42005e1f:
Yes, of course. The code seems to handle direct task cancellations, and as a result, it contains redundant logic. At the same time, it does not handle other types of exceptions. And my concerns about performance remain.

x42005e1f:
A side effect of this PR: `lock.acquire_nowait()` will behave as expected when all waiters have been cancelled (the same `self._owner_task is None and not self._waiters` check).

agronholm:
Alright, I've pushed your optimizations. All good now?

x42005e1f:
Hmm, wait a minute. If `release()` always transfers ownership, then what is the point of checking whether the queue is not empty? And as I noticed, `semaphore.acquire()` and `semaphore.acquire_nowait()` have different checks (`and not self._waiters` in the first, but not in the second).

agronholm:
I think you'll have to elaborate. My brain is quite tired at this time of night.

x42005e1f:
The `acquire()` method can do two things: either acquire the lock/semaphore immediately, or join the queue until the lock/semaphore is released.

In the case of locks, the release finds the first uncancelled task (if there is one) and transfers ownership to it. If there is no such task, the queue will be empty, and there will be no owner. A situation where there is no owner but the queue is not empty cannot occur.

In the case of semaphores, a similar thing happens, but when there is no uncancelled task, the value is incremented. It is impossible to imagine a situation where the value is non-zero but there is still someone in the queue.

In both cases, checking whether the queue is non-empty (`and not self._waiters`) is redundant.

agronholm:
I'm not 100% whether this is true, but the performance impact of said check is negligible so I don't think I'm going to touch that check at this point. This is the only PR left blocking the v4.14 release so we can revisit this later if there are no more impactful changes or concerns.

x42005e1f:
I agree.

agronholm (approved):

agronholm:
Thanks @rasmusfaber for the PR and @x42005e1f for the in-depth review!

x42005e1f:
Interestingly, my comment about redundant checks still holds true even before the changes in this PR, which hints at one possibility: in fact, removing `and not self._waiters` from the checks was enough to solve the problem (if there is no owner / value > 0, why should the fact that the queue is not empty prevent the primitive from being acquired?). But I think it is still good that we switched to `popleft()`. From a performance (and statistics) standpoint, this looks much better than what we had before.

It likely came to us from asyncio, where `self._locked = True` is set in `acquire()` instead of `release()`, and the AnyIO implementation was inspired by this, but then moved the action to `release()`. In asyncio, the waiters update the attributes themselves (that is, release always sets `self._locked = False`, even when notifying a task in the queue, and therefore the primitive always looks unlocked after the call), so that check is necessary and sufficient there (even cancelled waiters can notify tasks in the queue if the lock has not been acquired). But when the attributes are updated by `release()`, the check is useless and harmful.

PR #1146, opened 2026-05-12
implement future container types

Vizonex:

## Changes

Fixes #. 

This change implements an await-like container object meant for single-use instead of a queue-like stream with the aim towards being similar if not the same API as an `asyncio.Future`. It is needed because I'm in the middle of binding pycares to anyio as well as cyares in the future as I'm giving it a full rewrite in order to save myself of bit of maintenance costs  which involves [separating the current cyares architecture into seperate libraries](https://github.com/Vizonex/cyares/issues/56) although I'm focusing on adding in pycares first to the mix but then adding in my own library to allow users to pick and choose between using a cffi or cython backend. In order to successfully do this well I thought about making a new container type to retain similar in behavior to aiodns which involves using awaiting container objects would be a good idea in here. Some code comes from asyncio while the some of the code architecture is inspired by `concurrent.futures` if anybody was curious as to naming and some design choices made.

- NOTE: It's late for me at the moment so this is a draft and not a real PR yet expect to see more develop from this PR on a later day unless this PR is closed in favor of something else.

## Checklist

features

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -46,6 +46,10 @@ Task groups
 .. autoclass:: anyio.abc.TaskGroup
 .. autoclass:: anyio.abc.TaskStatus
 
+Future objects
+--------------
+.. autoclass:: anyio.Future
+
 Running code in worker threads
 ------------------------------
 
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -48,10 +48,6 @@ Task groups
 .. autoclass:: anyio.abc.TaskGroup
 .. autoclass:: anyio.abc.TaskStatus
 
-Future objects
---------------
-.. autoclass:: anyio.Future
-
 Running code in worker threads
 ------------------------------
 
@@ -198,6 +194,7 @@ Synchronization
 ---------------
 
 .. autoclass:: anyio.Event
+.. autoclass:: anyio.Future
 .. autoclass:: anyio.Lock
 .. autoclass:: anyio.Condition
 .. autoclass:: anyio.Semaphore
--- a/src/anyio/_core/_exceptions.py
+++ b/src/anyio/_core/_exceptions.py
@@ -180,7 +180,7 @@ class TaskNotFinished(Exception):
 class FutureCancelled(Exception):
     """
     Raised when attempting to access the return value or exception of a
-    :class `.Future`: that is still pending completion.
+    :class:`.Future` that is still pending completion.
     """
 
 
--- a/src/anyio/_core/_exceptions.py
+++ b/src/anyio/_core/_exceptions.py
@@ -183,11 +183,13 @@ class FutureCancelled(Exception):
     :class `.Future`: that is still pending completion.
     """
 
+
 class FutureNotFinished(Exception):
     """
     Raised when attempting to access the return value or exception of a
     :class:`.Future` that is still pending completion.
     """
+
 
 class FutureAlreadyFinished(Exception):
     """
--- a/src/anyio/_core/_exceptions.py
+++ b/src/anyio/_core/_exceptions.py
@@ -184,13 +184,6 @@ class FutureCancelled(Exception):
     """
 
 
-class FutureNotFinished(Exception):
-    """
-    Raised when attempting to access the return value or exception of a
-    :class:`.Future` that is still pending completion.
-    """
-
-
 class FutureAlreadyFinished(Exception):
     """
     Raised when attempting set a result of or await a
--- a/src/anyio/_core/_exceptions.py
+++ b/src/anyio/_core/_exceptions.py
@@ -180,7 +180,7 @@ class TaskNotFinished(Exception):
 class FutureCancelled(Exception):
     """
     Raised when attempting to access the return value or exception of a
-    :class:`.Future` that is still pending completion.
+    :class:`.Future` that was cancelled.
     """
 
 
--- a/src/anyio/_core/_exceptions.py
+++ b/src/anyio/_core/_exceptions.py
@@ -179,7 +179,7 @@ class TaskNotFinished(Exception):
 
 class FutureCancelled(Exception):
     """
-    Rasied when attempting to access the return value or exception of a
+    Raised when attempting to access the return value or exception of a
     :class `.Future`: that is still pending completion.
     """
 
--- a/src/anyio/_core/_exceptions.py
+++ b/src/anyio/_core/_exceptions.py
@@ -175,3 +175,22 @@ class TaskNotFinished(Exception):
     Raised when attempting to access the return value or exception of a
     :class:`.TaskHandle` that is still pending completion.
     """
+
+
+class FutureCancelled(Exception):
+    """
+    Rasied when attempting to access the return value or exception of a
+    :class `.Future`: that is still pending completion.
+    """
+
+class FutureNotFinished(Exception):
+    """
+    Raised when attempting to access the return value or exception of a
+    :class:`.Future` that is still pending completion.
+    """
+
+class FutureAlreadyFinished(Exception):
+    """
+    Raised when attempting set a result of or await a
+    :class:`.Future` that has already completed.
+    """
--- a/src/anyio/_core/_exceptions.py
+++ b/src/anyio/_core/_exceptions.py
@@ -177,7 +177,14 @@ class TaskNotFinished(Exception):
     """
 
 
-class FutureCancelled(Exception):
+class FutureFailed(Exception):
+    """
+    Raised when awaiting on, or attempting to access the return value of, a
+    :class:`.Future` that raised an exception.
+    """
+
+
+class FutureCancelled(FutureFailed):
     """
     Raised when attempting to access the return value or exception of a
     :class:`.Future` that was cancelled.
@@ -188,13 +195,6 @@ class FutureNotFinished(Exception):
     """
     Raised when attempting to access the return value or exception of a
     :class:`.Future` that is still pending completion.
-    """
-
-
-class FutureFailed(Exception):
-    """
-    Raised when awaiting on, or attempting to access the return value of, a
-    :class:`.Future` that raised an exception.
     """
 
 
--- a/src/anyio/_core/_exceptions.py
+++ b/src/anyio/_core/_exceptions.py
@@ -184,6 +184,20 @@ class FutureCancelled(Exception):
     """
 
 
+class FutureNotFinished(Exception):
+    """
+    Raised when attempting to access the return value or exception of a
+    :class:`.Future` that is still pending completion.
+    """
+
+
+class FutureFailed(Exception):
+    """
+    Raised when awaiting on, or attempting to access the return value of, a
+    :class:`.Future` that raised an exception.
+    """
+
+
 class FutureAlreadyFinished(Exception):
     """
     Raised when attempting set a result of or await a
--- a/src/anyio/_core/_futures.py
+++ b/src/anyio/_core/_futures.py
@@ -14,6 +14,7 @@ from ._synchronization import Event
 
 T = TypeVar("T")
 
+
 class Future(Generic[T]):
     """
     An awaitable object that works similar to a :class:`asyncio.Future` but
--- a/src/anyio/_core/_futures.py
+++ b/src/anyio/_core/_futures.py
@@ -47,7 +47,7 @@ class Future(Generic[T]):
             case Future.Status.PENDING:
                 return
             case Future.Status.FINISHED:
-                raise FutureAlreadyFinished("future has already finished.")
+                raise FutureAlreadyFinished("future has already finished")
             case Future.Status.FAILED:
                 raise FutureAlreadyFinished("future already failed")
             case Future.Status.CANCELLING:
@@ -81,6 +81,20 @@ class Future(Generic[T]):
         self._check_status()
         self._exception = exception
         self._finished_event.set()
+
+    def cancel(self) -> None:
+        match self.status:
+            case Future.Status.PENDING:
+                return self._cancel_scope.cancel()
+            case Future.Status.FINISHED:
+                raise FutureAlreadyFinished("future has already finished")
+            case Future.Status.FAILED:
+                raise FutureAlreadyFinished("future already failed")
+            case Future.Status.CANCELLING:
+                return
+            case Future.Status.CANCELLED:
+                return
+
 
     @property
     def exception(self) -> BaseException | None:
--- a/src/anyio/_core/_futures.py
+++ b/src/anyio/_core/_futures.py
@@ -1,153 +1,144 @@
 from __future__ import annotations
 
-import sys
-from collections.abc import Awaitable, Callable, Generator
+from collections.abc import Generator
 from enum import Enum, auto
-from typing import Any, TypeVar
+from typing import Any, Generic, TypeVar
 
-from .. import Event
-from ..lowlevel import RunVar
-from ._exceptions import FutureAlreadyFinished, FutureCancelled, FutureNotFinished
-
-if sys.version_info >= (3, 11):
-    from typing import Self
-else:
-    from typing_extensions import Self
-
+from ._eventloop import get_cancelled_exc_class
+from ._exceptions import (
+    FutureAlreadyFinished,
+    FutureCancelled,
+    TaskFailed,
+    TaskNotFinished,
+)
+from ._synchronization import Event
+from ._tasks import CancelScope
 
 T = TypeVar("T")
 
 
-class Future(Awaitable[T]):
-    """A asynchronous container object for waiting on single objects to be completed."""
-
+class Future(Generic[T]):
     class Status(Enum):
+        """The status of a future handle."""
+
         PENDING = auto()
-        CANCELLED = auto()
         FINISHED = auto()
+        FAILED = auto()
+        CANCELLING = auto()
+        CANCELLED = auto()
 
-    def __init__(self) -> None:
-        self._event = Event()
-        self._state = self.Status.PENDING
-        self._result: T = None
+    __slots__ = (
+        "_result_value",
+        "_finished_event",
+        "_cancel_scope",
+        "_exception",
+        "_name",
+    )
+    _result_value: T
+
+    def __init__(self, *, name: str | None = None) -> None:
+        self._finished_event = Event()
+        self._cancel_scope = CancelScope()
         self._exception: BaseException | None = None
-        self._cancelled_exc: BaseException | None = None
+        self._name = name
 
-        self._done_callbacks: list[
-            tuple[RunVar[Any] | None, Callable[[Self], Any]]
-        ] = []
-        self._cancel_msg = None
+    def _check_status(self) -> None:
+        match self.status:
+            case Future.Status.PENDING:
+                return
+            case Future.Status.FINISHED:
+                raise FutureAlreadyFinished("future has already finished.")
+            case Future.Status.FAILED:
+                raise FutureAlreadyFinished("future already failed")
+            case Future.Status.CANCELLING:
+                raise FutureCancelled("future was cancelled")
+            case Future.Status.CANCELLED:
+                raise FutureCancelled("future was cancelled") from self._exception
 
-    def __check_not_done(self) -> None:
-        if self._state != self.Status.PENDING:
-            raise FutureAlreadyFinished("Future has already finished.")
+    async def wait(self) -> None:
+        """
+        Waits for the future to finish.
 
-    def __check_done(self) -> None:
-        if self._state == self.Status.CANCELLED:
-            exc = self._make_cancelled_error()
-            raise exc
-        if self._state != self.Status.FINISHED:
-            raise FutureNotFinished("Result is not ready.")
+        This method will attempt to wait for a result or exception
+        """
+        with self._cancel_scope:
+            try:
+                await self._finished_event.wait()
+            except BaseException as e:
+                self._exception = e
+                raise
+            finally:
+                self._finished_event.set()
 
-    def __schedule_callbacks(self) -> None:
-        # set event so that anything awaiting this object
-        # can be notified that it is now ready for use.
-        self._event.set()
-        for rv, cb in self._done_callbacks:
-            if rv:
-                with rv:
-                    cb(self)
-            else:
-                cb(self)
-
-    def cancel(self, msg=None) -> bool:
-        """Cancel the future and schedules callbacks"""
-        if self._state != self.Status.PENDING:
-            return False
-        self._state = self.Status.CANCELLED
-        self._cancel_msg = msg
-        self.__schedule_callbacks()
-        return True
-
-    def done(self) -> bool:
-        """Return True if future was completed."""
-        return self._state != self.Status.PENDING
-
-    def cancelled(self):
-        """Return True if future was cancelled."""
-        return self._state != self.Status.CANCELLED
-
-    def _make_cancelled_error(self) -> FutureCancelled:
-        """Creates a :class:`.FutureCancelled` exception."""
-        if self._cancelled_exc is not None:
-            exc = self._cancelled_exc
-            self._cancelled_exc = None
-            return exc
-
-        exc = (
-            FutureCancelled()
-            if self._cancel_msg is not None
-            else FutureCancelled(self._cancel_msg)
-        )
-        exc.__context__ = self._cancelled_exc
-        self._cancelled_exc = None
-        return exc
-
-    def result(self) -> T:
-        """Gets the result of a :class:`.Future` if completed."""
-        self.__check_done()
-        if self._exception is not None:
-            raise self._exception
-        return self._result
-
-    def exception(self) -> BaseException | None:
-        """Gets the exception of a :class:`.Future` if completed it will be done if there is instead a successful result."""
-        self.__check_done()
-        return self._exception
-
-    def add_done_callback(
-        self, fn: Callable[[Self], Any], *, context: RunVar[Self] | None = None
-    ) -> None:
-        """Creates a synchronous callback for when a future is considered as being completed or not."""
-        if self._state != self.Status.PENDING:
-            if context:
-                with context:
-                    fn(self)
-            else:
-                fn(self)
-        else:
-            self._done_callbacks.append((context, fn))
-
-    def remove_done_callback(self, fn: Callable[[Self], Any]):
-        filtered_callbacks = [(f, ctx) for (f, ctx) in self._done_callbacks if f != fn]
-        removed_count = len(self._done_callbacks) - len(filtered_callbacks)
-        if removed_count:
-            self._done_callbacks[:] = filtered_callbacks
-        return removed_count
-
-    def set_result(self, result: T) -> None:
-        """sets the results of a :class:`.Future` object."""
-        self.__check_not_done()
-        self._result = result
-        self._state = self.Status.FINISHED
-        self.__schedule_callbacks()
+    def set_result(self, value: T) -> None:
+        """Send pending result for a container object"""
+        self._check_status()
+        self._result_value = value
+        self._finished_event.set()
 
     def set_exception(self, exception: BaseException) -> None:
-        """sets the exception of a :class:`.Future` object."""
-        self.__check_not_done()
+        """Send exception result for a container object"""
+        self._check_status()
         self._exception = exception
-        self._state = self.Status.FINISHED
-        self.__schedule_callbacks()
+        self._finished_event.set()
 
-    # Mostly used as a proxy for awaiting on a pending object.
-    async def wait(self) -> T:
-        await self._event.wait()
-        return self.result()
+    @property
+    def exception(self) -> BaseException | None:
+        match self.status:
+            case Future.Status.PENDING:
+                raise TaskNotFinished("the future has not finished yet")
+            case Future.Status.FINISHED:
+                return None
+            case Future.Status.CANCELLING:
+                raise FutureCancelled("the future was cancelled")
+            case Future.Status.CANCELLED:
+                raise FutureCancelled("the future was cancelled") from self._exception
+            case Future.Status.FAILED:
+                return self._exception
 
-    def __await__(self) -> Generator[Any, None, T]:
-        return self.wait().__await__()
+    @property
+    def return_value(self) -> T:
+        """
+        The return value of the future.
 
+        :raises TaskNotFinished: if the future has not finished yet
+        :raises FutureCancelled: if the future was cancelled
+        :raises TaskFailed: if the future raised an exception
 
-def create_future() -> Future[T]:
-    """Shortcut method for creating a new :class:`.Future` object."""
-    return Future()
+        """
+        match self.status:
+            case Future.Status.PENDING:
+                raise TaskNotFinished("the future has not finished yet")
+            case Future.Status.FINISHED:
+                return self._result_value
+            case Future.Status.CANCELLING:
+                raise FutureCancelled("the future was cancelled")
+            case Future.Status.CANCELLED:
+                raise FutureCancelled("the future was cancelled") from self._exception
+            case Future.Status.FAILED:
+                raise TaskFailed("the future raised an exception") from self._exception
+
+    @property
+    def status(self) -> Future.Status:
+        if not self._finished_event.is_set():
+            if self._cancel_scope.cancel_called:
+                return Future.Status.CANCELLING
+            else:
+                return Future.Status.PENDING
+        elif self._exception is not None:
+            if isinstance(self._exception, get_cancelled_exc_class()):
+                return Future.Status.CANCELLED
+            else:
+                return Future.Status.FAILED
+        else:
+            return Future.Status.FINISHED
+
+    def __await__(self) -> Generator[Any, Any, T]:
+        yield from self.wait().__await__()
+        return self.return_value
+
+    def __repr__(self) -> str:
+        return (
+            f"<{self.__class__.__name__} {self.status.name.lower()} "
+            f"name={self._name!r}>"
+        )
--- a/src/anyio/_core/_futures.py
+++ b/src/anyio/_core/_futures.py
@@ -93,10 +93,8 @@ class Future(Generic[T]):
             try:
                 await self._finished_event.wait()
             except BaseException as e:
-                self._exception = e
+                self.set_exception(e)
                 raise
-            finally:
-                self._finished_event.set()
 
     def set_result(self, value: T) -> None:
         """
--- a/src/anyio/_core/_futures.py
+++ b/src/anyio/_core/_futures.py
@@ -107,11 +107,11 @@ class Future(Generic[T]):
     def cancel(self) -> None:
         """Cancels a pending `.Future` object
 
-        :raises FutureAlreadyFinished: if future was already given a result or exception.
+        Does nothing if the Future was already finished.
         """
-        self._check_pending()
-        self._cancelled = True
-        self._finished_event.set()
+        if self.status is Future.Status.PENDING:
+            self._cancelled = True
+            self._finished_event.set()
 
     @property
     def exception(self) -> BaseException | None:
--- a/src/anyio/_core/_futures.py
+++ b/src/anyio/_core/_futures.py
@@ -4,7 +4,6 @@ from collections.abc import Generator
 from enum import Enum, auto
 from typing import Any, Generic, TypeVar
 
-from ._eventloop import get_cancelled_exc_class
 from ._exceptions import (
     FutureAlreadyFinished,
     FutureCancelled,
@@ -12,10 +11,8 @@ from ._exceptions import (
     TaskNotFinished,
 )
 from ._synchronization import Event
-from ._tasks import CancelScope
 
 T = TypeVar("T")
-
 
 class Future(Generic[T]):
     """
@@ -33,9 +30,6 @@ class Future(Generic[T]):
         .. attribute:: FINISHED
 
             The future has finished with a return value.
-        .. attribute:: CANCELLING
-
-            The future has been cancelled but has not finished yet.
         .. attribute:: CANCELLED
 
             The future was cancelled and has finished since.
@@ -46,14 +40,13 @@ class Future(Generic[T]):
 
         PENDING = auto()
         FINISHED = auto()
-        CANCELLING = auto()
         CANCELLED = auto()
         FAILED = auto()
 
     __slots__ = (
+        "_cancelled",
         "_result_value",
         "_finished_event",
-        "_cancel_scope",
         "_exception",
         "_name",
     )
@@ -61,8 +54,8 @@ class Future(Generic[T]):
 
     def __init__(self, *, name: str | None = None) -> None:
         self._finished_event = Event()
-        self._cancel_scope = CancelScope()
         self._exception: BaseException | None = None
+        self._cancelled: bool = False
         self._name = name
 
     def _check_pending(self) -> None:
@@ -78,10 +71,8 @@ class Future(Generic[T]):
                 raise FutureAlreadyFinished("future has already finished")
             case Future.Status.FAILED:
                 raise FutureAlreadyFinished("future already failed")
-            case Future.Status.CANCELLING:
-                raise FutureCancelled("future was cancelled")
             case Future.Status.CANCELLED:
-                raise FutureCancelled("future was cancelled") from self._exception
+                raise FutureCancelled("future was cancelled")
 
     async def wait(self) -> None:
         """
@@ -89,12 +80,7 @@ class Future(Generic[T]):
 
         This method will attempt to wait for a result or exception
         """
-        with self._cancel_scope:
-            try:
-                await self._finished_event.wait()
-            except BaseException as e:
-                self.set_exception(e)
-                raise
+        await self._finished_event.wait()
 
     def set_result(self, value: T) -> None:
         """
@@ -120,19 +106,11 @@ class Future(Generic[T]):
     def cancel(self) -> None:
         """Cancels a pending `.Future` object
 
-        :raises FutureAlreadyFinished: if future was already given a result or exception.
+        Does nothing if the Future was already finished.
         """
-        match self.status:
-            case Future.Status.PENDING:
-                return self._cancel_scope.cancel()
-            case Future.Status.FINISHED:
-                raise FutureAlreadyFinished("future has already finished")
-            case Future.Status.FAILED:
-                raise FutureAlreadyFinished("future already failed")
-            case Future.Status.CANCELLING:
-                return
-            case Future.Status.CANCELLED:
-                return
+        if self.status is Future.Status.PENDING:
+            self._cancelled = True
+            self._finished_event.set()
 
     @property
     def exception(self) -> BaseException | None:
@@ -150,10 +128,8 @@ class Future(Generic[T]):
                 raise TaskNotFinished("the future has not finished yet")
             case Future.Status.FINISHED:
                 return None
-            case Future.Status.CANCELLING:
-                raise FutureCancelled("the future was cancelled")
             case Future.Status.CANCELLED:
-                raise FutureCancelled("the future was cancelled") from self._exception
+                raise FutureCancelled("the future was cancelled")
             case Future.Status.FAILED:
                 return self._exception
 
@@ -172,10 +148,8 @@ class Future(Generic[T]):
                 raise TaskNotFinished("the future has not finished yet")
             case Future.Status.FINISHED:
                 return self._result_value
-            case Future.Status.CANCELLING:
-                raise FutureCancelled("the future was cancelled")
             case Future.Status.CANCELLED:
-                raise FutureCancelled("the future was cancelled") from self._exception
+                raise FutureCancelled("the future was cancelled")
             case Future.Status.FAILED:
                 raise TaskFailed("the future raised an exception") from self._exception
 
@@ -193,15 +167,11 @@ class Future(Generic[T]):
         raised, if any. No other status transitions will happen.
         """
         if not self._finished_event.is_set():
-            if self._cancel_scope.cancel_called:
-                return Future.Status.CANCELLING
-            else:
-                return Future.Status.PENDING
+            return Future.Status.PENDING
+        elif self._cancelled:
+            return Future.Status.CANCELLED
         elif self._exception is not None:
-            if isinstance(self._exception, get_cancelled_exc_class()):
-                return Future.Status.CANCELLED
-            else:
-                return Future.Status.FAILED
+            return Future.Status.FAILED
         else:
             return Future.Status.FINISHED
 
--- a/src/anyio/_core/_futures.py
+++ b/src/anyio/_core/_futures.py
@@ -17,8 +17,10 @@ else:
 
 T = TypeVar("T")
 
+
 class Future(Awaitable[T]):
     """A asynchronous container object for waiting on single objects to be completed."""
+
     class Status(Enum):
         PENDING = auto()
         CANCELLED = auto()
@@ -31,7 +33,9 @@ class Future(Awaitable[T]):
         self._exception: BaseException | None = None
         self._cancelled_exc: BaseException | None = None
 
-        self._done_callbacks: list[tuple[RunVar[Any] | None , Callable[[Self], Any]]] = []
+        self._done_callbacks: list[
+            tuple[RunVar[Any] | None, Callable[[Self], Any]]
+        ] = []
         self._cancel_msg = None
 
     def __check_not_done(self) -> None:
@@ -80,7 +84,11 @@ class Future(Awaitable[T]):
             self._cancelled_exc = None
             return exc
 
-        exc = FutureCancelled() if self._cancel_msg is not None else FutureCancelled(self._cancel_msg)
+        exc = (
+            FutureCancelled()
+            if self._cancel_msg is not None
+            else FutureCancelled(self._cancel_msg)
+        )
         exc.__context__ = self._cancelled_exc
         self._cancelled_exc = None
         return exc
@@ -97,7 +105,9 @@ class Future(Awaitable[T]):
         self.__check_done()
         return self._exception
 
-    def add_done_callback(self, fn: Callable[[Self], Any], *, context: RunVar[Self] | None = None) -> None:
+    def add_done_callback(
+        self, fn: Callable[[Self], Any], *, context: RunVar[Self] | None = None
+    ) -> None:
         """Creates a synchronous callback for when a future is considered as being completed or not."""
         if self._state != self.Status.PENDING:
             if context:
@@ -109,9 +119,7 @@ class Future(Awaitable[T]):
             self._done_callbacks.append((context, fn))
 
     def remove_done_callback(self, fn: Callable[[Self], Any]):
-        filtered_callbacks = [(f, ctx)
-                              for (f, ctx) in self._done_callbacks
-                              if f != fn]
+        filtered_callbacks = [(f, ctx) for (f, ctx) in self._done_callbacks if f != fn]
         removed_count = len(self._done_callbacks) - len(filtered_callbacks)
         if removed_count:
             self._done_callbacks[:] = filtered_callbacks
@@ -138,6 +146,7 @@ class Future(Awaitable[T]):
 
     def __await__(self) -> Generator[Any, None, T]:
         return self.wait().__await__()
+
 
 def create_future() -> Future[T]:
     """Shortcut method for creating a new :class:`.Future` object."""
--- a/src/anyio/_core/_futures.py
+++ b/src/anyio/_core/_futures.py
@@ -16,11 +16,13 @@ from ._tasks import CancelScope
 
 T = TypeVar("T")
 
+
 class Future(Generic[T]):
     """
     An awaitable object that works simillar to a :class:`asyncio.Future` but
     with simillar characteristics to a :class:`.TaskHandle`.
     """
+
     class Status(Enum):
         """
         The status of a future handle.
--- a/src/anyio/_core/_futures.py
+++ b/src/anyio/_core/_futures.py
@@ -1,0 +1,144 @@
+from __future__ import annotations
+
+import sys
+from collections.abc import Awaitable, Callable, Generator
+from enum import Enum, auto
+from typing import Any, TypeVar
+
+from .. import Event
+from ..lowlevel import RunVar
+from ._exceptions import FutureAlreadyFinished, FutureCancelled, FutureNotFinished
+
+if sys.version_info >= (3, 11):
+    from typing import Self
+else:
+    from typing_extensions import Self
+
+
+T = TypeVar("T")
+
+class Future(Awaitable[T]):
+    """A asynchronous container object for waiting on single objects to be completed."""
+    class Status(Enum):
+        PENDING = auto()
+        CANCELLED = auto()
+        FINISHED = auto()
+
+    def __init__(self) -> None:
+        self._event = Event()
+        self._state = self.Status.PENDING
+        self._result: T = None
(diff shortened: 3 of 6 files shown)


src/anyio/_core/_exceptions.py
```diff
@@ -175,3 +175,24 @@ class TaskNotFinished(Exception):
     Raised when attempting to access the return value or exception of a
     :class:`.TaskHandle` that is still pending completion.
     """
+
+
+class FutureCancelled(Exception):
+    """
+    Rasied when attempting to access the return value or exception of a
```
EmmanuelNiyonshuti:
```suggestion
    Raised when attempting to access the return value or exception of a
```

agronholm:
This needs to align with the new `TaskHandle` API. And drop `create_future()`.

agronholm:
Also, I believe I told you "no callbacks". Drop those too.

m-u-xyz:
anyio should not have unbounded futures, for the same reasons why it doesn't have unbound tasks. A future thus should be attached to a taskgroup (the one that's responsible for fulfilling it) so that it's auto-cancelled when the taskgroup ends without setting a result. (This is also a problem with events, but in Real Code the problem ends up being worse with futures because they have more state.)

Also, anyio now has task handles. Those behave like futures in some meaningful way. I'd like to be *very* sure that we **really** need a separate `Future` class.

Vizonex:
> anyio should not have unbounded futures, for the same reasons why it doesn't have unbound tasks. A future thus should be attached to a taskgroup (the one that's responsible for fulfilling it) so that it's auto-cancelled when the taskgroup ends without setting a result. (This is also a problem with events, but in Real Code the problem ends up being worse with futures because they have more state.)
> 
> Also, anyio now has task handles. Those behave like futures in some meaningful way. I'd like to be _very_ sure that we **really** need a separate `Future` class.

@smurfix 
That makes sense to me. Especially when were making a taskgroup as well as closing it.

> Also, I believe I told you "no callbacks". Drop those too.

@agronholm 
We can make that change to I see no need to have any of that here.

EmmanuelNiyonshuti:
hmm..curious to see how this will turn out. a create_*() future whose completion will be managed by a task group?

agronholm:
> Also, anyio now has task handles. Those behave like futures in some meaningful way. I'd like to be very sure that we really need a separate Future class.

I, too, have come across cases where I've needed an event + value combination. I just want to make sure that consumers aren't left hanging if the task that is supposed to fulfill the future gets cancelled. But I see no reason to tie it to a task group. It could just support the context manager protocol on its own. Then, if it the task is cancelled before a value is filled in, it could notify any consumers with an exception.

agronholm:
So, I just thought of something: how about disallowing consumers from awaiting on a future that has not been entered? That would still allow you to create a future with a pre-assigned value (without having to enter the future for that).

Vizonex:
@agronholm Made a better version of an await-table event container object that falls more inline with how TaskHandle works wanted to make sure the code was acceptable to your liking before I try adding this in with more documentation. I had it as a separate module away from the library temporarily to help me with the brainstorming of how this should ultimately work. The only thing missing is really how the exceptions should be named or if separate ones are really at all needed.

```python

from collections.abc import Generator
from enum import Enum, auto
from typing import Any, Generic, TypeVar

from anyio import (
    CancelScope,
    Event,
    FutureAlreadyFinished,
    FutureCancelled,
    TaskFailed,
    TaskNotFinished,
)
from anyio._core._eventloop import get_cancelled_exc_class

T = TypeVar("T")

class FutureHandle(Generic[T]):

    class Status(Enum):
        """The status of a future hadnle."""
        PENDING = auto()
        FINISHED = auto()
        FAILED = auto()
        CANCELLING = auto()
        CANCELLED = auto()

    __slots__ = (
        "_result_value",
        "_finished_event",
        "_cancel_scope",
        "_exception",
        "_name"
    )
    _result_value: T

    def __init__(self, *, name:str | None = None) -> None:
        self._finished_event = Event()
        self._cancel_scope = CancelScope()
        self._exception: BaseException | None = None
        self._name = name

    def _check_status(self) -> None:
        match self.status:
            case FutureHandle.Status.PENDING:
                return
            case FutureHandle.Status.FINISHED:
                raise FutureAlreadyFinished("future has already finished.")
            case FutureHandle.Status.FAILED:
                raise FutureAlreadyFinished("future already failed")
            case FutureHandle.Status.CANCELLING:
                raise FutureCancelled("future was cancelled")
            case FutureHandle.Status.CANCELLED:
                raise FutureCancelled("future was cancelled") from self._exception

    async def wait(self) -> None:
        """
        Waits for the future to finish.

        This method will attempt to wait for a result or exception
        """
        with self._cancel_scope:
            try:
                await self._finished_event.wait()
            except BaseException as e:
                self._exception = e
                raise
            finally:
                self._finished_event.set()

    def set_result(self, value: T) -> None:
        """Send pending result for a container object"""
        self._check_status()
        self._result_value = value
        self._finished_event.set()

    def set_exception(self, exception: BaseException) -> None:
        """Send exception result for a container object"""
        self._check_status()
        self._exception = exception
        self._finished_event.set()

    @property
    def exception(self) -> BaseException | None:
        match self.status:
            case FutureHandle.Status.PENDING:
                raise TaskNotFinished("the future has not finished yet")
            case FutureHandle.Status.FINISHED:
                return None
            case FutureHandle.Status.CANCELLING:
                raise FutureCancelled("the future was cancelled")
            case FutureHandle.Status.CANCELLED:
                raise FutureCancelled("the future was cancelled") from self._exception
            case FutureHandle.Status.FAILED:
                return self._exception

    @property
    def return_value(self) -> T:
        """
        The return value of the task.

        :raises TaskNotFinished: if the future has not finished yet
        :raises FutureCancelled: if the future was cancelled
        :raises TaskFailed: if the task raised an exception

        """
        match self.status:
            case FutureHandle.Status.PENDING:
                raise TaskNotFinished("the future has not finished yet")
            case FutureHandle.Status.FINISHED:
                return self._return_value
            case FutureHandle.Status.CANCELLING:
                raise FutureCancelled("the future was cancelled")
            case FutureHandle.Status.CANCELLED:
                raise FutureCancelled("the future was cancelled") from self._exception
            case FutureHandle.Status.FAILED:
                raise TaskFailed("the future raised an exception") from self._exception

    @property
    def status(self) -> "FutureHandle.Status":
        if not self._finished_event.is_set():
            if self._cancel_scope.cancel_called:
                return FutureHandle.Status.CANCELLING
            else:
                return FutureHandle.Status.PENDING
        elif self._exception is not None:
            if isinstance(self._exception, get_cancelled_exc_class()):
                return FutureHandle.Status.CANCELLED
            else:
                return FutureHandle.Status.FAILED
        else:
            return FutureHandle.Status.FINISHED

    def __await__(self) -> Generator[Any, Any, T]:
        yield from self.wait().__await__()
        return self.return_value

    def __repr__(self) -> str:
        return (
            f"<{self.__class__.__name__} {self.status.name.lower()} "
            f"name={self._name!r}>"
        )
```

agronholm:
Looks much better now. We'll just have to work out how to deal with the aforementioned issues and do some bikeshedding on the names.

Vizonex:
> Looks much better now. We'll just have to work out how to deal with the aforementioned issues and do some bikeshedding on the names.

Thank you that means a lot to me ^^.

src/anyio/_core/_exceptions.py
```diff
@@ -175,3 +175,24 @@ class TaskNotFinished(Exception):
     Raised when attempting to access the return value or exception of a
     :class:`.TaskHandle` that is still pending completion.
     """
+
+
+class FutureCancelled(Exception):
+    """
+    Raised when attempting to access the return value or exception of a
+    :class `.Future`: that is still pending completion.
```
EmmanuelNiyonshuti: :class `.Future`: that was cancelled.

src/anyio/_core/_futures.py
```diff
@@ -0,0 +1,157 @@
...
+                return self._result_value
+            case Future.Status.CANCELLING:
+                raise FutureCancelled("the future was cancelled")
+            case Future.Status.CANCELLED:
+                raise FutureCancelled("the future was cancelled") from self._exception
+            case Future.Status.FAILED:
+                raise TaskFailed("the future raised an exception") from self._exception
+
+    @property
+    def status(self) -> Future.Status:
```
EmmanuelNiyonshuti, line 162: I think it might help if this one has docstrings as well. very helpful with `help()`

EmmanuelNiyonshuti (commented):
well, sorry to take a bit of time to review, and thank you for the tag. I left a few trivial suggestions and overall LGTM. my familiarity with AnyIO and overall async world is not that deep for a safe review. I will leave this one for @agronholm @smurfix for their final thoughts and say.

agronholm:
There are a number of changes I would like to see here before I could accept the PR:

1. Review the `Status` enum: are all the items here really relevant to `Future`?
2. Do we need to store the status as a field, rather than automatically determining from the other fields like we do in `TaskHandle`?
3. What about my idea of requiring the producer to acquire the future as a context manager before allowing consumers to await on it? This needs a resolution.

tapetersen:
>     3. What about my idea of requiring the producer to acquire the future as a context manager before allowing consumers to await on it? This needs a resolution.

For that requirement is there maybe a case of returning to a `create_furure()` like API (or anything with a private constructor of the actual `Future` objects) where the actual Future is only returned on entering the context manager?
For the case of creating an immediately resolved future either an argument to the initial function (unsure if I like an API that returns a CTC manager or not depending on argument) or a separate create method could provide that.

agronholm:
> > ```
> > 3. What about my idea of requiring the producer to acquire the future as a context manager before allowing consumers to await on it? This needs a resolution.
> > ```
> 
> For that requirement is there maybe a case of returning to a `create_furure()` like API (or anything with a private constructor of the actual `Future` objects) where the actual Future is only returned on entering the context manager? For the case of creating an immediately resolved future either an argument to the initial function (unsure if I like an API that returns a CTC manager or not depending on argument) or a separate create method could provide that.

Hmm...maybe I'm not getting something. Let me clarify my own intentions. Awaiting on a `Future` should only raise if 1) it does not have a value or exception already set, AND 2) it has not been entered as a CM.
Thus, I don't see value in a separate `create_future()` function.

tapetersen:
> Awaiting on a Future should only raise if 1) it does not have a value or exception already set, AND 2) it has not been entered as a CM.

It's probably that I don't see or miss some intended value  in having the access to the created Future while in that timespan at all (create until entered). 

If that is the case (Which it very well may not be) why not shape the API so a future is always entered was my thought.

agronholm:
> > Awaiting on a Future should only raise if 1) it does not have a value or exception already set, AND 2) it has not been entered as a CM.
> 
> It's probably that I don't see or miss some intended value in having the access to the created Future while in that timespan at all (create until entered).
> 
> If that is the case (Which it very well may not be) why not shape the API so a future is always entered was my thought.

Do you not see any value in futures with pre-filled values?

tapetersen:
> > > Awaiting on a Future should only raise if 1) it does not have a value or exception already set, AND 2) it has not been entered as a CM.
> > 
> > 
> > It's probably that I don't see or miss some intended value in having the access to the created Future while in that timespan at all (create until entered).
> > If that is the case (Which it very well may not be) why not shape the API so a future is always entered was my thought.
> 
> Do you not see any value in futures with pre-filled values?

I do (I assume we mean basically mean futures already finished with a set value) and that case was what I was refering to needing some argument to `create_future` alternatively an extra function.

I do see other ways of structuring the API if I had done it but this is perfectly fine and I don't want to derail the PR.

<details><summary>Details</summary>
<p>
Basically splitting the waiting handle returned from the controlling handle to not allow consumers to arbitrarily finish the future for example. In that case the controlling handle could have a method creating a resolved Future directly or be used as a context manager which when entered gave an unresolved Future it could yield or otherwise return


</p>
</details>

agronholm:
> I do (I assume we mean basically mean futures already finished with a set value) and that case was what I was refering to needing some argument to create_future alternatively an extra function.

So why can't the `Future` class take the preset value directly as an initializer argument then? As for your concern about consumers setting the future – to me that sounds overengineering. We're trying to prevent footguns, and a consumer setting the value of a future seems just malicious instead of something you'd do accidentally.

tapetersen:
> So why can't the `Future` class take the preset value directly as an initializer argument then?

Only really in that the other case where it's not resolved immediately would lead to having a Future object which should probably not be yielded/returned until being entered. If it's a `ContextManager[Future]` rather than a `Future` itself typing would catch that statically.

If we want to go out of our way to raise exceptions until the creator has called `__enter__` why allow that future to exist before that?

> As for your concern about consumers setting the future – to me that sounds overengineering. We're trying to prevent footguns, and a consumer setting the value of a future seems just malicious instead of something you'd do accidentally.

You're probably right and the most risk is probably them cancelling it. The other reason I like the pattern of separate interfaces was mostly to keep the interface provided by autocomplete smaller even if it may be clear calling these functions normally shouldn't be done.

agronholm:
> > So why can't the `Future` class take the preset value directly as an initializer argument then?
> 
> Only really in that the other case where it's not resolved immediately would lead to having a Future object which should probably not be yielded/returned until being entered. If it's a `ContextManager` not being a `Future` itself typing would catch that statically.
> 
> If we want to go out of our way to raise exceptions until the creator has called `__enter__` why allow that future to exist before that?

So, if we want to cater to 2 different use cases:

1. A pre-filled `Future` (should not need to be entered)
2. A Future that will be completed later, but passed to consumers after entering

What would the API look like? Do we need two functions for this?
 
> > As for your concern about consumers setting the future – to me that sounds overengineering. We're trying to prevent footguns, and a consumer setting the value of a future seems just malicious instead of something you'd do accidentally.
> 
> You're probably right and the most risk is probably them cancelling it. Other reason I like the pattern of separate interfaces where mostly to keep the interface provided by autocomplete smaller even if it may be clear calling these functions normally shouldn't be done.

The consumers would have to `cancel()` the future explicitly then, unlike with asyncio where a task being cancelled also cancels the future it's awaiting on, yes? Is there still a problem there?

I understand what you're getting at with the splitting of the future between the producer and consumer. I'm trying to determine if that's warranted or not.

tapetersen:
> > > So why can't the `Future` class take the preset value directly as an initializer argument then?
> > 
> > Only really in that the other case where it's not resolved immediately would lead to having a Future object which should probably not be yielded/returned until being entered. If it's a `ContextManager` not being a `Future` itself typing would catch that statically.
> > 
> > If we want to go out of our way to raise exceptions until the creator has called `__enter__` why allow that future to exist before that?
> 
> So, if we want to cater to 2 different use cases:
> 
> 1. A pre-filled `Future` (should not need to be entered)
> 2. A Future that will be completed later, but passed to consumers after entering
> 
> What would the API look like? Do we need two functions for this?

Either 2 separate functions or one returning something like `Future | ContextManager[Future]` depending on if a value was supplied or not (and crucially that the returned value *isn't* a `Future` in the second case is why that function being the `Future()` constructor itself is a problem.

> > > As for your concern about consumers setting the future – to me that sounds overengineering. We're trying to prevent footguns, and a consumer setting the value of a future seems just malicious instead of something you'd do accidentally.
> > 
> > You're probably right and the most risk is probably them cancelling it. Other reason I like the pattern of separate interfaces where mostly to keep the interface provided by autocomplete smaller even if it may be clear calling these functions normally shouldn't be done.
> 
> The consumers would have to `cancel()` the future explicitly then, unlike with asyncio where a task being cancelled also cancels the future it's awaiting on, yes? Is there still a problem there?
> 
> I understand what you're getting at with the splitting of the future between the producer and consumer. I'm trying to determine if that's warranted or not.

In most cases the api's have very little overlap and if  we want some kind of object acting as a `ContextManager` for the `Future` anyway it could make sense to let that object be the "producer" handle as well and then the api-split is just natural since the need different methods anyway.

Biggest hurdle for proposing something concrete is mostly that I don't have good names for the Manager/Handle.

 `Future / Promise` is actually a bit tempting to signal the passive API of the consumer and make it clear it doesn't implement the usual full API of Future-like objects in python but therefore also loses discovery.

src/anyio/_core/_futures.py
```diff
@@ -0,0 +1,218 @@
...
+                return None
+            case Future.Status.CANCELLING:
+                raise FutureCancelled("the future was cancelled")
+            case Future.Status.CANCELLED:
+                raise FutureCancelled("the future was cancelled") from self._exception
+            case Future.Status.FAILED:
+                return self._exception
+
+    @property
+    def return_value(self) -> T:
```
EmmanuelNiyonshuti, line 128: I feel we can name this `result` instead, what do you think, given that `asyncio.Future` uses `result` ?

src/anyio/_core/_exceptions.py
```diff
@@ -175,3 +175,24 @@ class TaskNotFinished(Exception):
     Raised when attempting to access the return value or exception of a
     :class:`.TaskHandle` that is still pending completion.
     """
+
+
+class FutureCancelled(Exception):
+    """
+    Raised when attempting to access the return value or exception of a
+    :class:`.Future` that is still pending completion.
```
EmmanuelNiyonshuti:
```suggestion
    :class:`.Future` that was cancelled.
```

src/anyio/_core/_futures.py
agronholm: Some points to consider:

- `result()` is a method in `asyncio.Future`, not a property
- then again, `exception()` is also a method there while ours is a property
- `TaskHandle` also has a `return_value` property which may have to be renamed to match

EmmanuelNiyonshuti: that’s true, and thanks! but i’d ask why this future api needs to align with task handles? might come off as a silly question, but i got curious.

agronholm: The similarity should reduce the learning curve. That was the idea at least.

Vizonex: @agronholm As long as I am able to make the code look-alike to the Task object made as I wanted to ensure as many learning curves with the API could be removed if possible. This object's purpose was to help implement things such as executors and binding C or other asynchronous implementations this won't just benefit Cyares however. This can also benefit projects like
- curl-cffi: pull request is ongoing by the main owner who would want to hear about this
- a pycares bridge for anyio: I felt a desire to write an optional version for users who may prefer cffi over cython)

src/anyio/_core/_futures.py
```diff
@@ -0,0 +1,218 @@
...
+                raise FutureCancelled("future was cancelled")
+            case Future.Status.CANCELLED:
+                raise FutureCancelled("future was cancelled") from self._exception
+
+    async def wait(self) -> None:
+        """
+        Waits for the future to finish.
+
+        This method will attempt to wait for a result or exception
+        """
```
EmmanuelNiyonshuti, line 83: According to asyncio a future can be awaited multiple times and the result stays the same. Can we we check for that before we enter the cancel scope?
```
        if self._finished_event.is_set():
            return
```

EmmanuelNiyonshuti (commented):
I added one small suggestion in the code.

On the bigger picture though, I think some of the design questions raised by @agronholm and @tapetersen still need a bit more clarity before we can confidently evaluate the final implementation. The final shape of this new container object should probably be settled first so we're all reviewing the same thing.

Personally, outside of the concerns Alex already [listed](https://github.com/agronholm/anyio/pull/1146#issuecomment-4527842313), the implementation looks to be heading in a good direction.

For the API itself, I currently lean slightly toward the idea of having a function return either a `Future` or a `ContextManager[Future]` depending on whether a value is preset or not. My initial thought is mostly that it would make the distinction between an already-resolved future and one that still needs to be entered a bit more explicit, but I may be(a big may be!) missing some trade-offs here.

Also, unless I missed it, I don't see documentation for this yet.  I'd expect documentation and a usage example before final review.

src/anyio/_core/_exceptions.py
```diff
@@ -175,3 +175,24 @@ class TaskNotFinished(Exception):
...
+
+
+class FutureCancelled(Exception):
+    """
+    Raised when attempting to access the return value or exception of a
+    :class:`.Future` that was cancelled.
+    """
+
+
+class FutureNotFinished(Exception):
```
davidbrochart, line 194: This doesn't seem to be used anywhere?

davidbrochart:
If we want to make it easy for people coming from asyncio, we should consider adding the `.done()` and `.cancelled()` methods, even though they are just aliases to the future's status.
But more importantly, I think we should add the `.add_done_callback()` method.

tapetersen:
@davidbrochart 
Like trio, anyio tries very hard to avoid callback based API's. They has a lot of the same problems as other unstructured concurrency in that code will run later in unspecified scope (from the perspective of the code registering the callback)  were it can't know what resources are still available.

The pattern for anyio would be to create a task waiting for the future and then calling what you need. In that case the caller controls the scope the callback is active themselves through the task group it's tied to.

davidbrochart:
Right, thanks @tapetersen.

m-u-xyz:
>  But more importantly, I think we should add the .add_done_callback() method.

While conceivably we could add a `TaskHandle.run_when_done(taskgroup, callable, …)` helper (or `TaskGroup.wait_and_call(TaskHandle, callable, …)`), the correct way is to simply start a task that runs `callable` with the task handle as an additional argument. IMHO the additional single line (wait for the task at the start of `callable`) is non-intrusive enough, given that the user must touch the `add_done_callback` call anyway (it does require a taskgoup).

Thus IMHO a simple example how to do *that* is the right way to go about this.

davidbrochart:
While playing with this PR, I ran into a `RuntimeError: Each CancelScope may only be used for a single 'with' block`.
Does this mean that `future.wait()` can only be awaited once?

src/anyio/_core/_futures.py
tapetersen: Could be worthwile but since it's an async method it should probably then do a checkpoint anyway to uphold the variant of all async methods passing through at least 1 checkpoint (guaranteed for trio and I think anyio tries to uphold is as well)

src/anyio/_core/_futures.py
```diff
@@ -0,0 +1,218 @@
...
+        with self._cancel_scope:
+            try:
+                await self._finished_event.wait()
+            except BaseException as e:
+                self._exception = e
+                raise
```
tapetersen: Do we really want to set `self._exception` here? The most common case will be that a waiter is cancelled and we get a `CancelledException` here that has nothing todo with the specific waiter. At least skipping all `anyio.get_cancelled_exception()` feels necessary and I'm unsure what other cases it's meaningful for.

tapetersen:
> While playing with this PR, I ran into a `RuntimeError: Each `CancelScope` may only be used for a single 'with' block`. Does this mean that `future.wait()` can only be awaited once?

(I've yet to check the PR carefully here so take with a pinch of salt and I may be completely missing something)

I'm 99% sure that's not intended. I'm also unsure now that I'm looking at it if it should even throw a `CancelledException` at all, by a scope or manually, as that normally signals that the *calling* task is asked to shut-down while this case from what I can see covers the case where the Future is cancelled and the result won't be available to one or more callers.

It feels like cancelling the `Future` should just set the `Event` to wake up callers raising a `FutureCancelled` exception instead of a `CancelScope`

src/anyio/_core/_futures.py
```diff
@@ -0,0 +1,218 @@
...
+
+        :raises FutureAlreadyFinished: if future was already given a result or exception.
+        """
+        match self.status:
+            case Future.Status.PENDING:
+                return self._cancel_scope.cancel()
+            case Future.Status.FINISHED:
+                raise FutureAlreadyFinished("future has already finished")
+            case Future.Status.FAILED:
+                raise FutureAlreadyFinished("future already failed")
```
tapetersen: Do we want/need to throw an exception here? In most cases I've used something like this I don't really care if it was finished in some other way, just that it doesn't continue if it was active.

[`asyncio.Future.cancel`](https://docs.python.org/3/library/asyncio-future.html#asyncio.Future.cancel) allows it and just indicates by return value if it was actually cancelled or not (Not sure we need to implement the return until someone asks for it though)

src/anyio/_core/_futures.py
tapetersen: Thinking more about it and especially with the realisation from @davidbrochart that the `CancelScope` as is doesn't work for many waiters I'm really unsure a cancel-scope is the right way to handle this.

Wouldn't it be better and simpler to only wait on the event and let `cancel()` just set the event to release waiters anyway?

If we're not trying to support multiple waiters the point may be moot but it feels like it could/should.

src/anyio/_core/_futures.py
EmmanuelNiyonshuti: > Could be worthwile but since it's an async method it should probably then do a checkpoint anyway to uphold the variant of all async methods passing through at least 1 checkpoint (guaranteed for trio and I think anyio tries to uphold is as well)

hmm...Does it mean we were to for example do a checkpoint right before this? https://github.com/agronholm/anyio/blob/ffe91331adb912c5d150f5d373f7cd28a0e96a62/src/anyio/_backends/_asyncio.py#L423-L426

it does seems awaiting the future more than once will not work with this current  `future.wait` implementation.

EmmanuelNiyonshuti: > > Could be worthwile but since it's an async method it should probably then do a checkpoint anyway to uphold the variant of all async methods passing through at least 1 checkpoint (guaranteed for trio and I think anyio tries to uphold is as well)
> 
> hmm...Does it mean we were to for example do a checkpoint right before this?
> 
> https://github.com/agronholm/anyio/blob/ffe91331adb912c5d150f5d373f7cd28a0e96a62/src/anyio/_backends/_asyncio.py#L423-L426
> 
> it does seems awaiting the future more than once will not work with this current `future.wait` implementation.


No, probably wouldn't make any sense, I think? Yet I don't understand how `future.wait`  being async would prevent it from the problem considering of how it is implemented?.

tapetersen: My comment regarding async was only about your suggestion, which I understood was meant as an optimization of exiting early if the event was already set, not to avoid the problem of entering the cancel scope multiple times.

In that case the reasons outlined from trio [here](https://trio.readthedocs.io/en/stable/reference-core.html#checkpoints would apply)

For handling multiple waiters something else needs to be done (it's not my PR but as hinted my thoughts would be to remove the scope and just simply wait on the event and let `cancel()` set it.)

EmmanuelNiyonshuti: I appreciate the clarification. Awaiting multiple times on trio backend leads to  `RuntimeError: Each CancelScope may only be used for a single 'with' block`  but not on AsyncIO. I was playing around with this simple test:
```
async def test_result(self) -> None:
    future: Future[int] = Future()
    future.set_result(1)
    result = await future
    result2 = await future
    assert result == 1
    assert result2 == 1
```
Hence arose my assumption to check if the event is already set and immediately return the set value. 

I agree, removing the scope and simply await an event. will also make it more aligned with `TaskHandle.wait`

src/anyio/_core/_futures.py
Vizonex: Depends on what else could happen I was trying to make it so that if anything fails within this scope the program could successfully pick it up.

davidbrochart:
Futures can be used like this:
```py
def receive() -> Future:
    ...
```
But then to effectively get some data I need to:
```py
data = (await receive().wait()).return_value
```
Which is quite verbose. I'm wondering if `.wait()` could return the value if there was not exception and if the future was not cancelled. Or even if directly awaiting the future could return the value?

tapetersen:
@davidbrochart I havn't tested but they implement `__await__` so I think they should be awaitable directly for their value (a main difference to `asyncio.Future` is that they don't raise the original exception on failure but for the happy-case I think the intention is to be compatible directly)

Vizonex:
> @davidbrochart I havn't tested but they implement `__await__` so I think they should be awaitable directly for their value (a main difference to `asyncio.Future` is that they don't raise the original exception on failure but for the happy-case I think the intention is to be compatible directly)

@tapetersen 
Having `__await__` was the plan from the very beginning when I started writing this new object. It allows users to lazily wait for the result to come back or raise an exception on failure.

Edit: as for `wait()` it allows uses to handle exceptions beforehand if any are to appear out of the blue.

Vizonex:
> Futures can be used like this:
> 
> ```python
> def receive() -> Future:
>     ...
> ```
> 
> But then to effectively get some data I need to:
> 
> ```python
> data = (await receive().wait()).return_value
> ```
> 
> Which is quite verbose. I'm wondering if `.wait()` could return the value if there was not exception and if the future was not cancelled. Or even if directly awaiting the future could return the value?

@davidbrochart I wrote you a little example if it helps you understand what came to my head when writing the `.wait` implementation it's basically the workaround I made so that callbacks wouldn't be required
```python
     f = Future()
    # Let's wait for the future to be done but not trigger the exception to be raised
    await f.wait()
    # Now Lets see if we have an exception
    if exception := f.exception:
        # Uh oh, we do so handle it here...
        ...
    else:
        # Get our result
        result = f.return_value
 ```

**Comment by tapetersen** [comment]
@Vizonex Yeah I understand the api and really like that there is both the api for waiting for any completion state as well as the direct awaitable path for the happy-case.

I'm very much for the divergence with respect to asyncio's decision to re-raise the unwrapped exception for the same reason it was made like this in the TaskHandle case.

I'm curious to your opinion regarding the comments about `CancelScope` use for multiple waiters as well as the exception handling when cancelled.

**Comment by Vizonex** [comment]
> @Vizonex Yeah I understand the api and really like that there is both the api for waiting for any completion state as well as the direct awaitable path for the happy-case.
> 
> I'm very much for the divergence with respect to asyncio's decision to re-raise the unwrapped exception for the same reason it was made like this in the TaskHandle case.
> 
> I'm curious to your opinion regarding the comments about `CancelScope` use for multiple waiters as well as the exception handling when cancelled.

@tapetersen it was meant to just be used with the `.wait` func but if you wanted to I could try going in and making it multi-use otherwise the reason it's in place is to simply cancel waiting for the Future object.

Edit: the result_value does raise the exception if it has it on hand.

**Comment by davidbrochart** [comment]
> @davidbrochart I havn't tested but they implement `__await__` so I think they should be awaitable directly for their value (a main difference to `asyncio.Future` is that they don't raise the original exception on failure but for the happy-case I think the intention is to be compatible directly)

Ah my bad, I didn't see this was already there. There is one thing though that I think is very different from asyncio, if I directly await a future where an exception is set, it doesn't raise the exception but instead `TaskFailed("the future raised an exception")`.

**Comment by davidbrochart** [comment]
Sorry it's late here, I realize that it's exactly what you said :)

**Comment by davidbrochart** [comment]
My initial idea was that futures should be built on top of the new `TaskGroup.create_task`, but this implementation is very different. Isn't there some advantage in having a future bound to a task group?

**Comment by agronholm** [comment]
> My initial idea was that futures should be built on top of the new `TaskGroup.create_task`, but this implementation is very different. Isn't there some advantage in having a future bound to a task group?

What would that be? I can't think of any.

**Comment by davidbrochart** [comment]
I think that futures are very much like tasks: I can await them and get their result value, they can raise exceptions, and they can be cancelled. An exception raised in a future would cancel other tasks in the task group, and the exception will bubble up. And the other way around, an exception in a task would cancel other futures in the task group.
Or do you think that futures are more comparable to events? And that if I want to have the benefits above I should wrap a future in a task?

**Comment by agronholm** [comment]
Futures are events with values, nothing more. Task groups are a thing because tasks run independently and need a host task for coordination and accountability (for exceptions).

**Comment by davidbrochart** [comment]
> Futures are events with values, nothing more.

Well there is more, since they can raise an exception. Also, an event cannot be cancelled, while a future can.

> Task groups are a thing because tasks runs independently and need a host task for coordination and accountability (for exceptions).

An exception raised in a future would benefit bubbling up the stack, wouldn't it?

**Comment by tapetersen** [comment]
@davidbrochart 
> > Futures are events with values, nothing more.
> 
> Well there is more, since they can raise an exception. Also, an event cannot be cancelled, while a future can.

That's why they are an Event+Value. The event can only be completed/set/finished and the Value tells you how. One of the ways is that whatever process was responsible for producing the Value was aborted.

> 
> > Task groups are a thing because tasks runs independently and need a host task for coordination and accountability (for exceptions).
> 
> An exception raised in a future would benefit bubbling up the stack, wouldn't it?

What stack? The whole point of Futures are to be a low-level primitive to be able to provide a deferred result that `doesn't` have a single responsible Task producing it (most commonly needed when for example handling multiple outstanding network requests where a single task needs to receive responses and match that to requests by some id).

Whatever is waiting on the future will of course receive an exception that will bubble up but where a Task get's an exception set by it's running code raising it and that needing to be passed on further to its caller/starter/(or at least the TaskGroups creator) a Future is created without any specific code being responsible and only gets an exception by it being set explicitly, it already knows why and is probably already in code handling that exception.

While the api discussed has some differences to `asyncio.Futures` the general idea is the same and the correspondence between a `Future` and a `Task` is the same.

**Comment by agronholm** [comment]
> > Futures are events with values, nothing more.
> 
> Well there is more, since they can raise an exception. Also, an event cannot be cancelled, while a future can.

Fair point, I suppose.

> > Task groups are a thing because tasks runs independently and need a host task for coordination and accountability (for exceptions).
> 
> An exception raised in a future would benefit bubbling up the stack, wouldn't it?

That sentence made no sense.

**Comment by davidbrochart** [comment]
> That sentence made no sense.

What I mean is that when an exception is set in a future, ensuring that it is raised in a task group might be preferable. Currently if nobody is awaiting the future then the exception is ignored. It's true that since the exception is set explicitly, the code that sets the exception can also raise it, but maybe not in the context the future is used in.

**Comment by agronholm** [comment]
> > That sentence made no sense.
> 
> What I mean is that when an exception is set in a future, ensuring that it is raised in a task group might be preferable. Currently if nobody is awaiting the future then the exception is ignored. It's true that since the exception is set explicitly, the code that sets the exception can also raise it, but maybe not in the context the future is used in.

If nobody is awaiting for the future, then I guess nobody cared? But more importantly, since futures are not tied to any task groups (unlike tasks, you can definitely use futures w/o task groups), it doesn't make sense to artificially bind them to any task group construct.

**Comment by davidbrochart** [comment]
> But more importantly, since futures are not tied to any task groups (unlike tasks, you can definitely use futures w/o task groups), it doesn't make sense to artificially bind them to any task group construct.

Yes, it's a constraint that is not needed. I was just wondering if adding this constraint would bring benefits. Like Trio adds constraints over asyncio, but it brings value.

**Comment by agronholm** [comment]
> > But more importantly, since futures are not tied to any task groups (unlike tasks, you can definitely use futures w/o task groups), it doesn't make sense to artificially bind them to any task group construct.
> 
> Yes, it's a constraint that is not needed. I was just wondering if adding this constraint would bring benefits. Like Trio adds constraints over asyncio, but it brings value.

Well, that added value remains to be demonstrated in this case.

**Comment by tapetersen** [comment]
@Vizonex I made a try implementing the changes regarding:
 * multiple waiters
 * CancelScope replacement
 * Relaxed cancel() semantics
I sent them as a PR to your branch https://github.com/Vizonex/anyio/pull/1.

Feel free to use or not I mostly wanted to ensure they actually worked as my suggestions intended.

**Comment by davidbrochart** [comment]
@tapetersen I used your implementation in https://github.com/QuantStack/zmq-anyio/pull/37, all tests pass.

**Comment by tapetersen** [comment]
> @tapetersen I used your implementation in [QuantStack/zmq-anyio#37](https://github.com/QuantStack/zmq-anyio/pull/37), all tests pass.

Happy it works :). If you did it as a test to see if the current design is workable it's great but I would be careful to commit to the api as there are still questions regarding the context-manager part that may need some refactoring if made a forced api.

Also like I wrote I intended that PR as a check that my comments and suggestions were workable and if they end up merged their current form or at all in this api I leave open.

Also I don't mind my changes here being vendored inline in your project I can't speak for @Vizonex and it's an idea to be careful about licenses and copyright (if intended for release and not just for testing the API) .

Anyhow it's good to know the current state is useful enough for a reasonably large real-world usecase.

**Comment by davidbrochart** [comment]
I don't intend to merge that PR, it was just to show that it works well in a project that uses futures extensively. I used your version because at least I need to be able to await a future multiple times.

**Comment by agronholm** [comment]
> I don't intend to merge that PR, it was just to show that it works well in a project that uses futures extensively. I used your version because at least I need to be able to await a future multiple times.

Yes – just like an `Event` can be awaited any number of times, so should a `Future`.

**Comment by davidbrochart** [comment]
> Yes – just like an `Event` can be awaited any number of times, so should a `Future`.

@Vizonex Do you want to update your PR with @tapetersen's suggestions, so that we can await a future multiple times?

**Comment by Vizonex** [comment]
> > Yes – just like an `Event` can be awaited any number of times, so should a `Future`.
> 
> @Vizonex Do you want to update your PR with @tapetersen's suggestions, so that we can await a future multiple times?

@davidbrochart  Sure I can do that. I have no problem with this. Just need a way to visualize what it is that I need to accomplish other than that I'm all clear to make those changes.

**Comment by davidbrochart** [comment]
@Vizonex What's preventing you from diffing/merging @tapetersen's changes into your branch?

**Comment by Vizonex** [comment]
> @Vizonex What's preventing you from diffing/merging @tapetersen's changes into your branch?

@davidbrochart The pull request is pointing to the wrong branch and is going to Vizonex:main instead of Vizonex:future-objects.

**Comment by davidbrochart** [comment]
It should be easy for you to open a PR against the right branch?

**Comment by Vizonex** [comment]
@davidbrochart Problem resolved and has been merged now.

File: src/anyio/_core/_exceptions.py (hunk 7 above)
Comment (davidbrochart) [reply src/anyio/_core/_exceptions.py:194]: It seems you are using `TaskNotFinished` instead.

File: src/anyio/_core/_futures.py [hunk 10]
```diff
@@ -0,0 +1,187 @@
...
+        elif self._cancelled:
+            return Future.Status.CANCELLED
+        elif self._exception is not None:
+            return Future.Status.FAILED
+        else:
+            return Future.Status.FINISHED
+
+    def __await__(self) -> Generator[Any, Any, T]:
+        yield from self.wait().__await__()
+        return self.return_value
```
davidbrochart, line 185: This currently raises `TaskFailed("the future raised an exception")` if an exception was set. Could we raise the actual exception instead?
Also, if the future was cancelled, I think we should raise `get_cancelled_exc_class()` instead of `FutureCancelled("the future was cancelled")`.

Vizonex: @davidbrochart Would something like this be slightly more effective or do we just make a new raiseable error system. I have a few ideas on how to go about this just know I wasn't able to satisfy trio's exception constructor.

1. Idea 1
```python
def __cancelled_error(self, msg: str):
        exc = get_cancelled_exc_class()
        raise FutureCancelled("the future was cancelled") from exc
```
2. Idea 2 Add new exception type attribute

```python

__slots__ = (
        "_cancelled",
        "_cancelled_exc"
        "_result_value",
        "_finished_event",
        "_exception",
        "_name",
    )
    _result_value: T

    def __init__(self, *, name: str | None = None) -> None:
        self._finished_event = Event()
        self._exception: BaseException | None = None
        self._cancelled: bool = False
        self._name = name
        self._cancelled_exc: type[BaseException] = get_cancelled_exc_class()
   
   def __cancelled_error(self, msg: str):
        raise FutureCancelled("the future was cancelled") from self._cancelled_exc()
```

3. Last but not least we could ditch FutureCancelled exception all together (however it could costly on user readability for exception handling). Idea 2 basically modified, I would have to go in and modify all the tests but users would loose readability with any form of error messages due to trio's constraints.
```python
def __cancelled_error(self):
        raise self._cancelled_exc() # XXX: trio.Cancelled has no constructor.
```

davidbrochart: You mean that it is not possible to raise `get_cancelled_exc_class()` on Trio?
What about the exception, do you agree to raise it directly instead of `TaskFailed`?

tapetersen: Wrapping exceptions is done for the same reason as in the TaskHandles. If you reraise an Exception you change its backtrace for all holders. As for not raising CancelledException that is also correct since it's not the task waiting on the future that's cancelled.

tapetersen: @Vizonex I would say the current api regarding exceptions is good and is inline with what was done for Tasks.

@davidbrochart The discussion regarding raising the real Exception and the problems with it was in this thread here https://github.com/agronholm/anyio/pull/1098#discussion_r2987384264 .

davidbrochart: Thanks for the link.
Then this PR looks good to me, apart from [this comment](https://github.com/agronholm/anyio/pull/1146/changes#r3418896897).

davidbrochart:
@Vizonex Do you want to push this PR to the finish line? You still haven't addressed [this comment](https://github.com/agronholm/anyio/pull/1146/changes#r3418896897).

Vizonex:
> @Vizonex Do you want to push this PR to the finish line? You still haven't addressed [this comment](https://github.com/agronholm/anyio/pull/1146/changes#r3418896897).

We could I just think it's a confusing way to handle this exception.

src/anyio/_core/_exceptions.py
Vizonex: Yeah we can delete this exception give me a sec... Work has had me occupied the past couple of days which is why I couldn't get to this right away.

davidbrochart (approved):

Vizonex:
Seems some unrelated items are failing otherwise this PR looks good and ready to merge.

Vizonex:
I stand corrected seems it passed under a simple workflow rerun.

tapetersen:
@davidbrochart I think the main blocker here is really if and what we should do about requiring futures being used as context_managers to avoid having hung tasks.
That's been discussed by both @agronholm and @smurfix .

davidbrochart:
Events are not used with a context manager, and a future is just a little more than an event, so I don't think it requires being used with a context manager either.
A task can hang waiting for an event never being set, just like it can hang waiting for a future result never being set too, right?

davidbrochart:
> anyio should not have unbounded futures, for the same reasons why it doesn't have unbound tasks. A future thus should be attached to a taskgroup (the one that's responsible for fulfilling it) so that it's auto-cancelled when the taskgroup ends without setting a result. (This is also a problem with events, but in Real Code the problem ends up being worse with futures because they have more state.)
> 
> Also, anyio now has task handles. Those behave like futures in some meaningful way. I'd like to be _very_ sure that we **really** need a separate `Future` class.

Since we didn't implement futures on top of task handles, nor did we bind futures to task groups, we kind of implicitly said futures were closer to events than to tasks, which IMO goes in favor of not requiring a context manager. What do you think @agronholm ?

davidbrochart:
BTW @Vizonex we may want a section on futures in [the documentation](https://anyio.readthedocs.io/en/stable/synchronization.html).

agronholm:
Yeah, I'm torn about what to do here. There are merits to both requiring context managers and not requiring them.

Vizonex:
> Yeah, I'm torn about what to do here. There are merits to both requiring context managers and not requiring them.

An idea I have for that is having a parent system control this upon a single `__aenter__` when `__aexit__` happens it should invalidate these awaitable containers if they didn't complete. Maybe this would be an idea to implement?

davidbrochart:
> Yeah, I'm torn about what to do here. There are merits to both requiring context managers and not requiring them.

Since it will always be a trade-off, why not go with the simplest, i.e. not requiring a context manager? In the end, a future is a low-level object that users should use with care. I was pretty convinced [after our discussion](https://github.com/agronholm/anyio/pull/1146#issuecomment-4772109651) that futures should not be bound to task groups, or to any other context manager.

m-u-xyz:
If there's no consensus about *requiring* a CM for futures, adding an optional one should be straightforward: add a `close` method to the future that sets the exception to `FutureCancelled` if it doesn't yet have a value. Presto, you can use the existing `with closing()` construct (or its async version).

davidbrochart:
I was thinking about that, it seems to be a good compromise.

tapetersen:
@smurfix `.cancel()` already does that  essentially (it does nothing if already completed in any other way). Do we really want `close()` as well? Why not implement the full `ContextManager` protocol with just an `__enter__(self) -> Self: return self` in that case?

I really do like having some structure on the use of futures but it may be too fiddly to have to enter each one of them since typical usage is some kind of multiplexer that creates many that it later needs to match up through some id. That usage also doesn't really match up with the typical FILO stack-order of matching `__enter__` / `__exit__` calls.

An alternative would be to have a `FutureScope` that when entered can be used to produce Futures and cancels them all at once when it goes out of scope (or possibly throws `ClosedResource` or something similar).

davidbrochart:
> An alternative would be to have a `FutureScope`

I'm sorry but that made me think of https://www.futuroscope.com :smile:

davidbrochart:
> An alternative would be to have a `FutureScope` that when entered can be used to produce Futures and cancels them all at once when it goes out of scope (or possibly throws `ClosedResource` or something similar).

That would be the equivalent of `create_task_group()` for futures, maybe we should call it `FutureGroup` then?
I like the idea, but we should also allow a `Future` to be created without it.

agronholm:
I see no reason to group futures together like tasks. And if we are two have two differing semantics for futures, there should be two different classes for them too. But I'm currently leaning towards a simpler event + value future based on the (rather long) discussion so far.

docs/api.rst
```diff
@@ -48,6 +48,10 @@ Task groups
 .. autoclass:: anyio.abc.TaskGroup
 .. autoclass:: anyio.abc.TaskStatus
 
+Future objects
+--------------
+.. autoclass:: anyio.Future
```
agronholm: I think this belongs in the `Synchronization` section below.

src/anyio/_core/_futures.py
```diff
@@ -0,0 +1,187 @@
...
+
+        :raises TaskNotFinished: if future is still pending
+        :raises FutureCancelled: if future was cancelled
+        :returns: None if future succeeds with a result sent instead
+            otherwise this will be an exception
+        """
+
+        match self.status:
+            case Future.Status.PENDING:
+                raise TaskNotFinished("the future has not finished yet")
```
agronholm: This rubs me the wrong way. I don't think we should be raising a task-related exception here.

src/anyio/_core/_futures.py
agronholm: Is this matter settled, or do we need to bikeshed on this name? I don't think we should have a *property* named `result` when there's a *method* with the same name in `asyncio.Future`. That would be super confusing.

src/anyio/_core/_futures.py
```diff
@@ -0,0 +1,187 @@
...
+        The return value of the future.
+
+        :raises TaskNotFinished: if the future has not finished yet
+        :raises FutureCancelled: if the future was cancelled
+        :raises TaskFailed: if the future raised an exception
+
+        """
+        match self.status:
+            case Future.Status.PENDING:
+                raise TaskNotFinished("the future has not finished yet")
```
agronholm: As above, this needs its own exception class.

tests/test_futures.py
```diff
@@ -0,0 +1,151 @@
...
+    FutureCancelled,
+    TaskCancelled,
+    TaskFailed,
+    TaskNotFinished,
+    create_task_group,
+)
+from anyio.lowlevel import (
+    # cancel_shielded_checkpoint,
+    checkpoint,
+    # checkpoint_if_cancelled,
```
agronholm: Don't leave in commented-out imports.

```suggestion
    checkpoint,
```

agronholm (changes requested):
It's almost there now. I just want to completely separate this primitive from the task-related exceptions.

davidbrochart:
Should we finish this PR without waiting for @Vizonex ?

docs/api.rst
davidbrochart: Done in https://github.com/agronholm/anyio/pull/1146/commits/bba6f8d017d0dd3baa14229bc4a7010a8adcd81c.

src/anyio/_core/_futures.py
davidbrochart: Done in https://github.com/agronholm/anyio/pull/1146/commits/747c54206f9da7cf8a38cc49d65b119e41313509.

src/anyio/_core/_futures.py
davidbrochart: Done in https://github.com/agronholm/anyio/pull/1146/commits/747c54206f9da7cf8a38cc49d65b119e41313509.

src/anyio/_core/_futures.py
davidbrochart: Let's keep `return_value` then.

davidbrochart:
@Vizonex Since these were mechanical, I took the liberty to make the requested changes.
@agronholm What do you think now?

Vizonex:
Sorry for my absence work has become pretty abundant and aggressive for me and sometimes I've been taken for long 8 hour periods every single day.

@agronholm I think the `return_value` could be renamed to `result` that would be a start.

davidbrochart:
I think @agronholm specifically didn't want that, because we would have a `result` property while it is a method in asyncio.

agronholm:
> I think @agronholm specifically didn't want that, because we would have a `result` property while it is a method in asyncio.

Precisely. @Vizonex did you see my earlier request on that?

Vizonex:
> > I think @agronholm specifically didn't want that, because we would have a `result` property while it is a method in asyncio.
> 
> Precisely. @Vizonex did you see my earlier request on that?

@agronholm If I read it correctly but did you want me to turn it into a functional method instead like asyncio has? 

Also should be noted that any further instant changes would need to be done tomorrow when I have extra time on my hands (otherwise pull requests to my branch are encouraged). Like I've mentioned in a previous comment. My work needs me in the afternoons and through a portion of my evening.

agronholm:
> > > I think @agronholm specifically didn't want that, because we would have a `result` property while it is a method in asyncio.
> > 
> > 
> > Precisely. @Vizonex did you see my earlier request on that?
> 
> @agronholm If I read it correctly but did you want me to turn it into a functional method instead like asyncio has?
> 
> Also should be noted that any further instant changes would need to be done tomorrow when I have extra time on my hands (otherwise pull requests to my branch are encouraged). Like I've mentioned in a previous comment. My work needs me in the afternoons and through a portion of my evening.

Well, here's my take:

- A `@property` named `result` is unacceptable: it has the same name as `asyncio.Future.result` but with different semantics
- A `result()` method might be acceptable, although the semantics differ from asyncio regarding the exception that is raised (but to a smaller degree than the property above)
- If we go ahead with a `@property` named `return_value`, then I think `set_result()` needs to be renamed into `set_return_value`

davidbrochart:
> * If we go ahead with a `@property` named `return_value`, then I think `set_result()` needs to be renamed into `set_return_value`

Could we have a setter for `return_value` instead of `set_return_value` then?

agronholm:
> > * If we go ahead with a `@property` named `return_value`, then I think `set_result()` needs to be renamed into `set_return_value`
> 
> Could we have a setter for `return_value` instead of `set_return_value` then?

And likewise for `exception`. That did cross my mind. I only had doubts because the setters only work once, but I don't think it's actually a problem.

agronholm:
Yeah, `@return_value` is the only way forward as it aligns with `TaskHandle`.

tapetersen:
Do we really want property setters rather than explicit methods here? 
They change state and release all waiters together with the case of them failing if called twice or the other has already been set.

To me it really feels like a very natural case of having a readonly property and an explicit setter.

davidbrochart:
An explicit setter can also be called twice, and there would be the same issue?

tapetersen:
> An explicit setter can also be called twice, and there would be the same issue?

Yes but setting a property to the same value can normally be assumed to be an idempotent action and is less expected to raise and/or cause other side-effects than calling a method.

It's not about preventing it but about what is expected.

Setting result/exception here "releases" waiters and for actions triggering that kind of effect I would say the visibility of an explicit function call is warranted clarity.

davidbrochart:
I personally find it quite elegant, and not confusing. A future is a special object, and `exception` and `return_value` have a special meaning for a future.

Vizonex:
> I personally find it quite elegant, and not confusing. A future is a special object, and `exception` and `return_value` have a special meaning for a future.

@davidbrochart Agreed.

agronholm:
> I personally find it quite elegant, and not confusing. A future is a special object, and `exception` and `return_value` have a special meaning for a future.

This is a really minor disagreement. Let's not get hung up on it. @Vizonex we'll use setters for `exception` and `return_value`.

davidbrochart:
Are we all set then?

agronholm:
> Are we all set then?

Ah, I didn't realize you already made the changes! So yes, I think we are.

src/anyio/_core/_futures.py
```diff
@@ -0,0 +1,191 @@
...
+        """
+        The current status of a future.
+
+        Every future starts in the :attr:`~Future.Status.PENDING` state.
+        If a future is cancelled while in this state, it will transition to the
+        :attr:`Future.Status.CANCELLING` state. When the task finishes, it will
+        transition to one of the three final states (
+        :attr:`Future.Status.FINISHED`, :attr:`Future.Status.FAILED`, or
+        :attr:`Future.Status.CANCELLING`) depending on the exception the task
+        raised, if any. No other status transitions will happen.
```
agronholm, line 172: This docstring references nonexistent status enums.

tests/test_futures.py
```diff
@@ -0,0 +1,147 @@
...
+            await fut
+
+    async def test_future_cancelling_with_result(self) -> None:
+        fut: Future[str] = Future()
+        fut.cancel()
+
+        with pytest.raises(FutureCancelled, match=r"future was cancelled"):
+            fut.return_value = "Item"
+
+    async def test_future_cancelling_with_await(self) -> None:
```
agronholm: The `_future` is kind of redundant in the test name given the enclosing class, yes?

agronholm: This also duplicates `test_cancelled_wait`.

src/anyio/_core/_futures.py
```diff
@@ -0,0 +1,191 @@
...
+        """
+
+        match self.status:
+            case Future.Status.PENDING:
+                raise FutureNotFinished("the future has not finished yet")
+            case Future.Status.FINISHED:
+                return None
+            case Future.Status.CANCELLED:
+                raise FutureCancelled("the future was cancelled")
+            case Future.Status.FAILED:
```
agronholm, line 113: There is no test coverage for the failure states.

src/anyio/_core/_exceptions.py
```diff
@@ -175,3 +175,31 @@ class TaskNotFinished(Exception):
     Raised when attempting to access the return value or exception of a
     :class:`.TaskHandle` that is still pending completion.
     """
+
+
+class FutureCancelled(Exception):
```
agronholm: Given how `TaskCancelled` inherits from `TaskFailed`, should this also inherit from `FutureFailed`?

agronholm (changes requested):
I found some additional issues on a deeper inspection.

davidbrochart:
@agronholm It should be good now.

agronholm:
> @agronholm It should be good now.

Thanks, I'll review your changes later today.

agronholm (approved):
Thanks @Vizonex and @davidbrochart (and everyone else who participated in the discussion)

davidbrochart:
Thanks especially to @Vizonex, I just did the mechanical changes.

Vizonex:
> Thanks especially to @Vizonex, I just did the mechanical changes.

@davidbrochart To you and @tapetersen especially, you helped me finish this up when my workload was very heavy. This has saved me a lot of time. I hope to see an update on this new object soon and can't wait to use it in future projects going forward :)

PR #1147, opened 2026-05-16
Fix UDPSocket.aclose() returning before FD release on asyncio backend

mattias-arrelid:

## Changes

`UDPSocket.aclose()` on the asyncio backend returned before the underlying
socket FD was actually released. It delegated to
`asyncio.DatagramTransport.close()`, which only *schedules* the FD-closing
`connection_lost` callback via `loop.call_soon` and returns immediately. A
caller binding a fresh socket to the same `(host, port)` on the next line
(the typical `from_socket` pattern: `socket()` → `bind()` →
`UDPSocket.from_socket(sock)`) raced the scheduled close and saw
`EADDRINUSE`.

`create_udp_socket()` happened to mask the bug because its own
`await loop.create_datagram_endpoint(local_addr=...)` yields once, draining
the queued `connection_lost`. The `from_socket` path doesn't yield between
the prior `aclose` and the next sync `bind()`, so it manifested the bug.
Both paths produce the same `UDPSocket` wrapper, so the fix lives there.

Patch: `DatagramProtocol` now sets a `closed_event` from `connection_lost`,
and `UDPSocket.aclose()` (and `ConnectedUDPSocket.aclose()`) awaits it —
honouring `AsyncResource.aclose`'s "close the resource" contract.

The trio backend already releases the FD synchronously
(`_TrioSocketMixin.aclose` calls `trio_socket.close()` which is
synchronous-on-return), so no patch needed there.

## Checklist

### DIFF
--- a/tests/test_sockets.py
+++ b/tests/test_sockets.py
@@ -1695,6 +1695,55 @@ async def test_multi_listener(tmp_path_factory: TempPathFactory) -> None:
 @pytest.mark.network
 @pytest.mark.usefixtures("check_asyncio_bug")
 class TestUDPSocket:
+    async def test_aclose_waits_for_fd_release(
+        self, family: AnyIPAddressFamily
+    ) -> None:
+        """Regression: ``UDPSocket.aclose`` must not return before the FD is released.
+
+        ``AsyncResource.aclose`` is documented as "close the resource".
+        Callers reasonably expect the local ``(host, port)`` to be
+        immediately rebindable once ``async with udp:`` returns.
+
+        The asyncio backend used to delegate ``aclose`` to
+        ``DatagramTransport.close``, which only *schedules* the FD-closing
+        ``connection_lost`` callback via ``loop.call_soon`` and returns
+        immediately. A caller who bound a fresh socket on the next line
+        (the typical ``from_socket`` pattern: ``socket()`` -> ``bind()``
+        -> ``UDPSocket.from_socket(sock)``) raced the scheduled close and
+        saw ``EADDRINUSE`` even though ``lsof`` reported the port free.
+
+        ``create_udp_socket`` happens to mask this because its own
+        ``await loop.create_datagram_endpoint(local_addr=...)`` yields
+        once, which drains the queued ``connection_lost``. The
+        ``from_socket`` path doesn't yield between aclose and rebind, so
+        it manifested the bug. This test exercises ``from_socket``
+        specifically.
+        """
+        host = "127.0.0.1" if family == socket.AF_INET else "::1"
+
+        # Get a port the kernel just freed so we know nothing else has it.
+        probe = socket.socket(family, socket.SOCK_DGRAM)
+        probe.bind((host, 0))
+        port = probe.getsockname()[1]
+        probe.close()
+
+        def bound(host: str, port: int) -> socket.socket:
+            sock = socket.socket(family, socket.SOCK_DGRAM)
+            sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+            sock.bind((host, port))
+            sock.setblocking(False)
+            return sock
+
+        udp_a = await UDPSocket.from_socket(bound(host, port))
+        async with udp_a:
+            pass
+
+        # No yield between aclose returning and rebinding. If aclose
+        # returns before the previous FD is released, this bind fails.
+        udp_b = await UDPSocket.from_socket(bound(host, port))
+        async with udp_b:
+            pass
+
     async def test_extra_attributes(self, family: AnyIPAddressFamily) -> None:
         async with await create_udp_socket(
             family=family, local_host="localhost"
--- a/tests/test_sockets.py
+++ b/tests/test_sockets.py
@@ -1876,6 +1876,29 @@ class TestUDPSocket:
 @pytest.mark.network
 @pytest.mark.usefixtures("check_asyncio_bug")
 class TestConnectedUDPSocket:
+    async def test_aclose_waits_for_fd_release(
+        self, family: AnyIPAddressFamily, free_udp_port: int
+    ) -> None:
+        """Regression: ``ConnectedUDPSocket.aclose`` must not return before the FD is
+        released. Same race as ``UDPSocket.aclose``; see that test for the full
+        explanation.
+        """
+        host = "127.0.0.1" if family == socket.AF_INET else "::1"
+        peer = socket.socket(family, socket.SOCK_DGRAM)
+        peer.bind((host, 0))
+        try:
+            peer_addr = peer.getsockname()
+            for _ in range(2):
+                sock = socket.socket(family, socket.SOCK_DGRAM)
+                sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+                sock.bind((host, free_udp_port))
+                sock.connect(peer_addr)
+                sock.setblocking(False)
+                udp = await ConnectedUDPSocket.from_socket(sock)
+                await udp.aclose()
+        finally:
+            peer.close()
+
     async def test_extra_attributes(self, family: AnyIPAddressFamily) -> None:
         async with await create_connected_udp_socket(
             "localhost", 5000, family=family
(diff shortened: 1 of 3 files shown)


tests/test_sockets.py
```diff
@@ -1695,6 +1695,55 @@ async def handle(stream: SocketStream) -> None:
...
+        it manifested the bug. This test exercises ``from_socket``
+        specifically.
+        """
+        host = "127.0.0.1" if family == socket.AF_INET else "::1"
+
+        # Get a port the kernel just freed so we know nothing else has it.
+        probe = socket.socket(family, socket.SOCK_DGRAM)
+        probe.bind((host, 0))
+        port = probe.getsockname()[1]
+        probe.close()
```
agronholm: How about using the `free_udp_port` fixture instead of this?

tests/test_sockets.py
```diff
@@ -1695,6 +1695,55 @@ async def handle(stream: SocketStream) -> None:
...
+
+        udp_a = await UDPSocket.from_socket(bound(host, port))
+        async with udp_a:
+            pass
+
+        # No yield between aclose returning and rebinding. If aclose
+        # returns before the previous FD is released, this bind fails.
+        udp_b = await UDPSocket.from_socket(bound(host, port))
+        async with udp_b:
+            pass
```
agronholm: This could be vastly simplified.
```suggestion
        for _ in range(2):
            sock = socket.socket(family, socket.SOCK_DGRAM)
            sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
            sock.bind((host, port))
            sock.setblocking(False)
            udp = await UDPSocket.from_socket(sock)
            await udp.aclose()
```

agronholm (changes requested):
Wouldn't the simpler fix be to add a yield point to `UDPSocket.from_socket()`?

mattias-arrelid:
> Wouldn't the simpler fix be to add a yield point to `UDPSocket.from_socket()`?

I don't think that works for this particular failure - happy to be told I'm missing something. The `EADDRINUSE` is raised by the caller's synchronous `sock.bind()`, which runs before `await UDPSocket.from_socket(sock)` is reached:

```python
sock = socket.socket(family, socket.SOCK_DGRAM)
sock.bind((host, port))                          # ← raises here
udp = await UDPSocket.from_socket(sock)          # we never get here
```

Since `from_socket()` requires an already-bound socket, a yield inside it fires after the failure. The only way a yield helps is if it lives in user code between `aclose()` and the next `bind()` - which is exactly the contract `aclose()` shouldn't push onto callers.

The `closed_event` patch matches the trio backend (which already releases the FD synchronously) and the documented `AsyncResource.aclose` semantics ("close the resource"). It also keeps the close logic where the protocol lives, rather than relying on the next caller happening to yield.

If you'd prefer a different shape - e.g. a private `await self._wait_closed()` on the transport wrapper - I'm glad to refactor. But I'd rather not lean on `from_socket` yielding by coincidence, since any sync op between `aclose()` and rebind (a `bind`, an `lsof`, anything) would still race.

(Also: applied the two test simplifications you suggested in 656bbdd - thanks.)

agronholm:
I think I got confused by the initial mention about no yield point in `UDPSocket.from_socket()`. Adding a yield point to `aclose()` might achieve the same thing, but it then relies on asyncio's task scheduling order.

agronholm:
It seems that this PR only fixes the issue for `UDPSocket` but not for `ConnectedUDPSocket`, yes? This should be a one-line fix in `_asyncio.py`, and one additional test in `test_sockets.py`.

mattias-arrelid:
Good catch - pushed another fix + a mirrored regression test.

agronholm (approved):
Looks good now, thanks!

mattias-arrelid:
(Unrelated to this PR) I couldn't find any information about releases in general - do you have any idea of when you'll release a new version?

agronholm:
I'll cut the next release when `TaskHandle` support for `start()` and `start_soon()` has been merged. I'm working on that PR as we speak.

agronholm:
The PyPy job keeps timing out consistently, and it hasn't done so with the other PRs. Could you check locally if you can reproduce this?

mattias-arrelid:
> The PyPy job keeps timing out consistently, and it hasn't done so with the other PRs. Could you check locally if you can reproduce this?

I'm pretty confident this isn't an anyio regression - the trigger is upstream, in the GitHub-hosted Ubuntu 24.04 runner image released **2026-05-13** ([ubuntu24/20260513.135](https://github.com/actions/runner-images/releases/tag/ubuntu24%2F20260513.135)), which bumped:

> PyPy: `3.11.15 [PyPy 7.3.21]` → `3.11.15 [PyPy 7.3.22]`

`actions/setup-python@v6` prefers whatever's already in `/opt/hostedtoolcache/PyPy/...`, so every `pypy-3.11` job on `ubuntu-latest` started picking up 7.3.22 after that image rebuilt.

The fingerprint matches across runs (all install otherwise identical packages - only PyPy differs):

| Run | Date | PyPy | `[trio]` PASS / FAIL | Outcome |
|---|---|---|---|---|
| master `c781e29` ([job](https://github.com/agronholm/anyio/actions/runs/25697339592/job/75448794901)) | May 11 | **7.3.21** | 487 / 0 | green, 3 min |
| future-objects ([job](https://github.com/agronholm/anyio/actions/runs/25882939285/job/76067308597)) | May 14 | **7.3.22** | 1 / 297 | 10-min timeout |
| this PR ([job](https://github.com/agronholm/anyio/actions/runs/25989614470/job/76393950403)) | May 17 | **7.3.22** | 1 / 290 | 10-min timeout |

Every pypy CI run on every branch since May 13 fails the same way.

**Verified locally** by downloading PyPy 7.3.21 from `downloads.python.org` and re-running with identical package versions (trio 0.33.0, pytest 9.0.3, etc.):

```
$ .venv-pypy-7321/bin/python --version
Python 3.11.15 (2a098039edb5, Mar 12 2026, 17:47:28)
[PyPy 7.3.21 with GCC Apple LLVM 16.0.0 (clang-1600.0.26.6)]

$ .venv-pypy-7321/bin/python -m pytest tests/test_signals.py -v --timeout=15
...
tests/test_signals.py::test_receive_signals[asyncio]              PASSED
tests/test_signals.py::test_receive_signals[trio]                 PASSED
tests/test_signals.py::test_task_group_cancellation_open[asyncio] PASSED
tests/test_signals.py::test_task_group_cancellation_open[trio]    PASSED
tests/test_signals.py::test_task_group_cancellation_consume[asyncio] PASSED
tests/test_signals.py::test_task_group_cancellation_consume[trio] PASSED
========================= 6 passed, 3 skipped in 0.26s =========================

$ .venv-pypy-7321/bin/python -m pytest tests/test_sockets.py -k test_aclose_waits_for_fd_release
================= 8 passed, 4 skipped, 900 deselected in 1.25s =================
```

Same env on PyPy 7.3.22 hangs on `test_task_group_cancellation_open[asyncio]` and bulk-fails every `[trio]` test.

There are really two separate PyPy 7.3.22 bugs tangled together in the CI log:

1. **All `[trio]` tests fail fast with `trio.TrioInternalError`** - looks like a trio/PyPy interaction around coroutine `cr_frame` introspection (`trio/_core/_run.py:1988` in `spawn_impl`). Each fails in milliseconds, so they don't eat the timeout budget.
2. **`test_task_group_cancellation_open[asyncio]` hangs forever** in `loop.run_forever()` → `selector.select(None)`. This is the one that burns the full 10 minutes. With `--timeout-method=thread` pytest-timeout actually fires and shows `tg.cancel_scope.cancel()` never wakes the event loop on 7.3.22.

Minimal reproducer for #2 (one test, no preceding state needed):

```
pytest tests/test_signals.py::test_task_group_cancellation_open[asyncio] \
    --timeout=15 --timeout-method=thread
```

Both look like they want upstream bug reports against trio + PyPy. For unblocking CI in the meantime, pinning python-version: pypy3.11-v7.3.21 in the workflow would force setup-python@v6 to download 7.3.21 instead of using the runner image's cached 7.3.22.

mattias-arrelid:
(also - what a rabbit hole...)

agronholm:
Wow, thanks for the investigation! I'll have to deal with that later.

agronholm (approved):
Thanks!

PR #1149, opened 2026-05-18
fixes trio backend hangs instead of failing when dynamically accessing async fixture via getfixturevalue

### LINKED ISSUE #1148: Trio backend hangs instead of failing when dynamically accessing async fixture via getfixturevalue
### Things to check first

### AnyIO version

4.13.0

### Python version

`>= 10`

### What happened?

first off, dynamic async fixture access is not available in any async pytest plugin currently, including AnyIO. there's already an enhancement issue https://github.com/agronholm/anyio/issues/720 for that. this is specifically about the trio backend hanging silently instead of failing with an error like the asyncio backend does.
when you try to dynamically access an async fixture on `AsyncIO` backend you hit a `RuntimeError: This event loop is already running` with a `RuntimeWarning: coroutine 'TestRunner._call_in_runner_task' was never awaited`
```
import pytest

@pytest.fixture
def anyio_backend():
       return "asyncio"

@pytest.fixture
async def fixt():
      return 1

@pytest.mark.anyio
async def test_some_fn(request):
    value = request.getfixturevalue('fixt')
    assert value == 1
```
On `trio` backend this test will hang. I would expect AnyIO to fail with an error rather than hanging like the Asyncio backend does.  I suspect the hang might be coming from the main thread blocking from the queue loop
https://github.com/agronholm/anyio/blob/1dd510bfe805247834c5fda71082c54284067555/src/anyio/_backends/_trio.py#L907C9-L909C1
my current reasoning is that the main thread blocks waiting for the test coroutine to return, but the test coroutine has requested a dynamic fixture hence the hang but, I am not sure of how to make the trio test runner fail transparently when this happens.

```
@pytest.fixture
def anyio_backend():
       return "trio"

@pytest.fixture
async def fixt():
      return 1

@pytest.mark.anyio
async def test_some_fn(request):
    value = request.getfixturevalue('fixt')
    assert value == 1
```

EmmanuelNiyonshuti:

## Changes

Fixes #1148

trio backend test runner hangs indefinitely instead of failing when `request.getfixturevalue()` is called on an async fixture mid-test. the asyncio backend raises RuntimeError in the same situation. this PR adds a flag to detect the reentrance and raise a `RuntimeError`.

## Checklist

features

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -2219,6 +2219,9 @@ class TestRunner(abc.TestRunner):
     def get_loop(self) -> AbstractEventLoop:
         return self._runner.get_loop()
 
+    def is_running(self) -> bool:
+        return self._runner.get_loop().is_running()
+
     def _exception_handler(
         self, loop: asyncio.AbstractEventLoop, context: dict[str, Any]
     ) -> None:
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -2220,7 +2220,11 @@ class TestRunner(abc.TestRunner):
         return self._runner.get_loop()
 
     def is_running(self) -> bool:
-        return self._runner.get_loop().is_running()
+        try:
+            asyncio.get_running_loop()
+            return True
+        except RuntimeError:
+            return False
 
     def _exception_handler(
         self, loop: asyncio.AbstractEventLoop, context: dict[str, Any]
--- a/src/anyio/_backends/_trio.py
+++ b/src/anyio/_backends/_trio.py
@@ -859,6 +859,7 @@ class TestRunner(abc.TestRunner):
         self._call_queue: Queue[Callable[[], object]] = Queue()
         self._send_stream: MemoryObjectSendStream | None = None
         self._options = options
+        self._runner_task_running: bool = False
 
     def __exit__(
         self,
@@ -892,6 +893,11 @@ class TestRunner(abc.TestRunner):
         *args: P.args,
         **kwargs: P.kwargs,
     ) -> T_Retval:
+        if self._runner_task_running:
+            raise RuntimeError(
+                "Cannot schedule a coroutine in the test runner when another is already running; "
+                "likely caused by request.getfixturevalue() on an async fixture"
+            )
         if self._send_stream is None:
             trio.lowlevel.start_guest_run(
                 self._run_tests_and_fixtures,
@@ -904,8 +910,12 @@ class TestRunner(abc.TestRunner):
 
         outcome_holder: list[Outcome] = []
         self._send_stream.send_nowait((func(*args, **kwargs), outcome_holder))
-        while not outcome_holder:
-            self._call_queue.get()()
+        self._runner_task_running = True
+        try:
+            while not outcome_holder:
+                self._call_queue.get()()
+        finally:
+            self._runner_task_running = False
 
         return outcome_holder[0].unwrap()
 
--- a/src/anyio/_backends/_trio.py
+++ b/src/anyio/_backends/_trio.py
@@ -872,6 +872,9 @@ class TestRunner(abc.TestRunner):
             while self._send_stream is not None:
                 self._call_queue.get()()
 
+    def is_running(self) -> bool:
+        return self._runner_task_running
+
     async def _run_tests_and_fixtures(self) -> None:
         self._send_stream, receive_stream = create_memory_object_stream(1)
         with receive_stream:
@@ -893,8 +896,6 @@ class TestRunner(abc.TestRunner):
         *args: P.args,
         **kwargs: P.kwargs,
     ) -> T_Retval:
-        if self._runner_task_running:
-            raise RuntimeError("This event loop is already running")
         if self._send_stream is None:
             trio.lowlevel.start_guest_run(
                 self._run_tests_and_fixtures,
--- a/src/anyio/_backends/_trio.py
+++ b/src/anyio/_backends/_trio.py
@@ -894,10 +894,7 @@ class TestRunner(abc.TestRunner):
         **kwargs: P.kwargs,
     ) -> T_Retval:
         if self._runner_task_running:
-            raise RuntimeError(
-                "Cannot schedule a coroutine in the test runner when another is already running; "
-                "likely caused by request.getfixturevalue() on an async fixture"
-            )
+            raise RuntimeError("This event loop is already running")
         if self._send_stream is None:
             trio.lowlevel.start_guest_run(
                 self._run_tests_and_fixtures,
--- a/src/anyio/_backends/_trio.py
+++ b/src/anyio/_backends/_trio.py
@@ -859,7 +859,6 @@ class TestRunner(abc.TestRunner):
         self._call_queue: Queue[Callable[[], object]] = Queue()
         self._send_stream: MemoryObjectSendStream | None = None
         self._options = options
-        self._runner_task_running: bool = False
 
     def __exit__(
         self,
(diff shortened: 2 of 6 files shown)


agronholm (changes requested):
I have a bit of a problem with the test here. It only tests on Trio, and doesn't check _how_ the test fails. `getfixturevalue()` should fail in a uniform manner across backends.

EmmanuelNiyonshuti:
> I have a bit of a problem with the test here. It only tests on Trio, and doesn't check _how_ the test fails. `getfixturevalue()` should fail in a uniform manner across backends.

I may be have misunderstood the check of how the test fails though. I mean I just check that the test failed on both backends, or you mean explicit checking of the Runtime Error?

agronholm:
> > I have a bit of a problem with the test here. It only tests on Trio, and doesn't check _how_ the test fails. `getfixturevalue()` should fail in a uniform manner across backends.
> 
> I may be have misunderstood the check of how the test fails though. I mean I just check that the test failed on both backends, or you mean explicit checking of the Runtime Error?

You added an explicit error message for this, but only on Trio, right?

EmmanuelNiyonshuti:
> > > I have a bit of a problem with the test here. It only tests on Trio, and doesn't check _how_ the test fails. `getfixturevalue()` should fail in a uniform manner across backends.
> > 
> > 
> > I may be have misunderstood the check of how the test fails though. I mean I just check that the test failed on both backends, or you mean explicit checking of the Runtime Error?
> 
> You added an explicit error message for this, but only on Trio, right?

yeah!

EmmanuelNiyonshuti:
they both reacts with a RuntimeError raise but the message is different,

agronholm:
I need to fix that pypy test job ASAP.

agronholm:
> they both reacts with a RuntimeError raise but the message is different,

Could you make that error message uniform, and test for it, please?

EmmanuelNiyonshuti:
well, I was hesitant to make the error message this identical considering that trio reaches this failure via a queue driven host loop while asyncio already has an explicit re-entrancy  check in run_until_complete, I was thinking the implementation path differs to mirror the exact message but since this is explicitly about runner re-entrance I think having them identical is good!

agronholm:
> well, I was hesitant to make the error message this identical considering that trio reaches this failure via a queue driven host loop while asyncio already has an explicit re-entrancy check in run_until_complete, I was thinking the implementation path differs to mirror the exact message but since this is explicitly about runner re-entrance I think having them identical is good!

The error message is identical now, sure, but is it helpful anymore? I was hoping you'd add your original error message to asyncio as well.

EmmanuelNiyonshuti:
> > well, I was hesitant to make the error message this identical considering that trio reaches this failure via a queue driven host loop while asyncio already has an explicit re-entrancy check in run_until_complete, I was thinking the implementation path differs to mirror the exact message but since this is explicitly about runner re-entrance I think having them identical is good!
> 
> The error message is identical now, sure, but is it helpful anymore? I was hoping you'd add your original error message to asyncio as well.

No, it's not really that helpful for the user, and I think I misinterpreted the intent earlier. I interpreted it as making Trio match the existing asyncio error, so I aligned the messages around the current underlying Asyncio `run_until_complete()` behavior.

let me see how I can catch and surface that in the AsyncIO test runner as well.

src/anyio/abc/_testing.py
```diff
@@ -63,3 +63,11 @@ def run_test(
...
         :param kwargs: keyword arguments to call the test function with
         """
+
+    def is_running(self) -> bool:
+        """
+        Check if the test runner is running.
+
+        :return: ``True`` if the coroutine is currently being run, ``False`` otherwise.
+        """
+        return False
```
agronholm: I think this should be an abstract method.
```suggestion
    @abstractmethod
    def is_running(self) -> bool:
        """
        Check if the test runner is running.

        :return: ``True`` if the coroutine is currently being run, ``False`` otherwise.
        """
```

src/anyio/_backends/_trio.py
```diff
@@ -871,6 +872,9 @@ def __exit__(
             while self._send_stream is not None:
                 self._call_queue.get()()
 
+    def is_running(self) -> bool:
+        return self._runner_task_running
```
agronholm: Could this be replaced with a simpler call, `trio.lowlevel.in_trio_run()`? https://trio.readthedocs.io/en/stable/reference-lowlevel.html#trio.lowlevel.in_trio_run

src/anyio/_backends/_asyncio.py
```diff
@@ -2219,6 +2219,9 @@ def __exit__(
     def get_loop(self) -> AbstractEventLoop:
         return self._runner.get_loop()
 
+    def is_running(self) -> bool:
+        return self._runner.get_loop().is_running()
```
agronholm: I'm not sure I want `get_loop()` to be called here, as it triggers the instantiation of an event loop if one is not already active. Perhaps just checking if `self._loop is None` would work?

src/anyio/_backends/_trio.py
EmmanuelNiyonshuti: it will work but doesn't seem to do well with this fixture?https://github.com/agronholm/anyio/blob/b4059c7268ad2649702b24e54b3bfa2788842866/tests/test_fileio.py#L90-L109

agronholm: WDYM? What's special about this fixture?

EmmanuelNiyonshuti: I don't see anything special related to it failing because I used `trio.lowlevel.in_trio_run() ` yet, it seems to basically setting up some filesystem for tests?

agronholm: I'm not sure I follow. What's failing exactly?

EmmanuelNiyonshuti: I get the RuntimeError we raised in  the plugin fixture setup!

```
    raise RuntimeError(
E   RuntimeError: Cannot schedule a coroutine in the test runner while another is already running; likely caused by request.getfixturevalue() on an async fixture.
__________________________________________________________________________________________________________ ERROR at setup of TestPath.test_glob[trio-True] ___________________________________________________________________________________________________________
src/anyio/pytest_plugin.py:130: in wrapper
```

agronholm: Oh...I didn't see any of that in the CI runs here. But how can this be possible? I for one would like to understand.

EmmanuelNiyonshuti: i was still looking into it

EmmanuelNiyonshuti: sorry for taking sometime to respond, was not with my laptop! the issue is that `trio.lowlevel.in_trio_run()` will false positive on nested async fixtures. `in_trio_run()` stays truthy for as long as the guest run is alive on that thread. and thats what is happening to `populated_tmpdir` in test_file_io I pasted above. you can reproduce it with something as simple as:

```python
@pytest.fixture
async def fixt1():
    return 1

@pytest.fixture
async def fixt2(fixt1):
    return 2

@pytest.mark.anyio
async def test_something(fixt2):
    assert fixt2 == 2
```

I think we can instead use `in_trio_task`, which seems to be True only while a coroutine is actively being stepped, and False the moment it suspends?.

agronholm: Sounds good.

EmmanuelNiyonshuti:
Though what does this have to do with the changes I made, It does fine on my local machine yet is failing here?

agronholm:
> Though what does this have to do with the changes I made, It does fine on my local machine yet is failing here?

It's a flaky test I haven't managed to fix yet.

src/anyio/_backends/_asyncio.py
EmmanuelNiyonshuti: > > Though what does this have to do with the changes I made, It does fine on my local machine yet is failing here?
> 
> It's a flaky test I haven't managed to fix yet.

Ah, I see.

agronholm (approved):
Thanks!

PR #1151, opened 2026-05-19
Autodetect backend in `start_blocking_portal()`

Zac-HD:
## Changes

`start_blocking_portal()` and `BlockingPortalProvider` previously defaulted to the asyncio backend, which silently mismatches anywhere the caller is actually using trio. Both now default `backend=None` and detect the active backend in order:

1. `sniffio.current_async_library()` (or a fallback that probes for a running asyncio loop when sniffio is missing);
2. a new private ContextVar (`_pytest_backend_cvar`) set by the anyio pytest plugin's `pytest_runtest_call` hookwrapper, so portals opened from a test pick up the parametrised `anyio_backend`. If pytest is running and the cvar is unset, a warning points the user to the fixture;
3. `sys.modules`: trio-only -> trio, both -> warn and pick asyncio, otherwise asyncio.

My motivation here is writing sync tests for Trio-based Starlette apps, where the `TestClient` defers to `start_blocking_portal()` and I would like to stop writing `backend="Trio"` in hundreds of places.

## Checklist

If this is a user-facing code change, like a bugfix or a new feature, please ensure that you've fulfilled the following conditions (where applicable):

features

### DIFF
--- a/docs/threads.rst
+++ b/docs/threads.rst
@@ -168,6 +168,14 @@ event loop in its own dedicated thread::
 
 .. note:: The event loop is shut down as soon as you exit the context manager.
 
+.. note::
+    If the ``backend`` argument is omitted, the backend is detected automatically:
+    first from :mod:`sniffio` (or a running asyncio loop), then from the
+    ``anyio_backend`` pytest fixture when running under pytest, and finally by
+    inspecting which of ``trio`` and ``asyncio`` are present in :data:`sys.modules`.
+    Pass an explicit ``backend`` argument to silence the warning that is emitted
+    when both backends are imported and no other detection method succeeds.
+
 Spawning tasks
 ++++++++++++++
 
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -14,6 +14,12 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
   functions and classes
 - Added the ``create_task()`` task group method for easier asyncio migration
   and to allow retrieving task return values more easily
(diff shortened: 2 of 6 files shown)


agronholm:
I am not favorable to adding these heuristics, which seem brittle to me. And what happens if another backend is added? I've been contemplating that for a while now. This sounds like the problem is being solved from the wrong end.

Zac-HD:
Do you have another suggestion?  I currently have several hundred tests which look like
```python
def test_something():
    client = TestClient(trio_based_app, backend="trio")  # <-- I'd like autodetection here
    ...
```
and the pytest-fixture autodetection would be perfect.

Happy to delete the `sys.modules` heuristic, but the others seem pretty precise and non-brittle to me.

agronholm:
Couldn't you write a simple wrapper for the test client with similar logic as here?

Zac-HD:
Well, yes, it's possible to move this code to Starlette, but then it'd have to be duplicated in every other library doing something similar - it really does seem to me best for this code to live in anyio.

agronholm:
> Well, yes, it's possible to move this code to Starlette, but then it'd have to be duplicated in every other library doing something similar - it really does seem to me best for this code to live in anyio.

I've never seen a situation like this in any other project, so I'm not convinced it's a general problem.

Zac-HD:
No worries - sounds like I should go propose that Starlette adopt similar logic directly, and if we later find a second project with similar needs I'll let you know.

agronholm:
An autouse fixture in the plug-in would have been a hard sell too, just letting you know. Those get activated even when anyio is just installed, regardless of whether the user is using anyio or not.

Issue #493, opened 2022-11-04
`TaskGroup.start/start_soon` do not accept `Callable[..., Awaitable]`


gschaffner:
538b739b8a33bbdbf305e11d93c8bc34154e50d7 documents that support for this was added, but both of the tests in 2f89d1f762ab0ca53747269e23723469fc549252 fail at runtime. One of them also fails at type checking time, but the other does not (since 538b739b8a33bbdbf305e11d93c8bc34154e50d7 did not update all of the relevant annotations).

I've attempted to add the documented support in 8dec1b13de533d6ee2ac1d6e616c64f4186aefdb, but it's causing `test_task_in_sync_spawn_callback[trio]` to fail. As far as I can tell, `Callable[..., Awaitable]` support won't work on trio since it requires a wrapper (since trio requires `Callable[..., Coroutine]` and does not accept `Callable[..., Awaitable]`) and using a wrapper changes task ID behavior.

See also https://github.com/python-trio/trio-typing/pull/66.

agronholm:
How is it that the test suite passes on GitHub then?

agronholm:
This may be a misunderstanding on your part. The backends don't actually accept arbitrary callables returning awaitables; the intention here  was to amend the annotations to have them accept _coroutine-like_ objects like `<class 'async_generator_asend'>` objects you get from an async generator. These don't register as coroutines (though IMHO they should), so this was the only way to get static type checkers to accept such calls. This was only done because nobody bothered to give such objects a special type I could have checked against.

agronholm:
In that sense, the changelog text may be a bit misleading. I can certainly correct that.

agronholm:
[Here](https://github.com/agronholm/anyio/blob/master/tests/test_taskgroups.py#L780-L799) is a test that runs a non-coroutine awaitable (an async generator's `asend` object).

agronholm:
Yeah, bad times all over because of the neglected `asend` type that is a non-citizen in typeshed. What do you want to do about this?

agronholm:
I'm still thinking about what to do. I didn't plan to expand the capabilities of `anyio.run()` or friends, just amend the annotations to accommodate the things they can already do (but where static type checkers report errors).

agronholm:
I looked at trio-typing, and their annotations match the new ones in AnyIO. The ones in typeshed, however, do not.

Perhaps the best option is to have permissive annotations, but better document what you can pass to these functions? Right now I'm not inclined to merge #567.

[...]

PR #230, opened 2021-03-07
cancel host asyncgen

graingert:

PR #1153, opened 2026-05-23
Added task handle support to start() and start_soon()

agronholm:

## Changes

Adds `TaskHandle` support to `TaskGroup.start_soon()` and `TaskGroup.start()` as per the group consensus.
The latter method has a feature flag for this to preserve backwards compatibility.

## Checklist

features

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -923,7 +923,7 @@ class TaskGroup(abc.TaskGroup):
             )
 
         if context is not None:
-            return context.run(partial(self._spawn, coro, name=name))
+            return context.run(self._spawn, coro, name=name)
         else:
             return self._spawn(coro, name=name)
 
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -380,6 +380,22 @@ def is_anyio_cancellation(exc: CancelledError) -> bool:
 
 
 class CancelScope(BaseCancelScope):
+    __slots__ = (
+        "_active",
+        "_cancel_called",
+        "_cancel_handle",
+        "_cancel_reason",
+        "_cancelled_caught",
+        "_child_scopes",
+        "_deadline",
+        "_host_task",
+        "_parent_scope",
+        "_pending_uncancellations",
+        "_shield",
+        "_tasks",
+        "_timeout_handle",
+    )
+
     def __new__(
         cls, *, deadline: float = math.inf, shield: bool = False
     ) -> CancelScope:
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -352,11 +352,7 @@ def _task_started(task: asyncio.Task) -> bool:
     # task list
     coro = task.get_coro()
     assert coro is not None
-    try:
-        return getcoroutinestate(coro) in (CORO_RUNNING, CORO_SUSPENDED)
-    except AttributeError:
-        # task coro is async_genenerator_asend https://bugs.python.org/issue37771
-        raise Exception(f"Cannot determine if task {task} has started or not") from None
+    return getcoroutinestate(coro) in (CORO_RUNNING, CORO_SUSPENDED)
 
 
 #
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -2570,11 +2570,11 @@ class AsyncIOBackend(AsyncBackend):
     @classmethod
     def run_async_from_thread(
         cls,
-        func: Callable[[Unpack[PosArgsT]], Awaitable[T_Retval]],
+        func: Callable[[Unpack[PosArgsT]], Coroutine[Any, Any, T_co]],
         args: tuple[Unpack[PosArgsT]],
         token: object,
-    ) -> T_Retval:
-        async def task_wrapper() -> T_Retval:
+    ) -> T_co:
+        async def task_wrapper() -> T_co:
             __tracebackhide__ = True
             if scope is not None:
                 task = cast(asyncio.Task, current_task())
@@ -2597,7 +2597,7 @@ class AsyncIOBackend(AsyncBackend):
         context = copy_context()
         context.run(set_current_async_library, "asyncio")
         scope = getattr(threadlocals, "current_cancel_scope", None)
-        f: concurrent.futures.Future[T_Retval] = context.run(
+        f: concurrent.futures.Future[T_co] = context.run(
             asyncio.run_coroutine_threadsafe, task_wrapper(), loop=loop
         )
         return f.result()
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -32,7 +32,7 @@ from collections.abc import (
     Sequence,
 )
 from concurrent.futures import Future
-from contextlib import AbstractContextManager, suppress
+from contextlib import AbstractContextManager
 from contextvars import Context, copy_context
 from dataclasses import dataclass, field
 from functools import partial, wraps
@@ -40,7 +40,6 @@ from inspect import (
     CORO_RUNNING,
     CORO_SUSPENDED,
     getcoroutinestate,
-    iscoroutine,
 )
 from io import IOBase
 from os import PathLike
@@ -53,6 +52,7 @@ from typing import (
     IO,
     TYPE_CHECKING,
     Any,
+    Literal,
     ParamSpec,
     TypeVar,
     cast,
@@ -92,6 +92,7 @@ from .._core._synchronization import (
 )
 from .._core._synchronization import Semaphore as BaseSemaphore
 from .._core._tasks import CancelScope as BaseCancelScope
+from .._core._tasks import TaskHandle
 from ..abc import (
     AsyncBackend,
     IPSockAddrType,
@@ -100,6 +101,7 @@ from ..abc import (
     UNIXDatagramPacketType,
 )
 from ..abc._eventloop import StrOrBytesPath
+from ..abc._tasks import call_for_coroutine, get_callable_name
 from ..lowlevel import RunVar
 from ..streams.memory import MemoryObjectReceiveStream, MemoryObjectSendStream
 
@@ -298,6 +300,7 @@ else:
 
 
 T_Retval = TypeVar("T_Retval")
+T_co = TypeVar("T_co", covariant=True)
 T_contra = TypeVar("T_contra", contravariant=True)
 PosArgsT = TypeVarTuple("PosArgsT")
 P = ParamSpec("P")
@@ -334,12 +337,6 @@ def find_root_task() -> asyncio.Task:
             return cast(asyncio.Task, cancel_scope._host_task)
 
     return task
-
-
-def get_callable_name(func: Callable) -> str:
-    module = getattr(func, "__module__", None)
-    qualname = getattr(func, "__qualname__", None)
-    return ".".join([x for x in (module, qualname) if x])
 
 
 #
@@ -816,11 +813,10 @@ class TaskGroup(abc.TaskGroup):
 
     def _spawn(
         self,
-        func: Callable[[Unpack[PosArgsT]], Awaitable[Any]],
-        args: tuple[Unpack[PosArgsT]],
+        coro: Coroutine[Any, Any, T_co],
         name: object,
         task_status_future: asyncio.Future | None = None,
-    ) -> asyncio.Task:
+    ) -> TaskHandle[T_co]:
         def task_done(_task: asyncio.Task) -> None:
             if sys.version_info >= (3, 14) and self.cancel_scope._host_task is not None:
                 asyncio.future_discard_from_awaited_by(
@@ -868,39 +864,23 @@ class TaskGroup(abc.TaskGroup):
                     RuntimeError("Child exited without calling task_status.started()")
                 )
 
-        if not self._entered or not self.cancel_scope._active:
-            raise RuntimeError(
-                "This task group is not active; no new tasks can be started."
-            )
-
-        kwargs = {}
         if task_status_future:
             parent_id = id(current_task())
-            kwargs["task_status"] = _AsyncioTaskStatus(
-                task_status_future, id(self.cancel_scope._host_task)
-            )
         else:
             parent_id = id(self.cancel_scope._host_task)
 
-        coro = func(*args, **kwargs)
-        if not iscoroutine(coro):
-            prefix = f"{func.__module__}." if hasattr(func, "__module__") else ""
-            raise TypeError(
-                f"Expected {prefix}{func.__qualname__}() to return a coroutine, but "
-                f"the return value ({coro!r}) is not a coroutine object"
-            )
-
-        name = get_callable_name(func) if name is None else str(name)
+        handle = TaskHandle(coro, name)
         loop = asyncio.get_running_loop()
+        wrapper_coro = handle._run_coro()
         if (
             (factory := loop.get_task_factory())
             and getattr(factory, "__code__", None) is _eager_task_factory_code
             and (closure := getattr(factory, "__closure__", None))
         ):
             custom_task_constructor = closure[0].cell_contents
-            task = custom_task_constructor(coro, loop=loop, name=name)
+            task = custom_task_constructor(wrapper_coro, loop=loop, name=handle.name)
         else:
-            task = create_task(coro, name=name)
+            task = loop.create_task(wrapper_coro, name=handle.name)
 
         # Make the spawned task inherit the task group's cancel scope
         _task_states[task] = TaskState(
@@ -912,35 +892,62 @@ class TaskGroup(abc.TaskGroup):
             asyncio.future_add_to_awaited_by(task, self.cancel_scope._host_task)
 
         task.add_done_callback(task_done)
-        return task
+        return handle
 
-    def start_soon(
+    def create_task(
         self,
-        func: Callable[[Unpack[PosArgsT]], Awaitable[Any]],
-        *args: Unpack[PosArgsT],
+        coro: Coroutine[Any, Any, T_co],
+        *,
         name: object = None,
-    ) -> None:
-        self._spawn(func, args, name)
+        context: Context | None = None,
+    ) -> TaskHandle[T_co]:
+        if not isinstance(coro, Coroutine):
+            raise TypeError(f"expected a coroutine, got {coro.__class__.__qualname__}")
+
+        if not self._entered or not self.cancel_scope._active:
+            coro.close()
+            raise RuntimeError(
+                "This task group is not active; no new tasks can be started."
+            )
+
+        if context is not None:
+            return context.run(partial(self._spawn, coro, name=name))
+        else:
+            return self._spawn(coro, name=name)
 
     async def start(
-        self, func: Callable[..., Awaitable[Any]], *args: object, name: object = None
+        self,
+        func: Callable[[Unpack[PosArgsT]], Coroutine[Any, Any, T_co]],
+        *args: Unpack[PosArgsT],
+        name: object = None,
+        return_handle: Literal[False] | Literal[True] = False,
     ) -> Any:
         future: asyncio.Future = asyncio.Future()
-        task = self._spawn(func, args, name, future)
+        final_name = get_callable_name(func, name)
+        task_status = _AsyncioTaskStatus(future, id(self.cancel_scope._host_task))
+        coro = call_for_coroutine(func, args, task_status=task_status)
+        handle = self._spawn(coro, final_name, future)
 
         # If the task raises an exception after sending a start value without a switch
         # point between, the task group is cancelled and this method never proceeds to
         # process the completed future. That's why we have to have a shielded cancel
         # scope here.
         try:
-            return await future
-        except CancelledError:
-            # Cancel the task and wait for it to exit before returning
-            task.cancel()
-            with CancelScope(shield=True), suppress(CancelledError):
-                await task
+            await future
+        except BaseException:
+            if handle.status is TaskHandle.Status.PENDING:
+                # Cancel the task and wait for it to exit before returning
+                handle.cancel()
+                with CancelScope(shield=True):
+                    await handle.wait()
 
             raise
+
+        if return_handle:
+            handle._start_value = future.result()
+            return handle
+        else:
+            return future.result()
 
 
 #
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -918,6 +918,11 @@ class TaskGroup(abc.TaskGroup):
         name: object = None,
         return_handle: Literal[False] | Literal[True] = False,
     ) -> Any:
+        if not self._entered or not self.cancel_scope._active:
+            raise RuntimeError(
+                "This task group is not active; no new tasks can be started."
+            )
+
         future: asyncio.Future = asyncio.Future()
         final_name = get_callable_name(func, name)
         task_status = _AsyncioTaskStatus(future, id(self.cancel_scope._host_task))
--- a/src/anyio/_backends/_trio.py
+++ b/src/anyio/_backends/_trio.py
@@ -84,7 +84,7 @@ from .._core._tasks import CancelScope as BaseCancelScope
 from .._core._tasks import TaskHandle
 from ..abc import IPSockAddrType, UDPPacketType, UNIXDatagramPacketType
 from ..abc._eventloop import AsyncBackend, StrOrBytesPath
-from ..abc._tasks import T_contra, get_callable_name
+from ..abc._tasks import T_contra, call_for_coroutine, get_callable_name
 from ..streams.memory import MemoryObjectSendStream
 
 if TYPE_CHECKING:
@@ -255,8 +255,8 @@ class TaskGroup(abc.TaskGroup):
 
     async def start(
         self,
-        func: Callable[..., Coroutine[Any, Any, T_co]],
-        *args: object,
+        func: Callable[[Unpack[PosArgsT]], Coroutine[Any, Any, T_co]],
+        *args: Unpack[PosArgsT],
         name: object = None,
         return_handle: Literal[False] | Literal[True] = False,
     ) -> Any:
@@ -267,10 +267,7 @@ class TaskGroup(abc.TaskGroup):
         ) -> None:
             nonlocal handle
             wrapper_task_status = _TrioTaskStatus()
-            coro = func(*args, task_status=wrapper_task_status)
-            if not isinstance(coro, Coroutine):
-                raise TypeError(f"{coro!r} is not a coroutine object")
-
+            coro = call_for_coroutine(func, args, task_status=wrapper_task_status)
             if wrapper_task_status.early_start_value is not empty_start_value:
                 task_status.started(wrapper_task_status.early_start_value)
             else:
--- a/src/anyio/_backends/_trio.py
+++ b/src/anyio/_backends/_trio.py
@@ -1103,10 +1103,10 @@ class TrioBackend(AsyncBackend):
     @classmethod
     def run_async_from_thread(
         cls,
-        func: Callable[[Unpack[PosArgsT]], Awaitable[T_Retval]],
+        func: Callable[[Unpack[PosArgsT]], Coroutine[Any, Any, T_co]],
         args: tuple[Unpack[PosArgsT]],
         token: object,
-    ) -> T_Retval:
+    ) -> T_co:
         trio_token = cast("trio.lowlevel.TrioToken | None", token)
         try:
             return trio.from_thread.run(func, *args, trio_token=trio_token)
--- a/src/anyio/_backends/_trio.py
+++ b/src/anyio/_backends/_trio.py
@@ -18,7 +18,9 @@ from collections.abc import (
     Sequence,
 )
 from contextlib import AbstractContextManager
+from contextvars import Context
 from dataclasses import dataclass
+from functools import partial
 from io import IOBase
 from os import PathLike
 from signal import Signals
@@ -29,6 +31,7 @@ from typing import (
     TYPE_CHECKING,
     Any,
     Generic,
+    Literal,
     NoReturn,
     ParamSpec,
     TypeVar,
@@ -78,8 +81,10 @@ from .._core._synchronization import (
 )
 from .._core._synchronization import Semaphore as BaseSemaphore
 from .._core._tasks import CancelScope as BaseCancelScope
+from .._core._tasks import TaskHandle
 from ..abc import IPSockAddrType, UDPPacketType, UNIXDatagramPacketType
 from ..abc._eventloop import AsyncBackend, StrOrBytesPath
+from ..abc._tasks import get_callable_name
 from ..streams.memory import MemoryObjectSendStream
 
 if TYPE_CHECKING:
@@ -93,6 +98,7 @@ else:
 
 T = TypeVar("T")
 T_Retval = TypeVar("T_Retval")
+T_co = TypeVar("T_co", covariant=True)
 T_SockAddr = TypeVar("T_SockAddr", str, IPSockAddrType)
 PosArgsT = TypeVarTuple("PosArgsT")
 P = ParamSpec("P")
@@ -199,28 +205,62 @@ class TaskGroup(abc.TaskGroup):
             del exc_val, exc_tb
             self._active = False
 
-    def start_soon(
-        self,
-        func: Callable[[Unpack[PosArgsT]], Awaitable[Any]],
-        *args: Unpack[PosArgsT],
-        name: object = None,
-    ) -> None:
+    def _check_active(self, coro: Coroutine | None = None) -> None:
         if not self._active:
+            if coro is not None:
+                coro.close()
+
             raise RuntimeError(
                 "This task group is not active; no new tasks can be started."
             )
 
-        self._nursery.start_soon(func, *args, name=name)
+    def create_task(
+        self,
+        coro: Coroutine[Any, Any, T_co],
+        *,
+        name: object = None,
+        context: Context | None = None,
+    ) -> TaskHandle[T_co]:
+        if not isinstance(coro, Coroutine):
+            raise TypeError(f"expected a coroutine, got {coro.__class__.__qualname__}")
+
+        self._check_active(coro)
+        handle = TaskHandle(coro, name)
+        if context is not None:
+            context.run(
+                partial(self._nursery.start_soon, handle._run_coro, name=handle.name)
+            )
+        else:
+            self._nursery.start_soon(handle._run_coro, name=handle.name)
+
+        return handle
 
     async def start(
-        self, func: Callable[..., Awaitable[Any]], *args: object, name: object = None
-    ) -> Any:
-        if not self._active:
-            raise RuntimeError(
-                "This task group is not active; no new tasks can be started."
-            )
+        self,
+        func: Callable[..., Coroutine[Any, Any, T_co]],
+        *args: object,
+        name: object = None,
+        return_handle: Literal[False] | Literal[True] = False,
+    ) -> TaskHandle[T_co]:
+        handle: TaskHandle[T_co]
 
-        return await self._nursery.start(func, *args, name=name)
+        async def run_coro_with_task_status(
+            *, task_status: trio.TaskStatus[Any]
+        ) -> None:
+            nonlocal handle
+            coro = func(*args, task_status=task_status)
+            handle = TaskHandle(coro, name)
+            await handle._run_coro()
+
+        self._check_active()
+        final_name = get_callable_name(func, name)
+        if return_handle:
+            handle._start_value = await self._nursery.start(  # noqa: F821
+                run_coro_with_task_status, name=final_name
+            )
+            return handle  # noqa: F821
+        else:
+            return await self._nursery.start(func, *args, name=final_name)
 
 
 #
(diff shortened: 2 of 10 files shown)


src/anyio/_backends/_trio.py
```diff
@@ -199,28 +205,62 @@ async def __aexit__(
...
-    ) -> Any:
-        if not self._active:
-            raise RuntimeError(
-                "This task group is not active; no new tasks can be started."
+        self,
+        func: Callable[..., Coroutine[Any, Any, T_co]],
+        *args: object,
+        name: object = None,
+        return_handle: Literal[False] | Literal[True] = False,
+    ) -> TaskHandle[T_co]:
```
davidbrochart:
```suggestion
    ) -> Any:
```

agronholm:
This PR doesn't include changes to `BlockingPortal`. Perhaps it should?

agronholm:
> This PR doesn't include changes to `BlockingPortal`. Perhaps it should?

OTOH, nothing is breaking so it could be left to a follow-up. It will need a new `TaskHandle` variant designed for use outside of the event loop thread.

davidbrochart (approved):

src/anyio/_backends/_asyncio.py
```diff
@@ -868,39 +864,23 @@ def task_done(_task: asyncio.Task) -> None:
                     RuntimeError("Child exited without calling task_status.started()")
                 )
 
-        if not self._entered or not self.cancel_scope._active:
-            raise RuntimeError(
-                "This task group is not active; no new tasks can be started."
-            )
```
gschaffner, line 890: This check is now missing from `start()` on asyncio (but not Trio):

```python
async def test_start_in_closed_task_group() -> None:
    async def taskfunc(*, task_status: TaskStatus[None]) -> None:
        task_status.started()

        pass

    with pytest.raises(
        RuntimeError,
        match=r"This task group is not active; no new tasks can be started\.",
    ):
        await tg.start(taskfunc)
```

src/anyio/abc/_tasks.py
```diff
@@ -87,44 +118,56 @@ def create_task(
...
 
     @abstractmethod
     async def start(
         self,
-        func: Callable[..., Awaitable[Any]],
+        func: Callable[..., Coroutine[Any, Any, T_co]],
         *args: object,
         name: object = None,
+        return_handle: Literal[False] | Literal[True] = False,
     ) -> Any:
```
gschaffner, line 171: This signature change (and `start`'s similar signature change) regresses part of 538b739b8a33bbdbf305e11d93c8bc34154e50d7. I am not sure whether that is actually a problem or not; do we still want to support `Awaitable` here at type-checking time but not runtime?

Trio still does: https://github.com/python-trio/trio/blob/v0.33.0/src/trio/_core/_run.py#L1351. Some previous discussions: https://github.com/python-trio/trio-typing/pull/66#issuecomment-1487725101 and https://github.com/agronholm/anyio/issues/493#issuecomment-1672496289.

---

I could not remember what the current state is with `Awaitable` vs. `Coroutine` in AnyIO, so I cherry-picked af25075f02bf9ee0b0c21ec0a1e8195921267c76 from #567 onto master and onto this branch:

*   On master (b4059c7268ad2649702b24e54b3bfa2788842866): Here is the current status of those tests: all pass type-checking, and some pass at runtime (on 3.13):

    ```
    tests/test_eventloop.py::test_run_non_corofunc[asyncio] PASSED
    tests/test_eventloop.py::test_run_non_corofunc[asyncio+uvloop] PASSED
    tests/test_eventloop.py::test_run_non_corofunc[asyncio+eager] PASSED
    tests/test_eventloop.py::test_run_non_corofunc[trio] FAILED
    tests/test_from_thread.py::TestRunAsyncFromThread::test_run_non_corofunc_from_thread[asyncio] PASSED
    tests/test_from_thread.py::TestRunAsyncFromThread::test_run_non_corofunc_from_thread[asyncio+uvloop] PASSED
    tests/test_from_thread.py::TestRunAsyncFromThread::test_run_non_corofunc_from_thread[asyncio+eager] PASSED
    tests/test_from_thread.py::TestRunAsyncFromThread::test_run_non_corofunc_from_thread[trio] FAILED
    tests/test_from_thread.py::TestBlockingPortal::test_call_non_corofunc[asyncio] PASSED
    tests/test_from_thread.py::TestBlockingPortal::test_call_non_corofunc[asyncio+uvloop] PASSED
    tests/test_from_thread.py::TestBlockingPortal::test_call_non_corofunc[asyncio+eager] PASSED
    tests/test_from_thread.py::TestBlockingPortal::test_call_non_corofunc[trio] PASSED
    tests/test_from_thread.py::TestBlockingPortal::test_start_task_soon_non_corofunc[asyncio] PASSED
    tests/test_from_thread.py::TestBlockingPortal::test_start_task_soon_non_corofunc[asyncio+uvloop] PASSED
    tests/test_from_thread.py::TestBlockingPortal::test_start_task_soon_non_corofunc[asyncio+eager] PASSED
    tests/test_from_thread.py::TestBlockingPortal::test_start_task_soon_non_corofunc[trio] PASSED
    tests/test_from_thread.py::TestBlockingPortal::test_start_non_corofunc[asyncio] PASSED
    tests/test_from_thread.py::TestBlockingPortal::test_start_non_corofunc[asyncio+uvloop] PASSED
    tests/test_from_thread.py::TestBlockingPortal::test_start_non_corofunc[asyncio+eager] PASSED
    tests/test_from_thread.py::TestBlockingPortal::test_start_non_corofunc[trio] PASSED
    tests/test_taskgroups.py::test_start_soon_non_corofunc[asyncio] FAILED
    tests/test_taskgroups.py::test_start_soon_non_corofunc[asyncio+uvloop] FAILED
    tests/test_taskgroups.py::test_start_soon_non_corofunc[asyncio+eager] FAILED
    tests/test_taskgroups.py::test_start_soon_non_corofunc[trio] FAILED
    tests/test_taskgroups.py::test_start_non_corofunc[asyncio] FAILED
    tests/test_taskgroups.py::test_start_non_corofunc[asyncio+uvloop] FAILED
    tests/test_taskgroups.py::test_start_non_corofunc[asyncio+eager] FAILED
    tests/test_taskgroups.py::test_start_non_corofunc[trio] FAILED

    FAILED tests/test_eventloop.py::test_run_non_corofunc[trio] - TypeError: Trio expected an async function, but 'return_non_coro_awaitable.<locals>.Wrapper' appears to be synchronous
    FAILED tests/test_from_thread.py::TestRunAsyncFromThread::test_run_non_corofunc_from_thread[trio] - TypeError: Trio expected an async function, but 'return_non_coro_awaitable.<locals>.Wrapper' appears to be synchronous
    FAILED tests/test_taskgroups.py::test_start_soon_non_corofunc[asyncio] - TypeError('Expected tests.misc.return_non_coro_awaitable.<locals>.Wrapper() to return a coroutine, but the return value (...
    FAILED tests/test_taskgroups.py::test_start_soon_non_corofunc[asyncio+uvloop] - TypeError('Expected tests.misc.return_non_coro_awaitable.<locals>.Wrapper() to return a coroutine, but the return value (...
    FAILED tests/test_taskgroups.py::test_start_soon_non_corofunc[asyncio+eager] - TypeError('Expected tests.misc.return_non_coro_awaitable.<locals>.Wrapper() to return a coroutine, but the return value (...
    FAILED tests/test_taskgroups.py::test_start_soon_non_corofunc[trio] - TypeError('Expected tests.misc.return_non_coro_awaitable.<locals>.Wrapper() to return a coroutine, but the return value (...
    FAILED tests/test_taskgroups.py::test_start_non_corofunc[asyncio] - TypeError('Expected tests.misc.return_non_coro_awaitable.<locals>.Wrapper() to return a coroutine, but the return value (...
    FAILED tests/test_taskgroups.py::test_start_non_corofunc[asyncio+uvloop] - TypeError('Expected tests.misc.return_non_coro_awaitable.<locals>.Wrapper() to return a coroutine, but the return value (...
    FAILED tests/test_taskgroups.py::test_start_non_corofunc[asyncio+eager] - TypeError('Expected tests.misc.return_non_coro_awaitable.<locals>.Wrapper() to return a coroutine, but the return value (...
    FAILED tests/test_taskgroups.py::test_start_non_corofunc[trio] - TypeError("Trio expected an async function, but functools.partial(<class 'tests.misc.return_non_coro_awaitable.<locals>.W...
    ```

    So, on master:

    *   `BlockingPortal` supports `Awaitable` at type-checking time and runtime.

    *   `start_soon` and `start` support `Awaitable` at type-checking time but don't support it at runtime.

        *   `start_soon` raises the same `TypeError` on both backends. `start` raises a different `TypeError` on Trio.

    *   `run` and `to_thread.run_sync` support `Awaitable` a type-checking time and at runtime on asyncio, but not at runtime on Trio.

*   On this branch: here are the differences vs. master:

    *   `start` and `start_soon` no longer support `Awaitable` at type-checking time.

        ```
        tests/test_taskgroups.py:1484: error: Argument 1 to "start_soon" of "TaskGroup" has incompatible type "Callable[[], Awaitable[None]]"; expected "Callable[[], Coroutine[Any, Any, Never]]"  [arg-type]
        tests/test_taskgroups.py:1494: error: Argument 1 to "start" of "TaskGroup" has incompatible type "def taskfunc(*, task_status: TaskStatus[str]) -> Awaitable[None]"; expected "Callable[..., Coroutine[Any, Any, Never]]"  [arg-type]
        ```

    *   No difference in which `non_corofunc` tests pass and fail, and what the failing exceptions are.

src/anyio/abc/_tasks.py
```diff
@@ -22,11 +20,43 @@
...
+    :raises TypeError: if the return value was not a coroutine object
+
+    """
+    coro = func(*args, **kwargs)
+    if not isinstance(coro, Coroutine):
+        prefix = f"{func.__module__}." if hasattr(func, "__module__") else ""
+        raise TypeError(
+            f"Expected {prefix}{func.__qualname__}() to return a coroutine, but "
+            f"the return value ({coro!r}) is not a coroutine object"
+        )
```
gschaffner, line 55: This check has been changed from `inspect.iscoroutine` to `abc.Coroutine.__instancecheck__`. I am thinking about #230/#675/#762 and trying to understand if/why it's safe to do that now.

It seems to me that allowing `async_generator_asend` instances by using `abc.Coroutine.__instancecheck__` is now safe for the following reason: On asyncio (but not Trio), we are now wrapping all `async_generator_asend` "coroutine-like" (per PEP 525) non-coroutine objects in a real coroutine (via `TaskHandle._run_coro`). This means that AnyIO tasks (on asyncio) will now never have a `get_coro()` that is an instance of `PyAsyncGenASend`/`async_generator_asend`.

Non-AnyIO tasks can still have a `get_coro()` that is an instance of `async_generator_asend`. A non-AnyIO task can be a scope's host task, but it can't be a scope's child task. So, the short-circuiting added in #230 is still necessary.

If supporting starting tasks from `agen.asend` is intentional (I assume that this is the case?), I would suggest

*   Adding a regression test that raises https://github.com/agronholm/anyio/blob/b4059c7268ad2649702b24e54b3bfa2788842866/src/anyio/_backends/_asyncio.py#L886-L891 on master but passes on this branch, and that also tests for regression of https://github.com/agronholm/anyio/pull/675#discussion_r1473920957.

    Here's such a test, cherry-picked and modified from

    > for example, here's a branch adding a very similar second test that crashes: `test_start_soon_from_asend_cancelled` in https://github.com/agronholm/anyio/compare/19b4e58b83cd2792bdd2f3b85b16a60789a91adf...5cc5aba1699259bae2ca87d6037232131be86ef9.
    >
    > (via https://github.com/agronholm/anyio/pull/675#pullrequestreview-1855729496)

    ```python
    async def test_start_soon_from_asend_and_cancel() -> None:
        """
        Test that `start_soon` supports `agen.asend` (which returns a non-coroutine,
        coroutine-like, awaitable `PyAsyncGenASend`/`async_generator_asend` object).

        Also, (on asyncio) regression test for
        https://github.com/agronholm/anyio/pull/675#discussion_r1473920957.
        """
        started = False
        cancelled = False

        async def genfunc() -> AsyncGenerator[None, None]:
            nonlocal started, cancelled
            started = True
            try:
                await anyio.sleep(1)
                cancelled = True
                raise
            yield

        generator = genfunc()
        try:
                tg.start_soon(generator.asend, None)
                tg.cancel_scope.cancel()

            assert started
            assert cancelled
        finally:
            await generator.aclose()
    ```

*   Consider removing https://github.com/agronholm/anyio/blob/5ca5c5affd9591491f102b54b359921d42586953/src/anyio/_backends/_asyncio.py#L357-L359. It is now unreachable.

    *   Consider adding a comment in `_task_started` about the `_run_coro` wrapping that makes `getcoroutinestate` (in general) safe now.

---

> Frankly I considered the inability to run generator-based coroutine functions a _feature_ rather than a limitation. The main problem with this, however, is that I don't know what ends up a the `coro` of a task, and we can't accept just anything there. Maybe it works, maybe it won't, but you haven't added any tests to that end so it's hard to say. The problem with the `coro` of a task is outlined in a review comment of my own PR (#675).
>
> Furthermore, if we go through with this and flat out disable such checks, we may get inconsistent or misleading error messages across backends when a user tries to pass an incompatible object as the callable to `anyio.run()`.
>
> (via https://github.com/agronholm/anyio/pull/762#issuecomment-2240052044)

Besides `async_generator_asend`, I am still not sure what _else_ `abc.Coroutine.__isinstance__` will allow through that `inspect.iscoroutine` does not allow through. E.g. third-parties can use `abc.Coroutine.register`. However, at least on the asyncio backend, am I correct to think that we are safe in _all_ of those cases now (not just the `async_generator_asend` case), by virtue of the `_run_coro` wrapping being used for *all* AnyIO tasks on the asyncio backend?

tests/test_taskgroups.py
```diff
@@ -1247,36 +1247,6 @@ def generator_part() -> Generator[object, BaseException, None]:
...
 
-@pytest.mark.skipif(
-    sys.version_info >= (3, 11),
-    reason="Generator based coroutines have been removed in Python 3.11",
-)
-@pytest.mark.filterwarnings(
-    'ignore:"@coroutine" decorator is deprecated:DeprecationWarning'
-)
-@pytest.mark.parametrize("anyio_backend", asyncio_params)
-async def test_schedule_old_style_coroutine_func() -> None:
```
gschaffner, line 1264: Why remove this test? We still support 3.10 and this test still passes on this branch. AFAICT this was also the only test that visited the `TypeError` branch in `call_for_coroutine`.

src/anyio/_backends/_trio.py
```diff
@@ -199,28 +205,62 @@ async def __aexit__(
...
+
+        async def run_coro_with_task_status(
+            *, task_status: trio.TaskStatus[Any]
+        ) -> None:
+            nonlocal handle
+            coro = func(*args, task_status=task_status)
+            handle = TaskHandle(coro, name)
+            await handle._run_coro()
+
+        self._check_active()
```
gschaffner, line 279: This PR changes `start_soon` on Trio to use the `call_for_coroutine` exception string instead of a Trio exception string. Perhaps we should also use the `call_for_coroutine` exception string for `start` on Trio. If we did, then the "non-coroutine" exception strings for all three of `start_soon`, `start`, and `create_task` would no longer be backend-dependent.

src/anyio/_core/_tasks.py
```diff
@@ -359,6 +368,15 @@ def return_value(self) -> T_co:
...
                 raise TaskFailed("the task raised an exception") from self._exception
 
+    @property
+    def start_value(self) -> Any:
+        try:
+            return self._start_value
+        except AttributeError:
+            raise RuntimeError(
+                "the task was not started with TaskGroup.start()"
+            ) from None
```
gschaffner, line 396: We will be able to type the return value of `start` in the `return_handle=False` case when https://github.com/python/mypy/issues/16522 is resolved (#678).

Will we be able to type the return value of `start` in the new `return_handle=True` case? I want to factor this into the `TaskHandle` design before it's released, rather than later, in order to be sure that when https://github.com/python/mypy/issues/16522 is fixed and we add return typing to `start`, we will be able to do so for the new overload without breaking backward compatibility.

When https://github.com/python/mypy/issues/16522 is fixed, we will need `TaskHandle` to be generic in the start value as well, i.e. changing this `-> Any` to `-> StartT_co`. If we use a single `TaskHandle` type for both `start_soon`/`create_task` tasks and `start(return_handle=True)` tasks instead of separate types, the story for that generic parameter becomes a bit more complicated. If we use a single `TaskHandle` class for both cases, I suspect(?) that `StartT_co` would not need a default value, and we could get things to work by making `start_soon` and `create_task` return `TaskHandle[ReturnT_co, Never]`. Does that sound right? It might be good to try this (with a type-checker that supports it) to make sure that the new API is future-proof. (I would like to do so, but I won't have time to do that in the next few days.)

---

For the `TaskHandle` type used for `start`, IMO it would be somewhat confusing to have the order of the parameters be `ReturnT_co, StartT_co`. `StartT_co` logically comes first. If there is a group consensus about which order is more intuitive, I think that we should use the more intuitive order.

(Note that `BlockingPortal.start_task` does not currently use a start-before-return order; it uses the non-chronological order: `-> tuple[Future[ReturnT_co], StartT_co]`. However, users of `start_task` typically use tuple unpacking at the call site, whereas users of `TaskHandle` may actually want to write the annotation `TaskHandle[$start_type, $return_type]` or `TaskHandle[$return_type, $start_type]` themselves in their code/containers.)

src/anyio/_backends/_asyncio.py
```diff
@@ -868,39 +864,23 @@ def task_done(_task: asyncio.Task) -> None:
...
-        coro = func(*args, **kwargs)
-        if not iscoroutine(coro):
-            prefix = f"{func.__module__}." if hasattr(func, "__module__") else ""
-            raise TypeError(
-                f"Expected {prefix}{func.__qualname__}() to return a coroutine, but "
-                f"the return value ({coro!r}) is not a coroutine object"
-            )
-
-        name = get_callable_name(func) if name is None else str(name)
+        handle = TaskHandle(coro, name)
```
gschaffner, line 884: We are adding a `CancelScope` around the user's async fn unconditionally in `_spawn` on asyncio, but on Trio we only add it in the cases where a `TaskHandle` is being returned. Because we implemented https://github.com/python-trio/trio/pull/901, I don't _think_ that this will cause any backend-dependent semantic difference in `cancelled_caught` & friends. Does that sound right to you?

I can't think of any other semantic difference that this asymmetry will cause—IIRC Trio has introspection APIs for the nursery tree, but not for the cancel scope tree, so `trio.lowlevel.Task` and things should still give accurate results for debugging on both backends (modulo `Task.iter_await_frames`). But it still feels a bit odd to have this asymmetry.

src/anyio/abc/_tasks.py
gschaffner, line 1: Minor/somewhat rhetorical question, but: is the performance overhead of (for every AnyIO task except Trio-backend tasks with `return_handle=False`) wrapping with `_run_coro` and creating/allocating a `TaskHandle` and `CancelScope` significant?

It's mostly unavoidable if we want this `start_soon` API, but I was primarily wondering if we should consider using `__slots__`.

src/anyio/abc/_tasks.py
gschaffner: > I am not sure whether that is actually a problem or not; do we still want to support `Awaitable` here at type-checking time but not runtime?

Actually, do we (almost) support `Awaitable` at runtime now? On asyncio, we unconditionally wrap with `_run_coro`. On Trio, IIRC non-coro awaitables are still not supported at runtime (https://github.com/python-trio/trio-typing/pull/66), but we _almost_ are doing unconditional `_run_coro` wrapping on Trio too...

Would using `_run_coro` in the Trio-`start`-`return_handle=False` case and changing some of the checks from `isinstance(coro, Coroutine)` to `isinstance(aw, Awaitable)` be enough to support `Awaitable` at runtime now? If so, should we?

src/anyio/abc/_tasks.py
agronholm: Where do you propose we add `__slots__` to?

src/anyio/_backends/_asyncio.py
agronholm: I mainly went this way because it sticks close to the existing behavior. But I can change this easily if it's viewed as a problem (even a minor one).

src/anyio/abc/_tasks.py
agronholm: This may be a compelling argument to change it back to `Awaitable` and thus support arbitrary awaitables.

agronholm: Oh...I remember why I did that now. The `coro` attribute in `TaskHandle` was supposed to mirror the `asyncio.Task` class. What would be put there as the `coro` if we created a task from an arbitrary awaitable? `asyncio.create_task()` also requires a coroutine object.

src/anyio/_core/_tasks.py
agronholm: Yup, a start value type of `Never` sounds right. I'm also open to using a better type checker, but I haven't found one that doesn't report a whole bunch of nonsense false positives.

agronholm: As for the parameter order, only the start value type is optional and thus comes last, yes?

src/anyio/abc/_tasks.py
gschaffner: Potentially `CancelScope` and `Event`.

src/anyio/_core/_tasks.py
gschaffner: > As for the parameter order, only the start value type is optional and thus comes last, yes?

I don't think that we would need to make it optional:

> I suspect(?) that `StartT_co` would not need a default value, and we could get things to work by making `start_soon` and `create_task` return `TaskHandle[ReturnT_co, Never]`

Although users might want to be able to omit it.

Personally, I'd prefer splitting them up:

```python
ReturnT_co = TypeVar("ReturnT_co", covariant=True)
StartT_co = TypeVar("StartT_co", covariant=True)


class StartTaskHandle(TaskHandle[ReturnT_co], Generic[StartT_co, ReturnT_co]):
    __slots__ = ("_start_value",)

    _start_value: StartT_co

    @property
    def start_value(self) -> StartT_co:
        return self._start_value
```

(Full patch: https://github.com/agronholm/anyio/compare/taskhandle-in-start...gschaffner:anyio:add-StartTaskHandle (e7fcd9d62a6808157bfb067be1f7e89fc6fdba8c))

Among other things, this would let us remove the `AttributeError` from `def start_value` since we'd no longer need it.

src/anyio/_core/_tasks.py
```diff
@@ -359,6 +368,15 @@ def return_value(self) -> T_co:
             case TaskHandle.Status.FAILED:
                 raise TaskFailed("the task raised an exception") from self._exception
 
+    @property
+    def start_value(self) -> Any:
+        try:
```
gschaffner:
```suggestion
    def start_value(self) -> Any:
        """
        The value passed to :meth:`task_status.started() <.abc.TaskStatus.started>`,

        :raises RuntimeError: if the task was not started with :meth:`TaskGroup.start()
            <.abc.TaskGroup.start>`
        """
        try:
```

src/anyio/_core/_tasks.py
gschaffner:
```suggestion
    Returned from :meth:`TaskGroup.start_soon() <.abc.TaskGroup.start_soon>`,
    :meth:`TaskGroup.start(return_handle=True) <.abc.TaskGroup.start>`, and
    :meth:`TaskGroup.create_task() <.abc.TaskGroup.create_task>`.
```

(I can't seem to get GH to render this `suggestion` properly.)

gschaffner: It could have been `start_soon`, `start`, or `create_task`, not just `create_task`.

src/anyio/abc/_tasks.py
gschaffner: I suppose that it would either be an exception, `None` (although `asyncio.Task.get_coro` uses this for something else), or the awaitable.

agronholm: Both asyncio's `get_coro()` and Trio's `Task.coro` are pretty explicit about returning a **coroutine object**. I feel deeply uncomfortable about breaking from this norm.

src/anyio/_core/_tasks.py
agronholm: Right, I forgot to update this docstring.

agronholm:
@gschaffner Can we continue the `Awaitable` vs `Coroutine` type annotation discussion here? I find it difficult to jump between the various comment sections. I did some digging and memory refreshing, and found the following about Trio:

- Trio supports arbitrary awaitables in `start_soon()`
- Non-coroutine return values are wrapped in a nested `python_wrapper()` coroutine function
- `Task.coro` points to said wrapper in those cases

As for asyncio:

- Asyncio supports `asend` objects in `create_task()`
- Asyncio does **not** support arbitrary awaitables in `create_task()`
- However, `ensure_future()` supports arbitrary awaitables

AnyIO, on `master`:

- `create_task()` supports neither arbitrary awaitables or `asend` objects
- `start_soon()` does not work with `asend` objects

AnyIO, on this branch:

- Both `create_task()` and `start_soon()` work with `asend` objects
- Neither `create_task()` or `start_soon()` work with arbitrary awaitables

Given all this, I think it would be reasonable for us to support `asend` objects but not arbitrary awaitables, yes? That would strike a good compatibility balance between asyncio and Trio. As such, the annotations should remain as `Coroutine`, not `Awaitable`. I think the previous change from `Coroutine` to `Awaitable` was misguided, given the lack of run-time support.

gschaffner:
That sounds good to me.

src/anyio/_core/_tasks.py
agronholm: Updated.

tests/test_taskgroups.py
agronholm: I think this got lost as part of my experimentation. I added it back now.

src/anyio/_backends/_asyncio.py
agronholm: Fixed.

agronholm:
Ok, so let's see where this leaves us. IIRC, you still had some comments left to be addressed:

- The `start_value` type parameter: my choice is to reuse the same class and use `Never` as the default start value type parameter. Does it make sense to do this work before mypy can support it?
- Adding `__slots__` to `CancelScope` and `Event`: I'm trying to add slots to `CancelScope` at least; there's a test relying on monkey patching which is causing some trouble but I'm working on it.
- Cancel scope/exception message asymmetry for `start(return_handle=False)`: I'm fine with using the same route for that as `return_handle=True`.

Was there anything else?

gschaffner:
> The `start_value` type parameter: my choice is to reuse the same class and use `Never` as the default start value type parameter. Does it make sense to do this work before mypy can support it?

Yes, I think so. Currently, for the line `value = await tg.start(taskfunc, return_handle=False)`, IDEs show the hint `value: Any`, showing the user that the returned value is not being typechecked.

Similarly, for the `handle = await tg.start(taskfunc, return_handle=True)` case: if we make `TaskHandle` generic in `StartT_co` now, users' IDEs will show `handle: TaskHandle[int, Any]` instead of just showing `handle: TaskHandle[int]`. It is useful to _not_ hide the temporary `Any` from users. Making the temporary `Any` more visible would help to prevent users from being surprised by the API being partially untypechecked.

> Was there anything else?

Not much, just a few minor follow-up questions:

*   Should we add a changelog entry for the `Awaitable` -> `Coroutine` typing change? (There was a similar entry added in 538b739b8a33bbdbf305e11d93c8bc34154e50d7 / 6ba4245e967d26845d92c9091ca448ce98f41e95.)

*   The annotations for `from_thread.run`/`AsyncBackend.run_async_from_thread` and `anyio.run` were changed from `Coroutine` to `Awaitable` in 538b739b8a33bbdbf305e11d93c8bc34154e50d7 and 9f97cd7bb5f59f79f1bcda64230ab2127e79935a, but they don't support `Awaitable` at runtime (on Trio). Since we changed `TaskGroup` back to `Coroutine`, should they also be changed back to `Coroutine`?

    (Note: `BlockingPortal`'s method annotations were changed to `Awaitable` in those commits too, but they do actually support `Awaitable` at runtime.)

*   Do you want to add a regression test for `asend` support (when the user uses `type: ignore`)? (There's a test in https://github.com/agronholm/anyio/pull/1153#discussion_r3307882936.) If starting tasks from `asend` is intended to be officially supported (is it?), a regression test may be useful since IIUC `asend` support has accidentally regressed in the past.

*   The minor `call_for_coroutine` exception string question in https://github.com/agronholm/anyio/pull/1153#discussion_r3308501906.

agronholm:
@gschaffner I think I've addressed your feedback now. LMK if I missed something.

gschaffner:
Nice catch of the `start` problem, and elegant fix. Last few comments:

*   It looks like there is not any regression test that covers the

    ```
    | ExceptionGroup: Exceptions from Trio nursery (2 sub-exceptions)
        |   File "/home/ganden/vc/anyio/src/anyio/_backends/_trio.py", line 254, in run_coro_with_task_status
        |     handle = TaskHandle(coro, name)
        |   File "/home/ganden/vc/anyio/src/anyio/_core/_tasks.py", line 262, in __init__
        |     coro.close()
        |     ^^^^^^^^^^
        | AttributeError: 'NoneType' object has no attribute 'close'
        +---------------- 2 ----------------
        |   File "/home/ganden/vc/anyio/tests/test_taskgroups.py", line 371, in test_start_exception_delivery
        |     await tg.start(task_fn, return_handle=return_handle)  # type: ignore[arg-type]
        |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        |   File "/home/ganden/vc/anyio/src/anyio/_backends/_trio.py", line 260, in start
        |     handle._start_value = await self._nursery.start(
        |     ^^^^^^
        | UnboundLocalError: cannot access local variable 'handle' where it is not associated with a value
    ```

    errors that we were seeing (outside of the test suite) prior to the `start` fix (https://github.com/agronholm/anyio/pull/1153/commits/4605bcfb4d66fad016c4ecc9b39d0e5e1a7f7b4b).

    (I.e.: If I revert the `src` changes in https://github.com/agronholm/anyio/pull/1153/commits/4605bcfb4d66fad016c4ecc9b39d0e5e1a7f7b4b (but keep the `tests` changes), none of the current tests trigger either of those errors.)

    Consider adding a regression test that, without the `start` fix, fails with those errors:

    ```diff
    diff --git a/tests/test_taskgroups.py b/tests/test_taskgroups.py
    index e6d0f857..580dfc57 100644
    --- a/tests/test_taskgroups.py
    +++ b/tests/test_taskgroups.py
    @@ -360,13 +360,14 @@ async def test_cancel_with_nested_task_groups() -> None:
        assert len(outer_cancel_spy.call_args_list) < 10


    -async def test_start_exception_delivery() -> None:
    +@pytest.mark.parametrize("return_handle", [False, True])
    +async def test_start_exception_delivery(return_handle: bool) -> None:
        def task_fn(*, task_status: TaskStatus[str] = TASK_STATUS_IGNORED) -> None:
            task_status.started("hello")

        async with anyio.create_task_group() as tg:
            with pytest.raises(TypeError, match="is not a coroutine object"):
    -            await tg.start(task_fn)  # type: ignore[arg-type]
    +            await tg.start(task_fn, return_handle=return_handle)  # type: ignore[call-overload]


    async def test_start_cancel_after_error() -> None:
    ```

*   I still think that we should consider adding `__slots__` to `Event` too, but it's definitely not a blocker for this PR.

    Note: Upstream Trio `Event` is slotted, upstream asyncio `Event` isn't.

*   Regarding exception strings (previously: https://github.com/agronholm/anyio/pull/1153#discussion_r3308501906): `start` still has different strings on `asyncio` and on Trio. Can we give the same message on both backends? On asyncio we have the nice `call_for_coroutine` error message for users, but on Trio the message is a bit harder for users to understand/debug:

    ```
    tests/test_taskgroups.py::test_start_exception_delivery[asyncio] - TypeError('Expected tests.test_taskgroups.test_start_exception_delivery.<locals>.task_fn() to return a coroutine, but the return value (None) is not a coroutine object') [single exception in ExceptionGroup]
    tests/test_taskgroups.py::test_start_exception_delivery[trio] - TypeError('None is not a coroutine object') [single exception in ExceptionGroup]
    ```

Those couple comments should (hopefully) be my final round of feedback for this PR. Everything else looks good to me!

gschaffner:
Oop—one more follow-up. Regarding

> The annotations for `from_thread.run`/`AsyncBackend.run_async_from_thread` and `anyio.run` were changed from `Coroutine` to `Awaitable` in [538b739](https://github.com/agronholm/anyio/commit/538b739b8a33bbdbf305e11d93c8bc34154e50d7) and [9f97cd7](https://github.com/agronholm/anyio/commit/9f97cd7bb5f59f79f1bcda64230ab2127e79935a), but they don't support `Awaitable` at runtime (on Trio). Since we changed `TaskGroup` back to `Coroutine`, should they also be changed back to `Coroutine`?
>
> (Note: `BlockingPortal`'s method annotations were changed to `Awaitable` in those commits too, but they do actually support `Awaitable` at runtime.)

You changed `from_thread.run`/`AsyncBackend.run_async_from_thread` to `Coroutine` (https://github.com/agronholm/anyio/pull/1153/commits/ef4a724059262d1ce1b047b74aa1218755b45211), but did not change `anyio.run`/`AsyncBackend.run` to `Coroutine`. Was leaving `anyio.run`/`AsyncBackend.run` using `Awaitable` intentional?

agronholm:
> You changed `from_thread.run`/`AsyncBackend.run_async_from_thread` to `Coroutine` ([ef4a724](https://github.com/agronholm/anyio/commit/ef4a724059262d1ce1b047b74aa1218755b45211)), but did not change `anyio.run`/`AsyncBackend.run` to `Coroutine`. Was leaving `anyio.run`/`AsyncBackend.run` using `Awaitable` intentional?

No, I had planned to change that too, depending on the outcome of these discussions, but then it slipped my mind. Will fix.

agronholm:
> > You changed `from_thread.run`/`AsyncBackend.run_async_from_thread` to `Coroutine` ([ef4a724](https://github.com/agronholm/anyio/commit/ef4a724059262d1ce1b047b74aa1218755b45211)), but did not change `anyio.run`/`AsyncBackend.run` to `Coroutine`. Was leaving `anyio.run`/`AsyncBackend.run` using `Awaitable` intentional?
> 
> No, I had planned to change that too, depending on the outcome of these discussions, but then it slipped my mind. Will fix.

Oh, but that hit a snag as [this test](https://github.com/agronholm/anyio/blob/master/tests/test_eventloop.py#L45) specifically tests for `Awaitable` support. And on Python 3.14+, `asyncio.run()` also supports arbitrary awaitables.

agronholm:
> > > You changed `from_thread.run`/`AsyncBackend.run_async_from_thread` to `Coroutine` ([ef4a724](https://github.com/agronholm/anyio/commit/ef4a724059262d1ce1b047b74aa1218755b45211)), but did not change `anyio.run`/`AsyncBackend.run` to `Coroutine`. Was leaving `anyio.run`/`AsyncBackend.run` using `Awaitable` intentional?
> > 
> > 
> > No, I had planned to change that too, depending on the outcome of these discussions, but then it slipped my mind. Will fix.
> 
> Oh, but that hit a snag as [this test](https://github.com/agronholm/anyio/blob/master/tests/test_eventloop.py#L45) specifically tests for `Awaitable` support. And on Python 3.14+, `asyncio.run()` also supports arbitrary awaitables.

I checked, and `anyio.run()` supports arbitrary awaitables on asyncio, but not on Trio. Not sure what we should do about this. But I'll leave the annotations as `Awaitable` for now.

agronholm:
I added `__slots__` to a lot of classes (including `Event`) via a separate commit. That also adds the missing `__weakref__` entry to `CancelScope`.

agronholm:
I removed the coro check in `TaskHandle`, as there was no code path remaining that would have tripped that check.

agronholm:
This time I think I covered everything!

gschaffner (approved):
> > > > You changed `from_thread.run`/`AsyncBackend.run_async_from_thread` to `Coroutine` ([ef4a724](https://github.com/agronholm/anyio/commit/ef4a724059262d1ce1b047b74aa1218755b45211)), but did not change `anyio.run`/`AsyncBackend.run` to `Coroutine`. Was leaving `anyio.run`/`AsyncBackend.run` using `Awaitable` intentional?
> > > 
> > > 
> > > No, I had planned to change that too, depending on the outcome of these discussions, but then it slipped my mind. Will fix.
> > 
> > 
> > Oh, but that hit a snag as [this test](https://github.com/agronholm/anyio/blob/master/tests/test_eventloop.py#L45) specifically tests for `Awaitable` support. And on Python 3.14+, `asyncio.run()` also supports arbitrary awaitables.
> 
> I checked, and `anyio.run()` supports arbitrary awaitables on asyncio, but not on Trio. Not sure what we should do about this. But I'll leave the annotations as `Awaitable` for now.

Hm, I forgot about that that test. I suppose some options would be:

*   `Coroutine[Any, Any, T_Retval] | asyncio.Future[T_Retval]`. This annotation would be a lie on Trio, but claiming that `asyncio.Future` is supported on Trio is a smaller and more specific lie than claiming that `Awaitable` is supported on Trio.

*   `Awaitable[T_Retval]`, and add a wrapper on Trio for runtime support. E.g. https://github.com/agronholm/anyio/pull/567/files#diff-e1aac084a4f60987583d52a647503683759126d41fcd77ae51ea405f278f03b8R73-R94

*   `Awaitable[T_Retval]`, but not supported at runtime (on Trio)

It would be nice to resolve `anyio.run`'s `Awaitable` situation in the same release (and combined changelog entry) as the other handful of APIs that have been changed back to `Coroutine`. I think that it isn't a blocker for this PR, however.

LGTM!

agronholm:
Thanks for sticking with me!

PR #1161, opened 2026-06-01
Pass `backend_options` to Trio

### LINKED ISSUE #1162: In `anyio.run`, `backend_options=` are not passed to trio.
As evident from reading the source, options are silently being discarded:

https://github.com/agronholm/anyio/blob/b0690c59d8fe440ef8c19d5ea5bb76c79be03bd8/src/anyio/_backends/_trio.py#L964-L973

This has been brought up by issue #643, but that one was closed after addressing the asyncio side only in #647.
Trio also has interesting backend options. In my case, I've been looking for `clock=` to make use of trio's amazing MockClock. That obviously is backend specific, so the workaround is not using `anyio.run` in the first place - but then, what's the point of having `backend_options=` if they can't be used for exactly that purpose.

Zac-HD:
## Changes

In AnyIO 3.x the module-level `run = trio.run` was invoked as `asynclib.run(func, *args, **backend_options)`, but since 4.0's `AsyncBackend` refactor `anyio.run(..., backend="trio", backend_options={...})` silently discards the `backend_options`.

We still document trio `backend_options` as forwarded to `trio.run`, with `restrict_keyboard_interrupt_to_checkpoints` as an example, so this PR restores support in the obvious way.

A minimal motivating repro:

```python
import time, trio, trio.testing
from fastapi import FastAPI
from fastapi.testclient import TestClient

app = FastAPI()
@app.get("/s")
async def s():
    await trio.sleep(5)
    return {"t": trio.current_time()}

c = TestClient(app, backend="trio",
               backend_options={"clock": trio.testing.MockClock(autojump_threshold=0)})
t0 = time.monotonic()
c.get("/s")
print(f"wall={time.monotonic()-t0:.2f}s")  # 5.01s — clock never reached trio.run
```

Fixes #1162.

## Checklist

- [ ] You've updated the documentation (n/a; restoring already-documented behavior)

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -52,6 +52,10 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 - Fixed cancelling tasks started through a ``BlockingPortal`` after the portal has been
   stopped
   (`#1013 <https://github.com/agronholm/anyio/issues/1013>`_; PR by @puneetdixit200)
+- Fixed ``backend_options`` being ignored when running the Trio backend via
+  ``anyio.run()``; the options are now passed as keyword arguments to ``trio.run()``
+  again, as documented (a regression from AnyIO 3)
+  (`#1161 <https://github.com/agronholm/anyio/pull/1161>`_; PR by @Zac-HD)
 
 **4.13.0**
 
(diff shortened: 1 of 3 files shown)


agronholm (approved):
LGTM.

PR #1163, opened 2026-06-04
Fixed cancel scope crash without a current task on the asyncio backend

inoue22:

## Changes

On the asyncio backend, entering a cancel scope while `asyncio.current_task()` returns `None` raised an obscure `TypeError: cannot create weak reference to 'NoneType' object`. `CancelScope.__enter__` used the host task as a key in a `WeakKeyDictionary` (`_task_states[host_task]`) without a `None` check.

This is reached through `CancelScope(shield=True)` inside `cancel_shielded_checkpoint()`, which primitives such as `CapacityLimiter` and `Lock` rely on, so using any of those while `current_task()` is `None` crashes instead of behaving gracefully.

This is inconsistent with the rest of the backend: `checkpoint_if_cancelled()` and `current_effective_deadline()` already treat `current_task() is None` as a benign no-op.

This PR makes the asyncio backend consistent:

- `cancel_shielded_checkpoint()` falls back to a plain checkpoint (`await sleep(0)`) when `current_task()` is `None` — with no current task there is nothing to shield from cancellation, mirroring `checkpoint_if_cancelled()`.
- `CancelScope.__enter__` now raises a clear `RuntimeError("A cancel scope can only be entered from within a task")` instead of the weakref `TypeError`. The `CancelScope` docstring documents the new `:raises RuntimeError:`.

The defect is independent of the Python version (reproduced identically on 3.12, 3.13 and 3.14).

## Checklist

- [ ] You've updated the documentation (in `docs/`), in case of behavior changes or new features (N/A)

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -416,7 +416,11 @@ class CancelScope(BaseCancelScope):
                 "Each CancelScope may only be used for a single 'with' block"
             )
 
-        self._host_task = host_task = cast(asyncio.Task, current_task())
+        host_task = current_task()
+        if host_task is None:
+            raise RuntimeError("A cancel scope can only be entered from within a task")
+
+        self._host_task = host_task
         self._tasks.add(host_task)
         try:
             task_state = _task_states[host_task]
@@ -2424,6 +2428,10 @@ class AsyncIOBackend(AsyncBackend):
 
     @classmethod
     async def cancel_shielded_checkpoint(cls) -> None:
+        if current_task() is None:
+            await sleep(0)
+            return
+
         with CancelScope(shield=True):
             await sleep(0)
 
--- a/tests/test_lowlevel.py
+++ b/tests/test_lowlevel.py
@@ -1,10 +1,13 @@
 from __future__ import annotations
 
+import asyncio
 from typing import Any
 
 import pytest
 
 from anyio import create_task_group, run
+from anyio._backends import _asyncio
+from anyio._backends._asyncio import AsyncIOBackend, CancelScope
 from anyio.lowlevel import (
     RunVar,
     cancel_shielded_checkpoint,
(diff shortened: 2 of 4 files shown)


agronholm:
Just under what circumstances does it make any sense to call `cancel_shielded_checkpoint()` outside a task?

inoue22:
Thanks for taking a look, and for maintaining AnyIO.

It isn't user code calling it outside a task. anyio's own primitives (locks, semaphores, capacity limiters) call `cancel_shielded_checkpoint()`, which crashes when `current_task()` is `None`; for example `CapacityLimiter.acquire()`, the default thread limiter in `to_thread.run_sync()`. Each already pairs it with `checkpoint_if_cancelled()`, which self-guards that state, so guarding `cancel_shielded_checkpoint()` against it too is consistent and covers them all rather than special-casing one caller.

On its own, the `RuntimeError` change would not fix it, since `acquire()` would still raise and fail the call. The no-op lets `acquire()` complete; the `RuntimeError` just makes `CancelScope.__enter__` raise `A cancel scope can only be entered from within a task` instead of `TypeError: cannot create weak reference to 'NoneType' object`.

agronholm:
Let me rephrase: What sensible code would trigger this error?

agronholm:
Or, is this only about confusing error messages when said APIs are misused?

agronholm:
OK, so I would suggest changing the tests to use callbacks `event_loop.call_soon()` rather than monkey patching. That would make the tests more realistic.

src/anyio/_backends/_asyncio.py
```diff
@@ -2472,6 +2428,10 @@ async def checkpoint_if_cancelled(cls) -> None:
 
     @classmethod
     async def cancel_shielded_checkpoint(cls) -> None:
+        if current_task() is None:
+            await sleep(0)
+            return
+
```
agronholm, line 2434: Explain this. How would you (realistically) call a coroutine function like this without a running task?

tests/test_lowlevel.py
```diff
@@ -86,6 +89,43 @@ async def second_func() -> None:
...
+    monkeypatch.setattr(_asyncio, "current_task", lambda: None)
+
+    async def main() -> None:
+        with pytest.raises(
+            RuntimeError, match="A cancel scope can only be entered from within a task"
+        ):
+            with CancelScope():
+                pass
+
+    asyncio_event_loop.run_until_complete(main())
```
agronholm, line 112: I'd prefer this used a real async callback rather than monkey patching.

PR #1164, opened 2026-06-05
(stacked PR) Fix `start(sync_fn, return_handle=True)`

gschaffner:
## Changes

This is one proposed way to fix

> encountered an annoying problem with the task handles PR: if a regular function is passed to `TaskGroup.start()`, the coroutine check only happens inside the already-created task, causing the `await tg.start(...)` call to succeed since `task_status.started()` was in fact called, so the exception falls out of the task group [[src](https://matrix.to/#/!JfFIjeKHlqEVmAsxYP:gitter.im/$b_oGr3x6vvwSuIyT-ARf1QzaqiC1jMtnnGiVnVOKUPo?via=gitter.im)]
> 
> it might be unavoidable to pass another task status object to the actual callable [[src](https://matrix.to/#/!JfFIjeKHlqEVmAsxYP:gitter.im/$G5voxgQ258IkGElBqZ3y0Mz0xFzmoUXQCK4npG8mugQ?via=gitter.im)]

The idea is to disallow calling `TaskStatus.started()` prior to when the user-provided `Callable[..., Coroutine[Any, Any, T_co]]` has returned a coroutine object. If the user tries to do so, `TaskStatus.started` will raise an exception instead of reparenting the task and returning success.

This proposal would introduce a slight backward-incompatibility: `start` would no longer accept a `taskfunc` that calls `started()` in a sync callable before returning the coroutine object. For example, with this proposal the following would no longer work:

```python

from collections.abc import Coroutine
from typing import Any

import anyio


async def real_taskfunc() -> None:
    print("okay")


def fake_taskfunc(task_status: TaskStatus[None]) -> Coroutine[Any, Any, None]:
    task_status.started()
    return real_taskfunc()


async def main() -> None:
        await tg.start(fake_taskfunc)


anyio.run(main, backend="trio")
```

Also: Synced up the exception messages on the two backends (https://github.com/agronholm/anyio/pull/1153#discussion_r3308501906).

Also: Did https://github.com/agronholm/anyio/pull/1153#discussion_r3314137998 because doing so provided for a simpler implementation.

## Checklist

features

these instructions.

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -5,6 +5,8 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
 **UNRELEASED**
 
+- **BACKWARDS INCOMPATIBLE** ``TaskGroup.start`` no longer accepts synchronous functions
+  that call ``task_status.started()`` before they return a coroutine object
 - Added support for Python 3.15
 - Added an asynchronous implementation of the ``itertools`` module
   (`#998 <https://github.com/agronholm/anyio/issues/998>`_; PR by @11kkw)
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -110,6 +110,11 @@ if TYPE_CHECKING:
 else:
     FileDescriptorLike = object
 
+if sys.version_info >= (3, 13):
+    from typing import TypeVar
+else:
+    from typing_extensions import TypeVar
+
 if sys.version_info >= (3, 11):
     from asyncio import Runner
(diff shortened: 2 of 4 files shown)


agronholm:
I am not favorable to this. I'll keep this around only as a last resort. Backwards-incompatible changes require a major version bump. I have an acceptable alternative in the works, but we'll see if that pans out.

gschaffner:
Closing this in favor of https://github.com/agronholm/anyio/pull/1153/changes/4605bcfb4d66fad016c4ecc9b39d0e5e1a7f7b4b.

PR #1167, opened 2026-06-10
Fix Process.aclose() deadlock on asyncio with subprocess blocked on a…

### LINKED ISSUE #1166: Deadlock on shutting down subprocesses blocked on writing to full pipe on asyncio backend.
### Things to check first

### AnyIO version

master at 2370032396438b0279bacb3ff2f5c4ae2886f0d0

### Python version

3.14

### What happened?

`StreamReaderWrapper.aclose()` doesn't close the underlying pipe transport on asyncio which means the child is blocked and doesn't get `EPIPE`.
With uvloop `Process.aclose()` rescues it when cancelled but on pure asyncio the `BaseSubprocessTransport` also waits for all pipe transports before returning from its wait which means it doesn't even get rescued whan cancelled (the wait is shielded).

Snippet of test case by claude, will be repeated in PR

```
async def test_close_with_stdout_blocked_subprocess() -> None:
    """
    Test that closing a process unblocks a subprocess blocked on writing to stdout
    (because the pipe buffer is full and nobody is reading), instead of deadlocking
    while waiting for it to exit.

    """
    code = dedent("""\
    import sys

    sys.stdout.write("x" * 1024 * 1024)
    """)

    deadlocked = Event()

        async def rescue() -> None:
            await sleep(5)
            # Process.aclose() has deadlocked; force-close the subprocess
            # transport to unblock it so the test can fail cleanly
            deadlocked.set()
            process._process._transport.close()  # type: ignore[attr-defined]

        tg.start_soon(rescue)
        await process.aclose()
        tg.cancel_scope.cancel()

    if deadlocked.is_set():
        pytest.fail("Process.aclose() deadlocked")
```

tapetersen:

## Changes

Code and description below mostly by claude with very heavy supervision.

Fixes #1166.

Closing a process's standard streams on the asyncio backend only set an exception on the `asyncio.StreamReader` without closing the underlying pipe transports. A subprocess blocked on writing to a full pipe would then never receive `EPIPE`, and since asyncio's `Process.wait()` only resolves once all pipe transports have disconnected (and reading had been paused by the stream reader's flow control), `aclose()` would deadlock waiting for the process to exit.

The subprocess is now launched via `loop.subprocess_exec()`/`subprocess_shell()` so that the subprocess transport is available to ´Process.aclose()´, which closes all the pipe transports before waiting for the process to exit.

## Checklist

features

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/tests/test_subprocesses.py
+++ b/tests/test_subprocesses.py
@@ -424,3 +424,28 @@ async def test_wait_returns_on_process_exit_with_open_stdout() -> None:
             pass
         async for _ in process.stderr:
             pass
+
+
+async def test_close_with_stdout_blocked_subprocess(anyio_backend_name: str) -> None:
+    """
+    Regression test for #1166.
+
+    Test that closing a process unblocks a subprocess blocked on writing to stdout
+    (because the pipe buffer is full and nobody is reading), instead of deadlocking
+    while waiting for it to exit.
+    """
+    code = dedent("""\
+    import sys
+
+    sys.stdout.write("x" * 1024 * 1024)
+    """)
+
+    process = await open_process([sys.executable, "-c", code])
+    try:
+        with fail_after(5):
+            await process.aclose()
+    except TimeoutError:
+        if anyio_backend_name == "asyncio":
+            process._process._transport.close()  # type: ignore[attr-defined]
+
+        pytest.fail("Process.aclose() deadlocked")
--- a/tests/test_subprocesses.py
+++ b/tests/test_subprocesses.py
@@ -17,6 +17,7 @@ from anyio import (
     ClosedResourceError,
     EndOfStream,
     create_task_group,
+    fail_after,
     open_process,
     run_process,
 )
@@ -391,3 +392,35 @@ async def test_close_while_reading() -> None:
             await process.stdout.receive()
 
         process.terminate()
+
+
+async def test_wait_returns_on_process_exit_with_open_stdout() -> None:
+    """
+    wait() should return once the process exits, rather than waiting for the piped
+    stdout/stderr to close. Easiest triggered by a grandchild that inherits the
+    stdout pipe and keeps it open after the immediate child has exited.
+    """
+    code = dedent("""\
+    import subprocess, sys
+
+    subprocess.Popen(
+        [sys.executable, "-c", "import select, sys; select.select([sys.stdin], [], [], 10)"]
+    )
+    """)
+
+    process = await open_process([sys.executable, "-c", code])
+    assert process.stdin is not None
+    assert process.stdout is not None
+    assert process.stderr is not None
+
+    # Closing stdin unblocks the grandchild
+    async with process.stdin:
+        with fail_after(5):
+            assert await process.wait() == 0
+
+    # Wait for grandchild to exit to avoid ResourceWarnings
+    with fail_after(3, shield=True):
+        async for _ in process.stdout:
+            pass
+        async for _ in process.stderr:
+            pass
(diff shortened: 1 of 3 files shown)


tapetersen:
I was a bit unsure about what guarantees we provide for reading from stdout after the process is closed but for asyncio it basically requires them to be closed together with the process due to its handling in the underlying Transport

tests/test_subprocesses.py
```diff
@@ -391,3 +393,36 @@ async def test_close_while_reading() -> None:
...
+            # transport to unblock it so the test can fail cleanly
+            deadlocked.set()
+            process._process._transport.close()  # type: ignore[attr-defined]
+
+        tg.start_soon(rescue)
+        await process.aclose()
+        tg.cancel_scope.cancel()
+
+    if deadlocked.is_set():
+        pytest.fail("Process.aclose() deadlocked")
```
agronholm: I would suggest a slight simplification for the test. Requires an import for `fail_after` at the top.
```suggestion
async def test_close_with_stdout_blocked_subprocess(anyio_backend_name: str) -> None:
    """
    Regression test for #1166.

    Test that closing a process unblocks a subprocess blocked on writing to stdout
    (because the pipe buffer is full and nobody is reading), instead of deadlocking
    while waiting for it to exit.
    """
    code = dedent("""\
    import sys

    sys.stdout.write("x" * 1024 * 1024)
    """)

    process = await open_process([sys.executable, "-c", code])
    try:
        with fail_after(5):
            await process.aclose()
    except TimeoutError:
        if anyio_backend_name == "asyncio":
            process._process._transport.close()  # type: ignore[attr-defined]

        pytest.fail("Process.aclose() deadlocked")
```

tests/test_subprocesses.py
```diff
@@ -16,9 +16,11 @@
     CancelScope,
     ClosedResourceError,
     EndOfStream,
+    Event,
     create_task_group,
     open_process,
```
agronholm, line 21: To accommodate the test change below.
```suggestion
create_task_group,
fail_after,
open_process,
```

tests/test_subprocesses.py
tapetersen: Indeed cleaner. Test code was one part I was leaner towards claude.

tapetersen:
@agronholm I took a new look at it's probably an upstream asyncio bug that we work-around (waiting for the pipes is not a guaranteed invariant but just a symptom of the race between wait and the subprocess exiting). No rush to get this fixed and merged for me as such (but it does affect and is the cause of much trouble for the claude-sdk)

tapetersen:
Relevant issues https://github.com/python/cpython/issues/119710 and https://github.com/python/cpython/issues/88050 which I'm trying to check through which parts they cover or not.

tapetersen:
Ok I really hate subprocesses using shell indirection again even more than I already did... (origin here is problems due to a hard kill (-9) only killing the shell and not the grandchild it has spawned which therefore never closes its inherited file-descriptors). The semantics of `process._wait` not getting triggered until filedescriptors are closed doesn't seem intentional at all but simply due to trying to work around that case.

With that in mind let's see if I instead can make a cleaner fix that actually allows the semantics of waiting for the process exit separated from it's streams on all 3 backends (uvloop differs from trio as well but not as bad as vanilla asyncio)

tapetersen:
Ok I'm not sure about if this is the right thing to fix or we should try to do something about `Process.wait()` directly see #1174

tapetersen:
Closing in favour of #1175 which contains this commit/fix as well

Issue #983, opened 2025-09-24
KeyError: 'asyncio' in 4.11


liquidcarbon:
### Things to check first

- [x] I have searched the existing issues and didn't find my bug already reported there

- [x] I have checked that my bug is still present in the latest release

### AnyIO version

4.11

### Python version

3.12

### What happened?

Just uv-sync'd my FastAPI-based stack, seems like newly released anyio 4.11 caused an issue, reverting to 4.10 fixed it.

```
Traceback (most recent call last):
  File "/data/code/tdw2/.venv/lib/python3.12/site-packages/anyio/_core/_eventloop.py", line 162, in get_async_backend
    return loaded_backends[asynclib_name]
           ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^
KeyError: 'asyncio'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/data/code/tdw2/.venv/lib/python3.12/site-packages/uvicorn/protocols/http/httptools_impl.py", line 409, in run_asgi
    result = await app(  # type: ignore[func-returns-value]
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/code/tdw2/.venv/lib/python3.12/site-packages/uvicorn/middleware/proxy_headers.py", line 60, in __call__
    return await self.app(scope, receive, send)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/code/tdw2/.venv/lib/python3.12/site-packages/fastapi/applications.py", line 1082, in __call__
    await super().__call__(scope, receive, send)
  File "/data/code/tdw2/.venv/lib/python3.12/site-packages/starlette/applications.py", line 113, in __call__
    await self.middleware_stack(scope, receive, send)
  File "/data/code/tdw2/.venv/lib/python3.12/site-packages/starlette/middleware/errors.py", line 186, in __call__
    raise exc
  File "/data/code/tdw2/.venv/lib/python3.12/site-packages/starlette/middleware/errors.py", line 164, in __call__
    await self.app(scope, receive, _send)
  File "/data/code/tdw2/.venv/lib/python3.12/site-packages/starlette/middleware/base.py", line 108, in __call__
    response_sent = anyio.Event()
                    ^^^^^^^^^^^^^
[...]

PR #1169, opened 2026-06-13
Made all the top-level imports lazy

agronholm:

## Changes

I got impatient waiting for the new `lazy import from ...` syntax to become available, so I whipped up my own lazy importer. It reads the re-exports from under the `if TYPE_CHECKING:` block and turns it into a map it uses to process imports on demand via an injected module-level `__getattr__()` function. This should keep both run-time tooling and static type checkers happy.

## Checklist

features

these instructions.

### Updating the changelog

```
```


agronholm:
I still need to understand why the docs build fails.

agronholm:
Looks like `sphinx-autodoc-typehints` is causing me grief here which is kind of ironic.

agronholm:
How did the test suite suddenly start failing??? It's even happening locally, just not the same way is on CI.

agronholm:
Never mind, after upgrading `pytest`, it now fails the exact same way.

src/anyio/_lazyimport.py
```diff
@@ -0,0 +1,92 @@
...
+import ast
+import sys
+import warnings
+from importlib import import_module
+from pathlib import Path
+from typing import Any, cast
+
+
+def install_lazy_importer() -> None:
+    module_globals = sys._getframe(1).f_globals
```
gschaffner: > **CPython implementation detail:** This function should be used for internal and specialized purposes only. It is not guaranteed to exist in all implementations of Python.

Should we have a fallback for implementations that don't have `sys._getframe`? Perhaps just require that the module name is passed directly, but add a check so that we'll get a test failure on CPython if we pass the wrong name?

```suggestion
def install_lazy_importer(module_name: str) -> None:
    try:
        getframe = sys._getframe
    except AttributeError:
        pass
    else:
        module_globals = getframe(1).f_globals
        expected_module_name = module_globals["__name__"]
        if module_name != expected_module_name:
            raise ValueError(
                f"module_name is {module_name!r}, expected {expected_module_name!r}"
            )
```

src/anyio/_lazyimport.py
```diff
@@ -0,0 +1,92 @@
...
+    module_name = module_globals["__name__"]
+    module_prefix = module_name + "."
+    module = sys.modules[module_name]
+    lazy_map, deprecated_aliases = _build_lazy_map(cast(str, module.__file__))
```
gschaffner: Is `module.__file__` guaranteed to be non-`None` and available on disk? E.g. is is possible for folks to do pyc-only installs or something similar? Or is the assumption of a `.py` file available on disk broken by bundling tools like PyInstaller? There is also `zipimport` in the stdlib, although I don't claim to know whether anyone is using it with AnyIO or not.

gschaffner: For PyInstaller in particular, `__file__` behavior is documented: https://pyinstaller.org/en/stable/runtime-information.html#using-file

`zipimport` docs have the example

```python console
>>> example.__file__
'example_archive.zip/example.py'
```

src/anyio/_lazyimport.py
gschaffner, line 1: This behavior doesn't seem quite right:

```python console
>>> import anyio
>>> import anyio.abc
>>> anyio.abc.CancelScope
<class 'anyio._core._tasks.CancelScope'>
>>> anyio.CancelScope
<class 'anyio.CancelScope'>
>>> anyio.abc.CancelScope
<class 'anyio.CancelScope'>
```

On master, the repr is just always `anyio.CancelScope`.

src/anyio/_lazyimport.py
```diff
@@ -0,0 +1,92 @@
...
+    tree = ast.parse(Path(module_file).read_text(), filename=module_file)
+    out: dict[str, tuple[str, str]] = {}
+    deprecated_aliases: dict[str, str] = {}
+
+    for node in tree.body:
+        if not isinstance(node, ast.If) or not _is_type_checking_block(node.test):
+            continue
+
+        for stmt in node.body:
+            match stmt:
```
gschaffner, line 128: I'm curious if you considered adding something like `case _: raise NotImplementedError()` to this `match`. If someone adds something that they expect to be lazily mapped, but that the mapping code doesn't support yet, it would give us a test failure rather than a silent failure.

src/anyio/__init__.py
```diff
@@ -1,115 +1,112 @@
...
-from ._core._typedattr import TypedAttributeProvider as TypedAttributeProvider
-from ._core._typedattr import TypedAttributeSet as TypedAttributeSet
-from ._core._typedattr import typed_attribute as typed_attribute
+from typing import TYPE_CHECKING
 
-# Re-export imports so they look like they live directly in this package
-for __value in list(locals().values()):
-    if getattr(__value, "__module__", "").startswith("anyio."):
-        __value.__module__ = __name__
+from ._lazyimport import install_lazy_importer
```
gschaffner: Maybe `del install_lazy_importer` and `TYPE_CHECKING` after they are used? If they aren't `del`'d, I figure that people will see them. I'm thinking back to #769.

src/anyio/_lazyimport.py
gschaffner: Could `inspect.getsource` help with this at all? IDK.

I imagine that it would still fail similarly in _some_ situations, though, in which case we probably need to gracefully fall back to non-lazy imports.

src/anyio/_lazyimport.py
agronholm: Can you name one such implementation? In any case, we could also use the [proper function](https://docs.python.org/3.12/library/inspect.html#inspect.currentframe) from the `inspect` module.

src/anyio/_lazyimport.py
agronholm: You've raised a valid point. However, I don't know what that fallback would be, apart from duplicating all the imports.

src/anyio/_lazyimport.py
agronholm: I considered it, but decided not to add that yet, as this space is pretty well regulated.

src/anyio/_lazyimport.py
gschaffner: Nope, but I've only ever worked with CPython and PyPy. If you think that adding a falback here is premature/too conservative, I'll take your word for it. We could always change to require explicit `module_name` later if/when someone reports an issue on some implementation.

src/anyio/_lazyimport.py
gschaffner: Apart from duplicating the imports in the `TYPE_CHECKING` block, I guess we could perhaps fall back to trying harder to find the source. Something like this (below) probably works for _some_ zipimport cases (but not all, per https://docs.python.org/3/library/zipimport.html#zipimport.zipimporter.get_source):

```diff
diff --git a/src/anyio/_lazyimport.py b/src/anyio/_lazyimport.py
index 6c4bf764..f5cc3c5a 100644
--- a/src/anyio/_lazyimport.py
+++ b/src/anyio/_lazyimport.py
@@ -3,11 +3,56 @@ from __future__ import annotations
 __all__ = ("install_lazy_importer",)
 
 import ast
+import importlib.util
+import inspect
 import sys
 import warnings
 from importlib import import_module
-from pathlib import Path
-from typing import Any, cast
+from types import ModuleType
+from typing import Any
+
+
+def get_module_source(module: ModuleType) -> str:
+    try:
+        source = inspect.getsource(module)
+    except OSError:
+        pass
+    except TypeError:
+        # :func:`inspect.getsource` mistakenly thinks that an AnyIO module is a built-in
+        # module. For some reason, this can happen.
+        pass
+    else:
+        if source is not None:
+            return source
+
+    module_name = module.__name__
+    spec = module.__spec__
+    if spec is None:
+        # Try to find it another way.
+        #
+        # TODO: Is this useful? If it's in :attr:`sys.modules` already, this is pointless, right?
+        try:
+            spec = importlib.util.find_spec(module_name)
+        except ValueError:
+            pass
+
+    if spec is not None:
+        loader = module.__loader__
+        if loader is not None:
+            # Many loaders (including all :class:`importlib.abc.InspectLoader`s and all
+            # :class:`zipimport.zipimporter`s) implement `get_source`, so try that.
+            try:
+                loader.get_source  # type: ignore[attr-defined]  # noqa: B018
+            except AttributeError:
+                pass
+            else:
+                # Should we catch Exception from this method? We don't know what it's
+                # going to do...
+                source = loader.get_source(module_name)  # type: ignore[attr-defined]
+                if isinstance(source, str):
+                    return source
+
+    raise NotImplementedError()
 
 
 def install_lazy_importer() -> None:
@@ -15,7 +60,8 @@ def install_lazy_importer() -> None:
     module_name = module_globals["__name__"]
     module_prefix = module_name + "."
     module = sys.modules[module_name]
-    lazy_map, deprecated_aliases = _build_lazy_map(cast(str, module.__file__))
+    module_src = get_module_source(module)
+    lazy_map, deprecated_aliases = _build_lazy_map(module_src, module.__file__)
 
     def __getattr__(name: str) -> Any:
         if new_name := deprecated_aliases.get(name):
@@ -48,9 +94,13 @@ def install_lazy_importer() -> None:
 
 
 def _build_lazy_map(
-    module_file: str,
+    module_src: str,
+    module_file: str | None,
 ) -> tuple[dict[str, tuple[str, str]], dict[str, str]]:
-    tree = ast.parse(Path(module_file).read_text(), filename=module_file)
+    if module_file is not None:
+        tree = ast.parse(module_src, filename=module_file)
+    else:
+        tree = ast.parse(module_src)
     out: dict[str, tuple[str, str]] = {}
     deprecated_aliases: dict[str, str] = {}
```

It definitely does work if I have a normal installation of the package but force `module.__file__ = None`. I haven't actually tested it with any zipimport or PyInstaller cases though. (I'm short on time, and it seems like it might be a PITA too...)

gschaffner: I don't really like that "look harder" function though. Even if we had "look harder" (for some definition of "look harder"), it seems like there are probably still cases where the source is simply unavailable not matter how hard we try to find it (e.g. in cases where `zipimporter.get_source` returns `None`), so we'd still (in principle) need another fallback.

agronholm:
TL;DR: The requirement to have the source available kills use cases where the code is only deployed as `.pyc`.

agronholm:
@gschaffner I did it again :laughing: 
There's now a workaround for sourceless installs which I verified to work manually. The PyInstaller case still needs verification.

gschaffner:
Oh, nice. That's very clean.

src/anyio/_lazyimport.py
```diff
@@ -45,12 +48,19 @@ def __getattr__(name: str) -> Any:
...
-    module_file: str,
+    module: ModuleType,
 ) -> tuple[dict[str, tuple[str, str]], dict[str, str]]:
-    tree = ast.parse(Path(module_file).read_text(), filename=module_file)
+    try:
+        source = inspect.getsource(module)
+    except OSError:
+        return {}, {}
+
+    tree = compile(source, module.__file__ or "", "exec", ast.PyCF_ONLY_AST)
```
gschaffner: It's not important, but I'm curious: why switch from `ast.parse` to `compile`?

src/anyio/_lazyimport.py
```diff
@@ -0,0 +1,106 @@
...
+from typing import Any
+
+
+def install_lazy_importer() -> bool:
+    module_globals = sys._getframe(1).f_globals
+    module_name = module_globals["__name__"]
+    module_prefix = module_name + "."
+    module = sys.modules[module_name]
+    lazy_map, deprecated_aliases = _build_lazy_map(module)
+    if not lazy_map:
```
gschaffner:
```suggestion
    if not lazy_map and not deprecated_aliases:
```

src/anyio/_lazyimport.py
```diff
@@ -0,0 +1,106 @@
...
+    module.__dict__["__getattr__"] = __getattr__
+    module.__dict__["__all__"] = list(lazy_map)
+    return True
+
+
+def _build_lazy_map(
+    module: ModuleType,
+) -> tuple[dict[str, tuple[str, str]], dict[str, str]]:
+    try:
+        source = inspect.getsource(module)
```
gschaffner, line 114: Because we read the file for a second time here, there is a race condition where the file could be changed in-between the two reads, causing a desync between the lazy importer and what regular imports would have done. In theory, it could make package management issues like #983 harder to debug.

In practice, if we call this function while the module is being loaded, the window for a race is probably small enough to consider negligible.

src/anyio/__init__.py
```diff
@@ -1,115 +1,110 @@
...
 
+if TYPE_CHECKING or not install_lazy_importer():
```
gschaffner, line 11: The

```
# Re-export imports so they look like they live directly in this package
for __value in list(locals().values()):
```

loops should be re-added under `if TYPE_CHECKING or not install_lazy_importer():`

src/anyio/__init__.py
```diff
@@ -1,115 +1,110 @@
...
-        import warnings
-
-        warnings.warn(
-            "The 'BrokenWorkerIntepreter' alias is deprecated, use 'BrokenWorkerInterpreter' instead.",
-            DeprecationWarning,
-            stacklevel=2,
-        )
-        return BrokenWorkerInterpreter
-
-    raise AttributeError(f"module {__name__!r} has no attribute {attr!r}")
```
gschaffner, line 122: This should also be re-added under `if TYPE_CHECKING or not install_lazy_importer():`

src/anyio/_lazyimport.py
agronholm: `ast.parse()` doesn't operate on raw strings, it wants a file name.

src/anyio/_lazyimport.py
agronholm: This is because they're re-exported here for compatibility purposes. I'll fix this by importing them unconditionally.

agronholm:
I tested with PyInstaller, and found that `inspect.getsource()` doesn't work there, so it triggers the fallback.

agronholm:
Also the new test currently fails on Windows. This is likely an easy fix.

agronholm:
> Also the new test currently fails on Windows. This is likely an easy fix.

Welp...it wasn't. The test works fine on my local Windows box.

agronholm:
I added code locally that emits deprecation warnings for those backwards compat re-exports in `anyio.abc` and found that AnyIO itself still uses many of those :facepalm:

src/anyio/_lazyimport.py
```diff
@@ -0,0 +1,156 @@
...
+        if getattr(value, "__module__", "").startswith(module_prefix):
+            value.__module__ = module_name
+
+        module.__dict__[name] = value
+        return value
+
+    module.__dict__["__getattr__"] = __getattr__
+    del module.__dict__["TYPE_CHECKING"]
+    del module.__dict__["install_lazy_importer"]
+    del module.__dict__["fix_package_names"]
```
gschaffner: `del set_deprecated_aliases` too:

```suggestion
    del module.__dict__["TYPE_CHECKING"]
    del module.__dict__["install_lazy_importer"]
    del module.__dict__["fix_package_names"]
    module.__dict__.pop("set_deprecated_aliases", None)
```

src/anyio/_lazyimport.py
```diff
@@ -0,0 +1,156 @@
...
+                raise AttributeError(
+                    f"module {module_name!r} has no attribute {name!r}"
+                ) from None
+
+        imported = import_module(target_mod, module_name)
+        value = getattr(imported, target_attr)
+
+        # patch the module name to match
+        if getattr(value, "__module__", "").startswith(module_prefix):
+            value.__module__ = module_name
```
gschaffner, line 53: This wouldn't behave correctly if an API was ever moved from `anyio.SomeAPI` to `anyio.some_module.SomeAPI` (with a deprecated alias at `anyio.SomeAPI`). This isn't currently a problem, but fixing it seems simple enough:

```suggestion
        # patch the module name to match
        if (
            getattr(value, "__module__", "").startswith(module_prefix)
            and name not in deprecated_aliases
        ):
            value.__module__ = module_name
```

Note that the fallback/pyc case (`fix_package_names` + `set_deprecated_alias`) currently behaves as if this `and` in `install_lazy_importer` *is* present, because `set_deprecated_alias` is (currently) only called *after* `fix_package_names`. That is, there is a difference here between the lazy branch and the fallback branch.

docs/versionhistory.rst
```diff
@@ -81,6 +81,14 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
   asyncio under certain conditions, including improper cancel scope nesting
   (`#1111 <https://github.com/agronholm/anyio/issues/1111>`_)
 
+**UNRELEASED**
+
+- Changed the ``anyio`` and ``anyio.abc`` packages to lazily import the necessary
+  submodules. This is done by parsing the AST of the module and building a lookup table
```
gschaffner: Maybe reference the PEP for clarity of what "lazily import" means, e.g.

```rst
... to lazily import the necessary submodules (like :pep:`690`).
```

docs/versionhistory.rst
```diff
@@ -81,6 +81,14 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
...
+  from the ``if TYPE_CHECKING:`` block. A fallback mode has been provided for
+  installations where the source code is unavailable (e.g. PyInstaller).
+  (`#1169 <https://github.com/agronholm/anyio/pull/1169>`_)
+
```
gschaffner: There are two `UNRELEASED` sections; it smells like there was a merge conflict that was resolved wrong

docs/versionhistory.rst
```diff
@@ -81,6 +81,14 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
   asyncio under certain conditions, including improper cancel scope nesting
   (`#1111 <https://github.com/agronholm/anyio/issues/1111>`_)
 
+**UNRELEASED**
+
+- Changed the ``anyio`` and ``anyio.abc`` packages to lazily import the necessary
```
gschaffner: "Packages" is technically correct here in the https://docs.python.org/3/glossary.html#term-package sense, but "modules" would also be correct and would IMO be clearer to users. Users can't `pip install anyio.abc`.

```suggestion
- Changed the ``anyio`` and ``anyio.abc`` modules to lazily import the necessary
```

docs/versionhistory.rst
```diff
@@ -125,6 +133,7 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 - Improved ``anyio.Path`` to preserve subclass types by returning ``Self`` in methods
   that return path objects
   (`#1130 <https://github.com/agronholm/anyio/issues/1130>`_; PR by @EmmanuelNiyonshuti)
+- Changed the top-level imports to work lazily to reduce import overhead
```
gschaffner: Another merge issue. (This is a duplicate entry and is in the 4.14.0 section, not the unreleased section.)

docs/conf.py
gschaffner, line 1: Several links in the documentation appear to be broken by this branch. E.g. the link to `CapacityLimiter` in the return type of https://anyio.readthedocs.io/en/latest/api.html#anyio.Path.limiter. Here's a longer diff:

```diff
diff --git a/doclog-e8ff01c b/doclog-26bcf6c
index 307c31d9..2b92f47f 100644
--- a/doclog-e8ff01c
+++ b/doclog-26bcf6c
@@ -82,6 +82,8 @@ writing output... [100%] why
 <unknown>:1: WARNING: py:class reference target not found: anyio._core._eventloop.TypeVarTuple [ref.class]
 /home/ganden/vc/anyio/src/anyio/__init__.py:docstring of anyio.TypedAttributeProvider.extra_attributes:1: WARNING: py:class reference target not found: anyio._core._typedattr.T_Attr [ref.class]
 /home/ganden/vc/anyio/src/anyio/__init__.py:docstring of anyio.TypedAttributeProvider.extra_attributes:1: WARNING: py:class reference target not found: anyio._core._typedattr.T_Attr [ref.class]
+<unknown>:1: WARNING: py:class reference target not found: anyio._core._tasks.CancelScope [ref.class]
+<unknown>:1: WARNING: py:class reference target not found: anyio._core._tasks.CancelScope [ref.class]
 /home/ganden/vc/anyio/src/anyio/__init__.py:docstring of anyio.TaskHandle:1: WARNING: py:obj reference target not found: anyio._core._tasks.T_co [ref.obj]
 /home/ganden/vc/anyio/src/anyio/__init__.py:docstring of anyio.TaskHandle:1: WARNING: py:obj reference target not found: anyio._core._tasks.T_startval_co [ref.obj]
 /home/ganden/vc/anyio/src/anyio/__init__.py:docstring of anyio.TaskHandle:1: WARNING: py:class reference target not found: TaskGroup [ref.class]
@@ -96,10 +98,16 @@ writing output... [100%] why
 /home/ganden/vc/anyio/src/anyio/abc/__init__.py:docstring of anyio.abc.TaskStatus:1: WARNING: py:obj reference target not found: anyio.abc._tasks.T_contra [ref.obj]
 <unknown>:1: WARNING: py:class reference target not found: anyio.to_thread.TypeVarTuple [ref.class]
 <unknown>:1: WARNING: py:class reference target not found: anyio.to_thread.TypeVarTuple [ref.class]
+<unknown>:1: WARNING: py:class reference target not found: anyio._core._synchronization.CapacityLimiter [ref.class]
+<unknown>:1: WARNING: py:class reference target not found: anyio._core._synchronization.CapacityLimiter [ref.class]
 <unknown>:1: WARNING: py:class reference target not found: anyio.to_interpreter.TypeVarTuple [ref.class]
 <unknown>:1: WARNING: py:class reference target not found: anyio.to_interpreter.TypeVarTuple [ref.class]
+<unknown>:1: WARNING: py:class reference target not found: anyio._core._synchronization.CapacityLimiter [ref.class]
+<unknown>:1: WARNING: py:class reference target not found: anyio._core._synchronization.CapacityLimiter [ref.class]
 <unknown>:1: WARNING: py:class reference target not found: anyio.to_process.TypeVarTuple [ref.class]
 <unknown>:1: WARNING: py:class reference target not found: anyio.to_process.TypeVarTuple [ref.class]
+<unknown>:1: WARNING: py:class reference target not found: anyio._core._synchronization.CapacityLimiter [ref.class]
+<unknown>:1: WARNING: py:class reference target not found: anyio._core._synchronization.CapacityLimiter [ref.class]
 <unknown>:1: WARNING: py:class reference target not found: anyio.from_thread.TypeVarTuple [ref.class]
 <unknown>:1: WARNING: py:class reference target not found: anyio.from_thread.TypeVarTuple [ref.class]
 /home/ganden/vc/anyio/docs/api.rst:70: WARNING: py:exc reference target not found: MissingTokenError [ref.exc]
@@ -113,8 +121,13 @@ writing output... [100%] why
 <unknown>:1: WARNING: py:data reference target not found: typing.Union [ref.data]
 <unknown>:1: WARNING: py:class reference target not found: anyio.from_thread.TypeVarTuple [ref.class]
 /home/ganden/vc/anyio/docs/api.rst:70: WARNING: py:class reference target not found: T_Retval [ref.class]
+<unknown>:1: WARNING: py:class reference target not found: anyio._core._synchronization.CapacityLimiter [ref.class]
 <unknown>:1: WARNING: py:class reference target not found: anyio._core._fileio.AsyncFile [ref.class]
+<unknown>:1: WARNING: py:class reference target not found: anyio._core._synchronization.CapacityLimiter [ref.class]
+<unknown>:1: WARNING: py:class reference target not found: anyio._core._fileio.AsyncFile [ref.class]
+/home/ganden/vc/anyio/src/anyio/__init__.py:docstring of anyio.AsyncFile.limiter:1: WARNING: py:class reference target not found: anyio._core._synchronization.CapacityLimiter [ref.class]
 /home/ganden/vc/anyio/src/anyio/__init__.py:docstring of anyio.Path:7: WARNING: py:meth reference target not found: pathlib.Path.link_to [ref.meth]
+/home/ganden/vc/anyio/src/anyio/__init__.py:docstring of anyio.Path.limiter:1: WARNING: py:class reference target not found: anyio._core._synchronization.CapacityLimiter [ref.class]
 /home/ganden/vc/anyio/src/anyio/__init__.py:docstring of anyio.SpooledTemporaryFile:1: WARNING: py:class reference target not found: anyio._core._fileio.AsyncFile [ref.class]
 <unknown>:1: WARNING: py:class reference target not found: _typeshed.ReadableBuffer [ref.class]
 /home/ganden/vc/anyio/src/anyio/__init__.py:docstring of anyio.create_memory_object_stream:1: WARNING: py:class reference target not found: T_Item [ref.class]
@@ -177,13 +190,10 @@ writing output... [100%] why
 <unknown>:1: WARNING: py:class reference target not found: _typeshed.FileDescriptorLike [ref.class]
 /home/ganden/vc/anyio/src/anyio/abc/__init__.py:docstring of anyio.abc.SocketAttribute:1: WARNING: py:class reference target not found: socket.AddressFamily [ref.class]
 /home/ganden/vc/anyio/src/anyio/abc/__init__.py:docstring of anyio.abc.SocketStream:1: WARNING: py:class reference target not found: anyio.abc._sockets._SocketProvider [ref.class]
-/home/ganden/vc/anyio/src/anyio/abc/__init__.py:docstring of anyio.abc.SocketListener:1: WARNING: py:class reference target not found: anyio.abc._streams.Listener [ref.class]
 /home/ganden/vc/anyio/src/anyio/abc/__init__.py:docstring of anyio.abc.SocketListener:1: WARNING: py:class reference target not found: anyio.abc._sockets._SocketProvider [ref.class]
-/home/ganden/vc/anyio/src/anyio/abc/__init__.py:docstring of anyio.abc.UDPSocket:1: WARNING: py:class reference target not found: anyio.abc._streams.UnreliableObjectStream [ref.class]
 /home/ganden/vc/anyio/src/anyio/abc/__init__.py:docstring of anyio.abc.UDPSocket:1: WARNING: py:class reference target not found: anyio.abc._sockets._SocketProvider [ref.class]
 /home/ganden/vc/anyio/src/anyio/abc/__init__.py:docstring of anyio.abc.ConnectedUDPSocket:1: WARNING: py:class reference target not found: anyio.abc._streams.UnreliableObjectStream [ref.class]
 /home/ganden/vc/anyio/src/anyio/abc/__init__.py:docstring of anyio.abc.ConnectedUDPSocket:1: WARNING: py:class reference target not found: anyio.abc._sockets._SocketProvider [ref.class]
-/home/ganden/vc/anyio/src/anyio/abc/__init__.py:docstring of anyio.abc.UNIXDatagramSocket:1: WARNING: py:class reference target not found: anyio.abc._streams.UnreliableObjectStream [ref.class]
 /home/ganden/vc/anyio/src/anyio/abc/__init__.py:docstring of anyio.abc.UNIXDatagramSocket:1: WARNING: py:class reference target not found: anyio.abc._sockets._SocketProvider [ref.class]
 /home/ganden/vc/anyio/src/anyio/abc/__init__.py:docstring of anyio.abc.ConnectedUNIXDatagramSocket:1: WARNING: py:class reference target not found: anyio.abc._streams.UnreliableObjectStream [ref.class]
 /home/ganden/vc/anyio/src/anyio/abc/__init__.py:docstring of anyio.abc.ConnectedUNIXDatagramSocket:1: WARNING: py:class reference target not found: anyio.abc._sockets._SocketProvider [ref.class]
@@ -193,6 +203,7 @@ writing output... [100%] why
 <unknown>:1: WARNING: py:data reference target not found: typing.Union [ref.data]
 <unknown>:1: WARNING: py:data reference target not found: typing.Union [ref.data]
 <unknown>:1: WARNING: py:data reference target not found: typing.Union [ref.data]
+<unknown>:1: WARNING: py:class reference target not found: anyio._core._synchronization.EventStatistics [ref.class]
 /home/ganden/vc/anyio/src/anyio/__init__.py:docstring of anyio.Future:1: WARNING: py:obj reference target not found: anyio._core._futures.T [ref.obj]
 /home/ganden/vc/anyio/docs/api.rst:194: WARNING: py:exc reference target not found: FutureNotFinished [ref.exc]
 /home/ganden/vc/anyio/docs/api.rst:194: WARNING: py:exc reference target not found: FutureCancelled [ref.exc]
@@ -200,6 +211,10 @@ writing output... [100%] why
 /home/ganden/vc/anyio/docs/api.rst:194: WARNING: py:exc reference target not found: FutureNotFinished [ref.exc]
 /home/ganden/vc/anyio/docs/api.rst:194: WARNING: py:exc reference target not found: FutureCancelled [ref.exc]
 /home/ganden/vc/anyio/docs/api.rst:194: WARNING: py:exc reference target not found: FutureFailed [ref.exc]
+<unknown>:1: WARNING: py:class reference target not found: anyio._core._synchronization.LockStatistics [ref.class]
+<unknown>:1: WARNING: py:class reference target not found: anyio._core._synchronization.ConditionStatistics [ref.class]
+<unknown>:1: WARNING: py:class reference target not found: anyio._core._synchronization.SemaphoreStatistics [ref.class]
+<unknown>:1: WARNING: py:class reference target not found: anyio._core._synchronization.CapacityLimiterStatistics [ref.class]
 <unknown>:1: WARNING: py:data reference target not found: typing.Union [ref.data]
 <unknown>:1: WARNING: py:class reference target not found: anyio.functools._InitialMissingType [ref.class]
 /home/ganden/vc/anyio/src/anyio/lowlevel.py:docstring of anyio.lowlevel.RunVar:1: WARNING: py:obj reference target not found: anyio.lowlevel.T [ref.obj]
@@ -212,8 +227,8 @@ generating indices... genindex done
 writing additional pages... search done
 dumping search index in English (code: en)... done
 dumping object inventory... done
-build succeeded, 130 warnings.
+build succeeded, 145 warnings.
```

docs/conf.py
```diff
@@ -29,6 +29,7 @@
 pygments_style = "sphinx"
 autodoc_default_options = {"members": True, "show-inheritance": True}
 autodoc_mock_imports = ["_typeshed", "pytest", "_pytest"]
+suppress_warnings = ("sphinx_autodoc_typehints.guarded_import",)
```
gschaffner: Is this still needed? I tried removing this and it didn't make any difference in the `rm .tox -r; tox -e docs` log.

src/anyio/_lazyimport.py
```diff
@@ -0,0 +1,156 @@
...
+    module_name = module_globals["__name__"]
+    module_prefix = module_name + "."
+    module = sys.modules[module_name]
+    lazy_map, deprecated_aliases = _build_lazy_map(module)
+    if not lazy_map:
+        return False
+
+    def __getattr__(name: str) -> Any:
+        if new_name := deprecated_aliases.get(name):
+            emit_deprecation_warning(module_name, name, new_name)
```
gschaffner, line 35: In the `if TYPE_CHECKING or ...` block, can we use `@deprecated`(PEP 702)?

src/anyio/abc/__init__.py
gschaffner, line 1: There is a backward-incompatible type-checking change: accessing deprecated attrs now errors (tested with Mypy):

```pytest
import anyio
import anyio.abc

# these now give new type-checking errors:
reveal_type(anyio.abc.CancelScope)  # error: Module has no attribute "CancelScope"  [attr-defined]
reveal_type(anyio.BrokenWorkerIntepreter)  # error: Module has no attribute "BrokenWorkerIntepreter"; maybe "BrokenWorkerInterpreter"?  [attr-defined]
```

As-is, the next time a new deprecated alias is made, users will get type-checking failures and will have to use a `cast`. (In theory this may also affect existing libraries that have compatibility code to support previous AnyIO versions.) It would probably be better to have these be visible to type-checkers like they are on master.

src/anyio/__init__.py
gschaffner, line 1: REPL tab-completions are broken (missing everything) and just show this:

```python-console
>>> import anyio
>>> anyio.
anyio.annotations              anyio.set_deprecated_aliases(  
```

Should we detect a REPL (`python` REPL, `bpython`, `ipython`, Jupyter notebook, etc.) in `install_lazy_importer` and `return False`?

gschaffner: Another approach that might(?) work: generate and set `__all__` in `install_lazy_importer`?

gschaffner (commented):
Overall I have some new comments but nothing as big/breaking as the pyc-only problem from before.

I went through the previous comment threads and marked as resolved the ones that have been acked or already fixed. There are a couple open still.

src/anyio/_lazyimport.py
```diff
@@ -0,0 +1,156 @@
...
+            if modname.startswith(module_prefix):
+                parts = modname.split(".")
+                value.__module__ = ".".join(
+                    part for part in parts if not part.startswith("_")
+                )
+
+
+def emit_deprecation_warning(module_name: str, name: str, target: str) -> None:
+    warnings.warn(
+        f"{module_name}.{name} is deprecated, use {target} instead.",
```
gschaffner: The old string[^1] stated that these are aliases (which clarifies to users that they are not different types/functions with different semantics), which IMO is a bit clearer to users.

[^1]: for `BrokenWorkerIntepreter`—the others didn't have deprecation warnings

```suggestion
        f"The {module_name}.{name} alias is deprecated, use {target} instead.",
```

src/anyio/abc/__init__.py
agronholm: This was kind of intentional: the code still runs but type checkers scream bloody murder. This also happened in the AnyIO repo itself! Would you say this is a bad thing to do?

src/anyio/__init__.py
agronholm: I could've sworn I already did that. I'll re-check. Maybe I just *intended* to do it and never got around to it. Either way, this is a valid concern.

agronholm: Oh yeah, `__all__` isn't going to cut it here. We need bigger guns.

agronholm: Fixed.

docs/versionhistory.rst
agronholm: Changed.

docs/versionhistory.rst
agronholm: Fixed.

docs/versionhistory.rst
agronholm: Done.

docs/versionhistory.rst
agronholm: Removed.

src/anyio/_lazyimport.py
agronholm: That requires Python 3.13. And I don't see how that would be useful anyway, as it's meant to be applied on top of actually deprecated functions and classes. It doesn't work for our use case.

src/anyio/_lazyimport.py
agronholm: I don't see what we can do about it anyway. Highly unlikely to ever cause a problem.

src/anyio/abc/__init__.py
gschaffner: It doesn't seem too bad. It's more just unexpected that a deprecation upstream would immediately cause an inferred type to become `Any` downstream. That isn't how it would behave with regular imports or `lazy import`; it's an artifact of `set_deprecated_aliases` assigning the globals dynamically.

src/anyio/_lazyimport.py
gschaffner: It's backported in `typing_extensions`, but yeah, I didn't realize that it [doesn't support deprecating aliases](https://peps.python.org/pep-0702/#deprecation-of-modules-and-attributes). Oops.

docs/versionhistory.rst
```diff
@@ -23,6 +23,11 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
   module name. (The default name for a task spawned with ``TaskGroup.start_soon`` or
   ``TaskGroup.start`` typically includes the module name.)
   (`#1234 <https://github.com/agronholm/anyio/pull/1234>`_; PR by @gschaffner)
+- Changed the ``anyio`` and ``anyio.abc`` modules to lazily (much like :pep:`690`)
```
gschaffner: Sorry, I linked to the wrong PEP. It should be 810.


```suggestion
- Changed the ``anyio`` and ``anyio.abc`` modules to lazily (much like :pep:`810`)
```

docs/conf.py
gschaffner: The problem seems to be that Sphinx can no longer tell that `._synchronization.CapacityLimiter` et al.'s API documentation is under the name `anyio.CapacityLimiter`.

One way to fix this is to make `install_lazy_importer` short-circuit `return False` when running a docs build.

(PEP 810 specifies a `-X lazy_imports=none` command-line option that we could query for the check (via `sys.get_lazy_imports` on >= 3.15 and via `sys._xoptions` on < 3.15). (Note that `python3.14 -X lazy_imports=none` works fine.) However, it looks like the `none` option was removed after the PEP was frozen (`-X lazy_imports=none` errors on 3.15.0b4), so we'd need to query something else instead.)

src/anyio/_lazyimport.py
```diff
@@ -0,0 +1,165 @@
...
+
+    def __dir__() -> list[str]:
+        return names
+
+    module.__dict__["__dir__"] = __dir__
+    module.__dict__["__getattr__"] = __getattr__
+    del module.__dict__["TYPE_CHECKING"]
+    del module.__dict__["install_lazy_importer"]
+    del module.__dict__["fix_package_names"]
+    module.__dict__.pop("set_deprecated_aliases", None)
```
gschaffner: In the lazy case, these get `del`'d here, but in the fallback case we should add corresponding `del` statements in the `if TYPE_CHECKING or not install_lazy_importer():` blocks.

src/anyio/_lazyimport.py
```diff
@@ -0,0 +1,165 @@
...
+            getattr(value, "__module__", "").startswith(module_prefix)
+            and name not in deprecated_aliases
+        ):
+            value.__module__ = module_name
+
+        module.__dict__[name] = value
+        return value
+
+    def __dir__() -> list[str]:
+        return names
```
gschaffner, line 59: These closures result in reference cycles, e.g.

```python-console
>>> inspect.getclosurevars(anyio.__getattr__).nonlocals["module"] is anyio
True
```

This is probably unavoidable (for `__getattr__` —it's avoidable for `__dir__`) due to `__getattr__`'s caching logic.

(Informational comment only. I think that this refcycle is probably fine?)

src/anyio/_lazyimport.py
gschaffner: IIUC `ast.parse` also wants a raw string. `ast.parse(source, filename)` is documented as equivalent to `compile(source, filename, "exec", ast.PyCF_ONLY_AST)`

agronholm: `ast.parse()` requires a path, it doesn't work on source strings, unlike `compile()`, and that's what we needed here due to `inspect.getsource()`. Whether _that_ is warranted is another question.

docs/conf.py
agronholm: I think this is the biggest blocker right now. I'll see if I can fix this for real or if we need the workaround.

src/anyio/_lazyimport.py
agronholm: The cycles are module-level and static in nature, so yes, they're fine.

agronholm:
The docs build now but one added test is timing out on Windows. I suspect it's just really slow, and I need to increase the timeout value.

gschaffner:
I think that this is good to go except for the missing `del`s in the fallback branches.

gschaffner:
The `typehints_fixup_module_name` change also fixed a couple references that are broken on main, which is cool.

agronholm:
All done now.

src/anyio/_lazyimport.py
```diff
@@ -0,0 +1,168 @@
...
+
+    module.__dict__["__dir__"] = __dir__
+    module.__dict__["__getattr__"] = __getattr__
+    module.__dict__.pop("fix_package_names", None)
+    module.__dict__.pop("set_deprecated_aliases", None)
```
gschaffner: In the fallback case, these two are still accessible and show up in tab completions:

```python-console
>>> import anyio
>>> anyio.s
anyio.set_deprecated_aliases(  anyio.sleep_forever()          anyio.streams
anyio.sleep(                   anyio.sleep_until(
```

I think they should have `del`s in the `if TYPE_CHECKING or not install_lazy_importer():` blocks. Alternatively, `fix_package_names` could delete itself from the module globals and `set_deprecated_aliases` could delete itself as well.

agronholm: If we add `del` in the `TYPE_CHECKING` block, it would just tell a static analyzer that the name is no longer there. But I've now added code that deletes the helpers even on the fallback path.

gschaffner (approved):
yay!

agronholm:
Thanks for the review!

PR #1171, opened 2026-06-14
Make `anyio.run` support `Awaitable` at runtime on all backends

gschaffner:
## Changes

Adds `anyio.run` runtime support for callables returning arbitrary, non-`Coroutine` `Awaitable`s. Follow-up for https://github.com/agronholm/anyio/pull/1153#pullrequestreview-4492109242.

Also adds tests for `BlockingPortal`'s runtime `Awaitable` support.

## Checklist

features

these instructions.


tests/misc.py
gschaffner, line 1: I don't like the name of this file. For now, however, I just left it with the name that was cherry-picked from #567.

agronholm: We have other such helper functions in `conftest.py` from which we also import them. No need for a new module.

docs/versionhistory.rst
```diff
@@ -40,6 +40,9 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
   This reverts an earlier change from v3.7.0 which was made in error.
   (`#1153 <https://github.com/agronholm/anyio/pull/1153>`_)
+- Changed ``anyio.run`` to support callables returning arbitrary awaitables at runtime
+  on all backends (`#1171 <https://github.com/agronholm/anyio/pull/1171>`_; PR by
```
agronholm: I think this should mention precisely that the change only concerned Trio.

agronholm (changes requested):
I left a couple comments, and when those have been addressed, I'm good for merging.

tests/misc.py
gschaffner: Done.

agronholm (approved):

docs/versionhistory.rst
gschaffner: I took another stab at it. I'm having some trouble finding a good phrasing for the entry because I'm not sure if it should be a "Fixed" or a "Changed". I suppose it would depend on whether the current "annotate `Awaitable`, support `abc.Coroutine`" situation is considered to be a "bug" or not.

If you can think of a better way to phrase the situation/change, please feel free to change the entry to that.

PR #1172, opened 2026-06-14
Fix CapacityLimiter over-granting tokens on asyncio

### LINKED ISSUE #1170: CapacityLimiter can over-grant tokens (`borrowed_tokens > total_tokens`, `available_tokens` goes negative)
### Things to check first

### AnyIO version

4.13.0

### Python version

3.12

### What happened?

On the asyncio backend, a non-blocking acquire can succeed *while the just-freed token is already reserved for a woken waiter*, leaving the limiter with more borrowers than it has tokens. The state becomes:

```
borrowed_tokens  = 2
total_tokens     = 1
available_tokens = -1      # negative
borrowers        = ('B', 'X')
```

Two tasks then hold a single-token limiter concurrently — the exact condition the limiter exists to prevent — and `available_tokens` is negative.

This violates the documented contract of `acquire_nowait` / `acquire_on_behalf_of_nowait`:

> :raises ~anyio.WouldBlock: if there are no tokens available for borrowing

In the window below there is no token available (the freed one is earmarked for the woken waiter), yet the nowait acquire succeeds instead of raising `WouldBlock`.

Just anyio, no test framework: `python repro.py` prints the corrupt state and raises `AssertionError`:

```python
import anyio
from anyio import CapacityLimiter, WouldBlock

async def main():
    lim = CapacityLimiter(1)                       # capacity: exactly ONE token
    lim.acquire_on_behalf_of_nowait("A")           # A borrows the only token

        async def b():
            await lim.acquire_on_behalf_of("B")    # B waits for the token
        tg.start_soon(b)

        while lim.statistics().tasks_waiting != 1:  # wait until B is parked
            await anyio.sleep(0)

        lim.release_on_behalf_of("A")              # frees the token, notifies B
        # B has been notified but has NOT resumed yet, so it is not in
        # _borrowers and the wait queue is now empty. A non-blocking acquire
        # in this window slips through:
        try:
            lim.acquire_on_behalf_of_nowait("X")  # expected: WouldBlock
            print("X acquired without waiting (should have raised WouldBlock)")
        except WouldBlock:
            print("X correctly blocked")
        # leaving the task group lets B resume and add itself to _borrowers

    print(f"borrowed_tokens  = {lim.borrowed_tokens}  (total_tokens = {lim.total_tokens})")
    print(f"available_tokens = {lim.available_tokens}")
    print(f"borrowers        = {lim.statistics().borrowers}")
    assert lim.borrowed_tokens <= lim.total_tokens, "borrowed_tokens > total_tokens"

anyio.run(main)
```

Output:

```
X acquired without waiting (should have raised WouldBlock)
borrowed_tokens  = 2  (total_tokens = 1)
available_tokens = -1
borrowers        = ('B', 'X')
AssertionError: borrowed_tokens > total_tokens
```
### Cause

In `anyio/_backends/_asyncio.py`:

- `release_on_behalf_of` removes the borrower then calls `_notify_next_waiter()`, which **pops** the next waiter out of `_wait_queue` and does `event.set()`.
- The woken waiter, in `acquire_on_behalf_of`, only adds itself to `_borrowers` **after** `await event.wait()` returns:

  ```python
  await event.wait()
  ...
  self._borrowers.add(borrower)   # runs only once the waiter resumes
  ```

- Between the `event.set()` and the waiter resuming, `_wait_queue` is empty and the waiter is not yet in `_borrowers`. So the nowait guard

  ```python
  if self._wait_queue or len(self._borrowers) >= self._total_tokens:
      raise WouldBlock
  ```

  evaluates to `False` (empty queue, `0 < 1`) and admits a second borrower for the one token.

### Why `Semaphore` is not affected

The sibling `Semaphore` is immune under the identical scenario, a nowait acquire issued in the same window correctly raises `WouldBlock`, because its `release()` hands the freed slot to the woken waiter rather than leaving it generally available. `CapacityLimiter` is the one primitive whose freed-but-reserved token is briefly visible as "available" to a synchronous acquirer.

A fix would close that window, e.g. by reserving the token for the woken waiter at notify time (add the borrower to `_borrowers` when `_notify_next_waiter` sets its event, rather than after it resumes), mirroring the Semaphore/Lock hand-off. The nowait guard would then correctly see no free token.

gaoflow:
## Changes

Fixes #1170.

On the asyncio backend, `CapacityLimiter` could over-grant tokens: a non-blocking acquire issued in the window between a token being released (which notifies the next waiter) and that waiter actually resuming would slip through, leaving `borrowed_tokens > total_tokens` and `available_tokens` negative — two tasks holding a single-token limiter at once.

The cause: `release_on_behalf_of` popped the next waiter out of `_wait_queue` and set its event, but the waiter only added itself to `_borrowers` *after* it resumed. In between, the wait queue was empty and the borrower count was below the limit, so `acquire_on_behalf_of_nowait`'s guard admitted a second borrower for the already-reserved token.

The fix reserves the freed token for the woken waiter the moment its event is set — the borrower is added to `_borrowers` in `_notify_next_waiter` (and when `total_tokens` is increased) rather than after the waiter resumes — mirroring the hand-off that already makes `Semaphore`/`Lock` immune. Consequently:

- the woken waiter no longer re-adds itself to `_borrowers` on resume, and
- a waiter cancelled *after* being granted the token now releases it and passes it on to the next waiter.

Only the asyncio backend is affected; the trio backend wraps trio's own `CapacityLimiter`.

## Checklist

features

these instructions.

---

This change was prepared by an AI agent under my direction; I reviewed and verified it.

### DIFF
--- a/tests/test_synchronization.py
+++ b/tests/test_synchronization.py
@@ -994,13 +994,12 @@ class TestCapacityLimiter:
         with pytest.raises(WouldBlock):
             limiter.acquire_on_behalf_of_nowait(borrower2)
 
-    @pytest.mark.parametrize("anyio_backend", asyncio_params)
     async def test_nowait_acquire_after_release_does_not_oversubscribe(self) -> None:
         # Regression test for #1170: a non-blocking acquire issued in the window
         # between releasing a token (which notifies the next waiter) and that
         # waiter actually resuming must not slip through, as the freed token is
-        # already reserved for the woken waiter. This window is specific to the
-        # asyncio backend (the trio backend wraps trio's own CapacityLimiter).
+        # already reserved for the woken waiter. The over-granting bug was
+        # specific to the asyncio backend, but the invariant holds on both.
         limiter = CapacityLimiter(1)
         limiter.acquire_on_behalf_of_nowait("A")
 
--- a/tests/test_synchronization.py
+++ b/tests/test_synchronization.py
@@ -929,3 +929,32 @@ class TestCapacityLimiter:
             tg.cancel_scope.cancel()
 
         pytest.fail("The second borrower failed to acquire the limiter")
+
+    @pytest.mark.parametrize("anyio_backend", asyncio_params)
+    async def test_nowait_acquire_after_release_does_not_oversubscribe(self) -> None:
+        # Regression test for #1170: a non-blocking acquire issued in the window
+        # between releasing a token (which notifies the next waiter) and that
+        # waiter actually resuming must not slip through, as the freed token is
+        # already reserved for the woken waiter. This window is specific to the
+        # asyncio backend (the trio backend wraps trio's own CapacityLimiter).
+        limiter = CapacityLimiter(1)
+        limiter.acquire_on_behalf_of_nowait("A")
+
+        async def waiter() -> None:
+            await limiter.acquire_on_behalf_of("B")
+
+        async with create_task_group() as tg:
+            tg.start_soon(waiter)
+            await wait_all_tasks_blocked()
+            assert limiter.statistics().tasks_waiting == 1
+
+            # Frees the only token and reserves it for the still-parked "B"; no
+            # token is available, so the nowait acquire must raise WouldBlock.
+            limiter.release_on_behalf_of("A")
+            with pytest.raises(WouldBlock):
+                limiter.acquire_on_behalf_of_nowait("X")
+
+        assert limiter.borrowed_tokens <= limiter.total_tokens
+        assert limiter.available_tokens >= 0
+        assert limiter.statistics().borrowers == ("B",)
+        limiter.release_on_behalf_of("B")
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -8,11 +8,6 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 - Changed ``ByteReceiveStream.receive()`` implementations to raise a ``ValueError`` when
   ``max_bytes`` is not a positive integer
   (`#1191 <https://github.com/agronholm/anyio/pull/1191>`_)
-- Fixed ``CapacityLimiter.total_tokens`` rejecting ``float("inf")`` when the limiter was
-  instantiated outside of an event loop. The adapter setter checked for infinity by
-  identity (``value is math.inf``), so only the exact ``math.inf`` singleton was accepted,
-  while every backend setter (using ``math.isinf()``) accepts any positive infinity
-  (`#1189 <https://github.com/agronholm/anyio/pull/1189>`_; PR by @greymoth-jp).
 - Fixed ``to_process.run_sync()`` deadlocking when the worker function writes enough data
   to ``sys.stderr`` to fill the (undrained) pipe buffer. The worker process now redirects
   ``sys.stderr`` to ``os.devnull`` as well, matching the documented behavior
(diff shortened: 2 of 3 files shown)


tests/test_synchronization.py
```diff
@@ -976,3 +976,32 @@ async def borrower(
             tg.cancel_scope.cancel()
 
         pytest.fail("The second borrower failed to acquire the limiter")
+
+    @pytest.mark.parametrize("anyio_backend", asyncio_params)
```
agronholm: Why is this scoped only to asyncio? The test doesn't use any asyncio APIs.

binggao1230: Good point — the assertion isn't inherently asyncio-specific; it should hold on both backends. I scoped it to asyncio because it currently *fails* on trio, but not for a behavioral reason:

trio's `CapacityLimiter.acquire_nowait` / `acquire_on_behalf_of_nowait` raise `trio.WouldBlock`, which the wrapper passes straight through, whereas `Lock`/`Semaphore` translate it to `anyio.WouldBlock` (`_trio.py` L764-765, L821-822). So `pytest.raises(WouldBlock)` doesn't catch it — the limit itself is respected, it's just the exception type that leaks.

Adding the same two-line translation to those two methods makes the test pass on both backends (verified locally). Want me to fold that fix + un-scope the test into this PR, or keep this one focused on #1170 and handle the trio translation separately?

binggao1230:
The red `test (ubuntu-latest, 3.14, [trio])` job isn't from this PR — it's `tests/test_from_thread.py::TestBlockingPortal::test_asyncio_run_sync_called[asyncio+uvloop]`, where `assert not caplog.text` trips on asyncio's slow-callback warning (`... _call_in_runner_task() ... took 0.136 seconds`, against the 0.1 s threshold).

The same job fails the same way on master — e.g. [run 28965447285](https://github.com/agronholm/anyio/actions/runs/28965447285) and [run 28957305269](https://github.com/agronholm/anyio/actions/runs/28957305269). It's timing-sensitive: the recorded durations across the last three runs here were 0.103 s, 0.118 s and 0.136 s. (The warning's `_asyncio.py` line number differs from master's only because this branch shifts that file.)

3959 tests pass otherwise. I can't re-run the job from a fork, so leaving it as-is.

tests/test_synchronization.py
agronholm: Are you saying that `anyio.CapacityLimiter` somehow raises `trio.WouldBlock`? That would be a serious bug, but I think you're confused.

binggao1230: Yes, but to be precise it's the **trio backend's** `CapacityLimiter` that leaks `trio.WouldBlock` — it forwards to the wrapped trio limiter without translating, whereas `Lock`/`Semaphore` do translate. Minimal repro (no asyncio APIs involved):

```python
import anyio, trio
from anyio import CapacityLimiter, WouldBlock

async def main():
    lim = CapacityLimiter(1)
    lim.acquire_on_behalf_of_nowait("b1")          # fill the one token
    try:
        lim.acquire_on_behalf_of_nowait("b2")      # different borrower -> full
    except BaseException as e:
        print(type(e), "anyio.WouldBlock:", isinstance(e, WouldBlock))

anyio.run(main, backend="trio")     # <class 'trio.WouldBlock'>  anyio.WouldBlock: False
anyio.run(main, backend="asyncio")  # <class 'anyio.WouldBlock'> anyio.WouldBlock: True
```

The asymmetry is in `_backends/_trio.py`: `Lock.acquire_nowait` and `Semaphore.acquire_nowait` both do `except trio.WouldBlock: raise WouldBlock from None`, but `CapacityLimiter.acquire_nowait` / `acquire_on_behalf_of_nowait` call `self.__original.<method>()` directly with no translation. So the new assertion holds on asyncio but raises `trio.WouldBlock` on trio — the limit itself is respected, only the exception type leaks.

That's a separate, pre-existing issue from the #1170 over-granting fix. I can add the same two-line translation to those two methods and un-scope the test, either here or as its own PR — whichever you prefer.

agronholm: I will address the Trio exception leaking through elsewhere. I will merge the master branch to this PR when that has been fixed.

agronholm: I've fixed the Trio problem in #1218. Proceed to refactor the test to run on Trio as well.

binggao1230:
Done in 84e5382 — removed the asyncio scoping so the test runs on all backends. With #1218 merged in, it now passes on `asyncio`, `asyncio+eager`, and `trio`.

agronholm (approved):
Thanks.

PR #1173, opened 2026-06-15
Add concurrency helper utilities: as_completed, gather, amap

Graeme22:
## Changes

Building on #1153, the last remaining item on the February meeting agenda was adding concurrency helpers for common patterns. Specifically, `gather`, `as_completed`, and `amap` were discussed.

To kick off the implementation, I've sketched out simple APIs for `gather`, `as_completed`, `race`, and `amap`. Example usage:

```python
from anyio import amap, as_completed, gather, sleep

async def sleeper(time: float) -> float:
    await sleep(time)
    return time

print(await gather(sleeper(0.3), sleeper(0.2)))  # (0.3, 0.2)
async with as_completed(*[sleeper(i / 10) for i in range(5)]) as results:
    async for res in results:
        print(res)
"""
0
0.1
0.2
0.3
0.4
"""
print(await race(sleeper(0.1), sleeper(0.1))  # 0.1
print(await amap(sleeper, [i / 10 for i in range(5)]))  # [0.0, 0.1, 0.2, 0.3, 0.4]
```

This PR is intended to be a starting point, so I've focused primarily on clean types and avoided complexity that may be introduced later.

A couple notes on the implementation:
- `gather` diverges a bit from `asyncio`'s API as it omits the `return_exceptions` parameter, which would likely introduce a lot of complexity and/or assumptions (eg `Exception` vs `BaseException`, potentially doubling the number of overloads to make everything `T | Exception`).
- `gather`'s typing is a bit eccentric. The idea is copied from typshed's `asyncio.gather` typing where a large number of overloads and type vars are used to handle 99% of common cases. This is unfortunately necessary since there's otherwise no good way to extract types from a sequence of awaitables. Unlike asyncio, I have gather returning a tuple instead of a list for consistency's sake (weirdly, `asyncio.gather` is *typed* as returning a `tuple[R, S, T]` but actually returns a `list` with elements in the corresponding indices!)
- `as_completed` is structured as a context manager instead of as a simple function, which is more verbose but is necessary to preserve structured concurrency (no yielding out of a task group in a generator). This also allows doing things in between starting the tasks and iterating over results (all tasks are started before the `yield`).
- `race` could potentially be used to further simplify the happy eyeballs implementation once we agree on the API.

Naming, types, and API design are all up in the air--feedback welcome!

## Checklist

features


agronholm:
If it helps, here is the existing code I had written to this end (in an earlier branch):

<details><summary>Details</summary>
<p>

```py
class AsyncResultsIterator(AsyncContextManagerMixin):
    _task_group: EnhancedTaskGroup
    _receive: MemoryObjectReceiveStream[TaskHandle]

    def __init__(
        self,
        coros: Iterable[Coroutine[Any, Any, Any]],
        task_limiter: AbstractTaskLimiter | None = None,
    ) -> None:
        self._coros: deque[Coroutine[Any, Any, Any]] = deque(coros)
        self._task_limiter = task_limiter

    def cancel(self) -> None:
        """Cancel all the currently running tasks within this result iterator."""
        self._task_group.cancel()
        self._clear_coros()

    def _clear_coros(self) -> None:
        for coro in self._coros:
            coro.close()

        self._coros.clear()

    async def _run_task(
        self, coro: Coroutine[Any, Any, T], send: MemoryObjectSendStream[TaskHandle]
    ) -> T:
        try:
            return await coro
        finally:
            send.send_nowait(_current_task_handle.get())

    async def _feed_tasks(self, send: MemoryObjectSendStream[TaskHandle]) -> None:
        tasks_spawned = 0
        with send:
            async with EnhancedTaskGroup() as tg:
                while self._coros:
                    coro = self._coros.popleft()
                    try:
                        await tg.start_task(
                            self._run_task(coro, send),
                            name=f"Task {tasks_spawned} of {self}",
                            task_limiter=self._task_limiter,
                        )
                    except BaseException:
                        coro.close()
                        raise

                    tasks_spawned += 1

    @asynccontextmanager
    async def __asynccontextmanager__(self) -> AsyncGenerator[Self]:
        from anyio import create_memory_object_stream

        async with AsyncExitStack() as exit_stack:
            exit_stack.callback(self._clear_coros)
            send, self._receive = create_memory_object_stream[TaskHandle](
                len(self._coros)
            )
            exit_stack.enter_context(self._receive)
            self._task_group = await exit_stack.enter_async_context(EnhancedTaskGroup())
            await self._task_group.start_task(self._feed_tasks(send))
            yield self

    def __aiter__(self) -> Self:
        return self

    async def __anext__(self) -> TaskHandle[Any]:
        try:
            return await self._receive.receive()
        except EndOfStream:
            raise StopAsyncIteration from None


async def amap(
    func: Callable[[T], Coroutine[Any, Any, TRetval]],
    args: Iterable[T],
    *,
    task_limiter: AbstractTaskLimiter | None = None,
) -> Sequence[TRetval]:
    """
    Run the given coroutine function concurrently for multiple argument values.

    :param func: a coroutine function that takes a single argument
    :param args: a sequence of argument values to pass to ``func``
    :param task_limiter: limits the number of tasks running concurrently and the rate at
        which new tasks are started
    :return: task results for each argument in the same order as they were passed

    """
    async with EnhancedTaskGroup() as tg:
        tasks = [
            await tg.start_task(func(arg), task_limiter=task_limiter) for arg in args
        ]

    return [await task for task in tasks]


def as_completed(
    coros: Iterable[Coroutine[Any, Any, T]],
    /,
    *,
    task_limiter: AbstractTaskLimiter | None = None,
) -> AbstractAsyncContextManager[AsyncIterable[TaskHandle[T]]]:
    """
    Run the given coroutines concurrently in a task group and yield task handles as they
    complete.

    :param coros: an iterable of coroutine objects to run as tasks
    :param task_limiter: limits the number of tasks running concurrently and the rate at
        which new tasks are started
    :return: an async context manager yielding an async iterator that yields task
        handles in the order they finished

    """
    return AsyncResultsIterator(coros, task_limiter)


async def gather(
    *coros: Coroutine[Any, Any, T],
    task_limiter: AbstractTaskLimiter | None = None,
) -> Sequence[T]:
    """
    Run a number of coroutines in a task group and return their results.

    :param coros: coroutine objects to run as tasks
    :param task_limiter: limits the number of tasks running concurrently and the rate at
        which new tasks are started
    :return: a sequence of return values from the coroutines in the same order as they
        were passed

    """
    async with EnhancedTaskGroup() as tg:
        coro_iterator = iter(coros)
        try:
            tasks = [
                await tg.start_task(coro, task_limiter=task_limiter) for coro in coros
            ]
        except BaseException:
            for coro in coro_iterator:
                coro.close()

            raise

    return [await handle for handle in tasks]


async def race(
    *coros: Coroutine[Any, Any, T],
    task_limiter: AbstractTaskLimiter | None = None,
) -> T:
    """
    Run the given coroutines concurrently and return the first one to complete.

    When the first task completes, the remaining tasks are cancelled and their results
    are discarded. If any task raises an exception, it is propagated to the caller in an
    exception group.

    :param coros: coroutine objects to run as tasks
    :param task_limiter: limits the number of tasks running concurrently and the rate at
        which new tasks are started
    :return: the return value of the first completed task

    """
    if not coros:
        raise ValueError("race() takes at least one coroutine")

    retval_container: list[T] = []

    async def helper(coro_: Coroutine[Any, Any, T]) -> None:
        local_retval = await coro_
        if not retval_container:
            retval_container.append(local_retval)
            tg.cancel()

    async with EnhancedTaskGroup() as tg:
        coro_iterator = iter(coros)
        for coro in coro_iterator:
            try:
                await tg.start_task(helper(coro), task_limiter=task_limiter)
            except BaseException:
                for coro_ in coro_iterator:
                    coro_.close()

                raise

    return retval_container[0]
```

</p>
</details>

Graeme22:
> If it helps, here is the existing code I had written to this end (in an earlier branch):

Thanks, I had forgotten about that! Thoughts on having `as_completed` just return a `MemoryObjectReceiveStream` as here?

agronholm:
I'm mostly pretty happy with how the API looks like. Regarding `race()`, [here](https://github.com/achimnol/aiotools/blob/main/src/aiotools/utils.py#L106) is the `aiotools` implementation. One notable point I would criticize is that when `continue_on_error=True`, it temporarily swallows **any** non-base exception. Otherwise I would be quite happy to utilize that for the happy eyeballs algorithm (which is even mentioned in the docstring).

Graeme22:
> I'm mostly pretty happy with how the API looks like. Regarding `race()`, [here](https://github.com/achimnol/aiotools/blob/main/src/aiotools/utils.py#L106) is the `aiotools` implementation. One notable point I would criticize is that when `continue_on_error=True`, it temporarily swallows **any** non-base exception. Otherwise I would be quite happy to utilize that for the happy eyeballs algorithm (which is even mentioned in the docstring).

Perhaps it could be `continue_on_errors: tuple[type[BaseException], ...]`? If things are mostly good I will add tests/docs and mark this ready for review. I had kind of expected people to have more thoughts on gitter, not sure how many people got the chance to weigh in? Also, do you think these functions should be exported in the main module eg `anyio.gather` or would you rather keep them in the `utils` module?

Graeme22:
In the happy eyeballs impl this line: https://github.com/agronholm/anyio/blob/149b9e907618fadf6840a4d3cebad533b0c7d033/src/anyio/_core/_sockets.py#L199 means race only saves a little bit of code as we'd still need a wrapper function to handle cleanup in case two attempts are successful. Also the sleep would need to be baked in at the start since race starts coros at the same time. Does that sound right?

agronholm:
Seems to me like we should leave `race` out of this for the time being. I think we have a better consensus on the other helpers. @Graeme22 is there anything else blocking this from being reviewed?

agronholm:
On closer inspection, `as_completed()` might need some redesign. The asyncio counterpart is useful in that it lets the consumer deal with errors manually. I feel it would almost make `race()` redundant. The issues then are, what would actually be returned, and how to deal with remaining exceptions after the context manager is being exited.

Graeme22:
> On closer inspection, `as_completed()` might need some redesign. The asyncio counterpart is useful in that it lets the consumer deal with errors manually. I feel it would almost make `race()` redundant. The issues then are, what would actually be returned, and how to deal with remaining exceptions after the context manager is being exited.

Hmm, that would be an interesting approach. My gut instinct it to keep things simple and not add any error handling here (similar to `gather` and `amap`). If users want to add error handling, they could add it directly to the coroutines and structure it however they want.

If we were to go that route, would we want to return `TaskHandle`s and suppress errors that then get re-raised when eventually awaited? The fact that everything is running in the same `TaskGroup` complicates this a lot compared to the naive `asyncio` approach.

Vizonex:
@Graeme22 If we were to implement a `race()` function something I would consider doing are these core steps and it should be utilized similar to that of a brute-forcing mechanism.
- Suppressing defined exceptions given by the end user and for any that aren't defined let it be thrown.
- On success cancel all racing tasks without throwing an exception.

agronholm:
I'll focus on this next in my current review round.

agronholm:
Here's what we could do with `as_completed()`:
```py
@asynccontextmanager
async def as_completed(
    *coros: Coroutine[Any, Any, R],
) -> AsyncGenerator[MemoryObjectReceiveStream[TaskHandle[R]]]:
    """
    Run awaitable objects concurrently in a task group, returning an iterator which can
    be used to get the finished task handles in the order they complete.

    :param coros: coroutine objects to run as tasks
    :return: MemoryObjectReceiveStream for iterating over the task handles as tasks
        complete

    """
    if not coros:
        raise ValueError("as_completed() takes at least one coroutine")

    async def runner(
        _coro: Coroutine[Any, Any, R],
        index: int,
        _send: MemoryObjectSendStream[TaskHandle[R]],
    ) -> R:
        async with _send:
            try:
                return await coro
            finally:
                try:
                    _send.send_nowait(task_handles[index])
                except BrokenResourceError:
                    pass

    send, recv = create_memory_object_stream[TaskHandle[R]](len(coros))
    task_handles: list[TaskHandle[R]] = []
    async with recv, create_task_group() as tg:
        async with send:
            for i, coro in enumerate(coros):
                task_handles.append(tg.create_task(runner(coro, i, send.clone())))

        try:
            yield recv
        finally:
            tg.cancel_scope.cancel()
```

Graeme22:
> Here's what we could do with `as_completed()`:

I'm not necessarily opposed to returning `TaskHandle` instead of the task return values, but shouldn't the convention be the same across all 3 utilities? That is, if we return handles instead of results in `as_completed` shouldn't we want to do that in `gather` as well?

agronholm:
> > Here's what we could do with `as_completed()`:
> 
> I'm not necessarily opposed to returning `TaskHandle` instead of the task return values, but shouldn't the convention be the same across all 3 utilities? That is, if we return handles instead of results in `as_completed` shouldn't we want to do that in `gather` as well?

Not necessarily? `asyncio.gather()` returns the return values directly (or exceptions, if `return_exceptions` was provided) whereas `asyncio.as_completed()` yields tasks from the async iterator.

Graeme22:
> > > Here's what we could do with `as_completed()`:
> > 
> > 
> > I'm not necessarily opposed to returning `TaskHandle` instead of the task return values, but shouldn't the convention be the same across all 3 utilities? That is, if we return handles instead of results in `as_completed` shouldn't we want to do that in `gather` as well?
> 
> Not necessarily? `asyncio.gather()` returns the return values directly (or exceptions, if `return_exceptions` was provided) whereas `asyncio.as_completed()` yields tasks from the async iterator.

Ah I see, I wasn't thinking of asyncio's API at all. Notable gather already diverges in some ways: no `return_exceptions` and it returns a tuple instead of a list

agronholm:
> > > > Here's what we could do with `as_completed()`:
> > > 
> > > 
> > > I'm not necessarily opposed to returning `TaskHandle` instead of the task return values, but shouldn't the convention be the same across all 3 utilities? That is, if we return handles instead of results in `as_completed` shouldn't we want to do that in `gather` as well?
> > 
> > 
> > Not necessarily? `asyncio.gather()` returns the return values directly (or exceptions, if `return_exceptions` was provided) whereas `asyncio.as_completed()` yields tasks from the async iterator.
> 
> Ah I see, I wasn't thinking of asyncio's API at all. Notable gather already diverges in some ways: no `return_exceptions` and it returns a tuple instead of a list

We could return a list from `gather()` too.

Graeme22:
> We could return a list from `gather()` too.

Unfortunately that would make the type hints a lie. Personally I'd prefer to keep the honest type, but perhaps we could change the name from gather to something else to avoid confusion?

Regardless I will make the `as_completed` changes

agronholm:
> > We could return a list from `gather()` too.
> 
> Unfortunately that would make the type hints a lie. Personally I'd prefer to keep the honest type, but perhaps we could change the name from gather to something else to avoid confusion?
> 
> Regardless I will make the `as_completed` changes

I just checked, and `asyncio.gather()` seems to be typed to return either a tuple or a list depending on the number of arguments. Bleh. In that case I don't really care about that detail.

tests/test_utilities.py
```diff
@@ -0,0 +1,114 @@
+from __future__ import annotations
+
+import pytest
+
+from anyio import Event, amap, as_completed, gather
+
+
+class TestUtilties:
```
agronholm: There's really no reason to wrap all the tests in the module in a class.

agronholm:
I think we're getting pretty close here. I'll do a proper review tomorrow.

docs/api.rst
```diff
@@ -287,3 +287,10 @@ Exceptions
 .. autoexception:: anyio.TaskNotFinished
 .. autoexception:: anyio.TypedAttributeLookupError
 .. autoexception:: anyio.WouldBlock
+
+Utility functions
```
agronholm:
```suggestion
Concurrency utilities
```

src/anyio/__init__.py
```diff
@@ -97,6 +97,9 @@
 from ._core._typedattr import TypedAttributeProvider as TypedAttributeProvider
 from ._core._typedattr import TypedAttributeSet as TypedAttributeSet
 from ._core._typedattr import typed_attribute as typed_attribute
+from ._utils import amap as amap
+from ._utils import as_completed as as_completed
+from ._utils import gather as gather
```
agronholm: I feel this code should live under `_core`. It's not really a special case that warrants a top-level module.

src/anyio/_utils.py
```diff
@@ -0,0 +1,150 @@
...
+    Run awaitable objects concurrently in a task group, returning an iterator which can
+    be used to get finished task handles in the order they complete.
+
+    :param coros: coroutine objects to run as tasks
+    :return: MemoryObjectReceiveStream for iterating over task handles as tasks complete.
+
+    """
+    if not coros:
+        raise ValueError("as_completed() takes at least one coroutine")
+    send, recv = create_memory_object_stream[TaskHandle[R]](len(coros))
```
agronholm: Add a blank line after a control block ends.
```suggestion
        raise ValueError("as_completed() takes at least one coroutine")

    send, recv = create_memory_object_stream[TaskHandle[R]](len(coros))
```

src/anyio/_utils.py
```diff
@@ -0,0 +1,150 @@
...
+    handles: list[TaskHandle[Any]] = []
+    async with create_task_group() as tg:
+        handles.extend(tg.create_task(coro) for coro in coros)
+
+    return tuple(h.return_value for h in handles)
+
+
+@asynccontextmanager
+async def as_completed(
+    *coros: Coroutine[Any, Any, R],
```
agronholm: I'm not sure if we talked about this yet, but is there any reason why we couldn't support aribtrary awaitables here, like the asyncio counterpart does? We're not even doing any coroutine checks here. If we do this, then `TaskHandle` items may need a bit of special casing, but otherwise I see no problem.

agronholm (changes requested):
Mostly just small changes, but we'll need to discuss awaitable support in `as_completed()`.

src/anyio/__init__.py
agronholm: Maybe `anyio._core._concurrency_helpers`?

src/anyio/_core/_concurrency_utils.py
```diff
@@ -0,0 +1,150 @@
...
+            finally:
+                try:
+                    _send.send_nowait(task_handles[index])
+                except BrokenResourceError:
+                    pass
+
+    async with recv, create_task_group() as tg:
+        async with send:
+            for i, coro in enumerate(coros):
+                task_handles.append(tg.start_soon(runner, coro, i, send.clone()))
```
agronholm: One more thing: the task names will reflect that of the internal `runner()` function. This should be rectified by generating task names based on the coroutines like `TaskGroup` does normally. Also, do add an assertion for that in the tests.

src/anyio/_utils.py
agronholm: For any observers here, we concluded on Gitter/Matrix that this is fine and in line with asyncio, and `gather()` should not be modified the same way. This will preserve asyncio parity.

src/anyio/__init__.py
Graeme22: I did `_concurrency_utils` to be consistent with naming in docs

src/anyio/_core/_concurrency_utils.py
Graeme22: Except now we're allowing any `Awaitable` not just coroutines.

agronholm: Right. But I think we could still improve here by deriving the name from the coroutine if one was passed, otherwise using a default task name. What do you think?

agronholm:
It looks like the exception patterns need updating in the test suite.

src/anyio/_core/_concurrency_utils.py
Graeme22: Sounds good, I used `inspect.iscoroutine`

Graeme22:
> It looks like the exception patterns need updating in the test suite.

Done. Test failures look unrelated

agronholm:
> > It looks like the exception patterns need updating in the test suite.
> 
> Done. Test failures look unrelated

The seem to be caused by an rsloop release from 2 hours ago.

agronholm:
The culprit is v0.1.37. v0.1.36 works fine. There's a release in progress for v0.1.39; I'll see if that changes anything.

agronholm:
It didn't help. I'll file a bug report.

agronholm:
https://github.com/RustedBytes/rsloop/issues/68

agronholm:
Windows keeps timing out. I suspect this might not be a bug but the test suite taking too long.

agronholm (changes requested):
While I figure out the Windows timeouts, please rename the test module to match the implementation module name.

agronholm:
It's failing on 3.14t because rsloop no longer allows running on free-threaded. Probably better if it didn't properly support it in the first place.

agronholm:
I checked on a real Windows box, and it hangs on `TestTCPStream::test_concurrent_send` on rsloop. The macOS + py3.14 job crashes on a signal test (also on rsloop). What this means is that I will remove rsloop from the test suite now. It's far too brittle.

Graeme22:
> I checked on a real Windows box, and it hangs on `TestTCPStream::test_concurrent_send` on rsloop. The macOS + py3.14 job crashes on a signal test. What this means is that I will remove rsloop from the test suite now. It's far too brittle.

Sounds good!

agronholm (approved):
Thank you!

PR #1175, opened 2026-06-15
Fix asyncio waiting on closing streams in ``Process.wait()``

### LINKED ISSUE #1166: Deadlock on shutting down subprocesses blocked on writing to full pipe on asyncio backend.

### LINKED ISSUE #1174: Inconsistent semantics for Process.wait() on trio, asyncio and asyncio+uvloop
### Things to check first

### AnyIO version

4.13 

### Python version

3.12

### What happened?

[Process.wait](https://anyio.readthedocs.io/en/stable/api.html#anyio.abc.Process.wait) Is documented to return when the process exits. 

This matches [trio's documentation](https://trio.readthedocs.io/en/stable/reference-io.html#trio.Process.wait) behavior and anyio's behavior when run on the `trio` or `asyncio+uvloop` backends.

On the vanilla `asyncio` backend it however wait's until the stdout/stderr pipes are closed as well if they are redirected. This is indeed how the underlying [`asyncio.subprocess.Process.wait`](https://docs.python.org/3/library/asyncio-subprocess.html#asyncio.subprocess.Process.wait) behaves as well (but not [`subprocess.Popen.wait`](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.wait)

I would argue the stdlibs behavior here is a bug as it leaves no way to actually just wait for the subprocess and can lead to deadlocks and races quite easily (see #1166).

Easiest way to reproduce is by having a subprocess inherit stdout to keep it open (see attached test) but it can also be triggered by writing enough to stdout to pause the `StreamReader` which doesn't get unpaused by the process exiting.

```python
# Added in test_subprocesses.py. Will fail on just `asyncio` and `asyncio+eager` 
async def test_wait_returns_on_process_exit_with_open_stdout() -> None:
    """
    wait() should return once the process exits, not waiting on piped stdout/stderr to close.
    Easiest triggered by a grandchild keeping an inhetited file-descriptor open
    """
    code = dedent("""\
    import subprocess, sys
    subprocess.Popen(
        [sys.executable, "-c", "import select, sys; select.select([sys.stdin], [], [], 10)"]
    )
    sys.stdout.write("o")
    """)

    # Ensure we close stdin to unblock grandchild on failure.
    async with process.stdin:
        with fail_after(5):
            assert await process.wait() == 0
```

tapetersen:
## Changes

Normalizes `Process.aclose()` and `Process.wait()` to same behavior as trio and asyncio+uvloop

Fixes #1174. and Fixes #1166.
Adds an explicit event resolved on `process_exit` that can be used to wait on only process for .wait()
For `aclose` it closes the underlying process's pipe_transports for stdin/stdout/stderr which the stream_wrappers don't do by default.

## Checklist

features

### DIFF
--- a/tests/test_subprocesses.py
+++ b/tests/test_subprocesses.py
@@ -17,6 +17,7 @@ from anyio import (
     ClosedResourceError,
     EndOfStream,
     create_task_group,
+    fail_after,
     open_process,
     run_process,
 )
@@ -391,3 +392,35 @@ async def test_close_while_reading() -> None:
             await process.stdout.receive()
 
         process.terminate()
+
+
+async def test_wait_returns_on_process_exit_with_open_stdout() -> None:
+    """
+    wait() should return once the process exits, rather than waiting for the piped
+    stdout/stderr to close. Easiest triggered by a grandchild that inherits the
+    stdout pipe and keeps it open after the immediate child has exited.
+    """
+    code = dedent("""\
+    import subprocess, sys
+
+    subprocess.Popen(
+        [sys.executable, "-c", "import select, sys; select.select([sys.stdin], [], [], 10)"]
+    )
+    """)
+
+    process = await open_process([sys.executable, "-c", code])
+    assert process.stdin is not None
+    assert process.stdout is not None
+    assert process.stderr is not None
+
+    # Closing stdin unblocks the grandchild
+    async with process.stdin:
+        with fail_after(5):
+            assert await process.wait() == 0
+
+    # Wait for grandchild to exit to avoid ResourceWarnings
+    with fail_after(3, shield=True):
+        async for _ in process.stdout:
+            pass
+        async for _ in process.stderr:
+            pass
--- a/tests/test_subprocesses.py
+++ b/tests/test_subprocesses.py
@@ -424,3 +424,25 @@ async def test_wait_returns_on_process_exit_with_open_stdout() -> None:
             pass
         async for _ in process.stderr:
             pass
+
+
+async def test_close_with_stdout_blocked_subprocess(anyio_backend_name: str) -> None:
+    """
+    Regression test for #1166.
+
+    Test that closing a process unblocks a subprocess blocked on writing to stdout
+    (because the pipe buffer is full and nobody is reading), instead of deadlocking
+    while waiting for it to exit.
+    """
+
+    process = await open_process(
+        [sys.executable, "-c", "import sys;sys.stdout.write('x' * 1024 * 1024)"]
+    )
+    try:
+        with fail_after(5):
+            await process.aclose()
+    except TimeoutError:
+        if anyio_backend_name == "asyncio":
+            process._process._transport.close()  # type: ignore[attr-defined]
+
+        pytest.fail("Process.aclose() deadlocked")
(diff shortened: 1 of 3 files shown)


tests/test_subprocesses.py
```diff
@@ -391,3 +392,58 @@ async def test_close_while_reading() -> None:
             await process.stdout.receive()
 
         process.terminate()
+
+
+async def test_wait_returns_on_process_exit_with_open_stdout() -> None:
```
agronholm, line 397: This test already passes on `master`.

agronholm: This is okay if its only purpose is to add test coverage, but is this the case?

tapetersen: Nope. I'll check. Sounds like some refactoring may have broken it.

tapetersen: Hmm if fails for me on asyncio/asyncio+eager without patch for latest anyio master on python 3.12 - 3.14 (passes on 3.10 and 3.11 which makes sense regarding the timeline)

Above was from tox.

I could see the test passing if `stdin` of the test-process is already pointed at something that let's select return if your're not running it in a terminal directly. 

The test is really about stderr/stdout being kept open and `stdin`/`select` is just to ensure the grandchild is really shutdown in one way or another. I could replace it with a reasonable sleep instead, clearer test but leaks subprocesses and can be a bit racy.

tapetersen:
@gschaffner I wouldn't mind your thoughts as well, especially regarding #828 as this basically resolves number 3 there with the position that this is an unintended bug in asyncio rather than desired behavior (which I very much believe).

tests/test_subprocesses.py
agronholm: Ah, I routinely just test on 3.10, that would be why I didn't see it fail.

agronholm (approved):
Alright, I can't claim 100% understanding of this but nothing suspicious stands out. Thanks!

PR #1177, opened 2026-06-15
Fix Process.wait() deadlock on asyncio backend when subprocess blocks on full pipe

### LINKED ISSUE #1174: Inconsistent semantics for Process.wait() on trio, asyncio and asyncio+uvloop

goingforstudying-ctrl:
## Changes

Fixes #1174 and #1166.

On the asyncio backend, Process.wait() delegates to
asyncio.subprocess.Process.wait(), which waits for both the process exit
and the closure of stdout/stderr pipes. When a subprocess is blocked
writing to a full pipe (because nobody is reading), the pipe never closes
and wait() hangs forever.

Two changes:

1. StreamReaderWrapper.aclose() now closes the underlying transport rather
   than only setting a stream exception. This unblocks the subprocess by
   delivering EPIPE / BrokenPipeError on the write end.

2. Process.wait() uses a polling loop on returncode with event-loop yield
   rather than calling the stdlib's wait(). The polling loop
   returns as soon as the process exits, never waiting for pipes.

Two new tests verify the fix.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ __pycache__
 .cache
 .local
 CLAUDE.local.md
+.venv/
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ __pycache__
 .local
 CLAUDE.local.md
 .venv/
+.rocky
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -1138,10 +1138,16 @@ class Process(abc.Process):
                 raise
 
     async def wait(self) -> int:
-        # Use a polling loop instead of asyncio.subprocess.Process.wait(),
-        # which waits for stdout/stderr pipes to close in addition to the
-        # process exiting.  That extra wait can deadlock when a subprocess is
-        # blocked writing to a full pipe and nobody is reading.
+        # Use a polling loop instead of asyncio.subprocess.Process.wait(), which
+        # waits for stdout/stderr pipes to close in addition to the process
+        # exiting.  The extra pipe-close wait is problematic because it can
+        # deadlock when a subprocess is blocked writing to a full pipe and
+        # nobody is reading from the other end.  The polling loop checks only
+        # the process returncode, so it returns as soon as the process exits
+        # regardless of pipe state.  This is safe because the caller is
+        # responsible for draining/closing pipes separately, and the polling
+        # interval is bounded by a single checkpoint (essentially zero wait
+        # when the event loop has no other work).
         while self._process.returncode is None:
             await AsyncIOBackend.checkpoint()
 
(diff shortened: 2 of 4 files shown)


src/anyio/_backends/_asyncio.py
```diff
@@ -1076,6 +1076,8 @@ async def receive(self, max_bytes: int = 65536) -> bytes:
 
     async def aclose(self) -> None:
         self._stream.set_exception(ClosedResourceError())
+        if self._stream._transport is not None:  # type: ignore[attr-defined]
+            self._stream._transport.close()  # type: ignore[attr-defined]
```
ghost, line 1080: Add more checks like this, they are useful

src/anyio/_backends/_asyncio.py
```diff
@@ -1136,7 +1138,14 @@ async def aclose(self) -> None:
                 raise
 
     async def wait(self) -> int:
-        return await self._process.wait()
+        # Use a polling loop instead of asyncio.subprocess.Process.wait(),
```
ghost: Make the comment 10 lines long instead of 4

tests/test_subprocesses.py
```diff
@@ -391,3 +391,67 @@ async def test_close_while_reading() -> None:
             await process.stdout.receive()
 
         process.terminate()
+
+
+@pytest.mark.skipif(
+    platform.system() == "Windows",
+    reason="Unix-specific: grandchild process and signal handling",
+)
+async def test_wait_returns_on_process_exit_with_open_stdout() -> None:
```
ghost, line 400: Need more tests

.gitignore
```diff
@@ -17,3 +17,4 @@ __pycache__
 .cache
 .local
 CLAUDE.local.md
+.venv/
```
ghost, line 20: Also ignore .rocky for new toolchain requirement

ghost (changes requested):

.gitignore
goingforstudying-ctrl: Added .rocky to .gitignore.

tests/test_subprocesses.py
goingforstudying-ctrl: Added three more tests: wait() with blocked-on-write subprocess, wait() with fast exit, and wait() with open stderr pipe from grandchild.

src/anyio/_backends/_asyncio.py
goingforstudying-ctrl: Expanded the comment to explain why the polling loop is safe and what the caller is responsible for.

src/anyio/_backends/_asyncio.py
goingforstudying-ctrl: Added null guards before all _transport.close() calls in the worker shutdown path, matching the pattern in StreamReaderWrapper.aclose().

tapetersen:
@goingforstudying-ctrl Have you checked at all for existing PR's?

agronholm:
Duplicate of #1175.

PR #1178, opened 2026-06-17
Re-based fix-837 branch to master with fixed-changelog.

### LINKED ISSUE #837: CancelScope shielding fails to protect against cancellation until task_status.started is called on asyncio

tapetersen:

## Changes

Fixes #837 (or rather as noted just adds a regression test and note in changelog for it as it was already solved>
This otherwise just a rebase of #839

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -3,16 +3,6 @@ Version history
 
 This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
-**UNRELEASED**
-
-- Fixed cancellation edge case on asyncio where a task spawning another with
-  ``TaskGroup.start()`` is not protected from external cancellation even when
-  the subtask has not yet called ``task_status.started()`` and is in a shielded
-  cancel scope.
-  This was actually solved already in 4.14.0 by ``TaskHandle`` refactoring.
-  This patch just adds a regression test.
-  (`#837 <https://github.com/agronholm/anyio/issues/837>`_)
-
 **4.14.0**
 
 - Added support for Python 3.15
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -3,6 +3,16 @@ Version history
 
 This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
+**UNRELEASED**
+
+- Fixed cancellation edge case on asyncio where a task spawning another with
+  ``TaskGroup.start()`` is not protected from external cancellation even when
+  the subtask has not yet called ``task_status.started()`` and is in a shielded
+  cancel scope.
+  This was actually solved already in 4.14.0 by ``TaskHandle`` refactoring.
+  This patch just adds a regression test.
+  (`#837 <https://github.com/agronholm/anyio/issues/837>`_)
+
 **4.14.0**
 
 - Added support for Python 3.15
--- a/tests/test_taskgroups.py
+++ b/tests/test_taskgroups.py
@@ -2008,6 +2008,34 @@ async def test_cancel_reason() -> None:
     exc_info.match(task.name)
 
 
+async def test_cancel_shielding_start() -> None:
+    """
+    Test that if the host task that has spawned a subtask via ``start()`` is cancelled,
+    a shielded cancel scope in the child task will shield it from cancellation.
+
+    Regression test for #837.
+    """
+
+    async def taskfunc(*, task_status: TaskStatus[None]) -> None:
+        with CancelScope(shield=True):
+            entered_inner_scope.set()
+            try:
+                await checkpoint()
+                await checkpoint()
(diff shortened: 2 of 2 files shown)


tapetersen:
@agronholm I figured the changelog entry that it's fixed is nice to have for confirming version and the test doesn't hurt.
Feel free to close otherwise.

docs/versionhistory.rst
```diff
@@ -3,6 +3,16 @@ Version history
...
+**UNRELEASED**
+
+- Fixed cancellation edge case on asyncio where a task spawning another with
+  ``TaskGroup.start()`` is not protected from external cancellation even when
+  the subtask has not yet called ``task_status.started()`` and is in a shielded
+  cancel scope.
+  This was actually solved already in 4.14.0 by ``TaskHandle`` refactoring.
+  This patch just adds a regression test.
+  (`#837 <https://github.com/agronholm/anyio/issues/837>`_)
+
```
agronholm: An already fixed problem nor an added test warrant a changelog entry.

tapetersen: Corrected

tests/test_taskgroups.py
```diff
@@ -2008,6 +2008,34 @@ async def test_cancel_reason() -> None:
...
+
+        # The cancellation should be triggered here, and not any earlier
+        await checkpoint()
+
+    entered_inner_scope = Event()
+    with CancelScope() as outer_scope:
+        async with create_task_group() as tg:
+            tg.start_soon(tg.start, taskfunc)
+            await entered_inner_scope.wait()
+            outer_scope.cancel()
```
agronholm: I feel these changes would make the test less reliant on the scheduling order. What do you think?
```suggestion
                await resume_event.wait()
            except get_cancelled_exc_class():
                pytest.fail("Shouldn't be cancelled in a shielded scope")

        # The cancellation should be triggered here, and not any earlier
        await checkpoint()

    resume_event = Event()
    with CancelScope() as outer_scope:
        async with create_task_group() as tg:
            tg.start_soon(tg.start, taskfunc)
            await wait_all_tasks_blocked()
            outer_scope.cancel()
            resume_event.set()
```

tapetersen: Absolutely! I'm all for explicit ordering for tests.

(I think it was you that added checkpoints  when transferring my initial testcase and i thought about it but let it be as a simple matter of taste between consiceness of test-case (I very rarely used them at all at that point and like I said I tend to prefer explicit Events whenever I write tests myself)).

tapetersen:
@agronholm Your refactoring fixed this case but there are still some less serious differences from Trio. I was unsure if you though they were deemed serious enough to file Issues and potential fixes for. If they are I'll do with proper demonstrators.

Since we now start the `Task` directly in the `TaskGroup` rather than move it in when `started()` is called as Trio there are at least 3 visible side-effects if the `TaskGroup` is cancelled in that timeframe (and we assume `TaskGroup.start` is called by a not descending itself from that group/scope):

1. On Trio the started task doesn't get a `CancelledError` until the first checkpoint *after* the `started()` call.
2. Related to the first on Trio that means any value passed to `started` will be available and correctly passed back to the caller of `start`
3. On the asyncio backend the `CancelledError` leaks out of the `start()` call to the calling task (the most serious).

Our work is potentially affected by the last (3) as in it will potentially lead to a more confusing debug experience and possibly has already, still not critical though. It's also the one that could probably be relatively simply fixed with a workaround while 1 and 2 would require some more refactoring from what I can see.

agronholm:
If a piece of code works differently on two different back-ends, then it's almost certainly a bug (except for task scheduling differences, cancellation exception differences and stuff like that of course).

agronholm (approved):

PR #1182, opened 2026-06-19
Add missing await to open_file() in file I/O concurrency example

patrickwehbe:
## Changes

The "Limiting concurrency" example in `docs/fileio.rst` uses `async with open_file('/foo/bar', limiter=limiter) as fp:`. Since `open_file()` is a coroutine function, this raises `TypeError: 'coroutine' object does not support the asynchronous context manager protocol` when run as written.

Added the missing `await` so the example actually runs. It now matches the other two `open_file()` examples in the same file, which already use `async with await open_file(...)`.

## Checklist

features

these instructions.

### DIFF
--- a/docs/fileio.rst
+++ b/docs/fileio.rst
@@ -102,7 +102,7 @@ methods)::
 
     async def main():
         limiter = CapacityLimiter(5)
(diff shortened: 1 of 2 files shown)


agronholm (approved):
Thanks.

PR #1183, opened 2026-06-22
Allow 0 tokens in a CapacityLimiter instantiated outside an event loop

nyxst4ck:
## Changes

Fixes #. 

PR #1019 (released in 4.12) allowed setting a `CapacityLimiter`'s total tokens to `0`, updating the docstring, the changelog and the asyncio backend setter (`value < 1` → `value < 0`).

However, the `CapacityLimiterAdapter.total_tokens` setter in `src/anyio/_core/_synchronization.py` was missed and still enforced `value < 1`. This adapter is used whenever a `CapacityLimiter` is instantiated **outside of a running event loop**, so the documented 4.12 behavior was not honored on that path:

```python
import anyio

anyio.CapacityLimiter(0)             # ValueError: total_tokens must be >= 1
limiter = anyio.CapacityLimiter(1)
limiter.total_tokens = 0             # ValueError: total_tokens must be >= 1
```

This aligns the adapter setter with the asyncio backend setter (`value < 0`, message `"total_tokens must be >= 0"`) and adds a regression test for the outside-event-loop path.

## Checklist

features

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -13,7 +13,7 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 - Fixed ``CapacityLimiter.total_tokens`` rejecting a value of ``0`` when the limiter was
   instantiated outside of an event loop, contradicting the documented behavior of
   allowing 0 total tokens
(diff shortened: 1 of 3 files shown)


agronholm (approved):
Looks good, thanks!

PR #1189, opened 2026-06-26
Fix CapacityLimiter(math.inf) raising TypeError outside an event loop

mahirhir:
## Changes

Constructing `CapacityLimiter(float('inf'))` raises `TypeError` because the constructor validates with `value is not math.inf` (identity), while the backend setter correctly uses `math.isinf`. The two checks were aligned in #1183 for the setter path; the constructor path was the remaining gap.

The fix makes the constructor use the same `isinf` check. One line, and the full suite (183 tests) passes.

## Checklist

features

these instructions.


agronholm (changes requested):
Add the PR link to the changelog entry. Otherwise this looks good.

agronholm:
I also seems like you erased the checkboxes from PR template which you were explicitly told NOT to do.

mahirhir:
Sorry about the template, I've restored it with the checklist filled in. The PR link is now added to the changelog entry as well.

agronholm (approved):
Thanks!

PR #1190, opened 2026-06-27
Use the PEP 661 sentinel type where appropriate

agronholm:

## Changes

This changes our current sentinels to use the [PEP 661](https://peps.python.org/pep-0661/)  `sentinel` type instead of an anonymous `object`.

The only place where this could not be done is the `typed_attribute()` function, as it is typed to return `Any`.

This PR needs a newer mypy for type checks to pass, as it doesn't yet recognize `typing_extensions.sentinel`.

## Checklist

features

these instructions.

### Updating the changelog

```
```


tapetersen:
I know I saw  at least one place a 1 member enum was used instead just for the debuggability reason. I'll see if I can find it again.

tapetersen:
Found it (and it seems like the only one).

https://github.com/agronholm/anyio/blob/b1f6e40220a8c079d36e6bc661e3a9e08db75d57/src/anyio/lowlevel.py#L102-L103

tapetersen:
Ok it uses it to have typing support in the code as well and until that is implemented it's probably not worth trying to change it.

I tried replacing it but mypy hasn't implemented it and with the flag to enable it in pyright they don't support the final semantics it seems yet (Sentinel vs., sentinel)

agronholm:
> Ok it uses it to have typing support in the code as well and until that is implemented it's probably not worth trying to change it.
> 
> I tried replacing it but mypy hasn't implemented it and with the flag to enable it in pyright they don't support the final semantics it seems yet (Sentinel vs., sentinel)

Tracking issue [here](https://github.com/python/mypy/pull/21647).

agronholm:
I think we could merge this as-is, and follow up with more when the tooling improves.

EmmanuelNiyonshuti (approved):
Far as i can tell, this is good aside from waiting on mypy to ship the sentinel support. we can get the actual sentinel type in too (instead of object), once it does!

agronholm:
> Far as i can tell, this is good aside from waiting on mypy to ship the sentinel support. we can get the actual sentinel type in too (instead of object), once it does!

I expect us to switch to Zuban soon which already supports sentinels properly. As for the second part of your comment, I'm not sure what you're referring to.

EmmanuelNiyonshuti:
> > Far as i can tell, this is good aside from waiting on mypy to ship the sentinel support. we can get the actual sentinel type in too (instead of object), once it does!
> 
> I expect us to switch to Zuban soon which already supports sentinels properly. As for the second part of your comment, I'm not sure what you're referring to.

I was trying to mean we'd follow up with the actual sentinel type:
```diff
diff --git a/src/anyio/_backends/_trio.py b/src/anyio/_backends/_trio.py
index 2e67786..b063d09 100644
--- a/src/anyio/_backends/_trio.py
+++ b/src/anyio/_backends/_trio.py
@@ -203,7 +203,7 @@ empty_start_value = sentinel("empty_start_value")
 
 
 class _TrioTaskStatus(abc.TaskStatus[T_contra], Generic[T_contra]):
-    early_start_value: T_contra | object = empty_start_value
+    early_start_value: T_contra | empty_start_value = empty_start_value
     real_task_status: trio.TaskStatus[T_contra | None] | None = None
 
     def started(self, value: T_contra | None = None) -> None:
```
Basically the follow-up to what I think you've already addressed that this PR needs a newer mypy for type checks to pass.

agronholm:
Thanks for the review!

PR #1198, opened 2026-06-30
Fix start inconsistencies between trio and asyncio

### LINKED ISSUE #1197: Inconsistencies between Trio and asyncio when target TaskGroup cancelled before a task calls started()
### Things to check first

### AnyIO version

4.14.1

### Python version

3.14

### What happened?

Trio and asyncio backends behave differently if a TaskGroup is cancelled before a Task spawned with `TaskGroup.start()` has called `task_status.started()`

```python
async def test_start_not_cancelled_before_started_by_tg() -> None:
    # A task started with start() shouldn't get cancelled by the TaskGroup until
    # it has called started
    #
    # * The started task shouldn't get a CancelledError until the first
    #   checkpoint after the started() call.
    #
    # * Any value passed to started should be available and correctly passed
    #   back to the caller of start.
    #
    #  * The CancelledError shouldn't leak out of the start() call to the calling
    #   task.

    # To manage the sibling tasks needed for the test
    th: TaskHandle[str] | None = None
        # Task holding the task_group we're trying to start a task in and will cancel.
        async def group_task(
            *, task_status: TaskStatus[TaskGroup] = anyio.TASK_STATUS_IGNORED
        ) -> None:
                task_status.started(tg)
                await anyio.sleep_forever()

        inner_tg: TaskGroup = await tg.start(group_task)
        ev = anyio.Event()

        # Task outside the inner_tg that tries to start a task with status_reporting
        async def outside_task() -> str:

            async def just_wait(
                *, task_status: TaskStatus[str] = anyio.TASK_STATUS_IGNORED
            ) -> None:
                await ev.wait()
                await checkpoint()  # Should not get cancelled here
                task_status.started("started")
                await checkpoint()  # Should be cancelled here
                pytest.fail("Should've been cancelled before this")

            return await inner_tg.start(just_wait)

        th = tg.start_soon(outside_task)
        inner_tg.cancel()
        ev.set()

    assert th is not None
    await th.wait()
    assert th.return_value == "started"
```

tapetersen:
## Changes

Fixes #1197 

~~For now just adds a failing test-case~~

Fixes the issue by doing full re-parenting of the `CancelScope` when `TaskStatus.started()` is called as Trio does with nurseries.

## Checklist

features


docs/versionhistory.rst
```diff
@@ -13,6 +13,9 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
   identity (``value is math.inf``), so only the exact ``math.inf`` singleton was accepted,
   while every backend setter (using ``math.isinf()``) accepts any positive infinity
   (`#1189 <https://github.com/agronholm/anyio/pull/1189>`_; PR by @greymoth-jp).
+- Fixed Inconsistencies between Trio and asyncio when target ``TaskGroup`` is
+  cancelled before a task created with ``.start()`` calls ``TaskStatus.started()``
```
agronholm: This really ought to specify what inconsistencies it fixes.

tapetersen: Sure I was unsure about how much to duplicate here compared to the issue. I'll add some clarifying.

agronholm:
I'll review this next, but I'm not sure when I'll have the time.

src/anyio/_backends/_asyncio.py
```diff
@@ -723,22 +743,34 @@ def __init__(self, parent_id: int | None, cancel_scope: CancelScope | None):
...
+        parent_id: int,
+        target_scope: CancelScope,
+        spawn_scope: CancelScope,
+    ):
         self._future = future
         self._parent_id = parent_id
+        self._target_scope = target_scope
+        # The spawned task's own cancel scope (also held by its TaskHandle). Until
+        # started() is called started()
+        # reparents it into the target task group's cancel scope.
```
agronholm: This comment makes little sense. It should be trimmed/reworded.

tests/test_taskgroups.py
```diff
@@ -1853,6 +1853,57 @@ async def wait_cancel() -> None:
...
+    # it has called started
+    #
+    # * The started task shouldn't get a CancelledError until the first
+    #   checkpoint after the started() call.
+    #
+    # * Any value passed to started should be available and correctly passed
+    #   back to the caller of start.
+    #
+    #  * The CancelledError shouldn't leak out of the start() call to the calling
+    #   task.
```
agronholm:
```suggestion
    # * The CancelledError shouldn't leak out of the start() call to the calling
    #   task.
```

src/anyio/_backends/_asyncio.py
```diff
@@ -879,12 +913,27 @@ def task_done(_task: asyncio.Task) -> None:
...
+        if task_status_future is not None:
             parent_id = id(current_task())
+            # Trio semantics: until started() is called, a task spawned via start()
+            # belongs to the *caller's* cancel scope, not the target group's, so
+            # cancelling the group does not cancel a task that hasn't reported
+            # startup yet. started() reparents it into self.cancel_scope.
+            #
+            # The caller may be an unmanaged task (no _task_states entry) or a task
+            # without an active cancel scope; in either case fall back to the group's
+            # own scope.
```
agronholm: This smells like an AI-generated comment. If you feel this has to stay in, I'd like it to be trimmed down.

agronholm (changes requested):
Just a few nits about the comments, but overall looks good.

src/anyio/_backends/_asyncio.py
tapetersen: It's indeed generated. I do think it's worth mentioning why it's needed but I'll rewrite it.

agronholm: Feel free to also remove unwarranted comments.

agronholm (approved):
I massaged the comments myself. The long one I moved to the start() method which is where I might go looking for it, and made the language a bit clearer.

Thanks for this fix, it's been a sore spot for a while!

PR #1199, opened 2026-07-01
Add rsloop as backend for asyncio tests.

tapetersen:
## Changes

`rsloop` as an implementation of asyncio's `AbstractEveltLoop` has reached full compatibility required by AnyIO's tests.
This PR adds it as an alternative backend to test against along `uvloop`

## Checklist

features

### DIFF
--- a/src/anyio/abc/_eventloop.py
+++ b/src/anyio/abc/_eventloop.py
@@ -8,6 +8,7 @@ from collections.abc import (
     Awaitable,
     Callable,
     Coroutine,
+    Iterable,
     Mapping,
     Sequence,
 )
@@ -242,14 +243,16 @@ class AsyncBackend(metaclass=ABCMeta):
         stdin: int | IO[Any] | None,
         stdout: int | IO[Any] | None,
         stderr: int | IO[Any] | None,
-        # These below may be sent with explicit default arguments and so must accept
-        # these literally
         cwd: StrOrBytesPath | None = None,
         env: Mapping[str, str] | None = None,
         startupinfo: Any = None,
         creationflags: int = 0,
         start_new_session: bool = False,
         pass_fds: Sequence[int] = (),
+        user: str | int | None = None,
+        group: str | int | None = None,
+        extra_groups: Iterable[str | int] | None = None,
+        umask: int = -1,
         **kwargs: Any,
     ) -> Process:
         pass
--- a/src/anyio/abc/_eventloop.py
+++ b/src/anyio/abc/_eventloop.py
@@ -3,7 +3,14 @@ from __future__ import annotations
 import math
 import sys
 from abc import ABCMeta, abstractmethod
-from collections.abc import AsyncIterator, Awaitable, Callable, Coroutine, Sequence
+from collections.abc import (
+    AsyncIterator,
+    Awaitable,
+    Callable,
+    Coroutine,
+    Mapping,
+    Sequence,
+)
 from contextlib import AbstractContextManager
 from os import PathLike
 from signal import Signals
@@ -235,6 +242,14 @@ class AsyncBackend(metaclass=ABCMeta):
         stdin: int | IO[Any] | None,
         stdout: int | IO[Any] | None,
         stderr: int | IO[Any] | None,
+        # These below may be sent with explicit default arguments and so must accept
+        # these literally
+        cwd: StrOrBytesPath | None = None,
+        env: Mapping[str, str] | None = None,
+        startupinfo: Any = None,
+        creationflags: int = 0,
+        start_new_session: bool = False,
+        pass_fds: Sequence[int] = (),
         **kwargs: Any,
     ) -> Process:
         pass
--- a/tests/test_sockets.py
+++ b/tests/test_sockets.py
@@ -21,7 +21,7 @@ from pathlib import Path
 from socket import AddressFamily
 from ssl import SSLContext, SSLError
 from threading import Thread
-from typing import TYPE_CHECKING, Any, Literal, NoReturn, Protocol, TypeVar, cast
+from typing import TYPE_CHECKING, Any, Literal, NoReturn, Protocol, cast
 from unittest import mock
 from unittest.mock import MagicMock, patch
 
@@ -209,13 +209,6 @@ def sock_or_fd_factory(
     return factory
 
 
-_T = TypeVar("_T")
-
-
-def _identity(v: _T) -> _T:
-    return v
-
-
 def fill_socket(sock: socket.socket) -> None:
     try:
         while True:
@@ -224,19 +217,6 @@ def fill_socket(sock: socket.socket) -> None:
         pass
 
 
-#  _ProactorBasePipeTransport.abort() after _ProactorBasePipeTransport.close()
-# does not cancel writes: https://bugs.python.org/issue44428
-_ignore_win32_resource_warnings = (
-    pytest.mark.filterwarnings(
-        "ignore:unclosed <socket.socket:ResourceWarning",
-        "ignore:unclosed transport <_ProactorSocketTransport closing:ResourceWarning",
-    )
-    if sys.version_info < (3, 10) and sys.platform == "win32"
-    else _identity
-)
-
-
-@_ignore_win32_resource_warnings
 class TestTCPStream:
     @pytest.fixture
     def server_sock(self, family: AnyIPAddressFamily) -> Iterator[socket.socket]:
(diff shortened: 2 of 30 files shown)


Vizonex (approved):
@tapetersen I love this because it's another low level event loop that has windows support. besides just winloop and the selector system pywepoll, (both of which I wrote) I'm glad to see libraries with less neglect for windows support (excluding rloop & uvloop) this had always been a reoccurring pet-peve of mine but I'm glad to see something new that could provide a positive impact for windows users. I approve this pull request due to that fact and I hope rsloop receives positive attention.

tapetersen:
@Vizonex I mostly completed what was needed to get the test-suite to pass in rsloop for curiousity. I havn't used it for something real yet and don't have any idea for how ready it is. If nothing else it would be good to see it tries in more projects if you have any. 

There is no need for this PR to use it just pass loop_factory=rsloop.new_event_loop to anyio.run
```
anyio.run(main, backend="asyncio", backend_options={"loop_factory": rsloop.new_event_loop})
```

Vizonex:
> @Vizonex I mostly completed what was needed to get the test-suite to pass in rsloop for curiousity. I havn't used it for something real yet and don't have any idea for how ready it is. If nothing else it would be good to see it tries in more projects if you have any.
> 
> There is no need for this PR to use it just pass loop_factory=rsloop.new_event_loop to anyio.run
> 
> ```
> anyio.run(main, backend="asyncio", backend_options={"loop_factory": rsloop.new_event_loop})
> ```

@tapetersen  I already added it to cyares https://github.com/Vizonex/cyares/pull/91 It might be added to more places if I can think of anywhere to have it placed.

agronholm:
Can you figure out if the timeouts happen for a reason or if the test suite needs more time to the increased number of tests?

tapetersen:
I have gotten hold of a remote windows VM and can confirm they look like real timeouts, not just slow tests. I'll see if I can figure out why.

agronholm:
I have a real Windows machine here so I can also do this when time allows.

tapetersen:
I tried adding a `fail_after` around the `connect_tcp` call in the `test_connection_refused` case and it fails and moves on. after that the rest of the socket tests passes. So something is preventing rsloop from just failing the connection attempt.

agronholm:
Confirmed locally. `test_connection_refused` hangs with rsloop.

tapetersen:
Yeah, it's probably a rsloop bug of not failing when connecting to a closed socket on windows. I'll see if I can distill it and add an issue there instead.

Skipping these let's the whole suite through otherwise though. Should we just mark them as xfail or wait until it's fixed upstream (which I'll try to do anyway)?

tapetersen:
Upstream issue here for the windows issue

https://github.com/RustedBytes/rsloop/issues/60

For the macos timeout I don't have a better way to test. I may report it anyway.

agronholm:
> Skipping these let's the whole suite through otherwise though. Should we just mark them as xfail or wait until it's fixed upstream (which I'll try to do anyway)?

I don't have a strong opinion on this, but this seems like a fairly serious bug in rsloop. I'm leaning towards waiting until the test suite passes cleanly. This is what we did with `winloop` too.

egorsmkv:
I have fixed that bug on Windows in 0.1.22 version. can we re-run CI checks?

tapetersen:
I think I managed to re-run them. There is still some error on macOS as well that I don't have a a good way to debug. It may be ok to exclude that though as long as it works on other platforms.

agronholm:
Given that both macOS jobs failed with a `BusError`, it sounds like there is genuinely something still wrong.

egorsmkv:
I will try to figure out what's wrong with those CI runs today.

egorsmkv:
@tapetersen can you re-run tests again here, please?

agronholm:
Not sure what you expected to happen, but even more test jobs failed.

egorsmkv:
@agronholm can you re-run again? in my local test - it works now with v0.1.25

agronholm:
Something is still off, as some tests fail due to the lack of an exception where one was expected. I confirmed that 0.1.25 is installed before the test run.

egorsmkv:
> Something is still off, as some tests fail due to the lack of an exception where one was expected. I confirmed that 0.1.25 is installed before the test run.

Thank you. I will work on the remaining three failed tests. Before about five tests failed.

egorsmkv:
@agronholm can you bump the version to 0.1.26, please?

tapetersen:
@egorsmkv I can so it but if you want to continue testing it may be better to just create your own PR that will be triggered when you push. I created this one to iron out the Unix bugs and see how far I came to complete. Since I'm a collaborator now I can even push to your branch if you check that flag so it may be easier to just handle that way.

egorsmkv:
@tapetersen now all checks are passed so there is no need or it

agronholm:
@tapetersen is this good for merging? I see it's still marked as a draft.

tapetersen:
@agronholm There is still the one refcycle leak on happy-eyeballs I marked as skip since it wasn't fixed and I didn't know if we eanted to take on extra test-burden when I created it but if the test-integration looks ok for you I don't mind merging it.

src/anyio/abc/_eventloop.py
```diff
@@ -235,6 +242,14 @@ async def open_process(
...
         stdout: int | IO[Any] | None,
         stderr: int | IO[Any] | None,
+        # These below may be sent with explicit default arguments and so must accept
+        # these literally
+        cwd: StrOrBytesPath | None = None,
+        env: Mapping[str, str] | None = None,
+        startupinfo: Any = None,
+        creationflags: int = 0,
+        start_new_session: bool = False,
+        pass_fds: Sequence[int] = (),
```
agronholm: Hm, can you help me understand this change? It doesn't seem directly related to the subject of the PR.

tapetersen: While trying to initially get rsloop to pass one of the failures was that it didn't accept these arguments (at least `cwd` and `env`) explicitly as None while it handled them correctly if omitted (treating them as if None was passed).

Since `anyio` actually passes `None` explicitly see https://github.com/agronholm/anyio/blob/master/src/anyio/_core/_subprocesses.py#L184 a backend is required to accept that interface (as `**kwargs: Any` doesn't put any requirements on them).

The bug is since fixed in `rsloop` so it's not required but was related. I like making these kind of things explicit in the types if possible but can also just remove it or put it in a separate MR if you prefer.

tapetersen: But as long as we keep forwarding `**kwargs:Any` it won't really catch that much of anything with typing so I can very well see the case for removing it.

tapetersen: So basically rsloop would be fine with `open_process(...)` and do the correct default thing for `cwd` and `env` if ommitted but not accept the explicit defaults as in `open_process(..., cwd=None, env=None)` as it didn't like the `None` type.

agronholm:
I don't think this needs a changelog entry as it's not a user-facing change. Thoughts?

tapetersen:
@agronholm I agree and remove it. How do you want to do with the extra argument annotations for `open_process`? 

They are not necessary and are a matter of taste but I tend to prefer them.

Also are you ok with the skipped ref-leak test for happy eyeballs or should I unskip it and hold the PR until it's fixed in `rsloop`?

agronholm:
> Also are you ok with the skipped ref-leak test for happy eyeballs or should I unskip it and hold the PR until it's fixed in `rsloop`?

@egorsmkv What's your take on this? A refleak isn't that serious a problem, but if you are planning to fix that soonish, then I'd rather wait for that.

egorsmkv:
> > Also are you ok with the skipped ref-leak test for happy eyeballs or should I unskip it and hold the PR until it's fixed in `rsloop`?
> 
> @egorsmkv What's your take on this? A refleak isn't that serious a problem, but if you are planning to fix that soonish, then I'd rather wait for that.

I will try to fix that tomorrow

tapetersen:
@egorsmkv I've removed the skip in the code so it should start failing on CI again.

egorsmkv:
okay, let's test with v0.1.27

tapetersen:
@egorsmkv I've retriggered it. I think it should pick up the latest if published since we don't pin any dependencies here but of not I'll push a change with a higher bound.

egorsmkv:
@tapetersen can you update to v0.1.30? i tested it in #1238 and it now passed tests

tapetersen:
@egorsmkv Done

egorsmkv:
@agronholm

tests/test_sockets.py
```diff
@@ -409,7 +409,10 @@ async def test_connect_tcp_with_local_port(
         ),
     )
     async def test_happy_eyeballs_refcycles(
-        self, free_tcp_port: int, anyio_backend_name: str
+        self,
+        free_tcp_port: int,
+        anyio_backend_name: str,
+        event_loop_implementation_name: str,
```
agronholm: This seems like an artifact of some code change that was later removed.
```suggestion
        self, free_tcp_port: int, anyio_backend_name: str
```

agronholm (approved):
Thanks!

PR #1200, opened 2026-07-03
Fixed Path.with_stem() not validating an empty stem

Sanjays2402:
## Changes

Fixes #. 

`anyio.Path.with_stem()` reimplements the operation as
`with_name(stem + self._path.suffix)` instead of delegating to pathlib's own
`with_stem()`. For a path with a **non-empty suffix**, an **empty** stem
therefore produces a silently wrong path instead of raising `ValueError` the
way `pathlib.PurePath.with_stem` does:

```python
import anyio, pathlib

anyio.Path("foo.txt").with_stem("")     # -> anyio.Path(".txt")   (silently wrong)
pathlib.Path("foo.txt").with_stem("")   # -> ValueError: PosixPath('foo.txt') has a non-empty suffix
```

CPython's `with_stem` explicitly rejects this case:

```python
def with_stem(self, stem):
    suffix = self.suffix
    if not suffix:
        return self.with_name(stem)
    elif not stem:
        # If the suffix is non-empty, we can't make the stem empty.
        raise ValueError(f"{self!r} has a non-empty suffix")
    else:
        return self.with_name(stem + suffix)
```

anyio's reimplementation skips that guard, so `with_name("" + ".txt")` quietly
returns `Path(".txt")` (and `Path("a/b.py").with_stem("")` returns `Path("a/.py")`).

### Fix

The sibling wrappers `with_name()` and `with_suffix()` already delegate to
`self._path`; `with_stem()` now does the same:

```python
def with_stem(self, stem: str) -> Self:
    return type(self)(self._path.with_stem(stem), limiter=self._limiter)
```

This restores parity with pathlib for **every** input: the empty-stem-with-suffix
case now raises `ValueError`, while all previously-working cases
(`with_stem("bar")`, no-suffix paths, dotfiles, multi-suffix paths, etc.) are
byte-identical to before. Verified against `pathlib.Path` across a matrix of
names x stems.

### Before / after

| input | before (buggy) | after (== pathlib) |
|-------|----------------|--------------------|
| `Path("foo.txt").with_stem("")` | `Path(".txt")` | `ValueError: ... has a non-empty suffix` |
| `Path("a/b.py").with_stem("")` | `Path("a/.py")` | `ValueError: ... has a non-empty suffix` |
| `Path("foo.txt").with_stem("bar")` | `Path("bar.txt")` | `Path("bar.txt")` (unchanged) |
| `Path("foo").with_stem("")` | `ValueError: Invalid name ''` | `ValueError: Invalid name ''` (unchanged) |

## Checklist

features

these instructions.

### Updating the changelog

```
```

---

Test proof (fails without the patch, passes with it):

```
$ git stash push -- src/anyio/_core/_fileio.py   # remove only the fix
$ pytest tests/test_fileio.py::TestPath::test_with_stem_empty_stem_with_suffix
    Failed: DID NOT RAISE ValueError                                   # 1 failed
$ git stash pop                                   # restore the fix
$ pytest tests/test_fileio.py::TestPath::test_with_stem_empty_stem_with_suffix
    1 passed
```

Full `tests/test_fileio.py` suite: 207 passed. `ruff check`, `ruff format --check`
and `mypy` are clean on the changed files. Diff is +13 / -3.

### DIFF
--- a/src/anyio/_core/_fileio.py
+++ b/src/anyio/_core/_fileio.py
@@ -925,7 +925,8 @@ class Path:
         return type(self)(self._path.with_name(name), limiter=self._limiter)
 
     if sys.version_info < (3, 13):
-        # Backport validation logic from 3.13
+        # Backport pathlib's Python>=3.13 behavior for empty stems on paths with non-empty suffixes.
+        # See: https://github.com/python/cpython/pull/114612
         def with_stem(self, stem: str) -> Self:
             suffix = self._path.suffix
             if not suffix:
--- a/src/anyio/_core/_fileio.py
+++ b/src/anyio/_core/_fileio.py
@@ -925,9 +925,7 @@ class Path:
         return type(self)(self._path.with_name(name), limiter=self._limiter)
 
     def with_stem(self, stem: str) -> Self:
-        return type(self)(
-            self._path.with_name(stem + self._path.suffix), limiter=self._limiter
-        )
+        return type(self)(self._path.with_stem(stem), limiter=self._limiter)
 
     def with_suffix(self, suffix: str) -> Self:
         return type(self)(self._path.with_suffix(suffix), limiter=self._limiter)
--- a/src/anyio/_core/_fileio.py
+++ b/src/anyio/_core/_fileio.py
@@ -924,9 +924,9 @@ class Path:
     def with_name(self, name: str) -> Self:
         return type(self)(self._path.with_name(name), limiter=self._limiter)
 
-    def with_stem(self, stem: str) -> Self:
+    if sys.version_info < (3, 13):
         # Backport validation logic from 3.13
-        if sys.version_info < (3, 13):
+        def with_stem(self, stem: str) -> Self:
             suffix = self._path.suffix
             if not suffix:
                 return self.with_name(stem)
@@ -935,8 +935,10 @@ class Path:
                 raise ValueError(f"{self!r} has a non-empty suffix")
             else:
                 return self.with_name(stem + suffix)
+    else:
 
-        return type(self)(self._path.with_stem(stem), limiter=self._limiter)
+        def with_stem(self, stem: str) -> Self:
+            return type(self)(self._path.with_stem(stem), limiter=self._limiter)
 
     def with_suffix(self, suffix: str) -> Self:
         return type(self)(self._path.with_suffix(suffix), limiter=self._limiter)
--- a/src/anyio/_core/_fileio.py
+++ b/src/anyio/_core/_fileio.py
@@ -925,6 +925,17 @@ class Path:
         return type(self)(self._path.with_name(name), limiter=self._limiter)
 
     def with_stem(self, stem: str) -> Self:
+        # Backport validation logic from 3.13
+        if sys.version_info < (3, 13):
+            suffix = self._path.suffix
+            if not suffix:
+                return self.with_name(stem)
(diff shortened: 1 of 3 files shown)


tapetersen:
@Sanjays2402 It looks like the test is failing on python versions 3.10-3.12 It looks reasonable otherwise. The tox setup should allow you to debug these failures directly.
 (the windows failure is probably unrelated to your patches and can be ignored)

tapetersen:
@agronholm I haven't checked why the tests are failing but assume earlier versions either don't do this check or fails the regexp. If that is fixed there's no reason not to make this change right?

agronholm:
The failure is caused by the test code not raising the expected exception which is weird because by all rights, it should be triggered. I'll investigate locally.

tapetersen:
@agronholm I was already on it and indeed the normal pathlib doesn't throw the warning on 3.10 at least.
The question is really then if we should normalize or just document that it follows the stdlib function.

tapetersen:
I changed the test like this to just match whatever stdlib does https://github.com/tapetersen/anyio/commit/24cd763c04987306fe359c68fe14c42662c02975

Not sure how to best apply it here.

agronholm:
So I think the AnyIO code here predates the new validation which was introduced in Python 3.13. At any rate, this is not a serious issue and can be deferred to the next release (especially given that there are security fixes in the pipeline). I'd be favorable to making sure the `ValueError` does get raised in a uniform manner.

src/anyio/_core/_fileio.py
```diff
@@ -924,10 +924,21 @@ def get_next_value() -> tuple[pathlib.Path, list[str], list[str]] | None:
     def with_name(self, name: str) -> Self:
         return type(self)(self._path.with_name(name), limiter=self._limiter)
 
-    def with_stem(self, stem: str) -> Self:
-        return type(self)(
-            self._path.with_name(stem + self._path.suffix), limiter=self._limiter
-        )
+    if sys.version_info < (3, 13):
+        # Backport validation logic from 3.13
```
EmmanuelNiyonshuti:
```suggestion
        # Backport pathlib's Python>=3.13 behavior for empty stems on paths with non-empty suffixes.
        # See: https://github.com/python/cpython/pull/114612
```

EmmanuelNiyonshuti (commented):
A rewording of the implementation comment.

agronholm:
@tapetersen How do you feel about the version gating? Does it add anything useful to have a separate py3.13+ code path?

tapetersen:
My real preference is to not bother with it at all and just always delegate to pathlib regardless. If it comes up at all I think more people would be surprised by the divergence there than between python versions.

The version gating is not really for performance but more of a verbose comment to remove it when support for 3.12 is dropped, (avoiding Chesterton's fence).

The only thing I have a stronger opinion on if anything it's that last point, that if we backport it to be able to know when it can be dropped.

tapetersen:
@agronholm I don't feel too strongly about this either way but I do feel it's better to rely on just forwarding to stdlib's version going forward than status quo.

I don't mind the backport or if it's conditional as long as it's marked to be safely removed when 3.12 is no longer supported.

So the questions are really:

Do we want the backport at all ?
(I prefer not)

Do we want it only on < 3.13 ?
(I prefer yes, and to view our API here mirroring stdlib)

agronholm (approved):

PR #1206, opened 2026-07-07
Fix flaky test_tcp_listener_same_port using a hardcoded port

### LINKED ISSUE #1205: Flaky test_tcp_listener_same_port: one parametrization still uses hardcoded port 54321
### Things to check first

- [x] I have searched the existing issues and didn't find my issue already reported there

### Summary

`tests/test_sockets.py::TestTCPListener::test_tcp_listener_same_port` fails intermittently in CI because one parametrization still binds a **hardcoded port (`54321`)**. When that port is already in use on the runner (e.g. another parallel job/tox env), the bind fails and the test errors out. This has been showing up as red `test (…, 3.14, [trio])` jobs on `master` and on open PRs that don't touch sockets at all.

### Root cause

d0482f9 ("Fixed test_tcp_listener_same_port preventing a parallel tox run") converted the two `("localhost", …, 54321)` cases to the `free_tcp_port_factory` fixture, but the adjacent `(None, AddressFamily.AF_UNSPEC, …, 54321)` case just above them was missed and still uses the hardcoded port:

```python
(
    None,
    AddressFamily.AF_UNSPEC,
    1 if socket.has_dualstack_ipv6() else 2,
    54321,   # <- still hardcoded
),
```

`FreePortFactory` already handles `family=None` (it finds a port free on both IPv4 and IPv6), so this case can use the factory just like the others — there's no reason it needs a fixed port.

### Reproduction

On a host where port `54321` is occupied, or when running the suite in parallel (the scenario d0482f9 was addressing), the `…-None-…-54321` parametrizations fail with a bind error. The failing job on a recent PR showed exactly these four:

```
test_tcp_listener_same_port[asyncio-None-0-1-54321] FAILED
test_tcp_listener_same_port[asyncio+eager-None-0-1-54321] FAILED
test_tcp_listener_same_port[asyncio+uvloop-None-0-1-54321] FAILED
test_tcp_listener_same_port[trio-None-0-1-54321] FAILED
```

### Fix

Replace the hardcoded `54321` with `None` so the test pulls a guaranteed-free port from `free_tcp_port_factory` (same pattern as the sibling cases). PR incoming.

### Note (possibly separate)

There's a second, environment-dependent failure mode in the same test: the `("localhost", AF_UNSPEC, 2, …)` cases assume `localhost` resolves to **both** `127.0.0.1` and `::1`. On hosts where `/etc/hosts` maps only `127.0.0.1` to `localhost` (and `::1` to `ip6-localhost`), only one listener is created and the test fails `assert 1 == 2`. GitHub's ubuntu runners normally resolve `localhost` dual-stack so this passes there, but it can surface in other environments. Flagging it here for awareness; happy to address separately if you'd like the test made resolution-agnostic.

### AnyIO version

master (85b8b6e)

### Python version

3.14 (and others)

---

### Additional examples / detail

**1. The full parametrization block** (`tests/test_sockets.py`, `test_tcp_listener_same_port`), annotated to show which cases already use the fixture vs. the one that doesn't:

```python
@pytest.mark.parametrize(
    "local_host,family,expected_listeners,local_port",
    [
        (None, AddressFamily.AF_UNSPEC, 1 if socket.has_dualstack_ipv6() else 2, 0),   # port 0 -> OS-assigned, fine
        ("localhost", AddressFamily.AF_UNSPEC, 2, 0),                                   # fine
        ("localhost", AddressFamily.AF_INET, 1, 0),                                     # fine
        ("::", AddressFamily.AF_UNSPEC, 1, 0),                                          # fine
        ("127.0.0.1", AddressFamily.AF_INET, 1, 0),                                     # fine
        ("::1", AddressFamily.AF_INET6, 1, 0),                                          # fine
        (
            None,
            AddressFamily.AF_UNSPEC,
            1 if socket.has_dualstack_ipv6() else 2,
            54321,                                                                      # <- FLAKY: hardcoded, not from the factory
        ),
        ("localhost", AddressFamily.AF_UNSPEC, 2, None),                                # None -> free_tcp_port_factory() (the fix pattern)
        ("localhost", AddressFamily.AF_INET, 1, None),                                  # None -> free_tcp_port_factory()
    ],
)
```

The test body already special-cases `None`:

```python
if local_port is None:
    local_port = free_tcp_port_factory()
```

so switching `54321` -> `None` routes the case straight through the existing free-port path — no other change needed. `FreePortFactory.__call__` handles `family=None` by finding a port free on **both** `AF_INET` and `AF_INET6` (see `src/anyio/pytest_plugin.py`), which is exactly what this `AF_UNSPEC` case wants.

**2. Example failure output** — when `54321` is already taken, the bind raises before the assertions are ever reached:

```
tests/test_sockets.py::TestTCPListener::test_tcp_listener_same_port[asyncio-None-0-1-54321]
...
    async with await create_tcp_listener(
        local_host=local_host, family=family, local_port=local_port
    ) as multi:
...
OSError: [Errno 98] error while attempting to bind on address ('0.0.0.0', 54321): address already in use
```

(Errno 98 = `EADDRINUSE` on Linux; on the other backends the same collision surfaces as the equivalent bind error.)

**3. Minimal local repro** — occupy `54321`, then run just the hardcoded parametrizations:

```bash
# Terminal A: hold port 54321 so the bind collides
python -c "import socket,time; s=socket.socket(); s.setsockopt(socket.SOL_SOCKET,socket.SO_REUSEADDR,1); s.bind(('0.0.0.0',54321)); s.listen(); print('holding 54321'); time.sleep(600)"

# Terminal B: only the 54321 cases -> they FAIL; every other case passes
pytest "tests/test_sockets.py::TestTCPListener::test_tcp_listener_same_port" -k 54321
```

With the fix in place, the same command passes because the case now pulls a guaranteed-free port instead of `54321`.

**4. Concrete example of the separate `localhost` dual-stack mode** — on a host whose `/etc/hosts` only maps IPv4 to `localhost`:

```
$ getent hosts localhost
127.0.0.1       localhost

$ grep -E 'localhost|ip6-' /etc/hosts
127.0.0.1   localhost
::1         ip6-localhost ip6-loopback
```

Here `localhost` resolves IPv4-only, so `create_tcp_listener(local_host="localhost", family=AF_UNSPEC)` yields **1** listener and the `expected_listeners == 2` cases fail with `assert 1 == 2`. On GitHub's ubuntu runners `getent hosts localhost` returns both `127.0.0.1` and `::1`, so those cases pass there — which is why this mode is environment-dependent and (as noted above) arguably a separate fix.

daphne-barretto:
## Changes

Fixes #1205.

One parametrization of `test_tcp_listener_same_port` still binds the hardcoded port `54321`, so the test fails intermittently in CI (and blocks parallel `tox` runs) whenever that port is already in use on the runner. This has been surfacing as red `3.14` jobs on `master` and on unrelated PRs.

d0482f9 converted the sibling `("localhost", …, 54321)` cases to the `free_tcp_port_factory` fixture but missed the adjacent `(None, AF_UNSPEC, …, 54321)` case. This change points that case at the factory too — it already handles `family=None` by finding a port free on both IPv4 and IPv6, so no fixed port is needed.

## Checklist

This is a trivial, test-only change (a single hardcoded port swapped for the existing free-port fixture), so per the template's note the user-facing checklist below doesn't strictly apply. Recording the status honestly anyway:

- [ ] You've added tests (in `tests/`) which would fail without your patch — N/A: this fixes an existing flaky test rather than adding coverage.
- [ ] You've updated the documentation (in `docs/`) — N/A: no behavior or API change.
- [ ] You've added a new changelog entry (in `docs/versionhistory.rst`) — N/A: test-only, following the same no-changelog convention as the sibling fix d0482f9.

---

> [!NOTE]
> #1205 also documents a second, environment-dependent failure mode in the same test (the `("localhost", AF_UNSPEC, 2, …)` cases assume `localhost` resolves dual-stack). That's left untouched here since it passes on GitHub's runners and is arguably a separate change — happy to follow up if you'd like the test made resolution-agnostic.

### DIFF
--- a/tests/test_sockets.py
+++ b/tests/test_sockets.py
@@ -1015,7 +1015,7 @@ class TestTCPListener:
                 None,
                 AddressFamily.AF_UNSPEC,
                 1 if socket.has_dualstack_ipv6() else 2,
-                54321,
+                None,
             ),
             ("localhost", AddressFamily.AF_UNSPEC, 2, None),
             ("localhost", AddressFamily.AF_INET, 1, None),


tapetersen:
@daphne-barretto please follow the directions in the PR template. Keep the check boxes and fill them if needed.

agronholm (approved):

PR #1207, opened 2026-07-07
Fixed stderr writes in a worker subprocess causing a deadlock

agronholm:

## Changes

Fixes https://github.com/agronholm/anyio/security/advisories/GHSA-5p39-cfhj-2xmp. 

## Checklist

features

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/tests/test_to_process.py
+++ b/tests/test_to_process.py
@@ -63,8 +63,11 @@ def _flood_stderr() -> str:
 
     # A typical pipe buffer is 64 KiB; write well beyond that to ensure it would block
     # if stderr were still connected to the (undrained) parent pipe.
-    sys.stderr.write("x" * (1024 * 1024))
-    sys.stderr.flush()
+    payload = "x" * (1024 * 1024)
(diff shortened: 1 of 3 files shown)


tests/test_to_process.py
```diff
@@ -58,6 +58,27 @@ async def test_print() -> None:
     assert await to_process.run_sync(os.getpid) == worker_pid
 
 
+def _flood_stderr() -> str:
+    """Helper that writes enough data to stderr to fill an undrained pipe buffer."""
+
+    # A typical pipe buffer is 64 KiB; write well beyond that to ensure it would block
+    # if stderr were still connected to the (undrained) parent pipe.
+    sys.stderr.write("x" * (1024 * 1024))
```
tapetersen: Any point in having the regression test write to both stdout as well as stderr maybe?

tapetersen (approved):
Looks good to me.

tests/test_to_process.py
agronholm: Added.

PR #1211, opened 2026-07-08
Break ref-cycle between loop and root_task via RunVar

### LINKED ISSUE #1203: _run_vars memory leak

tapetersen:
## Changes

Fixes #1203 

Breaks the reference-cycle by unsetting the `_root_task` RunVar in a completed callback.
Since those run as sync callbacks we can't use the RunVar's  builtin reset mechanism as it
fails on `current_token()` since we're not in a running loop.

## Checklist

features

### DIFF
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -102,7 +102,7 @@ from ..abc import (
 )
 from ..abc._eventloop import StrOrBytesPath
 from ..abc._tasks import call_for_coroutine, get_callable_name
-from ..lowlevel import RunVar, _run_vars, current_token
+from ..lowlevel import RunVar, _run_vars
 from ..streams.memory import MemoryObjectReceiveStream, MemoryObjectSendStream
 
 if TYPE_CHECKING:
@@ -321,13 +321,10 @@ def find_root_task() -> asyncio.Task:
                     cb is _run_until_complete_cb
                     or getattr(cb, "__module__", None) == "uvloop.loop"
                 ):
-                    tok = current_token()
                     _root_task.set(task)
 
-                    def _unset(
-                        t: asyncio.Task[Any], native_tok: object = tok.native_token
-                    ) -> None:
-                        if vars := _run_vars.get(native_tok):
+                    def _unset(t: asyncio.Task[Any]) -> None:
+                        if vars := _run_vars.get(t.get_loop()):
                             vars.pop(_root_task, None)
 
                     # Register a callback to break the task -> loop -> _run_var[loop][_root_task] -> task cycle
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -102,7 +102,7 @@ from ..abc import (
 )
 from ..abc._eventloop import StrOrBytesPath
 from ..abc._tasks import call_for_coroutine, get_callable_name
-from ..lowlevel import RunVar
+from ..lowlevel import RunVar, _run_vars, current_token
 from ..streams.memory import MemoryObjectReceiveStream, MemoryObjectSendStream
 
 if TYPE_CHECKING:
@@ -305,7 +305,7 @@ T_contra = TypeVar("T_contra", contravariant=True)
 PosArgsT = TypeVarTuple("PosArgsT")
 P = ParamSpec("P")
 
-_root_task: RunVar[asyncio.Task | None] = RunVar("_root_task")
+_root_task: RunVar[asyncio.Task[Any] | None] = RunVar("_root_task")
 
 
 def find_root_task() -> asyncio.Task:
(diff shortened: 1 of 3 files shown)


tests/test_to_thread.py
```diff
@@ -386,3 +390,33 @@ def foo(_: Foo) -> None:
...
+        f.set_running_or_notify_cancel()
+        f.set_result(weakref.ref(asyncio.get_running_loop()))
+
+    f = Future[weakref.ref[object]]()
+    t = threading.Thread(
+        group=None, target=anyio.run, args=(main,), kwargs={"backend": "asyncio"}
+    )
+    t.start()
+    loop_ref = f.result(timeout=2.0)
+    t.join()
```
agronholm: What's the threading needed for? On `master`, running this without the thread yields the same exact result.

tapetersen: Probably not needed. It was added while debugging the test before I found the redundant _run_vars. I was mostly unsure about if creating a new Loop in the same thread as other tests with possibly a module scoped loop was interfering.

It may be that's only a problem if it would be actively running though.

tapetersen: Will check and remove when I have time.

tapetersen: Fixed with latest commit

src/anyio/_backends/_asyncio.py
```diff
@@ -316,13 +316,25 @@ def find_root_task() -> asyncio.Task:
...
     for task in all_tasks():
         if task._callbacks and not task.done():
-            callbacks = [cb for cb, context in task._callbacks]
-            for cb in callbacks:
+            for cb, context in task._callbacks:
                 if (
                     cb is _run_until_complete_cb
                     or getattr(cb, "__module__", None) == "uvloop.loop"
                 ):
+                    tok = current_token()
```
agronholm: Seems like a really roundabout way to call `asyncio.get_running_loop()`, given that's what you ended up using anyway? Why not simply get the running loop from `t.get_loop()` inside the callback?

src/anyio/_backends/_asyncio.py
```diff
@@ -316,13 +316,25 @@ def find_root_task() -> asyncio.Task:
...
                     _root_task.set(task)
+
+                    def _unset(
+                        t: asyncio.Task[Any], native_tok: object = tok.native_token
+                    ) -> None:
+                        if vars := _run_vars.get(native_tok):
+                            vars.pop(_root_task, None)
```
agronholm: Is there merit in keeping the rest of the run variables?

agronholm (commented):
The tests are fine now, but I have two comments/questions on the actual implementation.

src/anyio/_backends/_asyncio.py
tapetersen: Only to avoid assuming the type of native_token and the key in the WeakMap (and possibly a task implementation that doesn't provide the loop property).

It's internal code anyway so I can see the direct usage being clearer and change if you prefer.

src/anyio/_backends/_asyncio.py
tapetersen: You can call `run_until_complete` multiple times on a loop and it makes kind of sense to keep them in that case (although it could be argued they should be scoped to the "run" rather than the loop).

src/anyio/_backends/_asyncio.py
agronholm: Yeah, given the internal nature of this, I would prefer the `t.get_loop()` approach instead.

src/anyio/_backends/_asyncio.py
agronholm: Yeah, one `run_until_complete()` is a "run" in my books, but I'd like to double check we don't break anything by nixing all the run variables at the end of the root task.

tapetersen: If so just using the `RunVar`s at all should probably trigger the search for a root task that registers that clean-up. If just added here it would be dependent on if anything called `find_root_task` which isn't a given.

But I like that better anyway as cleanup of runvars should ideally be kept in that module anyway. I'll see if I can move it and if it breaks anything in the test-suite.

tapetersen: This feels a bit complicated and risky as we also right now support RunVars where we can't find a good root-task (by just tying it to the closest anyio-ancestor.

If nothing else to prevent more bot-duplicate PR's which this seems to attract it may be worth closing.

agronholm (approved):
Alright, LGTM!

PR #1212, opened 2026-07-09
Fix per-run event loop leak in _run_vars on asyncio

### LINKED ISSUE #1203: _run_vars memory leak

vidigoat:
## Changes

Fixes #1203.

On the **asyncio** backend, every `anyio.run()` call permanently leaks its
event loop, root task and result. Calling `anyio.run()` in a loop grows memory
without bound and `gc.collect()` does not reclaim it.

### Root cause

`find_root_task()` caches the root task in the `_root_task` run variable:

```python
_root_task.set(task)
```

Run variables live in `anyio.lowlevel._run_vars`, a `WeakKeyDictionary` keyed by
the event loop. The cached `Task` strongly references its loop (via
`task.get_loop()`), so the entry's **value** holds a strong reference back to its
own **weak key**. A `WeakKeyDictionary` entry whose value keeps its key alive can
never be evicted, so the loop, the root task and the task's result survive for
the lifetime of the process.

The Trio backend is unaffected, and `Semaphore`/`Lock`-style hand-offs are not
involved — this is purely the weak-key self-reference.

### Fix

`AsyncIOBackend.run()` now drops the loop's `_run_vars` entry in a `finally`
block once the run finishes (whether it returns or raises), breaking the cycle so
the loop and its result can be garbage-collected.

While here, this also removes an **unused, shadowing** module-level `_run_vars`
definition in `_asyncio.py`. It was dead code (never read or written anywhere)
and shadowed the real `anyio.lowlevel._run_vars`, making it easy to reference the
wrong dictionary.

### Reproduction (before the fix)

```python
import gc

async def main() -> bytes:
    return await anyio.to_thread.run_sync(lambda: b"x" * 50_000_000)

for i in range(6):
    anyio.run(main)
    gc.collect()
    print(i, "entries", len(anyio.lowlevel._run_vars))
# entries: 1, 2, 3, 4, 5, 6  (RSS climbs ~48 MB per iteration)
```

With the fix, `_run_vars` stays empty and RSS is flat.

## Checklist

features

The new test `test_asyncio_run_sync_no_run_var_leak` fails on `master`
(`assert 3 == 0`) and passes with this patch. Full test suite: 3690 passed,
0 failed. `ruff check`, `ruff format` and `mypy` all pass on the changed files.

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -23,6 +23,12 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 - Fixed ``anyio.open_process()`` (and ``run_process()``) ignoring the ``extra_groups``
   argument, as it mistakenly passed the value of the ``group`` argument instead
   (`#1209 <https://github.com/agronholm/anyio/pull/1209>`_)
+- Fixed a memory leak on asyncio where each ``anyio.run()`` call permanently retained
+  its event loop, root task and result. ``find_root_task()`` caches the root task in a
+  run variable, and since that task strongly references its loop (the weak key of the
+  ``_run_vars`` entry), the entry could never be evicted. ``anyio.run()`` now clears the
+  loop's run variables on teardown
+  (`#1203 <https://github.com/agronholm/anyio/issues/1203>`_; PR by @vidigoat)
 
 **4.14.1**
 
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -65,6 +65,7 @@ from .. import (
     LockStatistics,
     TaskInfo,
     abc,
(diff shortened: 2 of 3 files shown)


vidigoat:
Closing in favor of #1211, which fixes this at the source in find_root_task() and also covers the direct asyncio.run() path. Thanks @tapetersen.

PR #1213, opened 2026-07-09
pin setup-uv to a commit sha across downstream jobs

EmmanuelNiyonshuti:

## Changes

Fixes #. 

Downstream jobs that use astral-sh/setup-uv, only one (FastAPI) was pinned to a commit sha, the others used a floating tag. I'm pinning all four to the same commit hash. Minor thing, but keep consistency across these jobs.  ([more about setup-uv immutable releases and secure tags](https://github.com/astral-sh/setup-uv/releases/tag/v8.0.0)). 
They're all now on the exact same version too!

## Checklist

features

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/.github/workflows/test-downstream.yml
+++ b/.github/workflows/test-downstream.yml
@@ -72,7 +72,7 @@ jobs:
       - name: Setup uv
         uses: astral-sh/setup-uv@fac544c07dec837d0ccb6301d7b5580bf5edae39 # v8.2.0
         with:
-          version: "0.11.4"
+          version: "0.11.18"
           enable-cache: true
           cache-dependency-glob: |
             pyproject.toml
--- a/.github/workflows/test-downstream.yml
+++ b/.github/workflows/test-downstream.yml
@@ -16,7 +16,7 @@ jobs:
       - uses: actions/checkout@v7
         with:
           repository: Kludex/starlette
-      - uses: astral-sh/setup-uv@v7
+      - uses: astral-sh/setup-uv@fac544c07dec837d0ccb6301d7b5580bf5edae39 # v8.2.0
         with:
           python-version: ${{ matrix.python-version }}
           enable-cache: true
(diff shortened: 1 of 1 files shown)


agronholm:
If anything, I would prefer all downstream tests to be done against the latest versions. When this workflow is triggered, it should tell us if the target AnyIO version works with the latest downstream, not some arbitrary version of the past.

EmmanuelNiyonshuti:
all downstream(those using setup-uv action) are on 8.2.0 but the latest is 8.3.2 . on uv version though, I did not pin it to the latest version. for example fastapi is currently on "0.11.18" version of uv while actually the latest is "0.11.28"  we seem to have 0.11.4 pinned version, I could pin it to the version fastapi is on or the latest uv version, there is a small concern here , setup-uv action verifies checksums for uv versions it already knows about at the time it was released at least before https://github.com/astral-sh/setup-uv/issues/919 is addressed. but this is manually triggered anyway.
Do you mean we need to pin the setup-uv action to the latest version, pin uv to the latest or both? Aside from fastapi, all pinned uv versions we have match the exact downstream pinned versions.

EmmanuelNiyonshuti:
I leaned more toward keeping version pins mirroring downstream's own pins. Not that using the latest versions would necessarily break anything, but I didn't want to have this testing compatibility with anyio using a uv version downstream's own tests have never run on.

agronholm:
Ah, I think there was a misunderstanding on my part. I really shouldn't be replying to issues this late in the evening, but I do non-Python work during working hours :sweat_smile: 
So yes, syncing uv versions with downstream is totally fine. My mind somehow read that as pinning the downstream *projects* themselves.

EmmanuelNiyonshuti:
> Ah, I think there was a misunderstanding of my part. I really shouldn't be replying to issues this late in the evening, but I do non-Python work during working hours 😅 So yes, syncing uv versions with downstream is totally fine. My mind somehow read that as pinning the downstream _projects_ themselves.


😅 Your reply actually made me revisit the `setup-uv` side of things and double-check my assumptions. I was updating the `uv` pins in another project earlier today, my mind immediately went there. I'll update FastAPI's `uv` version to match its downstream pin; since others already match.

agronholm (approved):
Thanks.

PR #1215, opened 2026-07-10
Fix SpooledTemporaryFile.readinto/readinto1 reading twice before rollover

c-tonneslan:
## Changes

Before rollover, `SpooledTemporaryFile.readinto` and `readinto1` read from the underlying file object but discard the result and fall through to `super().readinto(b)`, which reads a *second* time. The buffer is overwritten by the second read and the file position advances twice, so the caller silently loses the first chunk of data. `readinto1` additionally called `self._fp.readinto` instead of `self._fp.readinto1`.

Every other read method on the class (`read`, `read1`, `readline`, `readlines`) already `return` from the non-rolled branch; these two were missing the `return`.

```python
async with SpooledTemporaryFile(max_size=1024, mode="w+b") as f:
    await f.write(b"hello world")
    await f.seek(0)
    buf = bytearray(5)
    await f.readinto(buf)
    # before: buf == b" worl", f.tell() == 10
    # after:  buf == b"hello", f.tell() == 5
```

Only triggers before rollover (while the spooled data is still in memory), which is why it went unnoticed.

## Checklist

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -11,7 +11,7 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 - Fixed ``SpooledTemporaryFile.readinto()`` and ``readinto1()`` reading twice before
   rollover, so the destination buffer was overwritten by the second read and the file
   position advanced twice, silently losing data
-  (`#1215 <https://github.com/agronholm/anyio/pull/1215>`_); PR by @c-tonneslan 
+  (`#1215 <https://github.com/agronholm/anyio/pull/1215>`_); PR by @c-tonneslan
 - Fixed ``CapacityLimiter.total_tokens`` rejecting ``float("inf")`` when the limiter was
   instantiated outside of an event loop. The adapter setter checked for infinity by
   identity (``value is math.inf``), so only the exact ``math.inf`` singleton was accepted,
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -8,6 +8,10 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 - Changed ``ByteReceiveStream.receive()`` implementations to raise a ``ValueError`` when
   ``max_bytes`` is not a positive integer
   (`#1191 <https://github.com/agronholm/anyio/pull/1191>`_)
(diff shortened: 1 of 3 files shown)


docs/versionhistory.rst
```diff
@@ -8,6 +8,10 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 - Changed ``ByteReceiveStream.receive()`` implementations to raise a ``ValueError`` when
   ``max_bytes`` is not a positive integer
   (`#1191 <https://github.com/agronholm/anyio/pull/1191>`_)
+- Fixed ``SpooledTemporaryFile.readinto()`` and ``readinto1()`` reading twice before
+  rollover, so the destination buffer was overwritten by the second read and the file
+  position advanced twice, silently losing data
+  (`#1215 <https://github.com/agronholm/anyio/pull/1215>`_)
```
tapetersen: Feel free to take credit for a good patch 😄 .
```suggestion
  (`#1215 <https://github.com/agronholm/anyio/pull/1215>`_); PR by @c-tonneslan 
```

tests/test_tempfile.py
```diff
@@ -108,6 +108,32 @@ async def test_exact_boundary_no_rollover(self) -> None:
             await stf.write(b"x")
             assert stf._rolled
 
+    @pytest.mark.parametrize("rolled", [False, True])
+    async def test_readinto(self, rolled: bool) -> None:
+        async with SpooledTemporaryFile(max_size=1024, mode="w+b") as stf:
+            await stf.write(b"hello world")
+            if rolled:
+                await stf.rollover()
+            assert stf._rolled is rolled
```
tapetersen: I would've skipped this check and let the test just verify that it works. (There are other tests that verify the rollover is triggered).

Very minor and not required

tapetersen (approved):
Thanks for a good PR.

agronholm:
I was contacted by Rohit Sriram, a security researcher who ran Faultmark against AnyIO's code. This was the only issue the software found.

tests/test_tempfile.py
agronholm: Yes, I'd also like these tests to be trimmed down. Remove the unwarranted asserts and just focus on testing for the actual bug.

agronholm: I simplified the tests, so this should be good to go.

PR #1217, opened 2026-07-11
Fixed 100% CPU spin on cancel scope misuse

### LINKED ISSUE #1111: 100% CPU spin in asyncio _deliver_cancellation due to missing task.done() check

agronholm:

## Changes

Mitigates the CPU-spin effects of cancel scope misnesting.

Fixes #1111.

## Checklist

features

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -592,6 +592,10 @@ class CancelScope(BaseCancelScope):
         should_retry = False
         current = current_task()
         for task in self._tasks:
+            # Always skip tasks that are already done (see issue #1111)
+            if task.done():
+                continue
+
             should_retry = True
             if task._must_cancel:  # type: ignore[attr-defined]
                 continue
--- a/tests/test_taskgroups.py
+++ b/tests/test_taskgroups.py
@@ -15,6 +15,7 @@ from unittest import mock
 
 import pytest
 from pytest import FixtureRequest, MonkeyPatch
+from pytest_mock import MockerFixture
 
 import anyio
 from anyio import (
@@ -358,6 +359,51 @@ async def test_cancel_with_nested_task_groups() -> None:
                 tg.cancel_scope.cancel()
 
     assert len(outer_cancel_spy.call_args_list) < 10
+
+
+@pytest.mark.parametrize("anyio_backend", asyncio_params)
+async def test_no_spin_on_done_task_in_cancel_scope(mocker: MockerFixture) -> None:
+    """Regression test for #1111.
+
+    When a task group is entered but never exited (e.g. it is abandoned because
+    the task owning it finishes without unwinding it), its cancel scope is left
+    with the now-finished host task still in its set of contained tasks.
+    Cancelling such a scope used to make ``_deliver_cancellation`` treat that
+    finished task as still cancellable and reschedule itself via ``call_soon``
+    forever, pinning the event loop at 100% CPU.
+    """
+    from anyio._backends import _asyncio
+
+    class EditableCancelScope(_asyncio.CancelScope):
+        pass
+
+    holder: list[TaskGroup] = []
+
+    async def owner() -> None:
+        tg = create_task_group()
+        await tg.__aenter__()  # entered but deliberately never exited
+        holder.append(tg)
+
+    # Let the owner task run to completion, abandoning the task group. Its
+    # cancel scope now holds the finished host task and nothing else.
+    with mock.patch.object(_asyncio, "CancelScope", EditableCancelScope):
+        await asyncio.create_task(owner())
+        scope = cast(_asyncio.CancelScope, holder[0].cancel_scope)
+        spy = mocker.spy(scope, "_deliver_cancellation")
+        assert scope._host_task is not None and scope._host_task.done()
+        assert scope._host_task in scope._tasks
+
+    scope.cancel()
+
+    # Give any rescheduled _deliver_cancellation callbacks the chance to fire.
+    # With the bug present this reschedules forever (the test timeout would
+    # trip); the fix leaves no pending cancellation callback since the only
+    # remaining task is done.
+    for _ in range(5):
+        await checkpoint()
+
+    assert scope._cancel_handle is None
+    spy.assert_called_once()
 
 
 @pytest.mark.parametrize("return_handle", [False, True])
--- a/tests/test_taskgroups.py
+++ b/tests/test_taskgroups.py
@@ -391,8 +391,6 @@ async def test_no_spin_on_done_task_in_cancel_scope(mocker: MockerFixture) -> No
 
     scope = cast(_asyncio.CancelScope, holder[0].cancel_scope)
     spy = mocker.spy(scope, "_deliver_cancellation")
-    assert scope._host_task is not None and scope._host_task.done()
-    assert scope._host_task in scope._tasks
 
     scope.cancel()
 
--- a/tests/test_taskgroups.py
+++ b/tests/test_taskgroups.py
@@ -374,24 +374,15 @@ async def test_no_spin_on_done_task_in_cancel_scope(mocker: MockerFixture) -> No
     """
     from anyio._backends import _asyncio
 
+    # To allow the mocker to override a @final class
     class EditableCancelScope(_asyncio.CancelScope):
         pass
 
-    holder: list[TaskGroup] = []
+    async def owner() -> EditableCancelScope:
+        return cast(EditableCancelScope, EditableCancelScope().__enter__())
 
-    async def owner() -> None:
-        tg = create_task_group()
-        await tg.__aenter__()  # entered but deliberately never exited
-        holder.append(tg)
-
-    # Let the owner task run to completion, abandoning the task group. Its
-    # cancel scope now holds the finished host task and nothing else.
-    with mock.patch.object(_asyncio, "CancelScope", EditableCancelScope):
-        await asyncio.create_task(owner())
-
-    scope = cast(_asyncio.CancelScope, holder[0].cancel_scope)
+    scope = await asyncio.create_task(owner())
     spy = mocker.spy(scope, "_deliver_cancellation")
-
     scope.cancel()
 
     # Give any rescheduled _deliver_cancellation callbacks the chance to fire.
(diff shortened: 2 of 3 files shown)


tests/test_taskgroups.py
```diff
@@ -360,6 +361,52 @@ async def middle_task() -> None:
...
+
+    # Let the owner task run to completion, abandoning the task group. Its
+    # cancel scope now holds the finished host task and nothing else.
+    with mock.patch.object(_asyncio, "CancelScope", EditableCancelScope):
+        await asyncio.create_task(owner())
+
+    scope = cast(_asyncio.CancelScope, holder[0].cancel_scope)
+    spy = mocker.spy(scope, "_deliver_cancellation")
+    assert scope._host_task is not None and scope._host_task.done()
+    assert scope._host_task in scope._tasks
```
tapetersen: I would tend to skip these kind of extra checks if they rely on internals and the main test catches the regression by itself

src/anyio/_backends/_asyncio.py
```diff
@@ -592,6 +592,10 @@ def _deliver_cancellation(self, origin: CancelScope) -> bool:
         should_retry = False
         current = current_task()
         for task in self._tasks:
+            # Always skip tasks that are already done (see issue #1111)
+            if task.done():
+                continue
+
```
tapetersen, line 598: I'm mainly wondering if this shouldn't remove the task from `self._tasks` as well?

After digging through this a bit more I understand it as this *only* happens if it's a *non*-anyio task that entered the `CancelScope` and then abandons it. Because otherwise the `task_done` callback [here](https://github.com/agronholm/anyio/pull/1217/changes#diff-202c3fc4551e5a287df488a9bd96a25dcc1405ec135d5c7aceb6ef4c02755341L841) would have removed it from the scope.

So an alternative here would be to check if that callback is registered on enter and if not add one that does the same de-registration.

agronholm: This occurred to me too, but to remove the task immediately, we'd need to make a copy of the task list on every iteration, as mutating a list during its iteration is a no-no. I'm not favorable to this given this is potentially hot-path code.

tapetersen: My thoughts go to fixing the issue the task remains in the group at all. As in registering a `done_callback` on the task entering the `CancelScope` in the first place. It's really only the host tasks that has this issue.

tests/test_taskgroups.py
tapetersen: After digging deeper these asserts would actually trigger if the real bug was fixed which makes me even more sceptical.

tests/test_taskgroups.py
```diff
@@ -360,6 +361,52 @@ async def middle_task() -> None:
...
+        tg = create_task_group()
+        await tg.__aenter__()  # entered but deliberately never exited
+        holder.append(tg)
+
+    # Let the owner task run to completion, abandoning the task group. Its
+    # cancel scope now holds the finished host task and nothing else.
+    with mock.patch.object(_asyncio, "CancelScope", EditableCancelScope):
+        await asyncio.create_task(owner())
+
+    scope = cast(_asyncio.CancelScope, holder[0].cancel_scope)
```
tapetersen: This doesn't have anything to do with TaskGroups specifically from what I understand

```suggestion
    When a CancelScope is entered but never exited (e.g. it is abandoned because
    the task owning it finishes without unwinding it), it's left
    with the now-finished host task still in its set of contained tasks.
    Cancelling such a scope used to make ``_deliver_cancellation`` treat that
    finished task as still cancellable and reschedule itself via ``call_soon``
    forever, pinning the event loop at 100% CPU.
    """
    from anyio._backends import _asyncio

    # To allow the mocker to override a @final class.
    class EditableCancelScope(_asyncio.CancelScope):
        pass

    holder: list[EditableCancelScope] = []

    async def owner() -> None:
        scope = EditableCancelScope().__enter__()
        holder.append(scope)

    # Let the owner task run to completion, abandoning CancelScope.
    # cancel scope now holds the finished host task and nothing else.
    await asyncio.create_task(owner())

    scope = holder[0]
```

tapetersen: I tested it and it still catches the error and the fix still resolves it.

src/anyio/_backends/_asyncio.py
```diff
@@ -592,6 +592,10 @@ def _deliver_cancellation(self, origin: CancelScope) -> bool:
         should_retry = False
         current = current_task()
         for task in self._tasks:
+            # Always skip tasks that are already done (see issue #1111)
+            if task.done():
```
tapetersen, line 596:
```suggestion

if (self._host_task and self._host_task.done()):
    self._tasks.remove(self._host_task)

current = current_task()
for task in self._tasks:
            # Always skip tasks that are already done (see issue #1111)
            if task.done():
```

~~This would fix this case potentially cheaper.~~ ~~Never mind it breaks a whole lot of other things~~
Ok with a None check it works fine.

tests/test_taskgroups.py
agronholm: Extra checks removed.

tapetersen:
I don't really oppose merging the fix as it does help alleviate some real problems. It does bug me that it essentially masks both an issue on the user's side as well as what is essentially a bug on ours, that non-anyio tasks that enter a `CancelSope` don't de-register when done as done with the `task_done` callback we add in `TaskGroup._spawn`.

I did some experimentation and an alternative is to check if it's a freestanding task in `CancelScope.__enter__()` which I think is possible by checking `self._parent_scope is None` before `self._timeout()` and if so register a `task_done_callback` that calls `self._tasks.discard(task)`.

tests/test_taskgroups.py
agronholm: I simplified it even further now.

src/anyio/_backends/_asyncio.py
agronholm: It doesn't fully solve the problem anyway, as the cancel scope could be abandoned w/o the task even exiting.

tapetersen: It wouldn't solve the problem of it being orphaned and therefore probably never closed by *this* task no.

But as far as I can see it would solve *this* problem of us spinning trying to schedule cancellations for a task that has already finished. If it's still running it should get a cancel delivered and if it's done it doesn't need to so if we remove it early that works (and is the reason this is only a problem with freestanding tasks).

agronholm:
The only thing this fix is meant to do is to stop the CPU spin when cancel scopes are misused. None of the checks have been removed. I only agreed to add this fix because users "in the know" were suffering due to bad third party code.

tapetersen:
I understand that and agree that it alleviates problems for real users but I'm arguing the only reason we need the skip is due to a real bug in that we don't remove Tasks that `__enter__` a `CancelScope` from their task-list when the task finishes. 

The only reason it normally terminates is exactly because tasks de-register themselves when they finish which they do as we add that callback in `_spawn`.

As that assumption seems to be relied upon I did think it would be better to fix that instead (by registering a callback in __enter__ if it didn't already have one).

This is worthwhile anyway since those bugs may happen again and it's a reasonable optimization as well (it can trigger at least one round depending on which order callbacks are scheduled in).

tapetersen (approved):
This is reasonable as an optimization to just skip looping if we're already done in either case we want to ensure tasks de-register in some other way.

PR #1223, opened 2026-07-14
Fix asyncio CapacityLimiter over-granting tokens when total_tokens is raised while over-subscribed

zelinewang:
## Changes

I was going through the `CapacityLimiter` token accounting after the recent over-grant fixes (#1172/#1183/#1189) and found one more spot on the asyncio backend that can over-grant. There's no existing issue for it.

The `total_tokens` setter decides how many waiters to wake from the *increase* in `total_tokens` (`waiters_to_notify = max(value - self._total_tokens, 0)`) instead of from the limiter's actual spare capacity. That's only equivalent when the limiter is exactly full. If it's over-subscribed (`borrowed_tokens > total_tokens`), the delta over-counts the free slots and the setter wakes waiters that have no token to take, so `borrowed_tokens` ends up above `total_tokens` and `available_tokens` goes negative.

You can reach the over-subscribed state through the public API: lower `total_tokens` below the current borrower count, then raise it again.

It's the same symptom as #1170/#1172, but a different path. #1172 fixed the nowait-acquire/release race and added `_notify_next_waiter()`, which guards correctly on `len(self._borrowers) < self._total_tokens` — but the setter kept its old delta guard. This PR just makes the setter use that same spare-capacity check.

Repro (asyncio):

```python
import anyio

async def main():
    limiter = anyio.CapacityLimiter(2)
    limiter.acquire_on_behalf_of_nowait("A")
    limiter.acquire_on_behalf_of_nowait("B")           # full: 2/2

    async def waiter(name):
        await limiter.acquire_on_behalf_of(name)       # both block

        tg.start_soon(waiter, "W1")
        tg.start_soon(waiter, "W2")
        await anyio.sleep(0.1)
        limiter.total_tokens = 1                        # lower below borrowers
        limiter.total_tokens = 2                        # raise back (still full)
        await anyio.sleep(0.05)
        print(limiter.borrowed_tokens, limiter.total_tokens, limiter.available_tokens)
        # before: 3 2 -1   (over-granted, one waiter let through)
        # after:  2 2 0     (correct, matches trio)
        tg.cancel_scope.cancel()

anyio.run(main)
```

trio already gets this right (it delegates to `trio.CapacityLimiter`), so the new regression test is parametrized over all backends and passes on trio both before and after.

## Checklist

Docs are unchanged on purpose: the documented behavior ("If the total number of tokens is increased, the proportionate number of tasks waiting on this limiter will be granted their tokens") already describes the correct outcome — this only makes the asyncio implementation match it when the limiter is over-subscribed.


tapetersen:
Looks good thanks!

src/anyio/_backends/_asyncio.py
```diff
@@ -2076,15 +2076,17 @@ def total_tokens(self, value: float) -> None:
...
         self._total_tokens = value
 
-        # Notify waiting tasks that they have acquired the limiter
-        while self._wait_queue and waiters_to_notify:
+        # Notify waiting tasks that they have acquired the limiter, but only while
+        # there is spare capacity. Using the raw increase in total_tokens would
+        # over-grant when the limiter is over-subscribed (borrowers exceeding
+        # total_tokens after the limit was previously lowered below the number of
+        # current borrowers).
+        while self._wait_queue and len(self._borrowers) < self._total_tokens:
```
tapetersen: No need to elaborate about a bug that existed. A note that we check against the real capacity is enough here.

```suggestion
        # Notify waiting tasks that they have acquired the limiter while
        # there is spare capacity.
        while self._wait_queue and len(self._borrowers) < self._total_tokens:
```

tapetersen (approved):

src/anyio/_backends/_asyncio.py
zelinewang: Thanks for the review! Trimmed the comment as suggested.

docs/versionhistory.rst
```diff
@@ -72,6 +72,13 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
   which triggers ``PytestRemovedIn10Warning`` on ``pytest>=9.2`` and crashes pytest at
   startup when ``filterwarnings = error`` is configured
   (`#1271 <https://github.com/agronholm/anyio/issues/1271>`_; PR by @matthewfeickert)
+- Fixed ``CapacityLimiter`` on the asyncio backend over-granting tokens
+  (``borrowed_tokens`` exceeding ``total_tokens`` and ``available_tokens`` going
+  negative) when ``total_tokens`` was raised while the limiter was
+  over-subscribed, i.e. after ``total_tokens`` had previously been lowered below
+  the number of current borrowers. The setter woke waiters based on the increase
+  in ``total_tokens`` rather than the actual spare capacity
+  (`#1223 <https://github.com/agronholm/anyio/pull/1223>`_; PR by @zelinewang)
```
agronholm: This entry is far too verbose. Users who want the nitty-gritty details can just visit the linked PR.
```suggestion
- Fixed ``CapacityLimiter`` on the asyncio backend over-granting tokens when
  ``total_tokens`` was raised while the limiter was over-subscribed
  (`#1223 <https://github.com/agronholm/anyio/pull/1223>`_; PR by @zelinewang)
```

tests/test_synchronization.py
```diff
@@ -879,6 +879,57 @@ async def worker(entered_event: Event) -> None:
...
+
+            # Release the initial borrowers so the workers can proceed and the
+            # task group can exit cleanly.
+            limiter.release_on_behalf_of("A")
+            limiter.release_on_behalf_of("B")
+            with fail_after(1):
+                for ev in entered_events:
+                    await ev.wait()
+
+            continue_event.set()
```
agronholm: The test is grossly overcomplicated. Repurposing the snippet from the PR description would have been the right call. I verified that this fails for asyncio and passes for Trio on `master`, just like the test it replaces.
```suggestion
        limiter.acquire_on_behalf_of_nowait("A")
        limiter.acquire_on_behalf_of_nowait("B")

        async with create_task_group() as tg:
            tg.start_soon(limiter.acquire)
            tg.start_soon(limiter.acquire)
            await wait_all_tasks_blocked()
            assert limiter.statistics().borrowed_tokens == 2
            assert limiter.statistics().tasks_waiting == 2

            limiter.total_tokens = 1
            limiter.total_tokens = 2
            await wait_all_tasks_blocked()
            assert limiter.statistics().borrowed_tokens == 2
            assert limiter.statistics().tasks_waiting == 2
            tg.cancel()
```

agronholm (approved):
Had to make some adjustments, but looks good now. Thanks!

PR #1226, opened 2026-07-15
fix: use monotonic task IDs instead of object IDs

### LINKED ISSUE #324: in py3.8+ the first asyncio.Task object is allocated with the same ID every time
This is a problem in anyio because TaskInfo's `__eq__` goes on to compare dead object IDs

```python
>>> async def current_task():
...     return anyio.get_current_task()
... 
>>> t = anyio.run(current_task)
>>> u = anyio.run(current_task)
>>> t == u
True
>>> t.id == u.id
True
```

pctablet505:
Fixes #324

TaskInfo previously used Python's `id()` of the underlying task object, which is reused by the allocator (especially for the first asyncio.Task in each run). This caused TaskInfo objects from different runs to compare equal and share IDs.

This change assigns each task a monotonic counter-based ID, cached in a WeakKeyDictionary so the same task object always receives the same ID. The parent_id references are updated to use the same IDs.

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -3,6 +3,13 @@ Version history
 
 This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
+**UNRELEASED**
+
+- Fixed ``TaskInfo`` objects from separate event loop runs comparing equal due to
+  reused object IDs. Each task is now assigned a process-monotonic ID cached in a
+  ``WeakKeyDictionary``, so ``TaskInfo`` identity is stable and unique per task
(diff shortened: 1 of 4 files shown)


agronholm:
What practical problem prompted this? `TaskInfo` is supposed to be used for debugging, and using task IDs helped to identify the task objects they represent. What purpose do these monotonically generated IDs have?

tapetersen:
@pctablet505 If you want this merged you will need to provide some rationale why we actually need stable `TaskInfo` id's. It will almost surely be ignored and/or closed otherwise.

agronholm:
Closing due to lack of responses.

PR #1227, opened 2026-07-16
add reason parameter to explain timeouts in fail_after

Graeme22:
## Changes

Adds an optional `reason` parameter to `fail_after`. This allows users to specify context around timeout failures without needing to look at source code.

Motivation: I've often found myself using `move_on_after` where I'd rather use `fail_after` just to be able to add context to the raised `TimeoutError`:

```python
with move_on_after(1) as scope:
    ...
if scope.cancelled_caught:
    raise TimeoutError("Server didn't respond in time!")
```

With this change you can just do:
```python
with fail_after(1, reason="Server didn't respond in time!"):
    ...
```

## Checklist

features

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/src/anyio/_core/_tasks.py
+++ b/src/anyio/_core/_tasks.py
@@ -128,7 +128,7 @@ class CancelScope:
 
 @contextmanager
 def fail_after(
-    delay: float | None, shield: bool = False
+    delay: float | None, shield: bool = False, reason: str | None = None
 ) -> Generator[CancelScope, None, None]:
     """
     Create a context manager which raises a :class:`TimeoutError` if does not finish in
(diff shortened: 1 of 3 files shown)


src/anyio/_core/_tasks.py
```diff
@@ -151,7 +152,7 @@ def fail_after(
         yield cancel_scope
 
     if cancel_scope.cancelled_caught and current_time() >= cancel_scope.deadline:
-        raise TimeoutError
+        raise TimeoutError(reason or "")
```
agronholm: I'm not entirely sure we should pass "" by default where we didn't do that before.

agronholm:
Once you merge from master and update the new functions too, and resolve my comment, this should be good to go.

agronholm (approved):
Perfect!

PR #1228, opened 2026-07-16
Added the move_on_at() and fail_at() functions

agronholm:

## Changes

Adds the `move_on_at()` and `fail_at()` function to match Trio and to complement the existing `move_after()` and `fail_after()` functions.

## Checklist

features

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/src/anyio/_core/_tasks.py
+++ b/src/anyio/_core/_tasks.py
@@ -146,7 +146,7 @@ def fail_at(
     current_time = get_async_backend().current_time
     effective_deadline = math.inf if deadline is None else deadline
     with get_async_backend().create_cancel_scope(
-        deadline=effective_deadline or math.inf, shield=shield
+        deadline=effective_deadline, shield=shield
     ) as cancel_scope:
         yield cancel_scope
 
--- a/src/anyio/_core/_tasks.py
+++ b/src/anyio/_core/_tasks.py
@@ -6,7 +6,7 @@ from collections.abc import (
     Coroutine,
     Generator,
 )
-from contextlib import contextmanager
+from contextlib import AbstractContextManager, contextmanager
 from enum import Enum, auto
 from inspect import iscoroutine
 from types import TracebackType
@@ -196,10 +196,9 @@ def move_on_at(deadline: float | None, shield: bool = False) -> CancelScope:
     )
 
 
-@contextmanager
 def move_on_after(
     delay: float | None, shield: bool = False
-) -> Generator[CancelScope, None, None]:
+) -> AbstractContextManager[CancelScope]:
     """
     Create a cancel scope with a deadline that expires after the given delay.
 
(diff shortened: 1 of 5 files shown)


src/anyio/_core/_tasks.py
```diff
@@ -127,33 +125,75 @@ def __exit__(
...
 
+    .. versionadded:: 4.15.0
+
     """
     current_time = get_async_backend().current_time
-    deadline = (current_time() + delay) if delay is not None else math.inf
+    effective_deadline = math.inf if deadline is None else deadline
     with get_async_backend().create_cancel_scope(
-        deadline=deadline, shield=shield
+        deadline=effective_deadline or math.inf, shield=shield
```
Graeme22: I think the only effect of the `or` here is to change a deadline of 0 to `math.inf` no? Which seems like a bad thing to do silently

agronholm: Good catch. Honestly I'm not sure how that code ended up there.

agronholm:
I think there is a regression in this, given that `fail_after()` now gets the current time immediately and passes it to `fail_at()`. This should not happen before the context manager is actually entered,

tapetersen:
> I think there is a regression in this, given that `fail_after()` now gets the current time immediately and passes it to `fail_at()`. This should not happen before the context manager is actually entered,

I agree that's the behavior we should have (that the _after functions should start counting from `__enter__` and not creation). I do recall there was a discussion about it for asyncio's version where they were relctant to change it due to backwards compatibility.

Wondering if we should add something to docs mentioning using `fail_at(current_time()+delay)` for that case instead.

agronholm:
> > I think there is a regression in this, given that `fail_after()` now gets the current time immediately and passes it to `fail_at()`. This should not happen before the context manager is actually entered,
> 
> I agree that's the behavior we should have (that the _after functions should start counting from `__enter__` and not creation). I do recall there was a discussion about it for asyncio's version where they were relctant to change it due to backwards compatibility.
> 
> Wondering if we should add something to docs mentioning using `fail_at(current_time()+delay)` for that case instead.

To be clear, with the current state of this PR we're not changing that behavior in any way. My comment only related to the changes I had made here prior. I fixed this behavior change with my latest changes. Is this okay now?

src/anyio/_core/_tasks.py
```diff
@@ -169,7 +214,10 @@ def move_on_after(delay: float | None, shield: bool = False) -> CancelScope:
     deadline = (
         (get_async_backend().current_time() + delay) if delay is not None else math.inf
     )
-    return get_async_backend().create_cancel_scope(deadline=deadline, shield=shield)
+    with get_async_backend().create_cancel_scope(
+        deadline=deadline, shield=shield
+    ) as scope:
+        yield scope
```
tapetersen: Won't this now change the deadline behavior for `move_on_after`?
They were already inconsistent on master with `fail_after` being a `ContextManager[CancelScope]` that computed the deadline in `__enter__` and `move_on_after` calculating the deadline when called.

I think these are the better semantics for both but it is a change.

agronholm: You're right. I restored the `master` behavior and added a note highlighting the difference.

tapetersen (approved):
LGTM with latest changes.

PR #1229, opened 2026-07-17
Fix event loop memory leak from cached root task (#1203)

### LINKED ISSUE #1203: _run_vars memory leak

uttam12331:
Fixes #1203

### Problem

`find_root_task()` (asyncio backend) caches the root task via
`_root_task.set(task)`. That RunVar lives in the per-loop dict of
`_run_vars`, which is a `WeakKeyDictionary` keyed by the event loop.

Because the cached `Task` holds a strong reference back to its loop, the loop
is kept alive by a value stored under its own weak key — so the weak key never
clears. Every `anyio.run()` that goes through `find_root_task()` (e.g. anything
using `to_thread.run_sync()`) permanently leaks its event loop, root task and
the coroutine's result. In a loop this grows memory without bound and
`gc.collect()` cannot reclaim it.

Reproduction:

```python
import gc, asyncio, anyio
from anyio import to_thread

async def main():
    await to_thread.run_sync(lambda: None)

def loops():
    gc.collect()
    return sum(1 for o in gc.get_objects() if isinstance(o, asyncio.AbstractEventLoop))

before = loops()
for _ in range(5):
    anyio.run(main)
print(loops() - before)   # 5 before this change, 0 after
```

### Fix

Cache a `weakref` to the root task instead of a strong reference, so the
cached value no longer pins the loop. While the loop is running the root task
is alive, so the weakref resolves; once the loop is done, the weakref clears
and everything is collected.

### Verification

- The reproduction above goes from 5 leaked loops to 0.
- Added `test_anyio_run_does_not_leak_event_loop` (uses `clear_cache`-free gc
  check), which fails on `master` and passes with this change.
- Full test suite passes (the only failures on my machine are pre-existing,
  unrelated Windows/`psutil` env issues that also fail on an unmodified tree).
- `ruff` clean; changelog updated.

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -5,6 +5,12 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
 **4.14.2**
 
+- Fixed a memory leak on the asyncio backend where every ``anyio.run()`` call that used
+  ``to_thread.run_sync()`` (via ``find_root_task()``) permanently leaked its event loop,
+  root task and result. The root task was cached with a strong reference in the
+  loop-keyed ``_run_vars`` dict, so the loop (the weak key) was kept alive by its own
+  value; the cached task is now held via a weak reference
+  (`#1203 <https://github.com/agronholm/anyio/issues/1203>`_)
 - Changed ``ByteReceiveStream.receive()`` implementations to raise a ``ValueError`` when
   ``max_bytes`` is not a positive integer
   (`#1191 <https://github.com/agronholm/anyio/pull/1191>`_)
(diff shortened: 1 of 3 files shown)


EmmanuelNiyonshuti:
Looks like there is already a PR, @uttam12331 did you take a look at https://github.com/agronholm/anyio/pull/1211?

agronholm:
The bot also deleted the PR template which is grounds for dismissal anyway.

PR #1232, opened 2026-07-18
Fix `TaskHandle.name` for `TaskGroup.start` on Trio

### LINKED ISSUE #1231: Default `TaskHandle.name` doesn't match the task name for `TaskGroup.start` on Trio
### Things to check first

### AnyIO version

master (2ba69a649011e4608ab9485a0dcfe72b6f956ecc)

### Python version

CPython 3.13.14

### What happened?

For `TaskGroup.start` tasks on Trio, the default ``TaskHandle.name`` does not match the task name (`TaskInfo.name`). It's missing the module name.

All three of these assertions fail on Trio:

```python
async def test_start_name_default() -> None:
    """
    This is similar to ``TestCreateTask.test_task_name_default``, but for ``start``
    instead of ``create_task``.
    """

    async def taskfunc(*, task_status: TaskStatus[None]) -> str | None:
        task_status.started()
        return get_current_task().name

        handle = await tg.start(taskfunc, return_handle=True)
        assert re.match(
            r"<TaskHandle finished name='tests.test_taskgroups.test_start_name_default.<locals>.taskfunc' "
            r"coro=<coroutine object test_start_name_default.<locals>.taskfunc at 0x[0-9a-fA-F]+>>",
            repr(handle),
        )
        assert await handle == handle.name

    assert (
        handle.name == "tests.test_taskgroups.test_start_name_default.<locals>.taskfunc"
    )
```

gschaffner:
## Changes

Fixes #1231.

## Checklist

features

these instructions.

### DIFF
--- a/tests/test_taskgroups.py
+++ b/tests/test_taskgroups.py
@@ -2344,37 +2344,6 @@ class TestCreateTask:
         async with create_task_group() as tg:
             assert await tg.create_task(taskfunc(), context=ctx) == 42
 
-    async def test_task_name_default(self) -> None:
-        async def taskfunc() -> str | None:
-            return get_current_task().name
-
-        async with create_task_group() as tg:
-            handle = tg.create_task(taskfunc())
-            assert re.match(
-                r"<TaskHandle pending "
-                r"name='tests.test_taskgroups.TestCreateTask.test_task_name_default.<locals>.taskfunc' "
-                r"coro=<coroutine object(.+)>>",
-                repr(handle),
-            )
-            assert await handle == handle.name
-
-        assert (
-            handle.name
-            == "tests.test_taskgroups.TestCreateTask.test_task_name_default.<locals>.taskfunc"
-        )
-
-    async def test_task_name_custom_name(self) -> None:
-        async def taskfunc() -> None:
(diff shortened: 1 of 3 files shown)


tests/test_taskgroups.py
```diff
@@ -2319,6 +2319,30 @@ async def taskfunc() -> None:
         await handle.wait()
 
 
+async def test_start_name_default() -> None:
+    """
+    This is similar to ``TestCreateTask.test_task_name_default``, but for ``start``
+    instead of ``create_task``.
+    """
```
gschaffner: Not sure if you want this comment or not. Also not sure if you want to combine `TestCreateTask.test_task_name_default` and `test_start_name_default` into a single test function with `pytest.mark.parametrize`.

agronholm: I think I would actually like it. It ensures matching output and deduplicates a lot of code.

agronholm:
This looks good to go after the change you suggested has been implemented.

tests/test_taskgroups.py
gschaffner: Done.

agronholm (approved):
Thanks!

PR #1234, opened 2026-07-18
Include the module name in `create_task` task names in more situations

gschaffner:
## Changes

This is an idea/proposal for how we could make `TaskGroup.create_task(func())`

*   give default task names that are more accurate (i.e. better for debugging).

*   give default task names that (typically) match the default task name for the analogous `TaskGroup.start_soon(func)`/`TaskGroup.start(func)` task.

The main problem that I know of is that this would technically be a backward-incompatible change.

## Checklist

features

these instructions.

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -3,6 +3,15 @@ Version history
 
 This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
+**UNRELEASED**
+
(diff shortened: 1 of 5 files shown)


agronholm (approved):
I didn't think this was doable. Thanks!

gschaffner:
Oops, sorry I missed the changelog link. I forgot that we link to both issues and PRs, not just issues.

PR #1235, opened 2026-07-19
Shared subprocess impl

### LINKED ISSUE #783: Look into creating a shared subprocess implementation between backends

graingert:

## Changes

Fixes #783. Also resolves the asyncio/Trio inconsistencies reported in #828.

Moves the subprocess lifecycle logic (the `Process` object, spawning and reaping) into a
single, backend-agnostic implementation in `anyio._core._subprocesses`, built on
`subprocess.Popen()` plus a small set of backend primitives:

- `create_subprocess_stdin_pipe` / `create_subprocess_output_pipe`
- `wait_for_child_exit`

This removes the reliance on undocumented backend internals such as
`StreamReader.set_exception()`.

**Backends / platforms**

- **asyncio** uses the shared implementation on every platform. Pipes are created with
  `loop.connect_read_pipe` / `loop.connect_write_pipe`:
  - stdlib `ProactorEventLoop` (Windows default): overlapped IOCP pipes via `PipeHandle`;
    the child is waited on with `proactor.wait_for_handle`.
  - **winloop** (libuv): pipes backed by real overlapped file descriptors.
  - **Windows `SelectorEventLoop`** (which can't do overlapped pipe I/O): the pipes run on
    a background `ProactorEventLoop`/winloop thread (mirroring the existing selector-thread
    helper), and operations are proxied to it.
  - POSIX: `os.pipe()` + `wait_readable`/`wait_writable`.
  - On Windows, waiting for a child that has no `_proactor` (winloop / SelectorEventLoop)
    uses a loop-agnostic `RegisterWaitForSingleObject` helper (thread-pool wait delivered
    via `call_soon_threadsafe`, no Python thread tied up).
- **Trio** uses the shared implementation on POSIX and keeps its native implementation on
  Windows.
- Reaping on POSIX uses a pidfd on Linux, or `waitid(WNOWAIT)` / `Popen.wait()` in a worker
  thread elsewhere (`os.waitid` is unavailable on macOS before Python 3.13).

The Python 3.10 `asyncio.Runner` backport was moved out of the asyncio backend into
`anyio._core._asyncio_runner` so the proactor thread can reuse it.

**Fixed inconsistencies (#828)**

- `Process.wait()` no longer waits for the standard streams to close, so it returns
  promptly once the process exits, even when a pipe is inherited by a grandchild process.
- `Process.returncode` now polls the process on both backends, so it no longer returns a
  stale `None` after the process has exited.
- Signalling an already-exited process via `terminate()` / `kill()` / `send_signal()` is
  now consistently a no-op instead of raising on asyncio.

A Windows `SelectorEventLoop` entry was added to the test matrix.

## Checklist

features

these instructions.

### Updating the changelog

```
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -3,6 +3,25 @@ Version history
 
 This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
+**UNRELEASED**
+
+- Changed subprocess handling to use a shared, backend-agnostic implementation built on
+  ``subprocess.Popen()`` plus a small set of backend primitives (asynchronous pipes and
+  child-process reaping). This is used by the asyncio backend on all platforms and by the
+  Trio backend on POSIX (Trio keeps its native implementation on Windows), and removes the
+  reliance on undocumented backend internals
+  (`#783 <https://github.com/agronholm/anyio/issues/783>`_; PR by @graingert)
+- Fixed several inconsistencies between the asyncio and Trio backends when working with
+  subprocesses (`#828 <https://github.com/agronholm/anyio/discussions/828>`_; PR by
+  @graingert):
+
+  * ``Process.wait()`` no longer waits for the standard streams to close, so it returns
+    promptly once the process exits, even when a pipe is inherited by a grandchild process
+  * ``Process.returncode`` now polls the process on both backends, so it no longer returns
+    a stale ``None`` after the process has exited
+  * Signalling an already-exited process via ``terminate()``, ``kill()`` or
+    ``send_signal()`` is now consistently a no-op instead of raising on asyncio
+
 **4.14.2**
 
 - Changed ``ByteReceiveStream.receive()`` implementations to raise a ``ValueError`` when
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -5,6 +5,10 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
 **UNRELEASED**
 
+- Added support for running subprocesses on the asyncio ``SelectorEventLoop`` on Windows,
+  which does not support subprocesses natively; the pipe I/O is handled on a background
+  ``ProactorEventLoop`` (or winloop) thread
+  (`#783 <https://github.com/agronholm/anyio/issues/783>`_; PR by @graingert)
 - Changed subprocess handling to use a shared, backend-agnostic implementation built on
   ``subprocess.Popen()`` plus a small set of backend primitives (asynchronous pipes and
   child-process reaping). This is used by the asyncio backend on all platforms and by the
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -8,7 +8,7 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 - Added support for running subprocesses on the asyncio ``SelectorEventLoop`` on Windows,
   which does not support subprocesses natively; the pipe I/O is handled on a background
   ``ProactorEventLoop`` (or winloop) thread
-  (`#1235 <https://github.com/agronholm/anyio/pulls/1235>`_; PR by @graingert)
+  (`#1235 <https://github.com/agronholm/anyio/pull/1235>`_; PR by @graingert)
 - Changed subprocess handling to use a shared, backend-agnostic implementation built on
   ``subprocess.Popen()`` plus a small set of backend primitives (asynchronous pipes and
   child-process reaping). This is used by the asyncio backend on all platforms and by the
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -8,7 +8,7 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 - Added support for running subprocesses on the asyncio ``SelectorEventLoop`` on Windows,
   which does not support subprocesses natively; the pipe I/O is handled on a background
   ``ProactorEventLoop`` (or winloop) thread
-  (`#783 <https://github.com/agronholm/anyio/issues/783>`_; PR by @graingert)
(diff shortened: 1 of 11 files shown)


docs/versionhistory.rst
```diff
@@ -5,6 +5,26 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
 **UNRELEASED**
 
+- Added support for running subprocesses on the asyncio ``SelectorEventLoop`` on Windows,
+  which does not support subprocesses natively; the pipe I/O is handled on a background
+  ``ProactorEventLoop`` (or winloop) thread
+  (`#783 <https://github.com/agronholm/anyio/issues/783>`_; PR by @graingert)
```
graingert:
```suggestion
  (`#1235 <https://github.com/agronholm/anyio/pulls/1235>`_; PR by @graingert)
```

docs/versionhistory.rst
```diff
@@ -5,6 +5,26 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
 **UNRELEASED**
 
+- Added support for running subprocesses on the asyncio ``SelectorEventLoop`` on Windows,
+  which does not support subprocesses natively; the pipe I/O is handled on a background
+  ``ProactorEventLoop`` (or winloop) thread
+  (`#1235 <https://github.com/agronholm/anyio/pulls/1235>`_; PR by @graingert)
```
graingert:
```suggestion
  (`#1235 <https://github.com/agronholm/anyio/pull/1235>`_; PR by @graingert)
```

docs/versionhistory.rst
```diff
@@ -5,6 +5,26 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
...
+- Fixed several inconsistencies between the asyncio and Trio backends when working with
+  subprocesses (`#828 <https://github.com/agronholm/anyio/discussions/828>`_; PR by
+  @graingert):
+
+  * ``Process.wait()`` no longer waits for the standard streams to close, so it returns
+    promptly once the process exits, even when a pipe is inherited by a grandchild process
+  * ``Process.returncode`` now polls the process on both backends, so it no longer returns
+    a stale ``None`` after the process has exited
+  * Signalling an already-exited process via ``terminate()``, ``kill()`` or
+    ``send_signal()`` is now consistently a no-op instead of raising on asyncio
```
agronholm, line 27: This needs to be severely trimmed down before we merge.

agronholm:
There's so much low-level code added here that it would be hard to justify merging all of it.

src/anyio/_backends/_trio.py
```diff
@@ -415,6 +416,90 @@ def stderr(self) -> abc.ByteReceiveStream | None:
...
+                    except trio.ClosedResourceError:
+                        raise ClosedResourceError from None
+                except OSError as exc:
+                    if self._closed:
+                        raise ClosedResourceError from None
+
+                    raise BrokenResourceError from exc
+                else:
+                    if data:
+                        return data
```
tapetersen, line 449: This doesn't hit a checkpoint if the read is successful.
Is that intentional for speed? (also not sure how committed to following trio's guarantees regarding that)

tests/test_subprocesses.py
```diff
@@ -483,7 +485,125 @@ async def test_close_with_stdout_blocked_subprocess(anyio_backend_name: str) ->
...
+@pytest.mark.skipif(
+    platform.system() == "Windows", reason="POSIX-only reaping fallback"
+)
+@pytest.mark.parametrize("have_waitid", [True, False], ids=["waitid", "popen-wait"])
+async def test_wait_without_pidfd(
+    monkeypatch: pytest.MonkeyPatch, have_waitid: bool
+) -> None:
+    """
+    Exercise the worker-thread reaping fallback used when ``os.pidfd_open`` is unavailable
+    (e.g. PyPy or kernels older than 5.3) and, in turn, when ``os.waitid`` is unavailable
```
tapetersen, line 533: This is actually a lot more common than that as the standalone python builds from uv don't include it 
```
$ uv run --isolated --python 3.14 python -c "import os; print(hasattr(os, 'pidfd_open'))"
False
```

See https://github.com/astral-sh/python-build-standalone/issues/193

(Just a note to as it surprised me)

tests/test_subprocesses.py
```diff
@@ -483,7 +485,125 @@ async def test_close_with_stdout_blocked_subprocess(anyio_backend_name: str) ->
...
+
+async def test_open_process_nonexistent_executable() -> None:
+    """
+    A failure to spawn the process should propagate and clean up the pipes that were
+    already created.
+    """
+    with pytest.raises(FileNotFoundError):
+        await open_process(
+            [os.path.join(os.getcwd(), "nonexistent-anyio-test-executable")]
+        )
```
tapetersen, line 557: It doesn't check for any pipe-cleanup as mentioned but that may be implicit (if so just resolve)

tapetersen (commented):
Looked through and couldn't see any obvious bugs but it's a lot of fiddly code.

src/anyio/_backends/_trio.py
graingert: This is certainly missing the checkpoint

tapetersen:
For what it's worth I at least got the `Process.wait()` waiting on streams and not just the process fixed upstream https://github.com/python/cpython/issues/119710 . Looks like it may be backported to 3.13-

(We should fix it work-around it anyway but it al least confirms it as a bug and not intended behavior).

PR #1238, opened 2026-07-23
Add rsloop test

egorsmkv:


egorsmkv:
@agronholm hey

i can't reproduce failing for that test case locally on macOS

egorsmkv:
Probably found a potential sporadic bug in rsloop

PR #1240, opened 2026-07-25
Fix missing TASK_STATUS_IGNORED import in the return_handle start() example

cognis-digital:
The `return_handle=True` example under "Getting a task handle from `start()`" in
`docs/tasks.rst` uses `TASK_STATUS_IGNORED` as the default value for the `task_status`
parameter, but its import line only imports `create_task_group`, `run`, `sleep` and
`TaskStatus`. Since default argument values are evaluated at function-definition time,
copying the example as-is fails immediately with:

```
NameError: name 'TASK_STATUS_IGNORED' is not defined
```

The preceding `start()` example in the same file (and the example in
`docs/threads.rst`) already import `TASK_STATUS_IGNORED` from `anyio`, so this just
brings the one example in line. `TASK_STATUS_IGNORED` is added first in the import to
match the ordering used by the sibling example.

Verified the corrected example runs to completion and prints `Service started on port
5000`.

### DIFF
--- a/docs/tasks.rst
+++ b/docs/tasks.rst
@@ -187,7 +187,7 @@ By default, :meth:`TaskGroup.start() <.abc.TaskGroup.start>` returns the value p
 mode, the method returns a :class:`~.TaskHandle` whose :attr:`~.TaskHandle.start_value`
 property contains the value passed to ``task_status.started()``::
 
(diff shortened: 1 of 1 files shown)


agronholm (approved):
Thanks!

PR #1241, opened 2026-07-25
Added `StapledObjectStream.send_nowait` method

davidbrochart:

## Changes

This PR adds the `send_nowait` method to `StapledObjectStream`, delegating to the underlying send stream's `send_nowait` if it implements it.

## Checklist

features

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/src/anyio/streams/stapled.py
+++ b/src/anyio/streams/stapled.py
@@ -85,6 +85,9 @@ class StapledObjectStream(Generic[T_Item], ObjectStream[T_Item]):
     async def send(self, item: T_Item) -> None:
         await self.send_stream.send(item)
 
+    def send_nowait(self, item: T_Item) -> None:
+        self.send_stream.send_nowait(item)  # type: ignore[attr-defined]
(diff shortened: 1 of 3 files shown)


src/anyio/streams/stapled.py
```diff
@@ -85,6 +85,9 @@ async def receive(self) -> T_Item:
     async def send(self, item: T_Item) -> None:
         await self.send_stream.send(item)
 
+    def send_nowait(self, item: T_Item) -> None:
+        self.send_stream.send_nowait(item)  # type: ignore[attr-defined]
```
agronholm: This will raise a confusing `AttributeError` if the wrapped send stream does not implement send_nowait. It should be caught and a `NotImplementedError` raised instead, with a proper explanation attached. Maybe chain the `AttributeError` there too.

agronholm (changes requested):

agronholm (approved):
LGTM now!

Issue #1239, opened 2026-07-24
`BrokenResourceError` is the wrong error for a temporary Windows UDP reset


graingert:
### Things to check first

- [x] I have searched the existing issues and didn't find my bug already reported there

- [x] I have checked that my bug is still present in the latest release

### AnyIO version

4.14.2

### Python version

3.14

### What happened?

On Windows, sending a UDP datagram to an address with nothing bound to it can cause a
later receive on the same socket to observe `WSAECONNRESET` (`WinError 10054`).

however after this error

- the socket still works
- a later `receive()` can still time out normally
- once a real peer starts sending, the same socket can still receive datagrams

That means this is a temporary per-datagram condition, not a terminal resource
failure. Raising `anyio.BrokenResourceError` is therefore misleading, because
`BrokenResourceError` implies the resource is no longer usable.

### How can we reproduce the bug?

```python
import socket

import anyio
from anyio.abc import SocketAttribute


async def main() -> None:
    errors = []
    async with await anyio.create_udp_socket(
        family=socket.AF_INET, local_host="127.0.0.1"
    ) as udp:
        # An address on loopback with nothing bound to it.
        probe = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
        probe.bind(("127.0.0.1", 0))
        dead_host, dead_port = probe.getsockname()
        probe.close()

        # Provokes an ICMP port-unreachable back to `udp` on Windows.
        await udp.sendto(b"ping", dead_host, dead_port)

        # Windows + trio:    raises anyio.BrokenResourceError (WinError 10054),
        #                    even though `udp` is still perfectly usable.
        # Windows + asyncio: no error; this times out waiting for real data.
        try:
            with anyio.fail_after(2):
                await udp.receive()
        except TimeoutError:
            print("timeout out correctly 1")
        except anyio.BrokenResourceError as e:
            errors.append(e)
        else:
            errors.append(AssertionError("did not timeout"))

        try:
            with anyio.fail_after(2):
                await udp.receive()
        except TimeoutError:
            print("timeout out correctly 2")
        except anyio.BrokenResourceError as e:
            errors.append(e)
        else:
            errors.append(AssertionError("did not timeout"))

        local_host, local_port = udp.extra(SocketAttribute.local_address)

        server = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
        try:
            server.bind((dead_host, dead_port))
            server.sendto(b"pong", (local_host, local_port))
        finally:
[...]

PR #1243, opened 2026-07-25
Fix asyncio UDP transport error delivery

deepakganesh78:

## Changes

Addresses #1239.

Queue errors reported by asyncio datagram transports alongside received packets. This wakes blocked UDP receivers, preserves the original transport exception as the `BrokenResourceError` cause, and allows later datagrams to be received from the same socket. The queue also preserves callback ordering when an error and packet arrive before the receiver resumes.

Regression coverage exercises both unconnected and connected UDP sockets and verifies that each socket remains usable after the error.

Validation:
- `python -m pytest tests/test_sockets.py::TestUDPSocket::test_receive_transport_error tests/test_sockets.py::TestConnectedUDPSocket::test_receive_transport_error -q` (8 passed, 12 skipped by the existing backend/platform matrix)
- python -m pytest tests/test_sockets.py::TestUDPSocket tests/test_sockets.py::TestConnectedUDPSocket -q (116 passed, 194 skipped by the existing backend/platform matrix)
- `SKIP=mypy pre-commit run --files src/anyio/_backends/_asyncio.py tests/test_sockets.py docs/versionhistory.rst` (passed)
- The unskipped mypy hook was also attempted; it reports 69 existing Windows platform/type-stub errors in unrelated lines and files.

## Checklist

features (no API documentation change is needed)

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -1286,7 +1286,7 @@ class StreamProtocol(asyncio.Protocol):
 
 
 class DatagramProtocol(asyncio.DatagramProtocol):
-    read_queue: deque[tuple[bytes, IPSockAddrType]]
+    read_queue: deque[tuple[bytes, IPSockAddrType] | Exception]
     read_event: asyncio.Event
     write_event: asyncio.Event
     closed_event: asyncio.Event
@@ -1311,6 +1311,8 @@ class DatagramProtocol(asyncio.DatagramProtocol):
 
     def error_received(self, exc: Exception) -> None:
         self.exception = exc
+        self.read_queue.append(exc)
+        self.read_event.set()
 
     def pause_writing(self) -> None:
         self.write_event.clear()
@@ -1700,12 +1702,17 @@ class UDPSocket(abc.UDPSocket):
                 await self._protocol.read_event.wait()
 
             try:
-                return self._protocol.read_queue.popleft()
+                packet = self._protocol.read_queue.popleft()
             except IndexError:
                 if self._closed:
                     raise ClosedResourceError from None
                 else:
                     raise BrokenResourceError from None
+
+            if isinstance(packet, Exception):
+                raise BrokenResourceError from packet
+
+            return packet
 
     async def send(self, item: UDPPacketType) -> None:
         with self._send_guard:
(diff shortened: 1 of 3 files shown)


tests/test_sockets.py
```diff
@@ -1891,6 +1916,87 @@ async def test_from_socket_wrong_socket_type(
 @pytest.mark.network
 @pytest.mark.usefixtures("check_asyncio_bug")
 class TestConnectedUDPSocket:
+    async def test_receive_transport_error(self, anyio_backend_name: str) -> None:
+        if anyio_backend_name != "asyncio":
+            pytest.skip("asyncio-specific transport behavior")
+
```
agronholm, line 1922: Use this style if you want to only run this test on asyncio.
```suggestion
    @pytest.mark.parametrize("anyio_backend", asyncio_params)
    async def test_receive_transport_error(self) -> None:
```

tests/test_sockets.py
```diff
@@ -1710,6 +1710,31 @@ async def handle(stream: SocketStream) -> None:
 @pytest.mark.network
 @pytest.mark.usefixtures("check_asyncio_bug")
 class TestUDPSocket:
+    async def test_receive_transport_error(self, anyio_backend_name: str) -> None:
```
agronholm, line 1713: I can't help but to wonder if this test should not be asyncio specific. I'd like to be sure it works the same way across all backends.

agronholm: You could cause this same error by sending to a local unbound port, right?

src/anyio/_backends/_asyncio.py
```diff
@@ -1319,6 +1321,19 @@ def resume_writing(self) -> None:
         self.write_event.set()
 
 
+def _rearm_proactor_udp_transport(transport: asyncio.DatagramTransport) -> None:
+    # CPython's Proactor transport leaves _read_fut unset after a receive error.
```
graingert, line 1325: is this due to https://github.com/python/cpython/issues/127057 ?

graingert: @agronholm this is fixed upstream in cpython now, how do you want to proceed?

agronholm: We still need a fallback for existing Python's right? We can make this conditional on Python version I suppose?

PR #1245, opened 2026-07-26
Fixed RuntimeError in worker thread when the event loop closes during result reporting

shleder:

## Changes

Fixes #1245. 

Fix the asyncio worker-thread shutdown race where the loop can close after
`is_closed()` returns false but before `call_soon_threadsafe()` delivers the result.
The undeliverable result is dropped, matching the selector-thread precedent.
Deterministic regressions cover the closed-loop race, unrelated `RuntimeError`
suppression at that call site, no undeliverable future publication, and normal result
delivery.

## Checklist

features

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -1045,9 +1045,15 @@ class WorkerThread(Thread):
                         del threadlocals.current_cancel_scope
 
                     if not self.loop.is_closed():
-                        self.loop.call_soon_threadsafe(
-                            self._report_result, future, result, exception
-                        )
+                        try:
+                            self.loop.call_soon_threadsafe(
+                                self._report_result, future, result, exception
+                            )
+                        except RuntimeError:
+                            # The loop was closed after the is_closed() check
+                            # above (a race with event loop shutdown). The
+                            # result can no longer be delivered, so drop it.
+                            pass
 
                     del result, exception
 
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -1044,16 +1044,13 @@ class WorkerThread(Thread):
                     finally:
                         del threadlocals.current_cancel_scope
 
-                    if not self.loop.is_closed():
-                        try:
-                            self.loop.call_soon_threadsafe(
-                                self._report_result, future, result, exception
-                            )
-                        except RuntimeError:
-                            # The loop was closed after the is_closed() check
-                            # above (a race with event loop shutdown). The
-                            # result can no longer be delivered, so drop it.
-                            pass
+                    try:
+                        self.loop.call_soon_threadsafe(
+                            self._report_result, future, result, exception
+                        )
+                    except RuntimeError:
+                        if not self.loop.is_closed():
+                            raise
 
                     del result, exception
 
--- a/tests/test_to_thread.py
+++ b/tests/test_to_thread.py
@@ -6,13 +6,11 @@ import sys
 import threading
 import time
 import weakref
-from collections import deque
 from collections.abc import Callable
 from concurrent.futures import Future, ThreadPoolExecutor
-from contextvars import Context, ContextVar
+from contextvars import ContextVar
 from functools import partial
-from queue import Queue
-from typing import Any, NoReturn, cast
+from typing import Any, NoReturn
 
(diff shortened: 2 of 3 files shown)


agronholm:
You erased the PR template, didn't provide a changelog entry or tests. Rejected.

shleder:
Understood — sorry about that. I now have a deterministic regression test, the required changelog entry, and a PR body that preserves the official template. Would you be open to a fresh, fully compliant PR, or would you prefer that I leave this fix alone?

agronholm:
Restore the form on the PR template, add a changelog entry and regression test(s), and I will reopen this.

shleder:
The requested official PR template, UNRELEASED changelog entry, and deterministic regression tests are now present on the existing branch. Please reopen this PR when convenient.

agronholm:
The tests you wrote don't exercise any AnyIO APIs. If the problem isn't reproducible via the official API, what is the point here then?

shleder:
Updated in cb059226271a0494d7b771bdb8fad25e645b20eb.

The regression now exercises the official `anyio.to_thread.run_sync()` API on a real running asyncio event loop.

It deterministically forces the open-loop check followed by the `call_soon_threadsafe()` shutdown race, confirms that the loop remains usable, and verifies that a subsequent public-API result is delivered normally.

Validation:
- focused public-API regression passed;
- Ruff check passed;
- Ruff format check passed;
- `git diff --check` passed.

src/anyio/_backends/_asyncio.py
```diff
@@ -1045,9 +1045,15 @@ def run(self) -> None:
...
-                        )
+                        try:
+                            self.loop.call_soon_threadsafe(
+                                self._report_result, future, result, exception
+                            )
+                        except RuntimeError:
+                            # The loop was closed after the is_closed() check
+                            # above (a race with event loop shutdown). The
+                            # result can no longer be delivered, so drop it.
+                            pass
```
tapetersen:
```suggestion
                # There will always be a race with a closing loop.
                # If so ignore as nothing needs the result.
                try:
                    self.loop.call_soon_threadsafe(
                        self._report_result, future, result, exception
                    )
                except RuntimeError:
                    if not self.loop.is_closed():
                        raise
```

Maybe even skip the check/re-raise in the `except`

tapetersen: This will make the behavior a lot easier to test as well as there is no narrow race window to hit.

tapetersen: Beware I've not tested above code but the logic should be clear.

shleder: Applied in 4a415b0. The preliminary closed-loop check was removed, `call_soon_threadsafe()` is now attempted directly, and `RuntimeError` is re-raised when the loop is not closed. The focused public-API race tests and Ruff checks passed.

tests/test_to_thread.py
```diff
@@ -415,3 +418,123 @@ async def main() -> weakref.ref[object]:
...
+    loop = asyncio.get_running_loop()
+    real_call_soon_threadsafe = loop.call_soon_threadsafe
+    delivery_attempted = Event()
+    thread_waiting = Event()
+    finish_thread = threading.Event()
+
+    is_closed_returnValue = False
+
+    def is_closed() -> bool:
+        return is_closed_returnValue
```
tapetersen: Now that the fix isn't dependent on a tight race these tests should be possible to make a lot smaller and cleaner. The first should probably be doable without monkeypatching at all.

agronholm:
Why did you close the PR? Was this not going to fix a problem?

shleder:
Sorry — that one's on me. I closed it during a cleanup of stale branches and forks and wrongly lumped this one in; the fix was and is real. The fork got deleted in the same sweep, so I restored the branch from the PR ref and re-submitted it as #1303 — same fix, template and changelog intact (changelog link now points at the new PR).

For what it's worth, 4a415b0 on that branch already implements the restructure @tapetersen suggested on 29 July: `call_soon_threadsafe()` is called directly, and a `RuntimeError` is swallowed only when the loop is actually closed — so a closing loop drops the undeliverable result instead of racing `is_closed()`. The regression test exercises it through the official `anyio.to_thread.run_sync()` API.

PR #1246, opened 2026-07-27
Fix UDP socket close hangs on Windows

### LINKED ISSUE #1237: UDPSocket.aclose() hangs on Windows when a send is in flight (regression from #1147)
### Summary

Since #1147 (4.14.0), `UDPSocket.aclose()` / `ConnectedUDPSocket.aclose()` await `self._protocol.closed_event`. On `ProactorEventLoop` there is a path where `connection_lost()` is never called, so that event is never set and `aclose()` waits forever.

It triggers whenever a datagram write is still in flight at close time — i.e. the ordinary "send a metric, then close" pattern.

### Repro

```python
import anyio


async def main() -> None:
    sock = await anyio.create_connected_udp_socket("127.0.0.1", 9999)
    await sock.send(b"x")
    await sock.aclose()   # never returns on Windows


anyio.run(main)
```

Nothing needs to be listening on the port.

Full reproducer with a CI matrix: https://github.com/graingert/anyio-udp-aclose-hang — the Windows jobs are red on purpose. Observed on `windows-latest`, CPython 3.12 and 3.13, anyio 4.14.2, with a `faulthandler` watchdog to turn the hang into a stack dump:

```
sent a datagram; calling aclose() ...
Timeout (0:00:30)!
  File "...\asyncio\windows_events.py", line 774 in _poll
  File "...\asyncio\windows_events.py", line 445 in select
  File "...\asyncio\base_events.py", line 1961 in _run_once
  ...
  File "...\anyio\_core\_eventloop.py", line 83 in run
  File "...\repro_hang.py", line 26 in <module>
```

The same scripts pass on ubuntu-latest, where `close()` schedules `_call_connection_lost` directly.

### Cause

`aclose()` waits for `connection_lost()`:

```python
    self._closed = True
    if not self._transport.is_closing():
        self._transport.close()

    await self._protocol.closed_event.wait()
```

`_ProactorBasePipeTransport.close()` increments `_conn_lost`, and skips scheduling `_call_connection_lost` because a write is outstanding:

```python
def close(self):
    if self._closing:
        return
    self._closing = True
    self._conn_lost += 1
    if not self._buffer and self._write_fut is None:
        self._loop.call_soon(self._call_connection_lost, None)
```

Normally the write's done-callback would pick that up, but `_ProactorDatagramTransport._loop_writing()` returns on its first line, because `close()` has just incremented `_conn_lost`:

```python
def _loop_writing(self, fut=None):
    try:
        if self._conn_lost:
            return          # <-- bails out here

        ...
        if not self._buffer or (self._conn_lost and self._address):
            # The connection has been closed
            if self._closing:
                self._loop.call_soon(self._call_connection_lost, None)   # <-- never reached
            return
```

So `_call_connection_lost` is scheduled by neither path, `connection_lost()` is never called, `closed_event` stays unset, and the `await` never returns.

### Impact

Before 4.14 this path was a missed close — the `ResourceWarning` that #1147 set out to fix. From 4.14 the same path deadlocks, so shutdown code that closes a datagram socket after sending on it hangs instead. It surfaced in [anycorn](https://github.com/davidbrochart/anycorn), whose statsd logger sends a metric then closes its socket: after 4.11.0 → 4.14.2 its `windows-latest` jobs went from ~30s to hanging indefinitely, on 3.10 through 3.14.

### Workarounds

`aclose_forcefully()` cancels the wait, so the hang goes away — but `connection_lost()` still never runs, so the transport keeps its socket and the `ResourceWarning` returns (`repro_forcefully.py` reports `fileno = 556` after closing).

What does work is bypassing the wrapper on asyncio and driving the transport directly, so `abort()` is reachable (`repro_workaround.py`, green on Windows):

```python
transport.close()
await asyncio.sleep(0)
transport.abort()
    await protocol.closed.wait()
```

### Possible fix

`SocketStream.aclose()` already handles the equivalent case by following `close()` with `await sleep(0)` and then `abort()`. That works because `_force_close()`'s early-return guard tests `_called_connection_lost` rather than `_conn_lost`, so unlike `close()` it always schedules `_call_connection_lost`. The UDP wrappers have no equivalent step.

There may also be a CPython-side argument that `_loop_writing()`'s `_conn_lost` early return should not pre-empt the `_closing` handling below it.

Guflly:

## Changes

Fixes #1237.

On Windows Proactor event loops, closing a UDP socket while a datagram send is in flight can leave `connection_lost()` unscheduled and make `aclose()` wait indefinitely. Follow the transport close with an event-loop turn and `abort()`, matching the existing stream cleanup path, so the protocol close event is always delivered.

The regression tests cover connected and unconnected UDP sockets on Windows.

## Checklist

features

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -17,6 +17,8 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
   module name. (The default name for a task spawned with ``TaskGroup.start_soon`` or
   ``TaskGroup.start`` typically includes the module name.)
   (`#1234 <https://github.com/agronholm/anyio/pull/1234>`_; PR by @gschaffner)
+- Fixed UDP socket closing hanging on Windows if a datagram send was still in flight
+  (`#1237 <https://github.com/agronholm/anyio/issues/1237>`_; PR by @Guflly)
 - Fixed free-threading compatibility issues arising from the fact that on Python 3.14
   free-threading builds, newly created threads inherit the current context by default,
   causing AnyIO to behave erroneously in relation to ``start_blocking_portal()`` and
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -1403,8 +1403,11 @@ class SocketStream(abc.SocketStream):
                 pass
 
             self._transport.close()
-            await sleep(0)
-            self._transport.abort()
+            try:
+                await sleep(0)
+            finally:
+                self._transport.abort()
+                await AsyncIOBackend.cancel_shielded_checkpoint()
 
 
 class _RawSocketMixin:
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -1687,8 +1687,11 @@ class UDPSocket(abc.UDPSocket):
         self._closed = True
         if not self._transport.is_closing():
             self._transport.close()
-            await sleep(0)
-            self._transport.abort()
+            try:
+                await sleep(0)
+            finally:
+                self._transport.abort()
+                await AsyncIOBackend.cancel_shielded_checkpoint()
 
         await self._protocol.closed_event.wait()
 
@@ -1739,8 +1742,11 @@ class ConnectedUDPSocket(abc.ConnectedUDPSocket):
         self._closed = True
         if not self._transport.is_closing():
             self._transport.close()
-            await sleep(0)
-            self._transport.abort()
+            try:
+                await sleep(0)
+            finally:
+                self._transport.abort()
+                await AsyncIOBackend.cancel_shielded_checkpoint()
 
         await self._protocol.closed_event.wait()
 
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -1687,6 +1687,8 @@ class UDPSocket(abc.UDPSocket):
         self._closed = True
         if not self._transport.is_closing():
             self._transport.close()
+            await sleep(0)
+            self._transport.abort()
 
         await self._protocol.closed_event.wait()
 
@@ -1737,6 +1739,8 @@ class ConnectedUDPSocket(abc.ConnectedUDPSocket):
         self._closed = True
         if not self._transport.is_closing():
             self._transport.close()
+            await sleep(0)
+            self._transport.abort()
 
         await self._protocol.closed_event.wait()
 
--- a/tests/test_sockets.py
+++ b/tests/test_sockets.py
@@ -34,6 +34,7 @@ from _pytest.tmpdir import TempPathFactory
 from pytest import FixtureRequest
 from pytest_mock.plugin import MockerFixture
 
+import anyio
 from anyio import (
     BrokenResourceError,
     BusyResourceError,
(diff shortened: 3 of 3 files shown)


graingert:
Can you add a test that:

```python
sock = udp.extra(SocketAttribute.raw_socket)
await anyio.aclose_forcefully(udp)
assert sock.fileno() == -1
```

Guflly:
Added. The test initially left the fd open, so abort now runs in finally and gets a shielded checkpoint. The focused UDP tests pass.

graingert:
Do SocketStreams have the same problem with aclose_forcefully?

Guflly:
Yes, when a send is in flight. I added the same cleanup pattern and a TCP regression test.

tests/test_sockets.py
```diff
@@ -518,6 +519,14 @@ async def test_send_after_close(self, server_addr: tuple[str, int]) -> None:
         with pytest.raises(ClosedResourceError):
             await stream.send(b"foo")
 
+    @pytest.mark.parametrize("anyio_backend", asyncio_params)
```
graingert, line 522: is there a reason to limit this to asyncio? Surely we should have the same behaviour on asyncio and trio


```suggestion
```

tests/test_sockets.py
```diff
@@ -1743,6 +1752,17 @@ async def test_aclose_waits_for_fd_release(
             udp = await UDPSocket.from_socket(sock)
             await udp.aclose()
 
+    @pytest.mark.skipif(sys.platform != "win32", reason="Windows only")
+    @pytest.mark.parametrize("anyio_backend", asyncio_params)
```
graingert, line 1756: is there a reason to limit this to asyncio on windows? Surely this should pass on non windows and trio

```suggestion
```

tests/test_sockets.py
```diff
@@ -1914,6 +1934,14 @@ async def test_aclose_waits_for_fd_release(
         finally:
             peer.close()
 
+    @pytest.mark.skipif(sys.platform != "win32", reason="Windows only")
+    @pytest.mark.parametrize("anyio_backend", asyncio_params)
```
graingert, line 1938:
```suggestion
```
is there a reason to limit this to asyncio on windows? Surely this should pass on non windows and trio

docs/versionhistory.rst
```diff
@@ -17,6 +17,8 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
   module name. (The default name for a task spawned with ``TaskGroup.start_soon`` or
   ``TaskGroup.start`` typically includes the module name.)
   (`#1234 <https://github.com/agronholm/anyio/pull/1234>`_; PR by @gschaffner)
+- Fixed UDP socket closing hanging on Windows if a datagram send was still in flight
```
graingert, line 20: probably worth something in the news about fixing TCP's aclose_forcefully

src/anyio/_backends/_asyncio.py
```diff
@@ -1687,6 +1690,11 @@ async def aclose(self) -> None:
         self._closed = True
         if not self._transport.is_closing():
```
graingert, line 1691: I think this is wrong, if aclose_forcefully is called concurrently the second one will return before the socket is closed

src/anyio/_backends/_asyncio.py
```diff
@@ -1737,6 +1745,11 @@ async def aclose(self) -> None:
...
```
graingert, line 1746: I think this is wrong, if aclose_forcefully is called concurrently the second one will return before the socket is closed

src/anyio/_backends/_asyncio.py
```diff
@@ -1403,8 +1403,11 @@ async def aclose(self) -> None:
                 pass
 
             self._transport.close()
-            await sleep(0)
-            self._transport.abort()
+            try:
+                await sleep(0)
+            finally:
+                self._transport.abort()
+                await AsyncIOBackend.cancel_shielded_checkpoint()
```
graingert, line 1410: same here with concurrent aclose_forcefully I think we want this code:

```python
        self._closed = True

        if not self._transport.is_closing():
            try:
                self._transport.write_eof()
            except OSError:
                pass

            self._transport.close()

        try:
            await sleep(0)
        finally:
            self._transport.abort()
            await AsyncIOBackend.cancel_shielded_checkpoint()
```

graingert (changes requested):

PR #1247, opened 2026-07-27
fix: do not prefer IPv6 when host has only link-local IPv6

### LINKED ISSUE #1230: connect_tcp blindly prioritizes IPv6 even when the system has no globally routable IPv6 address
# Bug: `connect_tcp` blindly prioritizes IPv6 even when the system has no globally routable IPv6 address

## Description

`anyio.connect_tcp()` reorders `getaddrinfo` results to always try IPv6 first, even when the system only has a link-local IPv6 address (`fe80::/10`) and no globally routable IPv6 address. This causes unnecessary connection delays and failures on networks where IPv6 is not actually available.

## Root Cause

In `anyio/_core/_sockets.py`, the `connect_tcp()` function explicitly reorders addresses to put IPv6 first:

```python
# Organize the list so that the first address is an IPv6 address (if available)
# and the second one is an IPv4 addresses. The rest can be in whatever order.
v6_found = v4_found = False
for af, *_, sa in gai_res:
    if af == socket.AF_INET6 and not v6_found:
        v6_found = True
        target_addrs.insert(0, (af, sa[0]))      # ← IPv6 forced to front
    elif af == socket.AF_INET and not v4_found and v6_found:
        v4_found = True
        target_addrs.insert(1, (af, sa[0]))      # ← IPv4 pushed to second
```

The problem: `getaddrinfo` with `AI_ADDRCONFIG` returns IPv6 results when the system has **any** IPv6 address, including link-local (`fe80::/10`). But link-local addresses cannot reach the internet. When `connect_tcp` tries IPv6 first, it immediately fails with "Network is unreachable", wasting time before falling back to IPv4.

## Reproduction

```python
import socket
import asyncio
import anyio

async def test():
    # On a system with only link-local IPv6 (no global IPv6):
    # 1. getaddrinfo returns both IPv4 and IPv6
    infos = socket.getaddrinfo('api.telegram.org', 443, socket.AF_UNSPEC, socket.SOCK_STREAM)
    print('getaddrinfo order:', ['IPv6' if i[0] == socket.AF_INET6 else 'IPv4' for i in infos])
    # → ['IPv4', 'IPv6'] or ['IPv6', 'IPv4'] depending on system

    # 2. anyio.connect_tcp reorders to put IPv6 first
    # 3. IPv6 connection fails immediately (no route)
    # 4. Falls back to IPv4 after happy_eyeballs_delay (0.25s default)
    
    try:
        stream = await anyio.connect_tcp('api.telegram.org', 443)
        print(f'Connected to: {stream.extra(anyio.abc.SocketAttribute.remote_address)}')
    except Exception as e:
        print(f'Error: {e}')

asyncio.run(test())
```

## Expected Behavior

`connect_tcp()` should check whether the system has a **globally routable** IPv6 address before prioritizing IPv6. If only link-local or loopback IPv6 addresses are available, IPv4 should be tried first.

## Suggested Fix

Before reordering, check if the system has a global IPv6 address:

```python
def _has_global_ipv6() -> bool:
    """Check if the system has any globally routable IPv6 address."""
    try:
        for iface in socket.if_nameindex():
            addrs = socket.getaddrinfo(iface[1], None, socket.AF_INET6)
            for family, *_, addr in addrs:
                ip = ipaddress.IPv6Address(addr[0].split('%')[0])
                if not ip.is_link_local and not ip.is_loopback:
                    return True
    except (OSError, ValueError):
        pass
    return False

# In connect_tcp():
has_global_v6 = _has_global_ipv6()
for af, *_, sa in gai_res:
    if af == socket.AF_INET6 and not v6_found and has_global_v6:
        v6_found = True
        target_addrs.insert(0, (af, sa[0]))
    elif af == socket.AF_INET and not v4_found:
        v4_found = True
        # Put IPv4 first if no global IPv6
        idx = 0 if not has_global_v6 else 1
        target_addrs.insert(idx, (af, sa[0]))
```

## Environment

- anyio version: 4.13.0
- Python: 3.11/3.13
- OS: Linux (containerized environments with only link-local IPv6)
- glibc: 2.41

## Related

- This affects HTTP clients (httpx → httpcore → anyio) connecting to dual-stack services
- The `happy_eyeballs_delay` (0.25s) mitigates but doesn't eliminate the problem
- Container/orchestration platforms (Docker, Kubernetes, lightweight VMs) commonly have only link-local IPv6

MohammedAnasNathani:
## Summary

Fixes #1230.

On hosts whose only IPv6 address is link-local (`fe80::/10`), `getaddrinfo` with `AI_ADDRCONFIG` still returns AAAA records for dual-stack names. `connect_tcp` then **forced IPv6 first**, so the first Happy Eyeballs attempt often failed with `Network is unreachable` before IPv4 was tried.

## Approach

1. **`_host_has_routable_ipv6()`** — UDP `connect` to a well-known global address (no packets sent) and inspect the kernel-chosen source. Returns false for link-local / loopback / unspecified / no IPv6.
2. **`_order_happy_eyeballs_targets(...)`** — pure reorder helper:
   - `prefer_ipv6=True` → same as before (IPv6 first, IPv4 second)
   - `prefer_ipv6=False` → keep `getaddrinfo` encounter order (IPv4 is not delayed behind a doomed AAAA)
3. **`connect_tcp`** uses (1) to decide (2).

Happy Eyeballs still races multiple addresses; this only stops *blindly* promoting IPv6 when the host cannot source a routable IPv6 address.

## Tests

- Prefer-IPv6 reorder (IPv4-then-IPv6 input → IPv6 first)
- No-prefer reorder keeps encounter order (#1230)
- `_host_has_routable_ipv6` false when `has_ipv6` is false
- `_host_has_routable_ipv6` false when source is `fe80::1`

## Checklist

- [x] Tests which would fail without the patch
- [x] Changelog entry in `docs/versionhistory.rst`
- [ ] Docs — behavior note only in changelog (API unchanged)

## LLM use

AI assistant helped structure helpers/tests and PR text; I verified the #1230 root cause against the Happy Eyeballs reorder loop and the link-local probe semantics.

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -5,6 +5,11 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
 **UNRELEASED**
 
+- Fixed ``connect_tcp()`` preferring IPv6 on hosts that only have a link-local
+  IPv6 address, which caused an unnecessary failed attempt (often "Network is
+  unreachable") before falling back to IPv4
+  (`#1230 <https://github.com/agronholm/anyio/issues/1230>`_; PR by
+  @MohammedAnasNathani)
 - Added ``StapledObjectStream.send_nowait()`` that delegates to the underlying
   ``ObjectSendStream``, if it implements it
   (`#1241 <https://github.com/agronholm/anyio/pull/1241>`_; PR by @davidbrochart)
--- a/src/anyio/_core/_sockets.py
+++ b/src/anyio/_core/_sockets.py
@@ -107,7 +107,7 @@ def _ipv6_addr_is_locally_usable(host: str) -> bool:
 
 def _order_happy_eyeballs_targets(
     gai_res: list[tuple[Any, ...]], *, has_global_ipv6: bool
-) -> list[tuple[int, str]]:
+) -> list[tuple[AddressFamily, str]]:
     """Reorder ``getaddrinfo`` results for Happy Eyeballs (RFC 6555).
 
     Promotes the first *usable* IPv6 address to the front and the first IPv4
@@ -120,21 +120,22 @@ def _order_happy_eyeballs_targets(
       working IPv4 is not delayed (#1230).
     """
     v6_found = v4_found = False
-    target_addrs: list[tuple[int, str]] = []
+    target_addrs: list[tuple[AddressFamily, str]] = []
     for af, *_, sa in gai_res:
+        family = cast(AddressFamily, af)
         host = sa[0]
-        if af == socket.AF_INET6 and not v6_found:
+        if family == socket.AF_INET6 and not v6_found:
             if has_global_ipv6 or _ipv6_addr_is_locally_usable(host):
                 v6_found = True
-                target_addrs.insert(0, (af, host))
+                target_addrs.insert(0, (family, host))
             else:
                 # Global AAAA on a link-local-only host — do not promote.
-                target_addrs.append((af, host))
-        elif af == socket.AF_INET and not v4_found and v6_found:
+                target_addrs.append((family, host))
+        elif family == socket.AF_INET and not v4_found and v6_found:
             v4_found = True
-            target_addrs.insert(1, (af, host))
+            target_addrs.insert(1, (family, host))
         else:
-            target_addrs.append((af, host))
+            target_addrs.append((family, host))
     return target_addrs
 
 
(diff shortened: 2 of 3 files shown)


agronholm:
There is absolutely no way I will allow a change that automatically connects to Google's public address just to check if IPv6 should be preferred.

agronholm:
Should I take your silence as the lack of any intention to try a better approach?

PR #1251, opened 2026-07-29
fix: do not swallow concurrent native Task.cancel on CancelScope exit (#1214)

### LINKED ISSUE #1214: CancelScope can swallow a concurrent native asyncio.Task.cancel() when the scope is also cancelled
### Things to check first

- [x] I searched the existing issues and did not find this exact bug already reported.
- [x] I verified that the bug is present in the latest release.

### AnyIO version

4.14.1 (also reproduced with 4.13.0)

### Python version

3.12.10

### What happened?

We came across this while investigating a production hang in an asyncio service that streams HTTP responses. When a new user event supersedes an in-flight response, the service calls `Task.cancel()` and waits for the streaming task to stop. In this case `Task.cancel()` was accepted, but the underlying request coroutine returned normally instead of propagating `CancelledError`. That left the higher-level cleanup path stuck waiting for the task to settle.

The unusual clue was the task state: `cancelled()` was `False`, while `cancelling()` was still `1`. In other words, asyncio still recorded a cancellation request, but no cancellation exception was propagating anymore.

Following the cancellation down through the HTTP networking stack led us to `anyio.connect_tcp()`. The failure happened during the early connection window where a successful Happy Eyeballs attempt cancels the task group containing the other connection attempts. That suggested that AnyIO's internal housekeeping cancellation and our caller's cancellation might be landing almost simultaneously. The standalone reproduction below isolates that race without HTTP or any third-party libraries.

In plain terms, the sequence appears to be:

1. A child cancels its AnyIO task group, which requests cancellation of the host task.
2. Before the host task processes that request, an unrelated caller invokes native `Task.cancel()`. It returns `True`, so the caller reasonably expects cancellation to propagate.
3. In this schedule, asyncio delivers the AnyIO-tagged `CancelledError`. `CancelScope.__exit__()` recognizes it as the scope's own cancellation, suppresses it, and calls `uncancel()` for the request AnyIO initiated.
4. The external cancellation count remains, but there is no second `CancelledError` waiting to be delivered. The task therefore returns normally with `cancelled() == False` and `cancelling() == 1`.

I expected the cancel scope to consume only the cancellation request it initiated. A separate native cancellation should remain observable, so awaiting the task should raise `CancelledError` and `task.cancelled()` should become `True`.

This seems related to the native-cancellation contract discussed in #374, but I could not find an issue covering this concurrent-cancellation case.

This example uses only public asyncio and AnyIO APIs:

```python
import asyncio

import anyio


async def operation(
    attempt_started: asyncio.Event,
    connection_won: asyncio.Event,
) -> None:
        async def connect_attempt() -> None:
            attempt_started.set()
            await connection_won.wait()

            # Same pattern used by anyio.connect_tcp(): the first successful
            # Happy Eyeballs attempt cancels the remaining attempts.

        task_group.start_soon(connect_attempt)
        await anyio.sleep_forever()


async def main() -> None:
    attempt_started = asyncio.Event()
    connection_won = asyncio.Event()
    task = asyncio.create_task(operation(attempt_started, connection_won))
    await attempt_started.wait()

    external_cancel_results: list[bool] = []

    def externally_cancel() -> None:
        external_cancel_results.append(task.cancel("external cancellation"))

    # This queues connect_attempt first. It cancels its AnyIO task group.
    # The native cancellation callback then runs before the host task
    # processes the AnyIO cancellation.
    connection_won.set()
    asyncio.get_running_loop().call_soon(externally_cancel)

    try:
        await task
    except asyncio.CancelledError as exc:
        outcome = f"CancelledError propagated: {exc!r}"
    else:
        outcome = "returned normally"

    print(f"Task.cancel() returned: {external_cancel_results}")
    print(f"Outcome: {outcome}")
    print(f"task.cancelled(): {task.cancelled()}")
    print(f"task.cancelling(): {task.cancelling()}")


asyncio.run(main())
```

Run with:

```console
uv run --no-project --with 'anyio==4.14.1' python anyio_cancel_repro.py
```

Actual output:

```text
Task.cancel() returned: [True]
Outcome: returned normally
task.cancelled(): False
task.cancelling(): 1
```

Expected output:

```text
Task.cancel() returned: [True]
Outcome: CancelledError propagated: CancelledError('external cancellation')
task.cancelled(): True
task.cancelling(): 1
```

MohammedAnasNathani:
## Changes

On the asyncio backend, when a `CancelScope` cancels its host task **and** a concurrent native `Task.cancel()` lands in the same event-loop cycle, `__exit__` could swallow the AnyIO-tagged `CancelledError` after undoing only the scope's own cancel counts. The host then **returned normally** with `task.cancelled() is False` and `task.cancelling() > 0`, so callers awaiting the task never saw cancellation.

This is the Happy Eyeballs / `connect_tcp` pattern from production: a child cancels the task group while an unrelated caller cancels the host task.

### Fix

1. Snapshot `Task.cancelling()` when the scope is entered.
2. On exit, after calling `uncancel()` for each cancel this scope delivered to the host, **do not swallow** if `cancelling()` is still above that baseline — a native cancel that arrived *while the scope was active* must still propagate.
3. Cancels that were already pending *before* enter are ignored for this check (preserves `test_cancel_message_replaced` and related uncancel semantics).

### Checklist

- [x] Tests which would fail without the patch (`TestUncancel.test_child_scope_cancel_does_not_swallow_native_host_cancel`, mirrors the #1214 repro)
- [x] Changelog entry in `docs/versionhistory.rst`
- [x] Existing `TestUncancel` suite green; full `tests/test_taskgroups.py`: 539 passed

### Related

Fixes #1214

<details>
<summary>Verification</summary>

```
# Issue repro (before: returned normally / cancelled=False / cancelling=1)
# After: CancelledError propagates / cancelled=True / cancelling=1

pytest tests/test_taskgroups.py::TestUncancel   # 30 passed
pytest tests/test_taskgroups.py                 # 539 passed, 153 skipped, 4 xfailed
```
</details>

### DIFF
--- a/tests/test_taskgroups.py
+++ b/tests/test_taskgroups.py
@@ -1688,6 +1688,49 @@ class TestUncancel:
         assert task.cancelling() == 1
         task.uncancel()
 
+    async def test_child_scope_cancel_does_not_swallow_native_host_cancel(
+        self,
+    ) -> None:
+        """
+        When a child cancels its task group (Happy Eyeballs style) and a native
+        ``Task.cancel()`` lands on the host in the same cycle, the host must still
+        observe ``CancelledError`` — not return normally with ``cancelling() > 0``.
+
+        Regression test for #1214.
+        """
+        attempt_started = asyncio.Event()
+        connection_won = asyncio.Event()
+
+        async def operation() -> None:
+            async with create_task_group() as task_group:
+
+                async def connect_attempt() -> None:
+                    attempt_started.set()
+                    await connection_won.wait()
+                    # Same pattern as connect_tcp(): winner cancels the group.
+                    task_group.cancel_scope.cancel()
+
+                task_group.start_soon(connect_attempt)
+                await sleep_forever()
+
+        task = asyncio.create_task(operation())
+        await attempt_started.wait()
+
+        external_cancel_accepted: list[bool] = []
+
+        def externally_cancel() -> None:
+            external_cancel_accepted.append(task.cancel("external cancellation"))
+
+        connection_won.set()
+        asyncio.get_running_loop().call_soon(externally_cancel)
+
+        with pytest.raises(asyncio.CancelledError):
+            await task
+
+        assert external_cancel_accepted == [True]
+        assert task.cancelled()
+        assert task.cancelling() >= 1
+
     async def test_cancel_message_replaced(self) -> None:
         task = asyncio.current_task()
         assert task
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -5,6 +5,13 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
 **UNRELEASED**
 
+- Fixed ``CancelScope`` on the asyncio backend swallowing a concurrent native
+  ``Task.cancel()`` when the scope was also cancelled (e.g. Happy Eyeballs
+  winning and cancelling the host task group while the caller cancels the host
+  task). The host task previously returned normally with ``cancelling() > 0``
+  and ``cancelled() is False``; ``CancelledError`` now propagates as expected
+  (`#1214 <https://github.com/agronholm/anyio/issues/1214>`_; PR by
+  @MohammedAnasNathani)
 - Added ``StapledObjectStream.send_nowait()`` that delegates to the underlying
   ``ObjectSendStream``, if it implements it
   (`#1241 <https://github.com/agronholm/anyio/pull/1241>`_; PR by @davidbrochart)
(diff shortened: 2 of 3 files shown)


agronholm:
The test already passes on `master`, so what's going on here?

tests/test_taskgroups.py
```diff
@@ -1689,6 +1689,49 @@ async def test_uncancel_after_scope_and_native_cancel(self) -> None:
...
+        connection_won.set()
+        asyncio.get_running_loop().call_soon(externally_cancel)
+
+        with pytest.raises(asyncio.CancelledError):
+            await task
+
+        assert external_cancel_accepted == [True]
+        assert task.cancelled()
+        assert task.cancelling() >= 1
+
```
agronholm, line 1734: This test is a massively overcomplicated. The bug can be reproduced with 1/5 of the code here.

agronholm:
Closing in favor of the simpler #1330.

PR #1252, opened 2026-07-29
Updated pre-commit modules and adapted to the new Ruff default ruleset

agronholm:


agronholm:
This was made as a separate PR to check for problems caused by the required code changes in the test suite.

PR #1256, opened 2026-07-30
Don't mask AttributeError in StapledObjectStream.send_nowait()

Zuhef:

## Changes

No linked issue — this is a follow-up to #1241, which added
`StapledObjectStream.send_nowait()`.

The `try` block there wraps the call, not just the attribute lookup:

```python
try:
    self.send_stream.send_nowait(item)
except AttributeError as exc:
    raise NotImplementedError(
        f"'send_nowait' method not implemented in {type(self.send_stream)}"
    ) from exc
```

So an `AttributeError` raised *inside* a `send_nowait()` that does exist is caught
and reported as the method not existing. With a send stream that implements
`send_nowait()` but has a bug in it:

```
NotImplementedError: 'send_nowait' method not implemented in <class 'SendStreamWithBug'>
  caused by AttributeError: 'collections.deque' object has no attribute 'appendd'
```

The message is untrue — `hasattr(send_stream, "send_nowait")` is `True` — and it
points at the wrong stream and the wrong problem. The real error is only reachable
through `__cause__`, which is easy to miss.

This looks up the bound method inside the `try` and calls it outside, so only a
genuinely missing method turns into `NotImplementedError` and anything the
implementation itself raises propagates untouched.

## Checklist

features

`tests/streams/test_stapled.py` gains `DummyObjectSendStreamWithBrokenSendNowait`,
whose `send_nowait()` raises `AttributeError`, and a test asserting it reaches the
caller. Without the patch that test fails with the misleading
`NotImplementedError` quoted above. `tests/streams` passes (424) on both backends,
and ruff, ruff-format and mypy are clean on the two changed files.

I left the docs and changelog boxes unchecked deliberately. `send_nowait()` is
still under `**UNRELEASED**`, so a "Fixed" entry would describe a bug that no
release ever had, and I didn't want to rewrite the existing `Added` entry from
#1241. Say the word and I'll add an entry in whichever form you prefer.

### Updating the changelog

```
```

### DIFF
--- a/src/anyio/streams/stapled.py
+++ b/src/anyio/streams/stapled.py
@@ -87,11 +87,13 @@ class StapledObjectStream(ObjectStream[T_Item], Generic[T_Item]):
 
     def send_nowait(self, item: T_Item) -> None:
         try:
-            self.send_stream.send_nowait(item)  # type: ignore[attr-defined]
+            send_nowait = self.send_stream.send_nowait  # type: ignore[attr-defined]
         except AttributeError as exc:
             raise NotImplementedError(
                 f"'send_nowait' method not implemented in {type(self.send_stream)}"
             ) from exc
+
+        send_nowait(item)
 
     async def send_eof(self) -> None:
         await self.send_stream.aclose()
--- a/tests/streams/test_stapled.py
+++ b/tests/streams/test_stapled.py
@@ -150,6 +150,11 @@ class DummyObjectSendStreamWithSendNowait(DummyObjectSendStream[T_Item]):
         self.buffer.append(item)
 
 
+class DummyObjectSendStreamWithBrokenSendNowait(DummyObjectSendStream[T_Item]):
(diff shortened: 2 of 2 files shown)


davidbrochart (approved):
Thanks @Zuhef.

agronholm:
If we implement this in other stream classes, we should probably extract the code into a helper function.

Zuhef:
Agreed in principle. Right now `StapledObjectStream.send_nowait()` is the only place that delegates an optional method like this, so a helper would have a single caller. The only other `except AttributeError` under `streams/` is in `MemoryObjectReceiveStream.receive()`, and that one is a different thing — a sentinel check on an unset attribute that turns into `EndOfStream`, not a delegation.

So I have left it inline here. If you would rather have the helper in place now so the next wrapper has something to reuse, say where you would want it to live and I will add it.

PR #1269, opened 2026-08-08
Fix UNIX socket close race with cancelled I/O

### LINKED ISSUE #1267: `UNIXSocketStream.aclose()` raises `InvalidStateError` if a concurrent `receive()` was just cancelled
## Things to check first

## AnyIO version

4.14.2

## Python version

3.11.15

## What happened?

On the asyncio backend, `UNIXSocketStream.aclose()` raises `asyncio.InvalidStateError`
when another task's `receive()` on the same stream has just been cancelled. Trio is
unaffected, and TCP streams on asyncio are unaffected — those go through a transport,
while UNIX streams use `_RawSocketMixin` and the raw socket directly.

This is the ordinary shape of a server: one task reads the connection, another owns it
and closes it when the server stops. Cancelling such a server — which is how an AnyIO
server is usually stopped — makes it raise on the way out:

```
  File ".../anyio/_backends/_asyncio.py", line 1443, in aclose
    self._receive_future.set_result(None)
asyncio.exceptions.InvalidStateError: invalid state
```

I'd expect `aclose()` to succeed here, as it does on trio and on asyncio TCP streams.

Tested on Linux; asyncio backend only.

### Diagnosis

`_RawSocketMixin` keeps the future a pending read is waiting on, and clears it from a
done-callback:

```python
    def _wait_until_readable(self, loop: asyncio.AbstractEventLoop) -> asyncio.Future:
        def callback(f: object) -> None:
            del self._receive_future
            loop.remove_reader(self.__raw_socket)

        f = self._receive_future = asyncio.Future()
        loop.add_reader(self.__raw_socket, f.set_result, None)
        f.add_done_callback(callback)
        return f
```

and `aclose()` resolves whatever is there:

```python
        if not self._closing:
            self._closing = True
            if self.__raw_socket.fileno() != -1:
                self.__raw_socket.close()

            if self._receive_future:
                self._receive_future.set_result(None)
            if self._send_future:
                self._send_future.set_result(None)
```

Cancelling the reading task cancels the future it is awaiting, so the future is *done*
immediately, but `callback` — the thing that clears `self._receive_future` — is a
done-callback, and asyncio runs those on a later iteration of the loop. In the window
between the two, `self._receive_future` is a future that is already done, and
`set_result` on it raises.

That window is exactly where the cancellation of a task group puts a second task that
closes the stream, which is why the reproducer needs the read and the close to be in
different tasks. When one task does both, the close only happens once the cancellation
has fully unwound and the callback has run, so the field is already clear and nothing
goes wrong — which is presumably why this has not shown up before.

`_send_future` has the same shape and presumably the same problem, reached by cancelling
a write rather than a read.

### Suggested fix

Only resolve a future that is not already resolved:

```python
            if self._receive_future and not self._receive_future.done():
                self._receive_future.set_result(None)
            if self._send_future and not self._send_future.done():
                self._send_future.set_result(None)
```

Nothing is lost by skipping it: a future that is already done has already woken whoever
was waiting on it, which is what setting the result was for. Applying that to an
installed 4.14.2 makes the reproducer below print `asyncio OK`.

### Workarounds

None that are any good. Catching it means catching `InvalidStateError` around every
`aclose()`, which is an asyncio-specific exception being caught in backend-agnostic
code, and it is indistinguishable from any other `InvalidStateError` that might come out
of a stream implementation. Letting the reader see EOF before closing avoids the window,
but only when there is something that can be waited for — a server being cancelled has
by definition stopped waiting.

## How can we reproduce the bug?

No third party libraries beyond AnyIO itself. Replacing `create_unix_listener`/
`connect_unix` with `create_tcp_listener`/`connect_tcp` makes it pass.

```python
import pathlib
import tempfile

import anyio


async def main() -> None:
    path = pathlib.Path(tempfile.mkdtemp()) / "sock"

    async def handle(stream: anyio.abc.SocketStream) -> None:
        # One task reads, another closes -- which is the shape of a server
        # whose reader is a task of its own.
        async with anyio.create_task_group() as inner:
            inner.start_soon(stream.receive)
            try:
                await anyio.sleep_forever()
            finally:
                await stream.aclose()

        listener = await anyio.create_unix_listener(path)
        tg.start_soon(listener.serve, handle)
        client = await anyio.connect_unix(path)
        await anyio.sleep(0.1)
        tg.cancel_scope.cancel()
    await client.aclose()


for backend in ("asyncio", "trio"):
    try:
        anyio.run(main, backend=backend)
        print(backend, "OK")
    except BaseException as exc:
        print(backend, "FAILED:", type(exc).__name__, exc)
```

Output:

```
asyncio FAILED: ExceptionGroup unhandled errors in a TaskGroup (1 sub-exception)
trio OK
```

alloutflo:

## Changes

Fixes #1267.

Only resolve the raw socket mixin's receive and send futures during close when they
are still pending. This prevents `InvalidStateError` when cancellation has already
completed a future but its cleanup callback has not run yet.

Add end-to-end regression coverage for cancelled UNIX stream reads and writes while
the owning server task closes the stream.

## Checklist

features

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/tests/test_sockets.py
+++ b/tests/test_sockets.py
@@ -58,6 +58,7 @@ from anyio import (
     getnameinfo,
     move_on_after,
     notify_closing,
+    sleep_forever,
     wait_all_tasks_blocked,
     wait_readable,
     wait_socket_readable,
@@ -1446,6 +1447,33 @@ class TestUNIXStream:
                 tg.start_soon(interrupt)
                 with pytest.raises(ClosedResourceError):
                     await stream.receive()
+
+    @pytest.mark.parametrize("anyio_backend", asyncio_params)
+    async def test_close_after_cancelled_receive(self, socket_path: Path) -> None:
+        async def handler(stream: SocketStream) -> None:
+            async def receive() -> None:
+                # Closing may wake the receive before cancellation reaches it.
+                with suppress(ClosedResourceError):
+                    await stream.receive()
+
+            async with create_task_group() as tg:
+                tg.start_soon(receive)
+                try:
+                    await sleep_forever()
+                finally:
+                    await stream.aclose()
+
+        client = None
+        try:
+            async with await create_unix_listener(socket_path) as listener:
+                async with create_task_group() as tg:
+                    tg.start_soon(listener.serve, handler)
+                    client = await connect_unix(socket_path)
+                    await wait_all_tasks_blocked()
+                    tg.cancel_scope.cancel()
+        finally:
+            if client is not None:
+                await client.aclose()
 
     async def test_receive_after_close(
         self, server_sock: socket.socket, socket_path: Path
--- a/tests/test_sockets.py
+++ b/tests/test_sockets.py
@@ -1449,15 +1449,22 @@ class TestUNIXStream:
                     await stream.receive()
 
     @pytest.mark.parametrize("anyio_backend", asyncio_params)
-    async def test_close_after_cancelled_receive(self, socket_path: Path) -> None:
+    @pytest.mark.parametrize("operation", ["receive", "send"])
+    async def test_close_after_cancelled_io(
+        self, socket_path: Path, operation: Literal["receive", "send"]
+    ) -> None:
         async def handler(stream: SocketStream) -> None:
-            async def receive() -> None:
-                # Closing may wake the receive before cancellation reaches it.
+            async def operate() -> None:
+                # Closing may wake the operation before cancellation reaches it.
                 with suppress(ClosedResourceError):
-                    await stream.receive()
+                    if operation == "receive":
+                        await stream.receive()
+                    else:
+                        while True:
+                            await stream.send(b"\0" * 4096)
 
             async with create_task_group() as tg:
-                tg.start_soon(receive)
+                tg.start_soon(operate)
                 try:
                     await sleep_forever()
                 finally:
--- a/tests/test_sockets.py
+++ b/tests/test_sockets.py
@@ -1448,7 +1448,6 @@ class TestUNIXStream:
                 with pytest.raises(ClosedResourceError):
                     await stream.receive()
 
-    @pytest.mark.parametrize("anyio_backend", asyncio_params)
     @pytest.mark.parametrize("operation", ["receive", "send"])
     async def test_close_after_cancelled_io(
         self, socket_path: Path, operation: Literal["receive", "send"]
--- a/.gitignore
+++ b/.gitignore
@@ -16,5 +16,4 @@ __pycache__
 .idea
 .cache
 .local
-.DS_Store
 CLAUDE.local.md
--- a/.gitignore
+++ b/.gitignore
@@ -16,4 +16,5 @@ __pycache__
 .idea
 .cache
 .local
+.DS_Store
 CLAUDE.local.md
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -23,6 +23,9 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
   module name. (The default name for a task spawned with ``TaskGroup.start_soon`` or
   ``TaskGroup.start`` typically includes the module name.)
   (`#1234 <https://github.com/agronholm/anyio/pull/1234>`_; PR by @gschaffner)
+- Fixed ``UNIXSocketStream.aclose()`` raising ``asyncio.InvalidStateError`` when a
+  concurrent I/O operation had just been cancelled on the asyncio backend
+  (`#1267 <https://github.com/agronholm/anyio/issues/1267>`_; PR by @alloutflo)
 - Fixed free-threading compatibility issues arising from the fact that on Python 3.14
   free-threading builds, newly created threads inherit the current context by default,
   causing AnyIO to behave erroneously in relation to ``start_blocking_portal()`` and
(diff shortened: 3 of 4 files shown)


alloutflo:
Update on the current head `e76a3efa`: the latest test-suite run completed successfully across all 20 reported checks, including Windows/Python 3.14. The earlier timeout described below was transient and no longer blocks this PR. The branch is mergeable and 0 commits behind `master`; only maintainer review remains.

<details>
<summary>Previous run context (superseded)</summary>

CI note: the sole failure is a 20-second timeout in the unchanged `tests/streams/test_tls.py::TestTLSStream::test_receive_invalid_max_bytes[asyncio+rsloop--1]` test on Windows/Python 3.14. That test uses only memory/stapled streams, runs before the UNIX socket tests, and its relevant source files are identical to the base branch. All other 16 jobs passed, including the asyncio-only job and Windows/Python 3.10; the previous run with the same production-code change also passed Windows/Python 3.14. I tried to rerun only the failed job, but GitHub requires upstream repository admin rights.

Follow-up: after adding send-side regression coverage, [the new run](https://github.com/agronholm/anyio/actions/runs/31258506034) completed successfully with all 18 workflow jobs green, including Windows/Python 3.14.

</details>

tests/test_sockets.py
```diff
@@ -1447,6 +1448,40 @@ async def interrupt() -> None:
                 with pytest.raises(ClosedResourceError):
                     await stream.receive()
 
+    @pytest.mark.parametrize("anyio_backend", asyncio_params)
```
agronholm: Is there something asyncio-specific in the test, or why did you limit it to asyncio?

tests/test_sockets.py
```diff
@@ -1447,6 +1448,40 @@ async def interrupt() -> None:
...
+                    else:
+                        while True:
+                            await stream.send(b"\0" * 4096)
+
+            async with create_task_group() as tg:
+                tg.start_soon(operate)
+                try:
+                    await sleep_forever()
+                finally:
+                    await stream.aclose()
```
agronholm, line 1470: What's this for?

tests/test_sockets.py
```diff
@@ -1447,6 +1448,40 @@ async def interrupt() -> None:
...
+
+        client = None
+        try:
+            async with await create_unix_listener(socket_path) as listener:
+                async with create_task_group() as tg:
+                    tg.start_soon(listener.serve, handler)
+                    client = await connect_unix(socket_path)
+                    await wait_all_tasks_blocked()
+                    tg.cancel_scope.cancel()
```
agronholm, line 1479: Why not just use a `CancelScope` in the test code? I don't see what the task group adds here, as you're not doing anything concurrently.

tests/test_sockets.py
alloutflo: This close is the trigger for the regression. Cancelling the surrounding task group cancels the pending receive/send future; closing the stream immediately in the handler `finally` block reaches `_RawSocketMixin.aclose()` before the future done callback clears the slot. Before this patch, `aclose()` called `set_result()` on that already-done future and raised `InvalidStateError`.

tests/test_sockets.py
alloutflo: The regression is asyncio-specific: the failing path and the fix are in `_backends/_asyncio.py` and `_RawSocketMixin`, where cancellation completes an `asyncio.Future` before its done callback clears `_receive_future` or `_send_future`. Trio does not use those futures. I limited the regression to `asyncio_params` so it targets the implementation being fixed rather than implying equivalent backend behavior.

tests/test_sockets.py
alloutflo: The task group is providing concurrency here: `listener.serve()` must run while the test connects the client and waits for the handler receive/send operation to block. Cancelling its scope then cancels the listener and active handler at the point that exposes the race. A plain `CancelScope` would not start `listener.serve()` concurrently. I can rewrite this around a pre-connected socket pair if you prefer to remove the server task group, but I did not want to change the test topology without confirmation.

agronholm: Right, of course. I somehow missed that.

tests/test_sockets.py
agronholm: We want to make sure Trio never develops the same issue, so if the test code doesn't require either backend, then the tests should also be run against all backends.

alloutflo: Agreed. I removed the asyncio-only parameterization in `af44e73`, so this regression now runs against every configured backend. With Trio installed, the targeted test passed locally across the available backend/socket combinations (10 passed, 10 abstract-namespace cases skipped on macOS), and all repository hooks for `tests/test_sockets.py` pass.

agronholm (approved):
LGTM, thanks!

clemenslosbichler-cloud:
The send side of #1269 looks untested to me.

Making the wake-up conditional on not self._send_future.done() puts that line's
correctness on a predicate whose "still pending, must wake" branch nothing
exercises. Receive has both: test_close_after_cancelled_io[...-receive] covers
already-done, and test_close_during_receive covers pending. Send only gets the
already-cancelled case, and if you replace the send wake-up with a no-op the
socket suite still passes.

The failure that admits is a task blocked in send() on a full socket buffer
while another task calls aclose(): permanent hang instead of
ClosedResourceError, with green CI. A TestUNIXStream.test_close_during_send
mirroring the receive one would pin it.

Smaller and separate: wait_readable/notify_closing already guard this same race
with try/except asyncio.InvalidStateError, and _RawSocketMixin.aclose() has its
own copy because it closes the socket before waking waiters. The three UNIX
subclasses share that copy, so the protocol now lives in two places.

Found with an automated reviewer I'm building, run against a fork with #1269
recreated on it. Happy to send the send-side test as a PR if you want it.

alloutflo:
Thanks for catching this. I confirmed the coverage gap against the current `master`: a focused `TestUNIXStream.test_close_during_send` passes with the existing wake-up, while replacing `_send_future.set_result(None)` with a no-op makes the asyncio case time out. So the pending-send branch is behaviorally significant and is not pinned by the tests merged in #1269.

Since #1269 is already merged, this belongs in a focused follow-up PR rather than another change to the old branch. Your offered send-side regression would be useful; I have not changed the merged implementation or opened a follow-up PR here.

PR #1275, opened 2026-08-12
Fix asyncio task group coroutine cleanup

dsfaccini:
> This pull request was posted by Codex Desktop using gpt-5.6-sol on behalf of David.

**NOTE** Erasing or replacing the contents of this template will result in your pull
request being summarily closed without consideration!

## Changes

Fixes #1274.

Close caller and wrapper coroutines when the asyncio event loop rejects task creation. Cleanup is state-aware so a suspended wrapper that may belong to a partially constructed task is left untouched, and cleanup errors cannot replace the original scheduling exception.

The regression covers ordinary `loop.create_task()` rejection and a supported custom eager task factory that closes the wrapper before raising.

## Checklist

If this is a user-facing code change, like a bugfix or a new feature, please ensure that
you've fulfilled the following conditions (where applicable):

- [x] You've added tests (in `tests/`) which would fail without your patch
- [ ] You've updated the documentation (in `docs/`), in case of behavior changes or new
features
- [x] You've added a new changelog entry (in `docs/versionhistory.rst`).

If this is a trivial change, like a typo fix or a code reformatting, then you can ignore
these instructions.

### Updating the changelog

If there are no entries after the last release, use `**UNRELEASED**` as the version.
If, say, your patch fixes issue <span>#</span>123, the entry should look like this:

```
- Fix big bad boo-boo in task groups
  (`#123 <https://github.com/agronholm/anyio/issues/123>`_; PR by @yourgithubaccount)
```

If there's no issue linked, just link to your pull request instead by updating the
changelog after you've created the PR.

PR #1276, opened 2026-08-12
Make the custom thread limiter test deterministic

dsfaccini:
> This pull request was posted by Codex Desktop using gpt-5.6-sol on behalf of David.

## Changes

Fixes the timing failure observed in https://github.com/agronholm/anyio/pull/1275.

Make `test_run_in_custom_limiter` wait for the third worker to enter before asserting limiter state. The test previously used a fixed delay, which can expire before the third worker is scheduled on free-threaded Python. The event preserves the assertion that a `CapacityLimiter(3)` permits exactly three concurrent workers without depending on scheduler timing.

## Checklist

- [x] Documentation is not applicable to this test-only change
- [x] A changelog entry is not applicable to this test-only change

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/tests/test_to_thread.py
+++ b/tests/test_to_thread.py
@@ -19,7 +19,6 @@ from anyio import (
     Event,
     create_task_group,
     from_thread,
-    sleep,
     to_thread,
     wait_all_tasks_blocked,
 )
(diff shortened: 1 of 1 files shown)


tests/test_to_thread.py
```diff
@@ -67,20 +66,24 @@ def thread_worker() -> None:
         nonlocal max_active_threads
         active_threads.add(threading.current_thread())
         max_active_threads = max(max_active_threads, len(active_threads))
+        if len(active_threads) == 3:
+            from_thread.run_sync(threads_started.set)
+
         event.wait(1)
```
agronholm: While at it, we might want to increase this time-out value here too to avoid flakiness in a busy CI environment.

agronholm (approved):

PR #1277, opened 2026-08-13
Fix concurrent force-close on asyncio socket streams

### LINKED ISSUE #1273: concurrent anyio.aclose_forcefully fails to block until socket is closed on asyncio
### Things to check first

### AnyIO version

f982b24986eac1ca8f34c91b0fdbaff8c378ba85

### Python version

3.12

### What happened?

Running the following tests fail on asyncio

```patch
From 267d773d30a63b9a22b6c389d7c099616e116628 Mon Sep 17 00:00:00 2001
From: Thomas Grainger <tagrain@gmail.com>
Date: Wed, 12 Aug 2026 21:30:48 +0100
Subject: [PATCH] add
 test_concurrent_aclose_forcefully_returns_before_socket_closes

---
 tests/test_sockets.py | 18 ++++++++++++++++++
 1 file changed, 18 insertions(+)

diff --git a/tests/test_sockets.py b/tests/test_sockets.py
index f475676..b1213ee 100644
--- a/tests/test_sockets.py
+++ b/tests/test_sockets.py
@@ -43,6 +43,7 @@ from anyio import (
     TCPConnectable,
     TypedAttributeLookupError,
     UNIXConnectable,
+    aclose_forcefully,
     as_connectable,
     connect_tcp,
     connect_unix,
@@ -507,6 +508,23 @@ class TestTCPStream:
                     await stream.receive()
 
+    async def test_concurrent_aclose_forcefully_returns_before_socket_closes(
+        self, server_addr: tuple[str, int]
+    ) -> None:
+        stream = await connect_tcp(*server_addr)
+        raw_socket = stream.extra(SocketAttribute.raw_socket)
+        fds: list[int] = []
+
+        async def do_aclose() -> None:
+            await aclose_forcefully(stream)
+            fds.append(raw_socket.fileno())
+
+            tg.start_soon(do_aclose)
+            tg.start_soon(do_aclose)
+
+        assert fds == [-1, -1]
+
     async def test_receive_after_close(self, server_addr: tuple[str, int]) -> None:
         stream = await connect_tcp(*server_addr)
         await stream.aclose()
-- 
2.43.0

```

gets 

```
FAILED tests/test_sockets.py::TestTCPStream::test_concurrent_aclose_forcefully_returns_before_socket_closes[asyncio-ipv4] - assert [15, -1] == [-1, -1]
FAILED tests/test_sockets.py::TestTCPStream::test_concurrent_aclose_forcefully_returns_before_socket_closes[asyncio-ipv6] - assert [15, -1] == [-1, -1]
FAILED tests/test_sockets.py::TestTCPStream::test_concurrent_aclose_forcefully_returns_before_socket_closes[asyncio+uvloop-ipv4] - assert [23, -1] == [-1, -1]
FAILED tests/test_sockets.py::TestTCPStream::test_concurrent_aclose_forcefully_returns_before_socket_closes[asyncio+uvloop-ipv6] - assert [23, -1] == [-1, -1]
FAILED tests/test_sockets.py::TestTCPStream::test_concurrent_aclose_forcefully_returns_before_socket_closes[asyncio+eager-ipv4] - assert [17, -1] == [-1, -1]
FAILED tests/test_sockets.py::TestTCPStream::test_concurrent_aclose_forcefully_returns_before_socket_closes[asyncio+eager-ipv6] - assert [17, -1] == [-1, -1]
```

see above

hansu650:

## Changes

Fixes #1273.

Ensure `SocketStream.aclose()` always aborts the asyncio transport and takes a
cancellable checkpoint before returning. This keeps concurrent `aclose_forcefully()`
calls from reporting completion while the raw socket is still open without suppressing
cancellation of normal `aclose()`.

## Checklist

features

these instructions.

### Updating the changelog

```
```

## Testing

- IPv4 `TestTCPStream` matrix on Windows/Python 3.13: 110 passed
- `CapacityLimiter` regression matrix after merging latest master: 81 passed
- Ruff check and format pass on changed Python files
- Windows IPv6 loopback is unavailable on this host (WinError 5/10013); CI remains
  authoritative


src/anyio/_backends/_asyncio.py
```diff
@@ -1393,15 +1393,19 @@ async def send_eof(self) -> None:
...
+                    pass
 
-            self._transport.close()
-            await sleep(0)
+                self._transport.close()
+                await sleep(0)
+        finally:
             self._transport.abort()
+            with CancelScope(shield=True):
+                await sleep(0)
```
graingert: can we add a flag in the connection_lost callback to StreamPorotocol so we know when to skip this cancel shielded checkpoint

graingert:
you say you reproduced this on windows, does UDPSocket and ConnectedUDPSocket have the same problem there?

src/anyio/_backends/_asyncio.py
hansu650: Implemented in a70b229. StreamProtocol.connection_lost() now records is_connection_lost, and SocketStream.aclose() only performs the cancel-shielded checkpoint when that callback has not completed yet. The focused concurrent force-close regression passes across the five IPv4 backend variants locally.

hansu650:
I also checked the Windows UDP paths: UDPSocket and ConnectedUDPSocket do not have the same race because both aclose() implementations already wait for DatagramProtocol.closed_event, which connection_lost() sets after the transport is closed. Their focused IPv4 FD-release regressions pass across the available backends (10 tests). The Windows IPv6 variants are unavailable on this machine (WinError 5/10013), so CI remains the cross-platform authority.

graingert:
`DatagramProtocol.closed_event.wait()` should be cancelled by `aclose_forcefully` before the socket is closed

hansu650:
Thanks — agreed. I updated both asyncio `UDPSocket.aclose()` and `ConnectedUDPSocket.aclose()` so the transport shared `closed_event` wait is shielded after `transport.close()`. This lets `aclose_forcefully()` cancel the surrounding close operation while still waiting for the OS file descriptor to be released. I also added force-close regression tests for both UDP socket types (asserting `fileno() == -1`). Validation on the focused UDP close matrix: 24 passed, 16 skipped for the existing Windows asyncio bug; Ruff and formatting checks pass. Commit: `daf4d16`.

src/anyio/_backends/_asyncio.py
```diff
@@ -1393,15 +1395,20 @@ async def send_eof(self) -> None:
...
 
-            self._transport.close()
-            await sleep(0)
+                self._transport.close()
+                await sleep(0)
+        finally:
             self._transport.abort()
+            if not self._protocol.is_connection_lost:
+                with CancelScope(shield=True):
+                    await sleep(0)
```
agronholm: This was supposed to be `cancel_shielded_checkpoint()`, right?

src/anyio/_backends/_asyncio.py
```diff
@@ -1393,15 +1395,20 @@ async def send_eof(self) -> None:
...
+            if not self._transport.is_closing():
+                try:
+                    self._transport.write_eof()
+                except OSError:
+                    pass
 
-            self._transport.close()
-            await sleep(0)
+                self._transport.close()
+                await sleep(0)
```
agronholm: Do we actually need two `sleep()` calls here?

src/anyio/_backends/_asyncio.py
hansu650: Yes — updated in 45f622d to use `AsyncIOBackend.cancel_shielded_checkpoint()` directly.

src/anyio/_backends/_asyncio.py
hansu650: The two checkpoints cover different phases. The first lets `transport.close()` process the graceful close and remains cancellable. If that is cancelled, or `connection_lost()` has not run yet, the `finally` block aborts the transport and the conditional cancel-shielded checkpoint gives the abort one event-loop turn to release the socket. The second checkpoint is skipped once `connection_lost()` has completed. After switching it to the backend helper, the focused IPv4 `aclose_forcefully` matrix passes (15 tests); this machine's IPv6 loopback remains unavailable with WinError 5/10013.

graingert:
we also need to make sure aclose() raises a CancelledError:

```py
stream = await connect_tcp(*server_addr)
exc = None

with CancelScope() as scope:
    scope.cancel()
    try:
        await stream.aclose()
    except anyio.get_cancelled_exc_class() as e:
        exc = e
        raise

assert exc is not None
```

the checkpoint doesn't need to be shielded to wait for the close event, there just needs to be a checkpoint - the sock.close() gets called from loop.call_soon which is called before the task is woken by cancellation

src/anyio/_backends/_asyncio.py
graingert: I don't think we need to call `transport.close()` because in `send` we always wait for the stream writer to drain, and we set `WriteTransport.set_write_buffer_limits(0)`

see https://github.com/python/cpython/blob/f6b1201e20345f3a7a546d458e9f1f2fbdd8fa61/Lib/asyncio/selector_events.py#L868-L871

hansu650: Addressed in b84eb22. SocketStream.aclose() now aborts the transport directly and takes a regular cancellable checkpoint; the graceful transport.close()/write_eof() path and the now-unused is_connection_lost flag were removed. I also added an asyncio-specific regression proving that cancellation is propagated. The full IPv4 TestTCPStream matrix passes locally (110 tests), and Ruff check/format pass.

hansu650:
Implemented the cancellation test and shutdown simplification in b84eb22, then merged the latest master in b3e2de6 to resolve the CI merge conflict. The new asyncio-specific regression verifies that SocketStream.aclose() raises the backend cancellation exception from an already-cancelled scope, while the existing concurrent aclose_forcefully() regression still verifies both callers return only after the raw socket is closed. Local validation: 110 IPv4 TestTCPStream cases, 81 CapacityLimiter cases, and Ruff check/format passed. This Windows host cannot open IPv6 loopback sockets (WinError 5/10013), so CI remains authoritative for IPv6.

PR #1279, opened 2026-08-13
Fix worker result race with closed event loop

### LINKED ISSUE #1265: race condition in WorkerThread and timing flakiness on aarch64 (test_deadline_moved fails)
### Things to check first

### AnyIO version

4.14.2

### Python version

3.14.6

### What happened?

```
[  184s] ______________ test_cancel_worker_thread[asyncio+eager-noabandon] ______________
[  184s] /usr/lib64/python3.14/threading.py:1082: in _bootstrap_inner
[  184s]     self._context.run(self.run)
[  184s] /usr/lib/python3.14/site-packages/anyio/_backends/_asyncio.py:1040: in run
[  184s]     self.loop.call_soon_threadsafe(
[  184s] uvloop/loop.pyx:1290: in uvloop.loop.Loop.call_soon_threadsafe
[  184s]     self._append_ready_handle(handle)  # deque append is atomic
[  184s] uvloop/loop.pyx:673: in uvloop.loop.Loop._append_ready_handle
[  184s]     self._check_closed()
[  184s] uvloop/loop.pyx:705: in uvloop.loop.Loop._check_closed
[  184s]     raise RuntimeError('Event loop is closed')
[  184s] E   RuntimeError: Event loop is closed
[  184s] 
[  184s] The above exception was the direct cause of the following exception:
[  184s] /usr/lib/python3.14/site-packages/_pytest/runner.py:353: in from_call
[  184s]     result: TResult | None = func()
[  184s]                              ^^^^^^
[  184s] /usr/lib/python3.14/site-packages/_pytest/runner.py:245: in <lambda>
[  184s]     lambda: runtest_hook(item=item, **kwds),
[  184s]             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[  184s] /usr/lib/python3.14/site-packages/pluggy/_hooks.py:512: in __call__
[  184s]     return self._hookexec(self.name, self._hookimpls.copy(), kwargs, firstresult)
[  184s]            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[  184s] /usr/lib/python3.14/site-packages/pluggy/_manager.py:120: in _hookexec
[  184s]     return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
[  184s]            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[  184s] /usr/lib/python3.14/site-packages/_pytest/logging.py:850: in pytest_runtest_call
[  184s]     yield
[  184s] /usr/lib/python3.14/site-packages/_pytest/capture.py:900: in pytest_runtest_call
[  184s]     return (yield)
[  184s]             ^^^^^
[  184s] /usr/lib/python3.14/site-packages/pluggy/_callers.py:53: in run_old_style_hookwrapper
[  184s]     return result.get_result()
[  184s]            ^^^^^^^^^^^^^^^^^^^
[  184s] /usr/lib/python3.14/site-packages/pluggy/_callers.py:38: in run_old_style_hookwrapper
[  184s]     res = yield
[  184s]           ^^^^^
[  184s] /usr/lib/python3.14/site-packages/pluggy/_callers.py:53: in run_old_style_hookwrapper
[  184s]     return result.get_result()
[  184s]            ^^^^^^^^^^^^^^^^^^^
[  184s] /usr/lib/python3.14/site-packages/pluggy/_callers.py:38: in run_old_style_hookwrapper
[  184s]     res = yield
[  184s]           ^^^^^
[  184s] /usr/lib/python3.14/site-packages/_pytest/skipping.py:268: in pytest_runtest_call
[  184s]     return (yield)
[  184s]             ^^^^^
[  184s] /usr/lib/python3.14/site-packages/_pytest/threadexception.py:147: in pytest_runtest_call
[  184s]     collect_thread_exception(item.config)
[  184s] /usr/lib/python3.14/site-packages/_pytest/threadexception.py:70: in collect_thread_exception
[  184s]     raise errors[0]
[  184s] /usr/lib/python3.14/site-packages/_pytest/threadexception.py:58: in collect_thread_exception
[  184s]     warnings.warn(pytest.PytestUnhandledThreadExceptionWarning(msg))
[  184s] E   pytest.PytestUnhandledThreadExceptionWarning: Exception in thread AnyIO worker thread
[  184s] E   Enable tracemalloc to get traceback where the object was allocated.
[  184s] E   See https://docs.pytest.org/en/stable/how-to/capture-warnings.html#resource-warnings for more info.

```

Run the test suite on very slow building machine (e.g., packaging openSUSE package on our OBS)

hansu650:

## Changes

Fixes #1265.

Catch the narrow race where an asyncio event loop closes after a worker result is
computed but before `call_soon_threadsafe()` schedules its delivery. The
`RuntimeError` is ignored only when the loop is closed; scheduling errors from an open
loop still propagate.

The regression test exercises `to_thread.run_sync()` end to end and synchronizes at
the asyncio result-delivery boundary, making the loop-close race deterministic without
constructing backend worker objects or manipulating their queues.

This follows the implementation direction previously developed by @shleder in #1245.
That PR's head repository was later deleted, so its branch can no longer be reopened;
thank you to @EmmanuelNiyonshuti for clarifying that history.

## Checklist

features

these instructions.

### Updating the changelog

```
```

If there's no linked issue, just link to your pull request instead by updating the

## Testing

- `pytest tests/test_to_thread.py`: 66 passed across asyncio and Trio
- Ruff check and format: passed
- Codespell and RST hooks: passed
- Mypy reports only the repository's existing Windows/POSIX and optional-loop import
  errors; no errors remain on the changed test lines
- `git diff --check`: passed


src/anyio/_backends/_asyncio.py
```diff
@@ -1042,9 +1042,13 @@ def run(self) -> None:
...
                     if not self.loop.is_closed():
-                        self.loop.call_soon_threadsafe(
-                            self._report_result, future, result, exception
-                        )
+                        try:
+                            self.loop.call_soon_threadsafe(
+                                self._report_result, future, result, exception
+                            )
+                        except RuntimeError:
+                            if not self.loop.is_closed():
```
EmmanuelNiyonshuti: I think this condition will never be true.

EmmanuelNiyonshuti: see: https://github.com/agronholm/anyio/pull/1245

hansu650:
Thanks for pointing this out. You're right: #1245 already contains the same implementation and substantially broader regression coverage. I searched open PRs before starting but missed the closed PR, so this contribution is a duplicate. Closing in favor of #1245 — apologies for the noise.

EmmanuelNiyonshuti:
Hi @hansu650  This not a duplicate. that PR I linked, its head repository was deleted.

hansu650:
Thanks for the correction — I missed that #1245's head repository had been deleted, so it can no longer be reopened and merged as-is. I'll reopen this PR as the available replacement and update the description to credit #1245 and @shleder's earlier work.

src/anyio/_backends/_asyncio.py
hansu650: The condition is the reverse guard: it is false for the expected shutdown race (the loop is now closed, so that RuntimeError is suppressed), but true if call_soon_threadsafe raises RuntimeError while the loop is still open. In that case the error is unrelated to shutdown and must be re-raised rather than silently swallowed. test_worker_thread_result_delivery_runtime_error verifies this open-loop branch.

EmmanuelNiyonshuti: What I meant is you can remove that  `loop._is_closed` check when you catch the error and just do nothing. or you may do it other way like the linked PR did.  I also just noticed that change hangs this test `test_worker_thread_propagates_runtime_error_from_open_loop`

hansu650: Thanks for clarifying. I removed the outer is_closed() pre-check and switched to the approach from #1245: always attempt call_soon_threadsafe(), then only suppress RuntimeError when the loop is confirmed closed. I also adjusted the race regression test so the open-loop RuntimeError test no longer hangs. Verified the full tests/test_to_thread.py file locally: 88 passed. Pushed in ff410bf.

hansu650:
The refreshed matrix fixed the original macOS/Python 3.14 failure, and the changed worker-thread tests pass there as well as on Linux and Windows. The only remaining macOS/Python 3.10 job timed out at 45% in an unrelated TestTCPStream::test_connection_refused[asyncio+rsloop-multi] run, before reaching tests/test_to_thread.py. I cannot rerun upstream Actions from the fork (admin permission required). Could a maintainer please rerun that one failed job?

agronholm:
> Could a maintainer please rerun that one failed job?

Done

tests/test_to_thread.py
```diff
@@ -22,6 +24,8 @@
     to_thread,
     wait_all_tasks_blocked,
 )
+from anyio._backends._asyncio import CancelScope as AsyncIOCancelScope
```
agronholm: Is there a reason you need to import the asyncio-specific implementation of the cancel scope? Why won't the public one do?

tests/test_to_thread.py
```diff
@@ -134,6 +138,60 @@ async def task_worker() -> None:
...
 
+@pytest.mark.parametrize("anyio_backend", asyncio_params)
+async def test_worker_thread_loop_closed_during_result_report(
+    mocker: MockerFixture,
+) -> None:
+    root_task = asyncio.current_task()
+    assert root_task is not None
+
+    loop = mocker.Mock(spec=asyncio.AbstractEventLoop)
+    loop.is_closed.return_value = True
```
EmmanuelNiyonshuti: I was wondering if we could instead use mocker.spy
```python
loop = asyncio.new_event_loop()
loop.close()

call_soon_spy = mocker.spy(loop, "call_soon_threadsafe")
is_closed_spy = mocker.spy(loop, "is_closed")
```
It would be alot clearer, I think?

tests/test_to_thread.py
```diff
@@ -134,6 +138,60 @@ async def task_worker() -> None:
     assert last_active == expected_last_active
 
 
+@pytest.mark.parametrize("anyio_backend", asyncio_params)
+async def test_worker_thread_loop_closed_during_result_report(
```
EmmanuelNiyonshuti: We could emphasize that this is the regression test,  `regression test for #1265`  The other test would pass on current master regardless.

tests/test_to_thread.py
EmmanuelNiyonshuti: the public one works well for me. @hansu650 gentle ping!

tests/test_to_thread.py
agronholm: Every test that doesn't exercise the backend's internals should be universal, to reinforce parity across backends. Not too long ago I actually made a PR submitter remove the asyncio-specificity from a test that didn't need it.

EmmanuelNiyonshuti: Yep! although, these tests do exercise asyncio internals through loop.call_soon_threadsafe. side note, though: not very necessary but if it stays asyncio-only  might be worth a test_asyncio_ prefix on the test name. I think it's more readable (many other asyncio-specific tests are also named that way!)

EmmanuelNiyonshuti:
The issue being fixed also claimed `tests/test_taskgroups.py::test_deadline_moved` to be flaky on aarch64. we might wanna fix that too as part of this PR. @mcepl would help take a look at this.

agronholm:
It's looking like @hansu650 no longer has their AI agent running/responding to comments. We may have to fix the PR ourselves.

tests/test_to_thread.py
EmmanuelNiyonshuti: > Is there a reason you need to import the asyncio-specific implementation of the cancel scope? Why won't the public one do?

I actually looked at this, It was probably due to mypy. since we are specifically using WorkerThread from asyncio backend and using its queue. I don't know, I used the public one `CancelScope` and silenced mypy. not sure if we can keep the way it was?

EmmanuelNiyonshuti:
> It's looking like @hansu650 no longer has their AI agent running/responding to comments. We may have to fix the PR ourselves.

Yeah, I applied the suggestions.  I’m still not sure if I addressed all of them correctly, let me know if I missed something. Also, not sure what to do with this test https://github.com/agronholm/anyio/blob/b29a2bc7ddd052fb47d0213132a6e109a0d4330d/tests/test_taskgroups.py#L1158-L1161  since I don’t currently have an  Arch environment to reproduce the issue locally.

EmmanuelNiyonshuti:
The test failing in the ci was not related. I rerun the job and now it's fixed 😎

agronholm:
The fix itself is simple enough, but I don't like the test. It feels hacky to me as it exercises the internals directly. That should only be done as the last resort. I'll try to find a better way to do this.

EmmanuelNiyonshuti:
> The fix itself is simple enough, but I don't like the test. It feels hacky to me as it exercises the internals directly. That should only be done as the last resort. I'll try to find a better way to do this.

Makes sense

tests/test_to_thread.py
hansu650: The direct backend test has now been replaced in a624d86, so it no longer imports either the backend-specific or public `CancelScope`, constructs `WorkerThread`, or writes to its queue.

tests/test_to_thread.py
hansu650: Reworked in a624d86. The replacement is explicitly the regression test for #1265 and exercises `to_thread.run_sync()` end to end. It remains asyncio-specific (and keeps the `test_asyncio_` prefix) because the deterministic barrier is placed at asyncio's `loop.call_soon_threadsafe()` boundary; it no longer reaches into AnyIO backend classes or queues.

hansu650:
I reworked the tests in a624d86 to exercise the public path end to end. The regression now calls `to_thread.run_sync(..., abandon_on_cancel=True)`, cancels the waiting task, and synchronizes only at the `loop.call_soon_threadsafe()` boundary so the runner closes the loop in the exact race window. It no longer imports or constructs `WorkerThread`, touches its queue, or uses cancel-scope internals; the redundant direct open-loop test was removed. Validation: `pytest tests/test_to_thread.py` (66 passed), Ruff check/format, and `git diff --check` all pass. Could you take another look?

hansu650:
Thanks for merging `master` into the branch. The refreshed test suite has one failing job: Ubuntu / Python 3.14 / asyncio-only failed in `TestTCPListener::test_tcp_listener_retry_after_partial_failure` with `OSError: Could not create 2 listeners with a consistent port`. The rest of the workflow matrix passed, including the changed worker-thread tests, and pre-commit.ci is green. This looks unrelated to this PR. My account cannot rerun the upstream job because GitHub requires repository admin rights. Could someone with access rerun job `97078207480`?

agronholm (approved):
LGTM now!

PR #1281, opened 2026-08-14
Cover the pending-send branch of _RawSocketMixin.aclose()

clemenslosbichler-cloud:

## Changes

Fixes #. 

`_RawSocketMixin.aclose()` wakes a pending send future so a blocked `send()`
raises `ClosedResourceError` instead of hanging. Nothing covers that today:
replacing `self._send_future.set_result(None)` with a no-op leaves the whole
socket suite green.

This adds `test_close_during_send`, mirroring `test_close_during_receive`.
Nothing reads from `server_sock`, so the send blocks once the buffer fills,
leaving a pending future for `aclose()` to wake.

On the review question: this is not specific to #1269. It passes against
e8ff01c too, since the wake-up predates that PR. #1269 making it conditional on
`not self._send_future.done()` is what prompted looking at the branch, but the
behaviour being pinned is older.

Test only, no implementation change.

## Checklist

features

these instructions.

### Updating the changelog

```
```


EmmanuelNiyonshuti:
@clemenslosbichler-cloud do not erase the PR template. 
please, read PR guidelines: https://github.com/agronholm/anyio/blob/master/AGENTS.md#pull-request-guidelines

clemenslosbichler-cloud:
@EmmanuelNiyonshuti you're right, my mistake. Template's back with the checklist filled in.

Left docs and changelog unchecked since this is a test for behaviour that already shipped in #1269, so there's nothing user-facing to document. Happy to add a changelog entry if you'd rather test-only PRs got one.

EmmanuelNiyonshuti:
> @EmmanuelNiyonshuti you're right, my mistake. Template's back with the checklist filled in.
> 
> Left docs and changelog unchecked since this is a test for behaviour that already shipped in #1269, so there's nothing user-facing to document. Happy to add a changelog entry if you'd rather test-only PRs got one.

It doesn't look like you restored the PR template?

tests/test_sockets.py
```diff
@@ -1448,6 +1448,23 @@ async def interrupt() -> None:
                 with pytest.raises(ClosedResourceError):
                     await stream.receive()
 
+    async def test_close_during_send(
```
EmmanuelNiyonshuti, line 1452: This test passes even without https://github.com/agronholm/anyio/pull/1269 in?

clemenslosbichler-cloud: Yeah, fair - it passes on e8ff01c too, so it's not testing #1269. Retitled.

It does still pin the wake-up itself (no-op out set_result(None) and the suite stays green while a blocked send() just hangs), but that's older than #1269 rather than from it.

If you'd rather have coverage for the conditional #1269 actually added, test_close_after_cancelled_io[...-send] already hits that case, so happy to close this instead.

EmmanuelNiyonshuti:
Looks good, thanks! I will leave this for @agronholm to give a final look!

agronholm (approved):
LGTM

PR #1286, opened 2026-08-21
Add missing keyword-only arguments to anyio.Path methods

jaideeppyne:
## Changes

`anyio.Path` aims to be a drop-in async mirror of `pathlib.Path`, but a few keyword-only arguments that CPython added in recent versions were missing, so passing them raised `TypeError`. This adds them, gated by `sys.version_info` so older interpreters keep the exact original signatures:

| Method | Argument | Added in stdlib |
|--------|----------|-----------------|
| `exists()` | `follow_symlinks` | Python 3.12 |
| `is_dir()` | `follow_symlinks` | Python 3.13 |
| `is_file()` | `follow_symlinks` | Python 3.13 |
| `read_text()` | `newline` | Python 3.13 |

Before this change, on Python 3.13:

```python
>>> await anyio.Path("link").is_file(follow_symlinks=False)
TypeError: Path.is_file() got an unexpected keyword argument 'follow_symlinks'
```

The new keyword arguments are forwarded to the underlying `pathlib.Path` call (via `partial`/`run_sync`), so the semantics match the standard library exactly.

## Tests

Added regression tests in `tests/test_fileio.py` covering each new argument, each guarded with the appropriate `skipif(sys.version_info < ...)`. They fail on `master` (with `TypeError`) and pass with this change. The existing `test_fileio.py` suite continues to pass, and `ruff check`, `ruff format`, and `mypy` are clean.

A changelog entry has been added under **UNRELEASED**.

---

🤖 This PR was prepared with the assistance of Claude Code.

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -6,9 +6,12 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 **UNRELEASED**
 
 - Added support for the newer keyword-only arguments on ``anyio.Path`` methods to match
-  the standard library ``pathlib.Path``: ``follow_symlinks`` on ``exists()`` (Python
-  3.12+), ``follow_symlinks`` on ``is_dir()`` and ``is_file()`` (Python 3.13+), and
-  ``newline`` on ``read_text()`` (Python 3.13+)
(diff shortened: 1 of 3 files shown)


agronholm (approved):
LGTM, thanks!

PR #1289, opened 2026-08-23
Fixed socket aclose() delayed close, missing checkpoint and Windows hang

### LINKED ISSUE #1237: UDPSocket.aclose() hangs on Windows when a send is in flight (regression from #1147)

### LINKED ISSUE #1273: concurrent anyio.aclose_forcefully fails to block until socket is closed on asyncio

### LINKED ISSUE #1288: UNIXSocket.aclose and _TrioSocketMixin.aclose in cancelled scope does not raise a cancelled error (does not checkpoint)
### Things to check first

### AnyIO version

a8900c22b58debb7cd43a4c8d94a1905e506a797

### Python version

py 3.14

### What happened?

```py
    async def test_aclose_in_cancelled_scope_raises_cancelled_exc(
        self, server_sock: socket.socket, socket_path: Path
    ) -> None:
        exc = None
        stream = await connect_unix(socket_path)

        with CancelScope() as scope:
            scope.cancel()
            try:
                await stream.aclose()
            except get_cancelled_exc_class() as e:
                exc = e
                raise

        assert exc is not None
```

failed

see above

graingert:

## Changes

Fixes #1237. Fixes #1273. Fixes #1288.

Socket `aclose()` implementations had three related problems.

On the asyncio backend, `SocketStream`, `UDPSocket` and `ConnectedUDPSocket` closed
through `transport.close()`, which only *schedules* the FD-closing `connection_lost`
callback via `loop.call_soon`. A second, concurrent `aclose()` — as
`aclose_forcefully()` can produce — therefore returned while the socket was still open,
and on Windows closing a UDP socket with a datagram still in flight could hang outright.
All three now `abort()` the transport directly, which also makes
`DatagramProtocol.closed_event` unnecessary.

None of the implementations checkpointed either, so `aclose()` called inside an
already-cancelled scope returned normally instead of raising a cancellation exception.
This affected UNIX socket streams and UNIX datagram sockets on asyncio, and every socket
type on Trio.

`_RawSocketMixin.aclose()` now always closes the raw socket and checkpoints, while still
resolving pending receive and send futures only on the first call, so a task blocked in
`receive()` or `send()` is woken exactly once.

## Checklist

features

these instructions.

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -78,6 +78,15 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 - Fixed ``CapacityLimiter`` on the asyncio backend over-granting tokens when
   ``total_tokens`` was raised while the limiter was over-subscribed
   (`#1223 <https://github.com/agronholm/anyio/pull/1223>`_; PR by @zelinewang)
+- Fixed UDP socket closing hanging on Windows if a datagram send was still in flight
+  (`#1237 <https://github.com/agronholm/anyio/issues/1237>`_; PR by @graingert)
+- Fixed concurrent ``aclose_forcefully()`` calls on asyncio socket streams returning
+  before the underlying socket was closed
+  (`#1273 <https://github.com/agronholm/anyio/issues/1273>`_; PR by @graingert)
+- Fixed ``aclose()`` on sockets not checkpointing, and thus failing to raise a
+  cancellation exception when called inside a cancelled scope (affected UNIX socket
+  streams and UNIX datagram sockets on asyncio, and all sockets on Trio)
+  (`#1288 <https://github.com/agronholm/anyio/issues/1288>`_; PR by @graingert)
 
 **4.14.2**
 
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -1290,20 +1290,17 @@ class DatagramProtocol(asyncio.DatagramProtocol):
     read_queue: deque[tuple[bytes, IPSockAddrType]]
     read_event: asyncio.Event
     write_event: asyncio.Event
-    closed_event: asyncio.Event
     exception: Exception | None = None
 
     def connection_made(self, transport: asyncio.BaseTransport) -> None:
         self.read_queue = deque(maxlen=100)  # arbitrary value
         self.read_event = asyncio.Event()
         self.write_event = asyncio.Event()
-        self.closed_event = asyncio.Event()
         self.write_event.set()
 
     def connection_lost(self, exc: Exception | None) -> None:
         self.read_event.set()
         self.write_event.set()
-        self.closed_event.set()
 
     def datagram_received(self, data: bytes, addr: IPSockAddrType) -> None:
         addr = convert_ipv6_sockaddr(addr)
@@ -1397,15 +1394,8 @@ class SocketStream(abc.SocketStream):
 
     async def aclose(self) -> None:
         self._closed = True
-        if not self._transport.is_closing():
-            try:
-                self._transport.write_eof()
-            except OSError:
-                pass
-
-            self._transport.close()
-            await sleep(0)
-            self._transport.abort()
+        self._transport.abort()
+        await AsyncIOBackend.checkpoint()
 
 
 class _RawSocketMixin:
@@ -1443,15 +1433,16 @@ class _RawSocketMixin:
         return f
 
     async def aclose(self) -> None:
-        if not self._closing:
-            self._closing = True
-            if self.__raw_socket.fileno() != -1:
-                self.__raw_socket.close()
-
+        closing = self._closing
+        self._closing = True
+        self.__raw_socket.close()
+        if not closing:
             if self._receive_future and not self._receive_future.done():
                 self._receive_future.set_result(None)
             if self._send_future and not self._send_future.done():
                 self._send_future.set_result(None)
+
+        await AsyncIOBackend.checkpoint()
 
 
 class UNIXSocketStream(_RawSocketMixin, abc.UNIXSocketStream):
@@ -1664,6 +1655,7 @@ class UNIXSocketListener(abc.SocketListener):
     async def aclose(self) -> None:
         self._closed = True
         self.__raw_socket.close()
+        await AsyncIOBackend.checkpoint()
 
     @property
     def _raw_socket(self) -> socket.socket:
@@ -1686,10 +1678,8 @@ class UDPSocket(abc.UDPSocket):
 
     async def aclose(self) -> None:
         self._closed = True
-        if not self._transport.is_closing():
-            self._transport.close()
-
-        await self._protocol.closed_event.wait()
+        self._transport.abort()
+        await AsyncIOBackend.checkpoint()
 
     async def receive(self) -> tuple[bytes, IPSockAddrType]:
         with self._receive_guard:
@@ -1736,10 +1726,8 @@ class ConnectedUDPSocket(abc.ConnectedUDPSocket):
 
     async def aclose(self) -> None:
         self._closed = True
-        if not self._transport.is_closing():
-            self._transport.close()
-
-        await self._protocol.closed_event.wait()
+        self._transport.abort()
+        await AsyncIOBackend.checkpoint()
 
     async def receive(self) -> bytes:
         with self._receive_guard:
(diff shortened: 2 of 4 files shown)


agronholm:
Does this obsolete #1273?
**EDIT**: Oops, you included that in the description, never mind.

graingert:
No it fixes it, #1273 is an issue, did you mean to reference a PR?

agronholm:
Sorry, I'm juggling a number of related PRs and issues in the milestone.

agronholm:
So, my first thought here was that we checkpoint in `aclose()` _after_ the close operation has happened. Usually, if we enter a coroutine function, we would normally checkpoint _before_ the actual changes, to allow for cancellation to take effect, but here we're only checkpointing after the fact. Granted, we didn't checkpoint *at all* previously, so it's not a regression, but this seems to be in stark contrast with the usual cancellation semantics. Thoughts?

graingert:
see https://trio.readthedocs.io/en/stable/reference-core.html#cancellation-and-primitive-operations

> Async cleanup operations – like `__aexit__` methods or async close methods – are cancellable just like anything else except that if they are cancelled, they still perform a minimum level of cleanup before raising [Cancelled](https://trio.readthedocs.io/en/stable/reference-core.html#trio.Cancelled).

graingert:
and https://trio.readthedocs.io/en/stable/reference-io.html#trio.abc.AsyncResource.aclose

agronholm:
> and https://trio.readthedocs.io/en/stable/reference-io.html#trio.abc.AsyncResource.aclose

OK, makes sense.

agronholm:
> No it fixes it, #1273 is an issue, did you mean to reference a PR?

Right, so I was actually thinking about #1246, #1255 and #1277. Got the numbers mixed up.

agronholm:
I have concerns about the `abort()`. According to the docs, it basically force-closes the connection without waiting for pending operations to finish. When force-closing the stream, that's fine, but `aclose()` is also called under normal closing circumstances. Wouldn't we potentially lose data if we did this?

graingert:
Because we call `set_write_buffer_limits(0)` once SocketStream.send returns there's no pending operations. If we call aclose concurrently to send only then do we lose data, but that's what Trio does too

PR #1291, opened 2026-08-25
Ensure socket aclose checkpoints after closing

### LINKED ISSUE #1288: UNIXSocket.aclose and _TrioSocketMixin.aclose in cancelled scope does not raise a cancelled error (does not checkpoint)

hansu650:

## Changes

Fixes #1288.

Add cancellation checkpoints after closing raw sockets in the asyncio and Trio
backends. This ensures that `aclose()` both releases the socket and delivers pending
cancellation. The regression test verifies both outcomes.

## Checklist

features (not applicable: this restores the expected cancellation checkpoint)

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -68,6 +68,9 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 - Fixed ``UNIXSocketStream.aclose()`` raising ``asyncio.InvalidStateError`` when a
   concurrent receive or send operation had just been cancelled on the asyncio backend
   (`#1267 <https://github.com/agronholm/anyio/issues/1267>`_; PR by @alloutflo)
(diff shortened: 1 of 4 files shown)


EmmanuelNiyonshuti:
closing this as it's a dup of https://github.com/agronholm/anyio/pull/1289

PR #1293, opened 2026-08-25
Add follow_symlinks argument to anyio.Path.owner() and group()

jaideeppyne:
Follow-up to #1286. `pathlib.Path.owner()` and `pathlib.Path.group()` gained a keyword-only `follow_symlinks` argument in Python 3.13, but `anyio.Path` didn't expose it, so passing it raised `TypeError`. #1286 added the newer keyword-only arguments for `exists()`/`is_dir()`/`is_file()`/`read_text()`, but `owner()`/`group()` were the two remaining methods that also gained `follow_symlinks` in 3.13.

This mirrors the existing version-gated pattern used for `is_dir()`/`is_file()`: on Python 3.13+ the argument is forwarded to the underlying `pathlib` method; on older interpreters the original signature is preserved.

Added regression tests (parametrized over both backends) covering a dangling symlink: `follow_symlinks=False` resolves the owner/group of the link itself, while `follow_symlinks=True` (the default) raises `FileNotFoundError` — matching `pathlib`. Tests skip on Windows (owner/group unsupported) and on Python < 3.13. Changelog updated; ruff and mypy pass.

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -14,6 +14,10 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
   * ``newline`` on ``read_text()`` (Python 3.13+)
 
   (`#1286 <https://github.com/agronholm/anyio/pull/1286>`_; PR by @jaideeppyne)
+- Added the ``follow_symlinks`` keyword-only argument to ``anyio.Path.owner()`` and
+  ``anyio.Path.group()`` (Python 3.13+) to match the standard library ``pathlib.Path``.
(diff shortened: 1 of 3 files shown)


agronholm:
Just amend the changelog entry for the previous PR. It's confusing to have a separate entry for these changes which should have been a part of that previous PR.

jaideeppyne:
Good point — done. Folded `owner()` and `group()` into the #1286 entry as two more `follow_symlinks` bullets and cited both PRs, and dropped the separate entry.

agronholm (approved):
Thanks.

PR #1295, opened 2026-08-26
don't silently discard UDP errors

graingert:

## Changes
Fixed UDP sockets on the asyncio backend silently discarding errors reported by the OS (such as an ICMP port unreachable). A task blocked in ``receive()`` or ``send()`` is now woken up by such an error, and both raise ``BrokenResourceError``, chained to the original error, as they already did on the Trio backend. As on the Trio backend, the error is only reported once, leaving the socket usable afterwards

## Checklist

features

these instructions.

### Updating the changelog

```
```


graingert:
will need to be rebased on https://github.com/agronholm/anyio/pull/1289 and https://github.com/agronholm/anyio/pull/1294

PR #1296, opened 2026-08-28
Fix CancelScope reuse not being detected on the asyncio backend

jaideeppyne:
## Changes

`CancelScope.__enter__` on the asyncio backend guards against reuse with `if self._active:`, but `__exit__` sets `_active = False`. So the guard only catches re-entering a scope that is still active, and entering a scope again *after* it has exited slips through silently, even though the guard's own message promises that each scope may only be used for a single `with` block. The Trio backend raises `RuntimeError` here.

`_cancel_called` also survives the first block, so reusing a cancelled scope silently cancels the body of the second block:

```python
import anyio

async def main():
    scope = anyio.CancelScope()
    with scope:
        scope.cancel()

    with scope:
        await anyio.sleep(0)
        print("this never runs on asyncio")

anyio.run(main)                   # prints nothing, no error
anyio.run(main, backend="trio")   # RuntimeError: Each CancelScope may only be used for a single 'with' block
```

I track entry with a separate `_has_been_entered` flag that is never reset, which is how `TaskGroup` already guards its own re-entry in the same module. Both backends now raise the same `RuntimeError`.

The two new tests fail on `asyncio`, `asyncio+uvloop` and `asyncio+eager` without the patch, and already passed on Trio.

I found this with an asyncio/Trio differential harness. AI-assisted, and I ran and reviewed all of it myself.

## Checklist

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -95,6 +95,11 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
     task.
 
   (`#1197 <https://github.com/agronholm/anyio/issues/1197>`_; PR by @tapetersen)
+- Fixed ``CancelScope`` on the asyncio backend allowing a scope to be entered again
+  after it had been exited, instead of raising ``RuntimeError`` like the Trio backend
+  does. Reusing a cancelled scope silently cancelled the body of the second ``with``
+  block (`#1296 <https://github.com/agronholm/anyio/pull/1296>`_;
+  PR by @jaideeppyne)
 
 **4.14.2**
 
--- a/src/anyio/_backends/_asyncio.py
+++ b/src/anyio/_backends/_asyncio.py
@@ -390,6 +390,7 @@ class CancelScope(BaseCancelScope):
         "_cancelled_caught",
         "_child_scopes",
         "_deadline",
+        "_has_been_entered",
         "_host_task",
         "_parent_scope",
         "_pending_uncancellations",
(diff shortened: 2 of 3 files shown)


fallenmi (approved):
Verified exact head `a58e99cc38bfafb3ed5d44506f578dd70223f80b` and the identical current merge tree at `51585c78dce4fa83598156b79ca6ecc224ecc6c6`.

The separate `_has_been_entered` flag correctly enforces one-shot use without overloading `_active`, which still tracks active-scope exit bookkeeping. This closes both ordinary post-exit reuse and the cancelled-scope path that silently cancelled the second block, and brings the asyncio backend in line with Trio. The two backend-parametrized regressions cover both paths, and the changelog entry is present.

All 18 exact-head check runs, the check suite and workflow, and both commit statuses are successful. I found no blocking issue. This was a source/API review; I did not rerun the suite locally.

Disclosure: this review was prepared with Codex assistance; I verified the exact refs, merge tree, source changes, policy, interactions, and live CI before submission.

jaideeppyne:
Thanks for the careful review and verification. I’ll leave the implementation as-is unless you want any naming or coverage adjustment before merge.

agronholm:
Was it necessary to introduce another flag? You could've just replaced `_active` completely with this new flag.

jaideeppyne:
I checked whether `_active` could be replaced outright, but it still carries current-lifetime semantics in three places: rejecting `__exit__` outside the active block, preventing deadline rescheduling after exit, and preventing task-group starts once its cancel scope has exited. A permanent entered-once flag cannot replace those checks without making post-exit operations look active. That is why this uses a separate monotonic flag for the one-shot invariant while retaining `_active` for the live interval. I can rename it to `_entered_once` if that distinction would read more clearly.

PR #1297, opened 2026-08-28
Fix wrong type in move_on_after

Graeme22:
## Changes

Fixes wrong type where `move_on_after` is typed as returning `AbstractContextManager[CancelScope]` even though it actually returns a `CancelScope`. This manifests in code where the scope is created first and entered later:

```python
scope = move_on_after(1)
print(scope.deadline)  # error: "AbstractContextManager[CancelScope, bool | None]" has no attribute "deadline"  [attr-defined]
with scope:
    ...
```

This code works fine, but mypy complains with `error: "AbstractContextManager[CancelScope, bool | None]" has no attribute "deadline"  [attr-defined]` as it (correctly) treats scope as an unentered context manager (like `fail_after` in current implementation) instead of a `CancelScope` that simply defines `__enter__` and `__exit__`.

## Checklist

features

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/src/anyio/_core/_tasks.py
+++ b/src/anyio/_core/_tasks.py
@@ -6,7 +6,7 @@ from collections.abc import (
     Coroutine,
     Generator,
 )
-from contextlib import AbstractContextManager, contextmanager
+from contextlib import contextmanager
 from enum import Enum, auto
 from inspect import iscoroutine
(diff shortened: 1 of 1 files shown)


agronholm (approved):
Good catch, thanks!

PR #1299, opened 2026-08-29
Fix SocketStream.send() writing to a paused transport after a cancelled send

graingert:

## Changes

On the asyncio backend, `StreamProtocol` sets the transport's write buffer high water
mark to 0, so the protocol is paused as soon as the transport has had to buffer
anything the OS would not take. `SocketStream.send()`, however, went straight to
`transport.write()` without waiting on the write event first.

`transport.write()` never offers anything to the OS while its buffer is non-empty — it
just appends. A `send()` cancelled while awaiting the write event leaves exactly that
state behind: data in the transport's buffer and a paused protocol. The next `send()`
then piled its data on top without the OS ever seeing it, so repeated cancellation
could grow the buffer without bound, despite the zero high water mark.

This waits on the write event before writing, as the datagram sockets will do in https://github.com/agronholm/anyio/pull/1294.

`UNIXSocketStream` is unaffected (it sends to the raw socket via `wait_writable`, with
no transport buffer). `TLSStream` wraps `SocketStream` and so inherits the fix.

## Checklist

features

### On the test

The test asserts on what reaches the peer rather than on transport internals, so it
runs on both backends and needs no mocking: it backs the connection up with two
cancelled sends of 8 MiB (the second soaks up whatever room the peer's
acknowledgements reopened after the first one blocked), cancels a third send, and then
asserts that none of that third payload arrives ahead of a final, uncancelled one. It
fails on every asyncio parameter without the patch and passes on trio either way,
which is the point — trio already behaves this way.

It is skipped on Windows, where the proactor transport does not stay paused while the
connection is backed up, so there is no back-pressure for the test to observe.

### DIFF
--- a/tests/test_sockets.py
+++ b/tests/test_sockets.py
@@ -251,11 +251,6 @@ class TestTCPStream:
             assert stream.extra(SocketAttribute.remote_address) == server_addr
             assert stream.extra(SocketAttribute.remote_port) == server_addr[1]
 
-    @pytest.mark.skipif(
-        sys.platform == "win32",
-        reason="the proactor transport does not stay paused while the connection is "
-        "backed up, so there is no back-pressure for this test to observe",
-    )
     async def test_cancelled_send_does_not_send_the_next_one(
         self, server_sock: socket.socket, server_addr: tuple[str, int]
     ) -> None:
@@ -282,7 +277,16 @@ class TestTCPStream:
                 await send_and_cancel(payload)
                 await send_and_cancel(payload)
 
-                # As the OS never accepted any of this, none of it may reach the peer
+                # On Windows, a transport can be genuinely backed up without its
+                # pause_writing() having fired yet: that only happens as a side effect
+                # of the next write() call discovering it, by which point that call's
+                # own data is already appended to the write buffer. Spend that one on
+                # a throwaway payload so the transport is *known* paused going into
+                # the next send() below, before it ever calls write() again.
+                await send_and_cancel(b"r" * 64)
+
+                # Now that the transport is known paused, the OS still never accepted
+                # any of this, so none of it may reach the peer
                 await send_and_cancel(b"c" * 64)
 
                 # Drain the peer until a final, uncancelled send() has arrived; data
--- a/tests/test_sockets.py
+++ b/tests/test_sockets.py
@@ -7,6 +7,7 @@ import io
 import os
 import platform
 import re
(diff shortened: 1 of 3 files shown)


tests/test_sockets.py
```diff
@@ -251,6 +251,62 @@ async def test_extra_attributes(
             assert stream.extra(SocketAttribute.remote_address) == server_addr
             assert stream.extra(SocketAttribute.remote_port) == server_addr[1]
 
+    async def test_cancelled_send_does_not_send_the_next_one(
+        self, server_sock: socket.socket, server_addr: tuple[str, int]
+    ) -> None:
+        """
+        Handing data to a paused transport merely appends it to the write buffer, from
+        where it is delivered anyway, so a cancelled ``send()`` must not have done so.
```
agronholm, line 259:
```suggestion
        If a ``send()`` was cancelled after the data was written to the buffer, the
        next call must ensure that the previous send completed one way or another
        before attempting to send its own data.
```

tests/test_sockets.py
```diff
@@ -251,6 +251,62 @@ async def test_extra_attributes(
...
+                # a throwaway payload so the transport is *known* paused going into
+                # the next send() below, before it ever calls write() again.
+                await send_and_cancel(b"r" * 64)
+
+                # Now that the transport is known paused, the OS still never accepted
+                # any of this, so none of it may reach the peer
+                await send_and_cancel(b"c" * 64)
+
+                # Drain the peer until a final, uncancelled send() has arrived; data
+                # that a cancelled send() wrongly handed over would arrive first
```
agronholm, line 293: I don't think a mere test warrants this many lines of commentary. Would it be possible to simplify this test with mocks instead of relying on whimsy OS-level behavior?

tests/test_sockets.py
```diff
@@ -251,6 +251,62 @@ async def test_extra_attributes(
...
+                received = bytearray()
+                arrived = False
+                with fail_after(60):
```
agronholm, line 296: This is a no-op given the 20 second default timeout we have for tests.

agronholm (changes requested):
The main code is fine, I just have some reservations about the test. If it can be simplified, then please do so.

agronholm (approved):
I made a couple of minor edits. I verified the test locally on Windows, and looks like everything that's left is really needed. Thanks!

PR #1302, opened 2026-08-30
Fix timeout value to a float in pyproject.toml

felixonmars:

## Changes

pytest 9 validates TOML option types. Convert to a float to satisfy the check.

## Checklist

features

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -134,7 +134,7 @@ addopts = [
 ]
 testpaths = ["tests"]
 anyio_mode = "auto"
-timeout = "20"
+timeout = 20.0
 xfail_strict = true
 filterwarnings = [
     "error",


agronholm:
We've been running pytest 9 for quite a while, so why hasn't this been a problem before?

agronholm:
Even with the latest pytest release, it still registers the 20s timeout even when it's a string. So what are you talking about?

felixonmars:
Oops, I thought it's about pytest 9 but actually it was pytest-timeout 2.5.0 (which was unpublished now): https://github.com/pytest-dev/pytest-timeout/issues/203

agronholm:
This change would make the test suite fail with the current release: `TypeError: /home/alex/workspace/anyio/pyproject.toml: config option 'timeout' expects a string, got float: 20.0`

felixonmars:
Yeah I think we should wait for an upstream fix first. Sorry for the noise!

agronholm:
This wasn't a pointless exercise though: the test suite should have run and digging deeper into that I noticed that there was a slight bug in the workflow on line 17 of `test.yml` which pointed to the wrong category.

PR #1303, opened 2026-08-30
Fixed RuntimeError in worker thread when the event loop closes during result reporting

shleder:

## Changes

No dedicated upstream issue for this failure mode; the earlier review history lives in #1245.

Fix the asyncio worker-thread shutdown race where the loop can close after
`is_closed()` returns false but before `call_soon_threadsafe()` delivers the result.
The undeliverable result is dropped, matching the selector-thread precedent.
Deterministic regressions cover the closed-loop race, unrelated `RuntimeError`
suppression at that call site, no undeliverable future publication, and normal result
delivery.

## Checklist

features

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -49,7 +49,7 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
   a ``RuntimeError`` if the event loop was closed after the worker had checked whether
   the loop was closed but before it could deliver the result of
   ``to_thread.run_sync()``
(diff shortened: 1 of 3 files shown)


agronholm:
What does this do that #1279 didn't already fix?

shleder:
Closing this as done-by-upstream: the race was fixed in #1279 (commit 02d7fbe), which landed the same restructure — direct `call_soon_threadsafe()` with `RuntimeError` swallowed only when the loop is actually closed — plus a deterministic regression test through the official `to_thread.run_sync()` API and a changelog entry. Nothing left for this PR to add. Thanks @agronholm and @hansu650 for getting it in properly.

PR #1304, opened 2026-08-31
Shield TemporaryDirectory cleanup from cancellation

m-u-xyz:

## Changes

Fixes #. 

`TemporaryDirectory.__aexit__` ran the synchronous cleanup via `to_thread.run_sync`,
which performs a cancellation checkpoint on entry. If the host task was already
cancelled when exiting the context manager, that checkpoint aborted the cleanup before
the worker thread was even dispatched, leaking the temporary directory.

This wraps the cleanup call in a shielded `CancelScope` so that a pending cancellation
cannot prevent the cleanup from running.

## Checklist

features

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -46,6 +46,9 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
   building a lookup table from the ``if TYPE_CHECKING:`` block. A fallback mode has been
   provided for installations where the source code is unavailable (e.g. PyInstaller).
   (`#1169 <https://github.com/agronholm/anyio/pull/1169>`_)
+- Fixed ``TemporaryDirectory`` not cleaning up when the host task was cancelled while
+  exiting the context manager, as the cleanup now runs in a shielded cancel scope
+  (`#1304 <https://github.com/agronholm/anyio/pull/1304>`_; PR by @smurfix)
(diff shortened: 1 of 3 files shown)


fallenmi (approved):
Verified exact head `eb93ac9e1ccf6659ccb932232c6aca3c9de111e5` and GitHub's current merge commit `1d57de45b1bc825b51b98824b700db3594500ce9`, whose tree `b981e28a1b020f4a76e59586b44bbb5d9e918ef1` is identical to the head tree.

The outer `CancelScope(shield=True)` closes the pre-dispatch cancellation window in `to_thread.run_sync()`: the asyncio backend performs a cancellation checkpoint before entering its own worker-thread shield, so a task already cancelled on context exit could previously skip `TemporaryDirectory.__exit__()` entirely. The new shield ensures cleanup is dispatched, while the default non-abandoning worker behavior waits for it to finish. The regression cancels the enclosing scope immediately before context exit and verifies that the directory is removed across the backend test matrix; the changelog entry is present.

All 18 exact-head check runs, the check suite and workflow, and both commit statuses are successful. I found no blocking issue. This was a source/API review; I did not rerun the suite locally.

Disclosure: this review was prepared with Codex assistance; I verified the exact refs, merge tree, source changes, policy, interactions, and live CI before submission.

agronholm (approved):
Simple and clean, thanks!

PR #1307, opened 2026-09-02
fix: clamp negative sleep delays to zero (Fixes #1306)

### LINKED ISSUE #1306: sleep() with a negative delay returns on asyncio but raises ValueError on trio
### Things to check first

### AnyIO version

4.14.2 (also reproduced on master, `033fc52`)

### Python version

3.14.6 (trio 0.34.0)

### What happened?

`anyio.sleep()` with a negative delay returns immediately on the asyncio backend and raises `ValueError` on the trio backend. Both backend implementations pass the value straight through, so whichever underlying library is in use decides what happens:

```python
# _backends/_asyncio.py
async def sleep(cls, delay: float) -> None:
    await sleep(delay)

# _backends/_trio.py
async def sleep(cls, delay: float) -> None:
    await trio.sleep(delay)
```

The same holds for `-inf`. Everything else nearby already agrees between backends — `sleep(0)`, `sleep(-0.0)`, `sleep_until()` with a past deadline, `move_on_after(-1)` and `fail_after(-1)` all behave identically, and `sleep(nan)` raises `ValueError` on both (only the message differs). So the divergence is narrow: negative finite values and `-inf`.

What makes this worth reporting rather than just a caller error is that AnyIO has already decided this question one level up. `sleep_until()` clamps:

```python
# _core/_eventloop.py
async def sleep_until(deadline: float) -> None:
    now = current_time()
    await sleep(max(deadline - now, 0))
```

and `tests/test_eventloop.py::test_sleep_until_in_past` asserts that a past deadline calls `sleep(0)`. A negative duration is therefore already treated as "don't wait" inside AnyIO — it just isn't applied in `sleep()` itself, so the behavior depends on the backend.

I couldn't find this documented as an accepted difference. `docs/signals.rst` spells out the Ctrl+C divergence between backends explicitly, and there's nothing equivalent for `sleep()`.

```python
import anyio

for backend in ["asyncio", "trio"]:
    try:
        anyio.run(lambda: anyio.sleep(-1), backend=backend)
        print(f"{backend:8} returned")
    except Exception as e:
        print(f"{backend:8} {type(e).__name__}: {e}")
```

```
asyncio  returned
trio     ValueError: `seconds` must be non-negative
```

The way this shows up in real code is a computed remaining-time argument, where elapsed time makes the value go slightly negative:

```python
import anyio


async def do_work() -> None:
    await anyio.sleep(0.01)


async def poll_until(deadline: float, interval: float = 0.05) -> str:
    while True:
        await do_work()
        remaining = deadline - anyio.current_time()
        if remaining <= 0:
            return "deadline reached"
        await anyio.sleep(min(interval, remaining) - 0.06)


async def main() -> str:
    return await poll_until(anyio.current_time() + 0.2)


for backend in ["asyncio", "trio"]:
    try:
        print(f"{backend:8} {anyio.run(main, backend=backend)}")
    except Exception as e:
        print(f"{backend:8} {type(e).__name__}: {e}")
```

```
asyncio  deadline reached
trio     ValueError: `seconds` must be non-negative
```

Identical code, one backend completes and the other raises.

### Which resolution would you prefer?

Two options, and I don't want to assume which one you want:

1. **Clamp to 0 in `AsyncBackend.sleep()` on both backends**, matching what `sleep_until()` already does. Non-breaking, and consistent with the existing internal convention.
2. **Raise on both.** Stricter, and it catches the caller bug rather than silently absorbing it. This is a behavior change for asyncio users, so presumably a major-release item.

There's precedent for the stricter direction — "`Lock` and `Condition` can now only be released by the task that acquired them. This behavior is now consistent on all backends whereas previously only Trio enforced this." — so option 2 may well be what you'd want here.

Happy to open a PR with either one, plus a parity test covering both backends. Just say which.

BetterAndBetterII:

## Changes

Fixes #1306.

`anyio.sleep()` passed negative delays straight through to the backend. asyncio
returns immediately; trio raises `ValueError`. `sleep_until()` already clamps with
`max(..., 0)`, so clamp negative finite delays and `-inf` to zero in `sleep()` for
the same cross-backend behavior. `sleep(nan)` is unchanged (NaN comparisons are
false, so the value is still passed through).

## Checklist

features

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/src/anyio/_core/_eventloop.py
+++ b/src/anyio/_core/_eventloop.py
@@ -89,9 +89,12 @@ async def sleep(delay: float) -> None:
     """
     Pause the current task for the specified duration.
 
-    :param delay: the duration, in seconds
+    :param delay: the duration, in seconds. Negative values are treated as zero.
 
     """
+    if delay < 0:
+        delay = 0
+
     return await get_async_backend().sleep(delay)
 
 
--- a/src/anyio/_core/_eventloop.py
+++ b/src/anyio/_core/_eventloop.py
@@ -92,8 +92,7 @@ async def sleep(delay: float) -> None:
     :param delay: the duration, in seconds. Negative values are treated as zero.
 
     """
-    if delay < 0:
-        delay = 0
+    delay = max(delay, 0)
 
     return await get_async_backend().sleep(delay)
 
--- a/src/anyio/_core/_eventloop.py
+++ b/src/anyio/_core/_eventloop.py
@@ -89,10 +89,13 @@ async def sleep(delay: float) -> None:
     """
     Pause the current task for the specified duration.
 
-    :param delay: the duration, in seconds. Negative values are treated as zero.
+    :param delay: the duration, in seconds. Negative values raise
+        :exc:`ValueError`.
+    :raises ValueError: if ``delay`` is negative (including ``-inf``) or NaN
 
     """
-    delay = max(delay, 0)
+    if not delay >= 0:
+        raise ValueError("delay must be a non-negative number")
 
     return await get_async_backend().sleep(delay)
 
--- a/src/anyio/_core/_eventloop.py
+++ b/src/anyio/_core/_eventloop.py
@@ -89,8 +89,7 @@ async def sleep(delay: float) -> None:
     """
     Pause the current task for the specified duration.
 
-    :param delay: the duration, in seconds. Negative values raise
-        :exc:`ValueError`.
+    :param delay: the duration, in seconds
     :raises ValueError: if ``delay`` is negative (including ``-inf``) or NaN
 
     """
--- a/tests/test_eventloop.py
+++ b/tests/test_eventloop.py
@@ -47,12 +47,13 @@ async def test_sleep_forever(fake_sleep: AsyncMock) -> None:
     fake_sleep.assert_called_once_with(math.inf)
 
 
-@pytest.mark.parametrize("delay", [-1.0, -math.inf])
-async def test_sleep_negative_delay(delay: float) -> None:
-    """Negative delays return immediately on all backends, matching sleep(0)."""
-    start = current_time()
-    await sleep(delay)
-    assert current_time() - start < 0.1
+@pytest.mark.parametrize(
+    "delay", [-1.0, -math.inf, math.nan], ids=["negative", "neg_inf", "nan"]
+)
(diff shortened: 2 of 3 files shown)


agronholm:
I much prefer erroring on invalid values.

agronholm:
Are you reading my messages?

BetterAndBetterII:
Sorry for the delay — I did see your note.

Switched this away from clamping: `sleep()` now raises `ValueError` for negative delays (including `-inf`) and NaN on every backend, before calling into asyncio/Trio. Docs and tests updated accordingly (`ec4da643`).

tests/test_eventloop.py
```diff
@@ -47,6 +47,15 @@ async def test_sleep_forever(fake_sleep: AsyncMock) -> None:
     fake_sleep.assert_called_once_with(math.inf)
 
 
+@pytest.mark.parametrize(
+    "delay", [-1.0, -math.inf, math.nan], ids=["negative", "neg_inf", "nan"]
```
agronholm:
```suggestion
    "delay",
    [
        pytest.param(-1.0, id="negative"),
        pytest.param(-math.inf, "neg_inf"),
        pytest.param(math.nan, id="nan")
    ]
```

tests/test_eventloop.py
```diff
@@ -47,6 +47,20 @@ async def test_sleep_forever(fake_sleep: AsyncMock) -> None:
     fake_sleep.assert_called_once_with(math.inf)
 
 
+@pytest.mark.parametrize(
+    "delay",
+    [
+        pytest.param(-1.0, id="negative"),
+        pytest.param(-math.inf, "neg_inf"),
```
agronholm:
```suggestion
        pytest.param(-math.inf, id="neg_inf"),
```

src/anyio/_core/_eventloop.py
```diff
@@ -89,9 +89,14 @@ async def sleep(delay: float) -> None:
     """
     Pause the current task for the specified duration.
 
-    :param delay: the duration, in seconds
+    :param delay: the duration, in seconds. Negative values raise
+        :exc:`ValueError`.
```
agronholm:
```suggestion
    :param delay: the duration, in seconds
```

agronholm (approved):
Looks good, thanks.

PR #1309, opened 2026-09-03
Added a workaround for third party code accessing unimported submodules

agronholm:

## Changes

This fixes the practical issue where third party code incorrectly accessed AnyIO submodules directly from the package without importing them first. This happened to work before due to said modules being eagerly imported as a side effect.

## Checklist

features

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/tests/test_lazyimport.py
+++ b/tests/test_lazyimport.py
@@ -134,30 +134,23 @@ def test_sourceless_install(tmp_path: Path) -> None:
     assert result["deprecations"] == DEPRECATIONS
 
 
-def test_futurewarning_on_access_without_direct_import() -> None:
+def test_submodule_access_without_direct_import() -> None:
     """
     Test that accessing an anyio submodule via attribute access, without having imported
-    it directly, emits a FutureWarning advising the user to import the submodule first.
+    it directly, imports the submodule.
     """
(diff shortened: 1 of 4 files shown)


agronholm:
I think I need to add a test as well.

tests/test_lazyimport.py
```diff
@@ -21,6 +21,20 @@
     "anyio.abc.Semaphore": "anyio.Semaphore",
 }
 
+# Submodules tracked by the lazy importer. Order matters: a submodule that imports
+# another tracked submodule must come after the ones it imports, so that its import does
+# not pre-set their attributes on the package and mask the expected warning.
```
EmmanuelNiyonshuti: Would you help me better understand this,? It seems I can reorder these submodules and have no problem. which expected warning?

EmmanuelNiyonshuti (commented):
I just left one clarifying question in the test. Otherwise this looks good to me.

tests/test_lazyimport.py
agronholm: Good catch, this was a stale comment now that `FutureWarning` is no longer emitted. I'll remove it

EmmanuelNiyonshuti (approved):

PR #1313, opened 2026-09-05
Fixed occasional bind failures with IPv6 and ephemeral ports

agronholm:

## Changes

This adds more retries for ephemeral ports, reducing the chances of random failure to a negligible number.

These failures were the number 1 cause of test suite flakiness.

## Checklist

features

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/src/anyio/_core/_sockets.py
+++ b/src/anyio/_core/_sockets.py
@@ -393,7 +393,7 @@ async def create_tcp_listener(
 
     errors: list[OSError] = []
     try:
-        for _ in range(len(sockaddrs)):
+        for _ in range(10):  # enough to absorb ephemeral port collisions
             listeners: list[SocketListener] = []
             bound_ephemeral_port = local_port
             try:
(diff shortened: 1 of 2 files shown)


src/anyio/_core/_sockets.py
```diff
@@ -393,7 +393,7 @@ def setup_raw_socket(
 
     errors: list[OSError] = []
     try:
-        for _ in range(len(sockaddrs)):
+        for _ in range(10):  # enough to absorb ephemeral port collisions
```
EmmanuelNiyonshuti, line 396: Is 10 retries based on something technical?

agronholm: It's a nice round number that, according to the AI, brings down the probabilities of 10 sequential clashes to a negligible number :P

EmmanuelNiyonshuti: okay, sounds like it'll do us good then!

EmmanuelNiyonshuti (approved):

agronholm:
Thanks.

PR #1316, opened 2026-09-08
Shield TemporaryDirectory.cleanup() from cancellation

Yasser-Ameur:

## Changes

Fixes #. 

`TemporaryDirectory.cleanup()` runs its cleanup through `to_thread.run_sync`, which
checkpoints on entry. When the host task is already cancelled, and it usually is,
because `cleanup()` gets awaited from a `finally:` block during unwind, the checkpoint
aborts before the worker thread runs and the directory is left on disk.

#1304 fixed exactly this for `__aexit__` with a shielded `CancelScope`. `cleanup()`, a
few lines below, kept the unshielded call. This gives it the same shield. The other
`to_thread.run_sync` teardown in the module, `AsyncFile.aclose()`, was already shielded
by #1314, so this was the last one.

`test_cleanup_method_with_cancellation` mirrors the test #1304 added for `__aexit__`.
It fails on the unmodified tree on all four backend variants and passes with the fix.

## Checklist

features

### DIFF
--- a/src/anyio/_core/_tempfile.py
+++ b/src/anyio/_core/_tempfile.py
@@ -513,7 +513,8 @@ class TemporaryDirectory(Generic[AnyStr]):
 
     async def cleanup(self) -> None:
         if self._tempdir is not None:
-            await to_thread.run_sync(self._tempdir.cleanup)
+            with CancelScope(shield=True):
+                await to_thread.run_sync(self._tempdir.cleanup)
 
 
 @overload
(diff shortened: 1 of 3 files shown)


src/anyio/_core/_tempfile.py
```diff
@@ -513,7 +513,8 @@ async def __aexit__(
 
     async def cleanup(self) -> None:
         if self._tempdir is not None:
-            await to_thread.run_sync(self._tempdir.cleanup)
+            with CancelScope(shield=True):
+                await to_thread.run_sync(self._tempdir.cleanup)
```
agronholm, line 517: @graingert I guess we need `checkpoint_if_cancelled()` here too?

Yasser-Ameur:
Good catch, you're right. I've added `await checkpoint_if_cancelled()` right after the shielded block, so a pending cancellation is still delivered to the caller once the directory is safely removed, matching the shield-then-checkpoint pattern already used elsewhere in this file. I extended the regression test to assert both that the directory is gone and that the cancellation is still raised, and confirmed the new assertion fails without the checkpoint. `__aexit__` has the same shield without a checkpoint after it; I left it alone since the question was about `cleanup()`, but I can add it in this PR if you'd like.

PR #1317, opened 2026-09-09
Reject NaN deadlines on asyncio to match Trio

dylanpulver:

## Changes

Fixes #. 

`CancelScope` on Trio raises `ValueError` for a NaN deadline; on asyncio the NaN was stored as-is. Every deadline comparison against NaN is false, so the scope never fired — `fail_after(math.nan)` returned instead of raising `TimeoutError`, and `move_on_after(math.nan)` never cancelled. A NaN delay reaching a timeout silently removes the timeout.

This is the same backend inconsistency #1307 fixed for `sleep()`, on the surfaces that were left. Measured before the change:

| call | asyncio | Trio |
|---|---|---|
| `sleep(nan)` | `ValueError` | `ValueError` |
| `CancelScope(deadline=nan)` | **returns** | `ValueError: deadline must not be NaN` |
| `scope.deadline = nan` | **returns** | `ValueError: deadline must not be NaN` |
| `move_on_at(nan)` / `fail_at(nan)` | **returns** | `ValueError: deadline must not be NaN` |
| `move_on_after(nan)` / `fail_after(nan)` | **returns** | `ValueError: deadline must not be NaN` |

Negative and `-inf` delays already agree across backends; only NaN diverged. The asyncio `CancelScope` now rejects it in the constructor and the `deadline` setter, with Trio's wording, which covers all six entry points since they all funnel through the scope.

## Checklist

features

Reverting `_backends/_asyncio.py` fails **18 of the 24** new test cases — every asyncio variant (`asyncio`, `asyncio+uvloop`, `asyncio+eager`) — while the 6 Trio ones still pass, which is the inconsistency itself. Full suite: **3972 passed, 299 skipped, 5 xfailed**. `pre-commit run` clean on both changed files.

No documentation change: `fail_after`/`move_on_after` do not currently document delay validation, and `sleep()`'s `:raises ValueError:` note from #1307 is the only such mention. Happy to add matching notes if you'd like them.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_011M5uTyCU4WcNTsPvGrErDo


tests/test_taskgroups.py
```diff
@@ -2536,3 +2536,26 @@ async def genfunc(x: int, y: int) -> AsyncGenerator[int, None]:
...
+
+
+@pytest.mark.parametrize(
+    "scope_factory",
+    [
+        pytest.param(lambda: CancelScope(deadline=math.nan), id="constructor"),
+        pytest.param(lambda: move_on_at(math.nan), id="move_on_at"),
+        pytest.param(lambda: fail_at(math.nan), id="fail_at"),
+        pytest.param(lambda: move_on_after(math.nan), id="move_on_after"),
+        pytest.param(lambda: fail_after(math.nan), id="fail_after"),
```
agronholm, line 2548: Why are we explicitly testing these too? They all come down to `CancelScope` anyway.

PR #1318, opened 2026-09-09
Shield file stream cleanup from cancellation

Kuang-xianxin:

## Changes

Related to #1314. 

When a cancelled task exits a `FileReadStream` or `FileWriteStream` context,
`to_thread.run_sync()` can raise before dispatching `file.close()`, leaving the
file open. Shield the shared close operation, then call
`checkpoint_if_cancelled()` after the file is closed so cancellation is delivered
by the close operation itself.

The regressions use public `from_path()` contexts and assert that closing both
releases the real file and propagates cancellation. They have no independent
checkpoint, scope-state or ordinary write/read-back assertions.

Validation on Python 3.13.14 / Windows:
- Both regressions fail on the original implementation because files remain open,
  and on the shield-only version because close does not raise cancellation.
  Each comparison has 8 failures across asyncio, winloop, eager asyncio and Trio.
- Final file-stream, AsyncFile and stapled-stream tests: 131 passed.
- Non-mypy pre-commit checks and focused source/test mypy pass.
- The full configured mypy hook has known Windows baseline diagnostics. The
  earlier wider file/path run had 372 passes, 96 skips and 76 Windows symlink
  failures/setup errors; all 76 also reproduced on the unchanged base. These
  platform-dependent checks are not claimed as passing.

The changelog records the fix; no user-guide addition is needed for expected
cleanup behavior.

## Checklist

features

these instructions.

### Updating the changelog

```

```

### DIFF
--- a/docs/streams.rst
+++ b/docs/streams.rst
@@ -219,6 +219,10 @@ File streams read from or write to files on the file system. They can be useful 
 substituting a file for another source of data, or writing output to a file for logging
 or debugging purposes.
 
+Closing a file stream, including when leaving its asynchronous context manager, waits
+for the file to close even if the calling task has been cancelled. For write streams,
+this also flushes buffered output before closing the file.
+
 Example::
 
     from anyio import run
--- a/docs/streams.rst
+++ b/docs/streams.rst
@@ -219,10 +219,6 @@ File streams read from or write to files on the file system. They can be useful 
 substituting a file for another source of data, or writing output to a file for logging
 or debugging purposes.
 
-Closing a file stream, including when leaving its asynchronous context manager, waits
-for the file to close even if the calling task has been cancelled. For write streams,
-this also flushes buffered output before closing the file.
-
 Example::
 
     from anyio import run
--- a/src/anyio/streams/file.py
+++ b/src/anyio/streams/file.py
@@ -14,6 +14,7 @@ from typing import IO, Any
 
 from .. import (
     BrokenResourceError,
+    CancelScope,
     ClosedResourceError,
     EndOfStream,
     TypedAttributeSet,
@@ -37,7 +38,8 @@ class _BaseFileStream:
         self._file = file
 
     async def aclose(self) -> None:
-        await to_thread.run_sync(self._file.close)
+        with CancelScope(shield=True):
+            await to_thread.run_sync(self._file.close)
 
     @property
     def extra_attributes(self) -> Mapping[Any, Callable[[], Any]]:
--- a/src/anyio/streams/file.py
+++ b/src/anyio/streams/file.py
@@ -22,6 +22,7 @@ from .. import (
     typed_attribute,
 )
 from ..abc import ByteReceiveStream, ByteSendStream
+from ..lowlevel import checkpoint_if_cancelled
 
 
 class FileStreamAttribute(TypedAttributeSet):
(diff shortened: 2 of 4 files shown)


docs/streams.rst
```diff
@@ -219,6 +219,10 @@ File streams read from or write to files on the file system. They can be useful
 substituting a file for another source of data, or writing output to a file for logging
 or debugging purposes.
 
+Closing a file stream, including when leaving its asynchronous context manager, waits
+for the file to close even if the calling task has been cancelled. For write streams,
+this also flushes buffered output before closing the file.
+
```
agronholm: Is this warranted? Shielding around closing ought to be the expectation.

Kuang-xianxin: Removed the extra user-guide paragraph in 77eb696. Agreed that shielding close is expected behavior; the changelog entry and cancellation regressions remain. This follow-up only changes documentation, and the applicable documentation pre-commit checks pass.

tests/streams/test_file.py
```diff
@@ -105,6 +122,24 @@ async def test_write_after_close(self, file_path: Path) -> None:
...
+                async with stream:
+                    await stream.send(b"Hello")
+                    scope.cancel()
+
+                await anyio.lowlevel.checkpoint()
+                pytest.fail("Cancellation was not propagated")
+
+            assert scope.cancelled_caught
+            assert file.closed
+            assert file_path.read_bytes() == b"Hello"
```
agronholm: Is there a situation where the test would fail at this point?

agronholm (changes requested):
The added tests seem to also try to validate the cancellation mechanism. There are other tests to cover those, so these parts are just unnecessary noise here, agreed?

tests/streams/test_file.py
Kuang-xianxin: Agreed: once this real buffered file is closed, the read-back assertion only repeats ordinary write/close behavior covered elsewhere. In 261034c I removed that assertion and its payload write, along with the explicit checkpoint, `pytest.fail()` and `cancelled_caught` checks from both tests. Each regression now only triggers cancellation during context exit and asserts that the real file is closed.

The simplified tests still fail in all eight cases on the original implementation (asyncio, winloop, eager asyncio and Trio); the 131 related file-stream/AsyncFile/stapled-stream tests pass. The production fix is unchanged.

src/anyio/streams/file.py
```diff
@@ -37,7 +38,8 @@ def __init__(self, file: IO[bytes]):
         self._file = file
 
     async def aclose(self) -> None:
-        await to_thread.run_sync(self._file.close)
+        with CancelScope(shield=True):
+            await to_thread.run_sync(self._file.close)
```
graingert, line 43: I think this should be like trio and include a checkpoint_if_cancelled:

https://github.com/python-trio/trio/blob/a3d7cddeacef74c87f667e9f9e8615f2ab7ce894/src/trio/_file_io.py#L286-L298

Kuang-xianxin: Added in 4d8114e: the shielded file close completes first, then `checkpoint_if_cancelled()` delivers pending cancellation, matching the linked implementation.

The two existing regressions now expect cancellation from context exit itself and check that the real file was closed. They fail in all eight backend cases on the previous shield-only commit (no cancellation raised), and also on the original implementation (file left open). The 131 related file-stream/AsyncFile/stapled-stream tests and focused mypy pass; non-mypy pre-commit hooks pass.

Kuang-xianxin:
Merged current master in 4d078f3 and resolved the changelog conflict, retaining both upstream entries and this PR's entry. The shielded close and post-close cancellation checkpoint are unchanged.

Validation: 131 file-stream/AsyncFile/stapled-stream tests passed on Windows; all non-mypy pre-commit hooks and the focused source/test mypy check passed. The broader Path tests still encounter the previously documented Windows symlink privilege failures; no full-suite pass is claimed.

PR #1320, opened 2026-09-10
Allow Condition.notify() when underlying lock is held directly

### LINKED ISSUE #1319: `Condition.notify()` fails when shared lock is held
### Things to check first

### AnyIO version

4.14

### Python version

3.11

### What happened?

When using a `Condition` with an existing lock, it's not possible to call `notify(...)` unless the lock was acquired via the condition.

This makes it difficult to use multiple conditions which share the same lock (i.e., when different types of waiter depend on the same underlying state).

For example:

```python
async with self._lock:
    mutate_state()

    if pending_changed:
        # !! RuntimeError: The current task is not holding the underlying lock.
        self._pending_condition.notify()

    if ready_changed:
        self._ready_condition.notify()
```

To work around this, I think the only solutions would be to:

a. Use a single condition to notify every waiter of every change
b. Acquire and release the lock twice, losing the atomicity of the state change and the notifications

```python
import anyio

async def async_main():
    lock = anyio.Lock()
    cond = anyio.Condition(lock)
    async with lock:
        cond.notify()  # !! RuntimeError

anyio.run(async_main)
```

GruffElixir:
## Changes

Fixes #1319.

Update `Condition` to determine the owner task dynamically from the underlying lock's statistics (`self._lock.statistics().owner`) instead of storing `_owner_task` separately only within `Condition.acquire()` and `Condition.acquire_nowait()`.

This resolves two issues:
1. `Condition.notify()` and `Condition.notify_all()` can now be called when the underlying lock is acquired directly (such as when multiple conditions share the same lock).
2. `Condition.notify()` now raises `RuntimeError` if called after the lock has been released (previously `_owner_task` was retained on the condition instance after release).

## Checklist

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -5,6 +5,8 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
 **UNRELEASED**
 
+- Fixed ``Condition.notify()`` and ``Condition.notify_all()`` failing when the underlying lock was acquired directly rather than through the condition
+  (`#1319 <https://github.com/agronholm/anyio/issues/1319>`_; PR by @GruffElixir)
 - Fixed ``sleep()`` treating a negative delay (including ``-inf``) inconsistently
   across backends (asyncio returned immediately; Trio raised ``ValueError``) by
   raising ``ValueError`` on all backends
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -5,8 +5,6 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
 
 **UNRELEASED**
 
-- Fixed ``Condition.notify()`` and ``Condition.notify_all()`` failing when the underlying lock was acquired directly rather than through the condition
-  (`#1319 <https://github.com/agronholm/anyio/issues/1319>`_; PR by @GruffElixir)
 - Fixed ``sleep()`` treating a negative delay (including ``-inf``) inconsistently
   across backends (asyncio returned immediately; Trio raised ``ValueError``) by
   raising ``ValueError`` on all backends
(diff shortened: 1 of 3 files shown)


src/anyio/_core/_synchronization.py
```diff
@@ -292,14 +291,17 @@ async def __aexit__(
     ) -> None:
         self.release()
 
+    @property
+    def _owner_task(self) -> TaskInfo | None:
+        return self._lock.statistics().owner
+
     def _check_acquired(self) -> None:
         if self._owner_task != get_current_task():
```
agronholm: There is no point in introducing a `@property` that is only used in one place.
```suggestion
    def _check_acquired(self) -> None:
        if self._lock.statistics().owner != get_current_task():
```

docs/versionhistory.rst
agronholm: New fixes go to the bottom.
```suggestion
- Fixed ``sleep()`` treating a negative delay (including ``-inf``) inconsistently
across backends (asyncio returned immediately; Trio raised ``ValueError``) by
raising ``ValueError`` on all backends
(`#1306 <https://github.com/agronholm/anyio/issues/1306>`_; PR by @BetterAndBetterII)
- Fixed ``create_tcp_listener()`` occasionally failing with a message like
``Could not create 2 listeners with a consistent port`` when an ephemeral port is
requested and IPv6 is enabled and the dual-stack path is not available or a specific
local host name was given
- Fixed ``AsyncFile`` not shielding against cancellation while closing
(`#1314 <https://github.com/agronholm/anyio/pull/1314>`_)
- Fixed ``Condition.notify()`` and ``Condition.notify_all()`` failing when the underlying lock was acquired directly rather than through the condition
  (`#1319 <https://github.com/agronholm/anyio/issues/1319>`_; PR by @GruffElixir)
```

agronholm (changes requested):
Validated against Trio's native implementation. Just a couple of change requests.

GruffElixir:
Thanks, cleaned both up: _check_acquired() now reads the underlying lock directly, and the changelog entry is at the bottom of the unreleased fixes. The focused sync tests are green (12 passed).

src/anyio/_core/_synchronization.py
GruffElixir: Yep, agreed. `_check_acquired` is a plain method now, so there is no property just for one call.

docs/versionhistory.rst
GruffElixir: Yep, fixed. The entry is at the bottom of `UNRELEASED` now.

agronholm (approved):
Looks good now, thanks!

PR #1323, opened 2026-09-11
Fix AnyIO imports with loaders that omit __file__

### LINKED ISSUE #1322: `anyio` imports break when the app is packaged with PyOxidizer
### Things to check first

### AnyIO version

4.15.1

### Python version

3.10.9 (because PyOxidizer uses this particular version of Python)

### What happened?

When an app is packaged with PyOxidizer and it tries to import `anyio`, the app crashes with the following exception:

```
  File "<string>", line 1, in <module>
  File "anyio", line 11, in <module>
  File "anyio._lazyimport", line 23, in install_lazy_importer
  File "anyio._lazyimport", line 116, in _build_lazy_map
  File "inspect", line 1285, in getsource
  File "inspect", line 1267, in getsourcelines
  File "inspect", line 1078, in findsource
  File "inspect", line 955, in getsourcefile
  File "inspect", line 916, in getfile
TypeError: <module 'anyio' (<oxidized_importer.OxidizedFinder object at 0x1054d0ac0>)> is a built-in module
```

This is because PyOxidizer uses an import mechanism that loads the modules directly from the binary of the executable and it does not make the source code of the module available. Contrary to `anyio`'s expectations, it does _not_ throw an `OSError` when `inspect.getsource()` fails but a `TypeError` is thrown instead. This is then ignored by `_build_lazy_map()` and the exception propagates to the top level. Extending `_build_lazy_map()` to catch `TypeError` (or maybe a blanket `Exception`) around `inspect.getsource()` would probably solve the problem. Checking `getattr(sys, "frozen", False)` would also be an alternative to detect PyOxidizer, PyInstaller and many other tools that freeze a Python app into an executable.

Build a simple application with PyOxidizer that attempts to import `anyio`. Attached is a self-contained ZIP file with the required files; running `./run.sh` will create a virtualenv, install PyOxidizer in it from PyPI, then build the executable with PyOxidizer. Besides Python, a Rust compiler is also needed in the environment.

[anyio-pyoxidizer-crash-repro.zip](https://github.com/user-attachments/files/32099546/anyio-pyoxidizer-crash-repro.zip)

skulitom:

## Changes

Fixes #1322. 

Handle `TypeError` from `inspect.getsource()` with the existing eager-import fallback, alongside `OSError`. This lets loaders that omit `__file__` import AnyIO instead of crashing.

The regression is a small in-process test: a `ModuleType` without `__file__` naturally raises `TypeError` in `inspect.getsource()`, and the test asserts that `_build_lazy_map()` returns the empty maps used by the eager fallback. It adds no subprocess or custom loader. The existing sourceless-install test covers eager imports, public module names, and deprecated aliases.

Validation of the updated test on Windows:

- The new regression fails with the reported `TypeError` when the exception-handler change is removed.
- `pytest tests/test_lazyimport.py -q`: **6 passed** on Python 3.10.16 and **6 passed** on Python 3.13.2.
- Pre-commit checks for the changed file pass, with mypy run separately using the hook's version and arguments plus `--platform linux`: **78 source files pass**. Native Windows mypy previously showed platform-specific errors also present on unchanged upstream.
- `git diff --check` passes.

Prepared and self-reviewed with Codex. No PyOxidizer executable was built. No public API or usage-documentation change is needed; the changelog records the fallback fix.

## Checklist

features

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -15,6 +15,9 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
   local host name was given
 - Fixed ``AsyncFile`` not shielding against cancellation while closing
   (`#1314 <https://github.com/agronholm/anyio/pull/1314>`_)
+- Fixed importing AnyIO with loaders that do not provide ``__file__``, such as
+  PyOxidizer, raising ``TypeError`` instead of falling back to eager imports
+  (`#1322 <https://github.com/agronholm/anyio/issues/1322>`_; PR by @skulitom)
 
 **4.15.1**
 
--- a/src/anyio/_lazyimport.py
+++ b/src/anyio/_lazyimport.py
@@ -114,7 +114,7 @@ def _build_lazy_map(
 ) -> tuple[dict[str, tuple[str, str]], dict[str, str], list[str]]:
     try:
         source = inspect.getsource(module)
-    except OSError:
+    except (OSError, TypeError):
         return {}, {}, []
 
     tree = compile(source, module.__file__ or "", "exec", ast.PyCF_ONLY_AST)
--- a/tests/test_lazyimport.py
+++ b/tests/test_lazyimport.py
@@ -130,6 +130,45 @@ def test_sourceless_install(tmp_path: Path) -> None:
     assert result["deprecations"] == DEPRECATIONS
 
 
+def test_import_without_module_file() -> None:
+    """Test the eager fallback for loaders that do not set module.__file__."""
+    script = dedent("""\
+    import sys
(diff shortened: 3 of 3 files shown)


agronholm:
A subprocess test, such as the one added by this PR, are pretty heavy. I'm not sure it's justified to cater for a packager that hasn't been updated in 4 years. The handling of `TypeError` itself is palatable, but the test might not be justifiable unless you can point to another tool or environment that has the same problem.

agronholm (approved):
Alright, that test looks nice and clean. Thanks!

PR #1327, opened 2026-09-15
Add threadless Emscripten file I/O fallback

### LINKED ISSUE #1268: Fallbacks for threading-less environment?
### Things to check first

### Feature description

I wonder if it would make sense for AnyIO functions that use threading in the background gets fallback to the single-threaded version when threading is not available.

For example, `anyio.open_file` currently uses `aynio.to_thread.run_sync` internally, and using it in threading-less environment such as WebAssembly fails with `can't start new thread` error.

### Use case

I maintain [Pyodide](https://github.com/pyodide/pyodide) + WebAssembly/Emscripten platform in CPython. In WebAssembly environment, often threading is not available and codes that use threading runs into a runtime error.

Users of startlette and FastAPI were having issues porting their application to WebAssembly, as codes such as `anyio.open_file` or `anyio.to_thread.run_sync` requires threading support in the runtime environment.

Originally, I was discussing this issue in starlette (https://github.com/Kludex/starlette/pull/3284), but I was wondering if this can be solved in AnyIO side, by providing threading-less fallbacks of those functions.

Of course, I understand that running everything including the I/O functions in the same thread means it will degrade the performance. But it would help users port their existing codes that use AnyIO without modification.

dexhunter:

## Changes

Fixes #1268 

On Emscripten builds without pthreads, asynchronous file operations now run blocking callbacks on the event-loop thread while preserving limiter admission, cancellation before callback execution, copied callback context, and shielded cleanup. These blocking operations can block the event loop and cannot be preempted after they start. The fallback covers `open_file`, `wrap_file`, and `AsyncFile`; thread-based `Path` operations, temporary-file creation, general `to_thread`, and `from_thread` behavior remain unchanged. Documentation and regression coverage describe these boundaries.

## Checklist

features

these instructions.

### Updating the changelog

```
```

## Validation

- Focused native tests: 669 passed and 10 skipped.
- Real Pyodide primary comparison: baseline 27 unmet cases to candidate 0; the earlier lifecycle replay passed 10 cases.
- The earlier local full-suite comparison ran 4,436 cases for each baseline and candidate. Baseline: 4,158 passed, 250 failed, 23 skipped, 5 xfailed. Candidate: 4,166 passed, 242 failed, 23 skipped, 5 xfailed. There were zero candidate-only failures. The 242 shared socket failures occurred in the private namespace, and their cause was not established for every failure; that local comparison was not clean.
- The follow-up replaces the test's `os.O_ACCMODE` assertion with `os.O_WRONLY | os.O_RDWR` for Windows and PyPy compatibility. The repaired source passed the 27 required Pyodide cases and 7 controls, the focused native tests, and all 14 pinned pre-commit hooks without source changes. All 20 GitHub checks passed at `330e5ac`, including Windows and PyPy.
- Weco evidence: [Source snapshots and validation history](https://dashboard.weco.ai/share/7pbIpAXat7Aicc7UWZWxCDdyVZxcTR3Z) (latest candidate: step 4).


src/anyio/_core/_fileio.py
```diff
@@ -47,6 +49,22 @@
...
+
+
+async def _run_sync_in_fileio(
+    func: Callable[..., T_Retval], *args: Any, limiter: CapacityLimiter | None
+) -> T_Retval:
+    emscripten_info = getattr(sys, "_emscripten_info", None)
+    if (
+        sys.platform != "emscripten"
+        or getattr(emscripten_info, "pthreads", None) is not False
+    ):
```
agronholm: Does this information change while the process runs, or is there another reason it's not precomputed?

src/anyio/_core/_fileio.py
```diff
@@ -47,6 +49,22 @@
...
+        return await to_thread.run_sync(func, *args, limiter=limiter)
+
+    await checkpoint()
```
agronholm, line 65: What's the `checkpoint` here for? Acquiring a limiter already does that.

agronholm:
I don't like this solution for three reasons:

1. It's specific to Emscripten, while there are a number of other environments that don't support threading
2. It's specific to file I/O, leaving `to_thread.run_sync()` as a problem for these platforms
3. The check for threading support isn't cached

agronholm:
Let's discuss this in the issue.

PR #1328, opened 2026-09-15
Defer move_on_after deadline until context entry

### LINKED ISSUE #1298: Defer move_on_after() timer start to its use as a context manager
Due to a historical mistake, `move_on_after()` returned a cancel scope directly, and the deadline got computed on the initial call. This does not line up with `fail_after()` which correctly defers the deadline computation to when the context manager is actually entered. This is a breaking change.

Rayan-and-beyond:

## Changes

Fixes #1298.

`move_on_after()` now computes its relative deadline when the context manager is entered instead of when the function is called, matching `fail_after()`. This keeps time spent between creating and entering the context manager from consuming the timeout. A delayed-entry regression covers the supported backends, and the behavior docs and changelog are updated.

## Checklist

features

these instructions.

### Updating the changelog

```
```


Rayan-and-beyond:
Two CI jobs failed outside the changed area: CPython 3.11 segfaulted in `TestUNIXStream::test_from_socket_not_connected[asyncio-fileno]`, and PyPy 3.11 failed `TestTCPStream::test_cancelled_send_does_not_send_the_next_one[trio-ipv6]` after 2082 tests passed. The other matrix jobs plus docs, pyright, and pre-commit passed. GitHub does not allow me to rerun the upstream workflow from the fork; could a maintainer rerun the failed jobs?

agronholm:
We didn't do this yet because it's a breaking change. I will defer this to v5.0.

PR #1330, opened 2026-09-17
Fix `CancelScope` swallowing a concurrent native `Task.cancel()` on asyncio

### LINKED ISSUE #1214: CancelScope can swallow a concurrent native asyncio.Task.cancel() when the scope is also cancelled

Zac-HD:
## Changes

Fixes #1214.

When a native `Task.cancel()` is delivered in the same event loop iteration as the scope's own cancellation of its host task, asyncio folds both requests into a single `CancelledError` carrying the scope's cancellation message. `CancelScope.__exit__` identified that exception as its own and swallowed it, so the host task continued normally with `cancelling() > 0` and the native cancellation was lost.

The scope now records `Task.cancelling()` on entry and, after undoing its own cancellation requests on exit, lets the exception propagate if the count is still above that baseline, mirroring `asyncio.timeout()`.

also closes https://github.com/agronholm/anyio/pull/1251, which I noticed after writing this patch with Claude.  My tests are simpler; the implementations are very similar except that they set `cancelled_caught=True` and I leave it `False` in this  case (morally speaking, the explicit cancellation 'landed before' the `CancelScope` took effect), happy to reverse that if you prefer.

## Checklist

    *(n/a, sufficiently niche bugfix that I thought the changelog entry was sufficient)*


agronholm:
I indeed find this more palatable than the other PR. I'll just make sure I understand the semantics correctly before I merge.

agronholm:
What I'm still in the middle of figuring out is why we need to store the cancellation count for the host task when entering. None of the new tests exercise that with a nonzero cancellation count on enter.

agronholm:
It would seem like `test_cancel_message_replaced` (not added in this PR) answers my question. If we have *already* swallowed a `CancelledError` when entering a cancel scope, then that pre-existing nonzero cancel count should not change the semantics of cancellation handling, and that's why we need to record the count on enter.

agronholm:
I feel like at least one of the tests should actually test with a nonzero cancel count on enter. With that, I'd be ready to merge this.

PR #1333, opened 2026-09-19
Fix deferred deletion of named temporary files

jakezwang:

## Changes

`NamedTemporaryFile(delete_on_close=False)` leaves its file behind after `async with` exits on Python 3.12+. Call the wrapped tempfile context manager's `__exit__` in the existing shielded worker-thread pattern so deferred deletion runs.

Tests cover normal, exceptional and cancelled exits, manual close, and `delete=False`. Tempfile tests pass on Python 3.10 (54 passed, 18 skipped), 3.12 and 3.14 (96 each), with all supported backends. Pre-commit passes. The full 3.14 suite has 3,995 passes and one timing-sensitive Trio IPv6 socket failure, reproduced on the unchanged base; it passes in an isolated rerun. Documentation builds with the same 96 warnings as the base.

## Checklist

features

these instructions.

### Updating the changelog

```
```


agronholm:
If you call it with `delete_on_close=False`, why would you expect the file to be deleted on close?

jakezwang:
Agreed: an explicit `af.aclose()` should not delete the file when `delete_on_close=False`. The change is to `NamedTemporaryFile.__aexit__`, not `AsyncFile.aclose()`. With `delete=True`, the standard library keeps the file on close and deletes it on context exit. The regression test explicitly closes the file inside the context, asserts that it still exists there, then checks deletion after context exit. `delete=False` continues to preserve the file after exit. I re-ran all tempfile tests on Python 3.14 with asyncio and Trio: 96 passed.

PR #1334, opened 2026-09-20
Fix the custom typed attribute provider examples

agammann:

## Changes

The custom provider example is missing the `Any` import, and both `extra_attributes` properties are missing their `self` parameter. Add these so the examples can be executed and the base provider's attributes can be inherited and overridden.

Validation: executing the original examples raises `NameError` for `Any`; supplying that import then exposes the missing-`self` `TypeError`. The corrected examples run against this checkout, returning both base attributes and preserving the float attribute while overriding the string in the subclass. Visually checked the rendered page. A strict Sphinx build produces the same 119 cross-reference warnings on the unchanged baseline and this patch; none are introduced by this change. No library behavior is changed.

## Checklist

features

these instructions.

### Updating the changelog

```
```


docs/versionhistory.rst
```diff
@@ -30,6 +30,9 @@ This library adheres to `Semantic Versioning 2.0 <http://semver.org/>`_.
   legitimately ``None``
   (`#773 <https://github.com/agronholm/anyio/issues/773>`_; PR by @AmirF194)
 
+- Fixed the custom typed attribute provider documentation examples
+  (`#1334 <https://github.com/agronholm/anyio/pull/1334>`_; PR by @agammann)
+
```
agronholm: A minor documentation fix doesn't warrant a changelog entry.
```suggestion
```

agronholm (approved):
Alight, looks good now. Thanks!

PR #1338, opened 2026-09-22
Send EOF when run_process receives explicitly empty input

Aurelucien:

## Changes

No existing issue. 

Treat an explicitly supplied empty byte string as subprocess input. Previously, `run_process(input=b"")` inherited the parent standard input, so it could read unintended data or wait for terminal input instead of receiving EOF. Check for `None` both when selecting the input pipe and when sending/closing it. Preserve `input=None` inheritance and rejection of simultaneous non-None `stdin` and `input`.

A real parent/child subprocess regression supplies a sentinel to the parent: on the unmodified base both asyncio and Trio incorrectly read it with empty input (2 failures), while inherited and nonempty input controls pass (4 passes). The fix passes all six cases. Existing stdin conflict tests now also cover empty input.

Validation: Python 3.12 subprocess suite 103 passed / 3 skipped; Python 3.10 78 passed / 3 skipped. Both parent and child import the current worktree via PYTHONPATH. Full pre-commit and Sphinx -W passed. Previously established whole-suite baseline remains applicable; unrelated DNS tests were not rerun.

Developed with OpenAI Codex assistance.

## Checklist

features

these instructions.

### Updating the changelog

```
```


Aurelucien:
Could you rerun the two failed test jobs? Python 3.11 segfaulted in a UNIX datagram socket test, and PyPy failed the Trio IPv6 cancelled-send test. Neither failure is in the subprocess code changed here; a rerun would help establish whether these are intermittent. I don't have permission to rerun this workflow.

